PR #34, opened 2022-10-19
fix: dependency error when contributing

helloitsm3:
This PR contains fixes for issues faced when trying to install dependencies.

image


helloitsm3:
@helloanoop feel free to merge if this is okay

PR #36, opened 2022-10-20
Added use local storage hook

helloitsm3:


helloitsm3:
@helloanoop I'll be separating each features into their own individual PRs now

Issue #53, opened 2022-10-28
Search and Copy functionality need to be supported in response


vyskh2498:
Currently, the functionality to search is limited to the request body and not available in the response body. In case of large response received for certain requests the user is unable to search in the response body. Similarly in case of copying the response also the user is restricted.

helloanoop:
Thanks for reporting this @vyskh2498 
Will get this fixed this weekend.

helloanoop:
@vyskh2498 This has been fixed in the latest release.

Thanks for reporting this.
Closing this ticket now.

PR #55, opened 2022-10-30
fix(#53): fix for response editor search issue

anusreesubash:


helloanoop:
Thanks for this @anusreesubash !

Issue #59, opened 2022-10-31
Collections icons partially working


shash68i:
On clicking icons, the action should be the same as clicking on the text alongside the icon. 'Import Collection' and 'Load Sample Collection' icons are clickable, but the Icons of 'Create Collection' and 'Add Collection to Workspace' has no action.

On hovering the icons, cursor should become pointer.

![bruno-issue](https://user-images.githubusercontent.com/37452438/198947226-0d32bd61-b120-4248-b95e-379c64d8be53.png)

helloanoop:
Nice catch!
And I agree that cursor should become pointer as you hover over the icon.

The fix should be fairly easy. @shash68i Would you like to pick this up ?

shash68i:
Sure, will work on this.

shash68i:
Thanks!
Looking forward to contribute more.

helloanoop:
This has been released. Closing the issue.
@shash68i  you should checkout the latest version of bruno.

Lots of new features are available: https://docs.usebruno.com/

PR #61, opened 2022-10-31
Collections icons partially working #59 - resolved

shash68i:
Collection items whole div made as action button and cursor pointer added to the whole div, including icons.


helloanoop:
Merged!
Thanks for this @shash68i

PR #66, opened 2022-11-07
Bugfix/split line

### LINKED ISSUE #62: Split line should have max squeeze limit
The split line should only be allowed to go up to a certain limit on both sides. Allowing it to go till the end is making stuffs overflow and non-readable.

bruno-issue

bruno-issue-2

Anx450z:
Closes #62 

## Added limit to right pane also
- min-width of the right pane is set to 350px.


helloanoop:
Awesome @Anx450z !
Thanks for the attention of detail here and taking care of the right pane width!

PR #69, opened 2022-11-09
Set default theme to the user's browser theme

nashidahmed:
Set the default theme to the user's preferred browser theme.

[Browser support](https://caniuse.com/prefers-color-scheme) for `prefers-color-scheme`.

**Note**: There is a [known issue](https://bugs.chromium.org/p/chromium/issues/detail?id=998903) with Chrome in Linux which causes the `prefers-color-scheme` to always detect light mode.


helloanoop:
This is cool 😎
Merged! Thanks a lot @nashidahmed

PR #70, opened 2022-11-10
RequestPane body form input text color visibility fix in dark mode

Gvinod1991:
In dark mode body form input text color visibility issue fix. Implemented color style of input in both dark and light theme files.


helloanoop:
Merged!
Thanks for this @Gvinod1991 !

PR #90, opened 2023-02-11
Compiled and cached expressions

ajaishankar:
Enabled compiling and caching of expressions.

Currently expression evaluation is dynamic and a new function need to be created every time an expression is evaluated.

By identifying the variables used in an expression, we are able to create a function and cache it.

```js
res.data.pets.map(pet => pet.name.toUpperCase())
```
In the above expression there are two variables that are identified (`res` and `pet`). These are picked up from the context when the function is called.

```js
function(context) {
  const { res, pet } = context;
  return res.data.pets.map(pet => pet.name.toUpperCase())
}
```
Though pet is not really in context, it is harmless and does not affect evaluation in any way...

### DIFF
--- a/packages/bruno-js/package.json
+++ b/packages/bruno-js/package.json
@@ -9,6 +9,9 @@
   "peerDependencies": {
     "vm2": "^3.9.13"
   },
+  "scripts": {
+    "test": "jest --testPathIgnorePatterns test.js"
+  },
   "dependencies": {
     "atob": "^2.1.2",
     "ajv": "^8.12.0",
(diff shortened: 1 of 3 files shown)


helloanoop:
Sweet 👏, Thanks @ajaishankar !

ajaishankar:
Thanks for the merge @helloanoop! 

Could you add bruno-js package also to the github unit test checks?

helloanoop:
@ajaishankar Added the unit tests to github workflows
https://github.com/usebruno/bruno/commit/1a175e44498bcaf556694d57067aad864fba088e

image

PR #91, opened 2023-02-13
res.get : deep object navigation and filtering

ajaishankar:
This PR borrows ideas from lodash, jsonpath, jsonQuery and jsonata and augments the the Bruno response with a `get` method that provides a very powerful and familiar syntax to query data!

Something like this
```js
res.body.data[0].journeys[0].segments[0].legs[0].legKey
```
Can be expressed succinctly as
```js
res.get('..legs[0].legKey')
```
Or let's add a filter!
```js
res.get('..legs[?].legKey', leg => leg.origin === 'BLR')
```

Currently getting and filtering deeply nested properties from the response is quite verbose. To alleviate it (somewhat) Bruno allows jsonQuery using the `res(query)` method. The downside is user has to learn yet another query language. Real world API responses like the one above are complex and users often resort to writing code in script blocks.

Since Bruno embraces Javascript for expressions let's embrace it for querying too!


helloanoop:
@ajaishankar This is fantastic !!

The `res(expr)` currently accepts a jsonQuery.
We still have time to left to change this behavior, as this feature has not been rolled out in the UI, and is currently only used in the CLI that we are prototyping and experimenting on.

The initial plan was to provide the option to override the default query parser.

Now, I am thinking that we should use your implementation, (thinking of calling it as bru query), as the default query.
```js
res('..legs[0].legKey') // parses it using bru query
res.jq('data.orderNumber') // parses it using json query
res.jp('data.orderNumber') // parses it using json path
res.jsonata('data.orderNumber') // parses it using jsonata
```
Instead of having the user to configure/override a query parser, I am thinking of supporting the 3 common json querying libs by default via an api like `res.jq()`, `res.jp()`, `res.jsonata()`

I think this simplifies things. What are your thoughts ?
If you are in agreement, please go ahead and make the changes to use your implementation as the default and move the jsonQuery to be available using `res.jq()` api

ajaishankar:
Thanks @helloanoop!

Really liking the way we are able to separate *what* we want to get (the path) from the filters.

I agree instead of a plugin we can simply provide `res.jq()`, `res.jp()`, `res.jsonata()`.

Glad the res(...) has not been released in the UI yet!

Will update the PR today to default to the bru query implementation and also add `res.jq()`  👍

ajaishankar:
> I also think, this deserves its own package inside our monorepo. I suggest to call it as `bruno-query` This will get shipped to npm as `@usebruno/query` in our npmjs org https://www.npmjs.com/package/@usebruno/cli
> 
> This would help `bruno-query` to be useful outside bruno itself Thoughts ?
> 
> If you are in agreement, feel free to implement the new package using typescript. The `bruno-graphql-docs` [here](https://github.com/usebruno/bruno/tree/main/packages/bruno-graphql-docs) uses typescript. I have used rollup to build it, open to using any other builder of your choice.

Yes I agree, feel this is a clean and powerful json query mechanism. It's like lodash get on steroids 😄 

Will move it to a separate package, rollup should be fine...

ajaishankar:
@helloanoop have added the bruno-query package.

Since this is in typescript there is a build step, have added that also to the unit test github workflow.

Could you review?

Thanks!

helloanoop:
Awesome 🚀
Nice work, @ajaishankar !!

Merged.

ajaishankar:
Thanks @helloanoop!

helloanoop:
@ajaishankar 
Published `@usebruno/query` `0.1.0` to npm
Also published `@usebruno/cli` `0.3.0` to use bruno-query while evaluation vars.

ajaishankar:
Have a couple of further ideas for bruno-query that I'll try do a PR this weekend...

1. Simplified filter for equality comparisons

    ```js
     get(data, 'items[?].amount', i => i.category == 'electronics' && i.brand == 'apple')

     get(data, 'items[?].amount', { category: 'electronics', brand: 'apple' }
     ```
4. Type inference from the query expression

helloanoop:
Sounds good @ajaishankar !

helloanoop:
@ajaishankar I have added the documentation for the query here https://docs.usebruno.com/scripting/response-query.html

ajaishankar:
Looks nice! @helloanoop 👍

PR #99, opened 2023-02-18
Correct the tab order on the CreateCollection modal

dcoomber:
collectionFolderName `inputRef` was overriding the `inputRef` on collectionName


helloanoop:
Merged.
Thanks for this @dcoomber !

Issue #106, opened 2023-02-25
Unhandled exception when entering a search term that does not match any loaded requests


dcoomber:
## Description

Entering a search term that does not match any requests results in an unhandled exception:

```
Error: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

### Steps to reproduce

1. Open Bruno
2. Load at least one collection
3. Selecting the search entry box, enter multiple characters that do not match the name of any loaded request

Exception occurs

### Proposed solution

Removing the following piece of code fixes the error.  I'll submit a PR, but I am uncertain of the impact.

```javascript
// packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js:58
  if (searchText && searchText.length) {
    if (!doesCollectionHaveItemsMatchingSearchText(collection, searchText)) {
      return null;
    }
  }
```

[bruno-app/src/components/Sidebar/Collections/Collection/index.js#L58](https://github.com/usebruno/bruno/blob/37b1c043eb21350f97f9dd797c73649c84e3f092/packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js#L58)

### Evidence

https://user-images.githubusercontent.com/47242934/221352245-407690e0-11cc-475a-9160-9f22cf6fe28b.mov

dcoomber:
Thank you for the detailed explanation, @helloanoop (and for correcting the bugfix).

Retested and confirmed fixed.  Closing this issue.

PR #107, opened 2023-02-25
Fixed unhandled exception on collection search Re #106

dcoomber:
Removed code related to no collections / requests matching the collection search term.

I tried a few different return values (`return false`, `return collection.item`) - TBH I don't understand the object model being used (yet)... so please feel free to decline / amend my commit.


helloanoop:
@dcoomber I have fixed this here.
https://github.com/usebruno/bruno/commit/b1edaba1c6c7af0f3b0a3c7c25b86b33a680c2e5

PR #108, opened 2023-02-26
filter shortcut for scalar properties

ajaishankar:
Filter shortcut for scalar properties
```js
get(data, '..items[?]', { id: 2, amount: 20 })
```
instead of
```js
get(data, '..items[?]', i => i.id === 2 && i.amount === 20)
```


helloanoop:
Looks good @ajaishankar !
Merged!

helloanoop:
@ajaishankar When you have some time, it'd be great if you could raise a PR to the docs here updating this API https://docs.usebruno.com/scripting/response-query.html

ajaishankar:
Thanks, will update docs...

PR #130, opened 2023-03-18
Proposed adjustment to terminology on requests

dcoomber:
I propose the following UI terminology changes relating to the new request dialog:

1. `Http` --> `REST`
2. `Graphql` --> `GraphQL`
3. `Form Url Encoded` --> `Form URL Encoded`


helloanoop:
Hi @dcoomber,
Agree with everything except changing `Http` -> `REST`
The reason being that people also use Bruno to perform SOAP requests using http call.

The primary reason for keeping graphql separately (although it also uses rest underneath) was to  switch to a user interface that is custom made for working with graphql api's

Can you change back the `REST` -> `Http`, the rest of the changes look good.

dcoomber:
Hi @helloanoop 

Thank you for the additional information - I forgot completely about SOAP 🤦 .

I'll switch that option back to HTTP.

helloanoop:
Merged!
Thanks @dcoomber !

Issue #128, opened 2023-03-18
CMD+W (Mac) closes the whole app, not just the current request tab


dcoomber:
In Postman, pressing CMD+W closes the current tab however in Bruno this key combination closes the whole application.

helloanoop:
Makes sense @dcoomber 
Cmd+W is kind of the used universally as a key to close the tab (incl chrome and vscode).

A good first issue for anyone in the community to pick up.

We would need to add a hotkey listener [here](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/providers/Hotkeys/index.js)
And call the `closeTabs` redux action with the current `activeTabUid`

helloanoop:
This functionality is now available in v0.11.0

PR #134, opened 2023-03-19
Proposed addition of CMD+W hotkey Re #128

dcoomber:
This commit adds a CMD+W hotkey (see #128 for details).

Please point me in the right direction if more error handling is required.

Caveat:  closing a tab with unsaved changes is consistent with the app behaviour when clicking the `x` to close the tab.

### DIFF
--- a/packages/bruno-app/src/providers/Hotkeys/index.js
+++ b/packages/bruno-app/src/providers/Hotkeys/index.js
@@ -10,6 +10,7 @@ import NewRequest from 'components/Sidebar/NewRequest';
 import BrunoSupport from 'components/BrunoSupport';
 import { sendRequest, saveRequest } from 'providers/ReduxStore/slices/collections/actions';
 import { findCollectionByUid, findItemInCollection } from 'utils/collections';
+import { closeTabs } from 'providers/ReduxStore/slices/tabs';
 
 export const HotkeysContext = React.createContext();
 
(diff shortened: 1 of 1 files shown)


helloanoop:
>  closing a tab with unsaved changes is consistent with the app behaviour when clicking the x to close the tab.
Sure, in the future we will improve this behaviour where we display a popup before closing a request with unsaved changes.

@dcoomber I just have one comment. Once you make the change, this should be good to merge.

dcoomber:
> @dcoomber I just have one comment. Once you make the change, this should be good to merge.

@helloanoop The code comment isn't reflecting as yet under commits.  Please point me in the right direction if the comment was added elsewhere.

packages/bruno-app/src/providers/Hotkeys/index.js
```diff
@@ -144,6 +145,23 @@ export const HotkeysProvider = (props) => {
...
+        })
+      );
+
+      return false; // this stops the event bubbling
+    });
+
+    return () => {
+      Mousetrap.unbind(['command+w', 'ctrl+w']);
+    };
+  }, [activeTabUid, tabs, collections, setShowNewRequestModal]);
```
helloanoop: You only need to pass the dependencies to `useEffect` that are being used inside it.
In this case in line 163, we only need to pass `[activeTabUid]`

helloanoop:
Nice! Merged!

Issue #132, opened 2023-03-18
isJson assertion returns the error "Invalid Chai property"


dcoomber:
## Description

isJson assertion returns the error `Invalid Chai property: json. Did you mean "to"?`

## Test collection

Test collection:  https://github.com/dcoomber/exploring-bruno/tree/dpc_session_1/exploration/02-create/EE%20Bookings

Test request:  "Retrieve a booking"

## Screenshot

<img width="1279" alt="Screenshot 2023-03-18 at 11 54 53" src="https://user-images.githubusercontent.com/47242934/226098679-ec4d70f7-596a-4b8d-be29-78e74766c55d.png">

helloanoop:
Thanks for catching this @dcoomber !

Would you be able to work on fixing this?
The code is [here](https://github.com/usebruno/bruno/blob/main/packages/bruno-js/src/runtime/assert-runtime.js#L227)

dcoomber:
I have ended up proposing the use of https://github.com/chaijs/chai-http, but am uncertain if this is what you're looking for. 
 `chai-http` `.to.be.json` is merely a check that the response headers includes `Content-Type: application/json`.  In this form, the assertion would only ever be valid when comparing against `res` and would not be usable for comparisons of subsets of the response structure.

Please let me know if you had a different solution in mind.

dcoomber:
After reviewing https://www.chaijs.com/api/plugins/ along with stepping through some of the other assertions, I realised that we're looking to `addProperty` and not `addMethod`.

PR updated.

PR #135, opened 2023-03-19
Add chai-http to enable to.be.json assertions Re #132

dcoomber:
Proposed fix for #132 uses `chai-http` to enable the `.to.be.json` assertion.

### DIFF
--- a/packages/bruno-electron/package.json
+++ b/packages/bruno-electron/package.json
@@ -18,6 +18,7 @@
     "@usebruno/schema": "0.3.1",
     "axios": "^0.26.0",
     "chai": "^4.3.7",
+    "chai-http": "^4.3.0",
     "chokidar": "^3.5.3",
     "dotenv": "^16.0.3",
     "electron-is-dev": "^2.0.0",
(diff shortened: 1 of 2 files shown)


packages/bruno-electron/package.json
```diff
@@ -18,6 +18,7 @@
     "@usebruno/schema": "0.3.1",
     "axios": "^0.26.0",
     "chai": "^4.3.7",
+    "chai-http": "^4.3.0",
```
helloanoop: @dcoomber This dep can be removed here.

helloanoop:
> To me it looks as if the custom plugin method json isn't being used.

@dcoomber To confirm this, you can put `console.log()` statements inside the chaijs `json` function. These logs will be displayed in the terminal where the electron app is running.

dcoomber:
Thank you for the encouragement, @helloanoop 

After reviewing https://www.chaijs.com/api/plugins/ along with stepping through some of the other assertions, I realised that we're looking to `addProperty` and not `addMethod`.

I've tested the updated assertion and its passing the positive and negative cases I've used.

image

helloanoop:
Awesome @dcoomber !
I am sure you had had the high of "damn it, i finally cracked this" 😃

Can you remove the `chai-http` in package.json ?
Once done, this PR is good to merge.

dcoomber:
@helloanoop I am thrilled that I was able to figure this out!  Thank you for your help.

I've removed the `chai-http` package 😅 .

helloanoop:
Nice work @dcoomber ! Merged !

Issue #170, opened 2023-08-11
RFC: Extend Bru script to allow for get of process.env values


dcoomber:
## Problem statement

I have several collections that rely on the use of authentication tokens that are specific to each developer and may not be shared or committed to version control.

## Proposed solution

Being able to load these tokens from the system environment (`process.env`) would support my needs for local usage as well as future usage on CI/CD pipeline.

### Example

```javascript
// Pre request script
bru.setVar("esp_token", bru.getProcessEnv("ESP_TOKEN"));
```

1. Environment variable `ESP_TOKEN` set in my local machine `.bashrc`(e.g. `export ESP_TOKEN=SUPER_SECRET_KEY`)
1. Request header `token` set to a collection variable (e.g. `token: {{esp_token}}` )
1. Pre-request script sets the value using a new Bru method `getProcessEnv` to retrieve the value from local machine ENV (e.g. `bru.setVar("esp_token", bru.getProcessEnv("ESP_TOKEN"));`)

dcoomber:
Thank you, @helloanoop 

Closing off the issue.

PR #171, opened 2023-08-11
RFC: Support process.env in Bru script Re #170

dcoomber:
Please also see proposed documentation change in https://github.com/usebruno/bruno-docs/pull/7


helloanoop (approved):

helloanoop:
Thanks @dcoomber !

Issue #168, opened 2023-08-08
Option to ignore "self signed certificate in certificate chain"


Nosceteipsum:
An option to ignore **"self signed certificate in certificate chain"**.
Postman got this option and it would be nice to have this option when you trust the endpoint.
![image](https://github.com/usebruno/bruno/assets/7213547/5baa794d-ff73-4891-9f3e-fc5a5746da88)

Error from Bruno:
_Error invoking remote method 'send-http-request': Error: self signed certificate in certificate chain_

ace90210:
literally hit this today, is there no way to disable this?

helloanoop:
@ace90210 @Nosceteipsum 
I am working on this. Expecting to release this update this week.

Here is a screenshot of the ongoing work.
![image](https://github.com/usebruno/bruno/assets/9350733/4e18998c-2786-4c8e-8547-cd2d0f95b5f2)

mschojoh:
any progress on this?

helloanoop:
@mschojoh 
I plan to release this on Windows on Linux tomorrow.
I am resolving notarisation issues for Mac Builds (this will take some time).

Which OS are you currently using Bruno on ?

mschojoh:
i use it on windows and really appreciate your work.

mschojoh:
when will you publish a new release with the option to disable ssl verification?

helloanoop:
@mschojoh @Nosceteipsum @ace90210 
This issue has been fixed in v0.13.0 and is available for Windows and Linux. Please go to [downloads](https://www.usebruno.com/downloads) to get the latest version

Note:
> The latest update, version 0.13.0, is now accessible for Windows and Linux users. Unfortunately, Mac builds are facing a delay due to notarization issues. As of now, the most recent Mac build available is version 0.12.2. The notarization challenges are expected to be resolved by September 15th, 2023.

mschojoh:
> @mschojoh @Nosceteipsum @ace90210 This issue has been fixed in v0.13.0 and is available for Windows and Linux. Please go to [downloads](https://www.usebruno.com/downloads) to get the latest version
> 
> Click on the gear icon on the bottom left to open the preferences window where you can disable SSL verification
> 
> Note:
> 
[...]

PR #172, opened 2023-08-16
add ability to turn certificate validation off

BrentShikoski:
Certificate validation can be turned off by adding ```certificateValidation: false``` as an environment variable or as a collection variable: ```bru.setVar("certificateValidation", "false");```.

### DIFF
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -45,6 +45,14 @@ const runSingleRequest = async function (filename, bruJson, collectionPath, coll
       request.data = qs.stringify(request.data);
     }
 
+    if((collectionVariables.certificateValidation && collectionVariables.certificateValidation == 'false') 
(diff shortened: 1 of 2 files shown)


helloanoop:
@BrentShikoski Thanks for taking time to work on this.

Currently in the latest builds for Windows and Linux support this feature (Disabling SSL verification) in preferences.
I have added a screenshot in https://github.com/usebruno/bruno/issues/168

This is available in Windows and Linux. Mac builds are facing notarization issues which is expected to be resolve by Sep 15th 2023

I feel we should use the preferences to control this behaviour at the app level instead of using the collection variables.

BrentShikoski:
OK. So how are you going to handle it in the cli, using a command line option or something?

helloanoop:
Yes, Curl uses `-k` or `--insecure`

I was thinking of supporting `--insecure` as a CLI arg. If you concur, it'd be great if you could work on a PR to get this implemented in CLI

PR #183, opened 2023-09-07
support alternative cacert file in cli

BrentShikoski:
Add ability in the cli to specify a file from which to read cacerts to be used to validate TLS connections.


helloanoop:
Merged.
Thanks @BrentShikoski !

`v0.7.1` has been published to npm.

PR #196, opened 2023-09-16
check response type for ResponsePane CodeEditor

mirkogolze:
Hi,

currently XML was presented in the response Pane surrounded by quotation marks and so the CodeEditor could not highlight the XML.
I added a check for the response type to check if the result is a JSON. Only then Stringify the JSON.
Before the `JSON.stringify` ran on XML too and added the quotations marks.

second change:
set the request header `content-type` only if it is not set by the user in the header variables.

BR
Mirko


helloanoop:
Merged!

Many Thanks @mirkogolze !

Issue #199, opened 2023-09-18
Environment Secrets


helloanoop:
**Objective**
Enable developers to use secrets in environment variables without storing them in `{env}.bru` files
This is so that the secrets don't get committed to git.

Implementation is based on some ideas put forth by @brainomite and @fcr--

For earlier discussions - please check #122 and #109
Creating a new issue to track a specific solution approach

**Part 1**
The DSL (Bru Lang)

Choosing the below approach over saving the var as `token: $secret$`
I feel it better for the DSL to explicitly specify secrets as a separate block for the below reasons
* Accidental commits of secrets become easily visible
* A pre commit check (linter) can be implemented to ensure all secret vars are null before commiting
```groovy
vars {
  url: http://localhost:3000
  ~status: active
}

vars:secret {
  token: null
}
```

**Part 2**
UI Updates

We add another column called secret which has a checkbox to indicate whether a var is a secret or not.

**Part 3**
Interceptor to store the secret in a file (at the electron layer). This component will replace the secret vars as null before writing them to `{env}.bru` files and hydrate them back using the cached secret values while loading them in UI.

Store the secrets inside a file `preferences.json` (which we already use) via `electron-store` with secrets encrypted via [safe-storage](https://www.electronjs.org/docs/latest/api/safe-storage)
```json
{
  "secrets": { 
    "collections": [{
      "path": "/Users/anoop/Code/acme-acpi-collection",
      "environments" : [{
        "name": "Local",
        "secrets": [{
          "name": "token",
          "value": "abracadabra"  
        }]
      }]
    }]
  }
}
```

Tasks
- [x] Bru Lang Updates
- [x] UI Updates
- [x] Interceptor at Electron Layer to nullify (on write) and hydrate(on read) secrets 

All changes are tracked in `feature/env-secrets` branch. I have already completed implementation for the Bru Lang updates

@brainomite I saw that you already working on Tasks 2 and 3 Let me know if you'd like to make any changes on the implementation spec. I wrote this since I saw you already working on this feature so that we have a consensus on the implementation approach.

szarrougtw:
Thanks for making us a branch and for doing the changes for the bru lang updates!

My original thought was that we don't put secrets in the env files at all. So with you example above, the env file would be 
```groovy
vars {
  url: http://localhost:3000
  ~status: active
}
```

and the electon-store would have
```json
{
  "secrets": { 
    "collections": [{
      "path": "/Users/anoop/Code/acme-acpi-collection",
[...]

PR #202, opened 2023-09-21
#199 add CLI feature to use command line parameter '--env-var' secret=…

mirkogolze:
#199 add CLI feature to use command line parameter '--env-var secret=xzy123'

### DIFF
--- a/packages/bruno-cli/src/commands/run.js
+++ b/packages/bruno-cli/src/commands/run.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const chalk = require('chalk');
 const path = require('path');
-const { exists, isFile, isDirectory, getSubDirectories } = require('../utils/filesystem');
+const { exists, isFile, isDirectory } = require('../utils/filesystem');
 const { runSingleRequest } = require('../runner/run-single-request');
 const { bruToEnvJson, getEnvVars } = require('../utils/bru');
 const { rpad } = require('../utils/common');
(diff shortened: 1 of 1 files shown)


packages/bruno-cli/src/commands/run.js
```diff
@@ -175,6 +182,31 @@ const handler = async function (argv) {
...
+        processVars = [envVar];
+      } else if (typeof envVar === 'object' && Array.isArray(envVar)) {
+        processVars = envVar;
+      } else {
+        console.error(chalk.red(`overridable environment variables not parsable: use name=value`));
+        return;
+      }
+      if (processVars && Array.isArray(processVars)) {
+        processVars.forEach((value) => {
+          let parts = value.split('=');
```
helloanoop, line 197: @mirkogolze This way of doing it will fail if the value has a `=` sign which is not uncommon in access tokens

Can you update this code to handle the case.
Something like
```javascript
processVars.forEach((value) => {
  // split the string at the first equals sign
  const match = value.match(/^([^=]+)=(.*)$/);
  if (!match) {
    console.error(
      chalk.red(`Overridable environment variable not correct: use name=value - presented: `) +
        chalk.dim(`${value}`)
    );
    return;
  }
  const name = match[1];
  const val = match[2];
  envVars[name] = val;
});
```

helloanoop:
@mirkogolze I have one comment on the PR.  Everything else looks good.

helloanoop:
@mirkogolze The `feature/env-secrets` branch might take time to get it completed and merged.

Can you raise a new PR to the `main` branch with the fixes per my comment ?
Thanks.

PR #207, opened 2023-09-24
#199 improve code to check given env vars correctly

mirkogolze:
#199 improve code to check given env vars correctly


helloanoop:
Merged!

Thanks for handling the edge case @mirkogolze !

PR #215, opened 2023-09-26
Feature/xml format

Cibico99:
Added XML formatting and moved getContentType() to Util Common for reuse.


helloanoop:
Nice! Merged.

Thanks @Cibico99 !!

PR #218, opened 2023-09-26
Adds fallback when no environments are defined

tpyle:
This adds a fallback for when a user has no environment defined. Otherwise, there's a bug where the `environments.length` can't be computed and causes the screen to blank.


helloanoop:
Merged!

Thanks @tpyle !

This has been released in v0.15.3 🎉

Issue #191, opened 2023-09-14
Support for Collection or environment vars in Header Names as we support in Header Values like {{varName}}


yuvanakannan:

helloanoop:
@yuvanakannan 

This functionality is now available in v0.16.0 🎉

Thanks @mirkogolze for getting this completed.

yuvanakannan:
Thanks @helloanoop @mirkogolze 

For some reason, the env vars used inside the body is no longer getting replaced by the values. Example "{{env}}" within the body goes as "{{env}}" instead of "DEV" or "QA".

mirkogolze:
For me it is working with release 0.16.1
Its working with collection variables. In Header, Names and Values and in Body
Its working with environment variables on all places. Too with variables from .env-File

helloanoop:
@yuvanakannan I got the issue, please see https://github.com/usebruno/bruno/issues/233
This is being fixed.

PR #225, opened 2023-09-27
#191 interpolate header names with variables

mirkogolze:


helloanoop:
Awesome! Merged.

Thanks @mirkogolze !!

PR #241, opened 2023-09-29
Handle failed requests and reduce duplication

tpyle:
This PR addresses some issues I ran into when requests failed to run. Basically, if a request produced an error (not due to status code), then it would show not show up as failed and not contribute to the status of the run. It also was not reflected in show results.

This modification makes it so that request information is tracked, and so that this reflects in exit code. The JSON report (the `--output` flag) also shows information for the response. In this case, it shows `null` for all fields in `response` and adds an `error` field.

I also cleaned up `runSingleRequest` to reduce the amount of duplicated logic around handling non-OK status codes.

An example of new CLI output:

image

### DIFF
--- a/packages/bruno-cli/examples/report.json
+++ b/packages/bruno-cli/examples/report.json
@@ -1,8 +1,11 @@
 {
   "summary": {
+    "totalRequests": 10,
+    "passedRequests": 10,
+    "failedRequests": 0,
     "totalAssertions": 4,
-    "passedAssertions": 4,
-    "failedAssertions": 0,
+    "passedAssertions": 0,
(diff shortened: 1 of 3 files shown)


helloanoop:
Got tied up with other PR's. Will be taking a look and getting this merged tomorrow.

helloanoop:
Will take a look at this tomorrow @tpyle

tpyle:
> Will take a look at this tomorrow @tpyle

No problem, take your time! I know there's been a lot of PRs lately so no worries.

helloanoop:
Merged!

Many thanks @tpyle !!

Issue #219, opened 2023-09-27
Can't import a POSTMAN collection with a / char in the name


l0new4nder3r:
When attempting to import a postman collection with a name containing "/", the initial detection seems correct ("name" is correctly displayed), the import seems to be successful... 
"Collection imported successfully" is being displayed, but the collection is not imported at all (nothing in the destination folder, and no new collection opened in the main interface).

When looking at the dev console (postman collection called "MYAPP\SOMETHING ELSE", by example)

> Uncaught (in promise) Error: Error invoking remote method 'renderer:import-collection': Error: ENOENT: no such file or directory, mkdir 'C:\Users\MY_USER\Documents\Bruno collections\MYAPP\SOMETHING ELSE'

Removing or replacing the "/" char leads to a successful import

helloanoop:
@l0new4nder3r The fix done by @tpyle has been released in `v0.16.3`

Can you test and confirm this on your end ?

l0new4nder3r:
Hi, thanks @tpyle !

Successfully tested a few minutes ago with v0.16.4

<img width="288" alt="Capture d’écran 2023-09-30 à 21 43 38" src="https://github.com/usebruno/bruno/assets/27144372/761fa831-bad4-48ed-92d5-db6289e9beb1">

Testing with a postman collection containing a slash in its name property.


<img width="365" alt="Capture d’écran 2023-09-30 à 21 42 10" src="https://github.com/usebruno/bruno/assets/27144372/40c2425d-0f9b-4a0b-a06b-c33d0e3c450c">

Importing process (same as before, slash is still present)


<img width="502" alt="Capture d’écran 2023-09-30 à 21 42 40" src="https://github.com/usebruno/bruno/assets/27144372/0177e8d9-8ec2-4165-b202-8ad1b3dcb064">

The collection is imported without issue - the only thing is that now, its name has changed. 
The slash has become an hyphen. 
Consistent with what I see in the commit, and at least it does not break!

_Improvement with a much lesser priority : allow any character in the collection name?_

tpyle:
Hi @l0new4nder3r,

Unfortunately, since bruno stores collections (and requests) as files and directories, that requirement comes from the filesystem, not from bruno itself. Every filesystem that I'm aware of doesn't allow `/` in directory names or folder names, as it's used in UNIX-like OS's as a delimiter between directories.

[...]

PR #242, opened 2023-09-29
Corrects issue when collection names in imports have slashes

tpyle:
Corrects #219


helloanoop:
Merged. Thanks @tpyle !

PR #252, opened 2023-09-29
Changed Proxied Requests to Use https-proxy-agent.

acostinescu:
HTTP proxies typically make use of TCP tunnels to send proxied HTTPS requests. However, Axios makes use of https.Agent by default which does not support TCP tunneling.

This PR makes use of the https-proxy-agent package which replaces https.Agent for proxied requests. Non-proxied requests still make use of https.Agent.

Note that this can be tested on any machine by spinning up a simple HTTP proxy with the following docker-compose:

```
services:
  proxy:
    image: ubuntu/squid
    ports:
      - "3128:3128"
```

Then, direct Bruno to use an http proxy with the hostname localhost and port 3128.

Before my changes, directing Bruno to use the proxy at localhost:3128 will result in a TLS error from Squid. After my changes, a TCP tunnel is established and HTTPS requests are able to be made.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,5 @@
   },
   "overrides": {
     "rollup": "3.2.5"
-  },
-  "dependencies": {}
+  }
 }
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -4,12 +4,12 @@ const fs = require('fs');
 const { forOwn, each, extend, get } = require('lodash');
 const FormData = require('form-data');
 const axios = require('axios');
-const https = require('https');
 const prepareRequest = require('./prepare-request');
 const interpolateVars = require('./interpolate-vars');
 const { ScriptRuntime, TestRuntime, VarsRuntime, AssertRuntime } = require('@usebruno/js');
 const { stripExtension } = require('../utils/filesystem');
 const { getOptions } = require('../utils/bru');
+const { HttpsProxyAgent } = require('https-proxy-agent');
 
 const runSingleRequest = async function (
   filename,
@@ -65,31 +65,6 @@ const runSingleRequest = async function (
       );
     }
 
-    // set proxy if enabled
-    const proxyEnabled = get(brunoConfig, 'proxy.enabled', false);
-    if (proxyEnabled) {
-      const proxyProtocol = get(brunoConfig, 'proxy.protocol');
-      const proxyHostname = get(brunoConfig, 'proxy.hostname');
-      const proxyPort = get(brunoConfig, 'proxy.port');
-      const proxyAuthEnabled = get(brunoConfig, 'proxy.auth.enabled', false);
-
-      const proxyConfig = {
-        protocol: proxyProtocol,
-        hostname: proxyHostname,
-        port: proxyPort
-      };
-      if (proxyAuthEnabled) {
-        const proxyAuthUsername = get(brunoConfig, 'proxy.auth.username');
-        const proxyAuthPassword = get(brunoConfig, 'proxy.auth.password');
-        proxyConfig.auth = {
-          username: proxyAuthUsername,
-          password: proxyAuthPassword
-        };
-      }
-
-      request.proxy = proxyConfig;
-    }
-
     // interpolate variables inside request
     interpolateVars(request, envVariables, collectionVariables, processEnvVars);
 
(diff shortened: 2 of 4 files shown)


mirkogolze:
Hi @acostinescu, @helloanoop,

this is an interesting fact. Is there a way to handle this a bit more global?

We are using `axios` in pre-request-scripts. So my idea was to only set the proxy configuration to environment variables `http_proxy`, `http_proxy` and `no_proxy`. So `axios` will handle all the work with proxies with help of https://www.npmjs.com/package/proxy-from-env out of the environment variables.
And of course the NodeVM sandbox must get the information of the proxy configuration by the env variables too.

helloanoop:
> this is an interesting fact. Is there a way to handle this a bit more global?

@mirkogolze If I am understanding correctly, what you are saying is that despite setting the `http_proxy`, `https_proxy` and `no_proxy` environment variables, `axios` is still not using it right ?

This is connected to another  behaviour I observed. the `process.env` which contains system vars are available inside electron while running dev locally
image

But the same thing does not work in the released build
image

I am assuming if we fix this `process.env` issue in release build, you should be able to set the proxy globally.

Additionally, we should add a UI at the level preferences in GUI to configure proxy that will apply to all collections

helloanoop:
@mirkogolze 

I can confirm that `axios` respects the `http_proxy` variable in the CLI.
I can confirm that `axios` respects the `http_proxy` variable in the GUI while developing locally

I can confirm that all the env variables are available when you open the installed app from command line.
```bash
# on mac
open /Applications/Bruno.app
```
image

This should solve your use case at the moment right?

mirkogolze:
I think `axios` is handling the environment variables correct. But they are not available in all scripts running inside the a sandbox. 
In VM2 (https://www.npmjs.com/package/vm2) you have the option `env - Object to be passed to process.env.` to pass environment variables. There you must pass the proxy variables. Perhaps other env variables are interesting inside the sandbox.

helloanoop:
@mirkogolze 

I just tried added this in the pre-request script
```javascript
const axios = require('axios');

const response = await axios.get('http://localhost:6000/ping?pre-req-script');

console.log(response.data);
```

And tested both CLI, Local Dev GUI, Installed GUI and all were able to use the proxy inside scripting
Can you check once again?

My setup
1. Simple express server that exposes `http://localhost:6000/ping` api
```javascript
const express = require('express');

const app = new express();

app.get("/ping", function(req, res) {
  console.log("Received /ping request");
  return res.send("pong");
});

app.listen(6000, function() {
  console.log(`Server started on port: 6000`);
});
```
2. Proxy server running at port 4000
```javascript
const http = require('http');
const httpProxy = require('http-proxy');

const proxy = httpProxy.createProxyServer({});

const server = http.createServer((req, res) => {
  // Define the target server you want to proxy to
  const target = 'http://localhost:6000';

  console.log(`Proxying request for: ${req.url}`);

  // Proxy the incoming request to the target server
  proxy.web(req, res, { target });

  // Handle errors
  proxy.on('error', (err) => {
    console.error('Proxy Error:', err);
    res.statusCode = 500;
    res.end('Proxy Error');
  });
});

// Listen on a specific port
const port = 4000;
server.listen(port, () => {
  console.log(`Proxy server listening on port ${port}`);
});
```
4. My `.zshrc` file has the `HTTP_PROXY` defined
```bash
export HTTP_PROXY=http://localhost:4000
```

And I am seeing the proxy server receiving logs and forwarding them
image

mirkogolze:
Hi @helloanoop,

yes it is working. I'am a bit confused.
I added following line to show the env variables.

```js
const axios = require('axios');

const response = await axios.get('http://localhost:6000/ping?pre-req-script');

console.log(response.data);
console.log(JSON.stringify(process));


{"argv":[],"title":"Windows PowerShell","version":"v16.16.0","versions":{"node":"16.16.0","v8":"10.6.194.17-electron.0","uv":"1.43.0","zlib":"1.2.12.1-motley","brotli":"1.0.9","ares":"1.18.1","modules":"109","nghttp2":"1.47.0","napi":"8","llhttp":"6.0.7","openssl":"1.1.1","cldr":"41.0","icu":"71.1","tz":"2022b","unicode":"14.0","electron":"21.1.1","chrome":"106.0.5249.103"},"arch":"x64","platform":"win32","env":{},"pid":37292,"features":{"inspector":true,"debug":false,"uv":true,"ipv6":true,"tls_alpn":true,"tls_sni":true,"tls_ocsp":true,"tls":true,"cached_builtins":false}}
```

Because of the empty `env` variables I thought, the env variables for the proxy are not active in the sandbox.

When this is working as expected, the the implementation for the proxy should only be setting the three env variables (`http_proxy`, `https_proxy` and `no_proxy`) and all should work out of the box.
Only when anybody needs to use its own `ca` files then the additional `https-proxy-agent` is necessary. And this too for pre-request-requests using axios.  But this has to be done by everybody using it.

helloanoop:
@mirkogolze 

> I'am a bit confused. I thought, the env variables for the proxy are not active in the sandbox.

I don't think we are injecting the process variable itself in runtime. That's why env is `{}`
To get a process env you should be able to use
```javascript
cont httpProxy = bru.getProcessEnv("http_proxy");
console.log(httpProxy);
```

> When this is working as expected, the the implementation for the proxy should only be setting the three env variables (http_proxy, https_proxy and no_proxy) and all should work out of the box.

Yes, this currently is working out of the box. Let me know if you are still seeing issues.

> Only when anybody needs to use its own ca files then the additional https-proxy-agent is necessary. And this too for pre-request-requests using axios.

If my understanding is correct - the fix by @acostinescu uses`https-procy-agent` to solve TCP tunneling.problem, which is not related to ca files. The problem is that if the proxy server uses TCP tunneling, then we get an error since node `https.Agent` doesn't support it.

acostinescu:
As @helloanoop said, the https-proxy-agent was only added to internal requests made by Bruno to support TCP tunneling. This agent handles certificates the same way that https.Agent does and also does not change how environment variables are handled.

I think there's certainly a discussion to be had about globally configuring Axios so that it also uses https-proxy-agent in pre/post request scripts if a proxy is configured for the collection, but I think that discussion, as well as the discussion about environment variables, is outside of the scope of this PR.

helloanoop:
@acostinescu 

I had to revert the PR.

I was able to reproduce the issue you mentioned in the current build and it was throwing an error.
But I observed an issue after merging the PR where the call would succeed even if you give wrong proxy url

It appears like the proxy configuration was never being used at all.
Can you check it again when you have some time ?

acostinescu:
Hi @helloanoop,

It seems the https-proxy-agent can connect to both HTTP and HTTPS proxies but is only intended for sending HTTPS requests, not HTTP. In the case of HTTP requests I was able to replicate that it does not use the proxy settings. It may be falling back to https.Agent in this case, but I'm not sure.

I'll see if I can come up with a better approach which handles all cases.

acostinescu:
Hi @helloanoop,

I realized that I had replaced Axios' default httpsAgent with HttpsProxyAgent, but of course this only handles HTTPS requests. Since I removed the proxy settings from the request itself and moved them into the Agent, this meant that HTTP requests were, of course, unaware of the proxy.

I've created a new PR #420 which resolves this by also replacing the httpAgent with HttpProxyAgent. This way both HTTP and HTTPS requests will be proxied.

Cheers, Alex

PR #255, opened 2023-09-29
Bugfix: Create collection modal validation

Its-treason:
Currently:

https://github.com/usebruno/bruno/assets/39559178/9e9338f9-95b6-462c-bbee-fd705b29d739

With my fix:

https://github.com/usebruno/bruno/assets/39559178/49f87574-7d89-442b-8744-c5a14ffe34cc

---
Also, as an Idea: I think the name tooltip could be removed as it doesn't add any information and the Folder Name tooltip could be expanded to clarify that the folder will be created inside the selected location. To avoid mistakes like: `~/projects/my-api/docs/docs`.

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
+++ b/packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
@@ -28,7 +28,9 @@ const CreateCollection = ({ onClose }) => {
         .min(1, 'must be atleast 1 characters')
         .max(50, 'must be 50 characters or less')
         .required('folder name is required'),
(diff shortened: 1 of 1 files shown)


helloanoop:
Nice. Merged !

Many thanks @Its-treason !

> I think the name tooltip could be removed as it doesn't add any information and the Folder Name tooltip could be expanded to clarify that the folder will be created inside the selected location. To avoid mistakes like: ~/projects/my-api/docs/docs (Wanted to save the collection inside ~/projects/my-api/docs).

I agree with you, please feel free to raise a PR with this change.

Issue #113, opened 2023-02-28
tooltip on mouse hover over environment variable value does not work in case of more than 1 variable


sanbhattdeep:
when constructing an URL for an API request using environment variables, hovering the mouse over the variable shows the variable value

This works only if the variable expression contains only 1 variable for e.g. https://{{base_url}}

if the variable expression contains more than 1 variable it only shows the last variable value and not for both for e.g. 
hovering over base_url or resource under https://{{base_url}}{{resource}} shows the resource variable only

barathrajramachandran:
@helloanoop I am picking up this issue.

helloanoop:
This has been fixed in `v0.16.5` Thanks to @brahma-dev !

PR #256, opened 2023-09-29
Ensure that adjacent variables are rendered in separate spans.

brahma-dev:
Currently adjacent variables like `{{var1}}{{var2}}` are rendered by CM in one `<span>` because they had the same classname; and likewise considered as one token in apis like `getTokenAt`

Fixes: https://github.com/usebruno/bruno/issues/113


helloanoop:
Great!

Codemirror issues are always difficult to debug.
Thanks for taking care of this @brahma-dev !

Issue #195, opened 2023-09-14
Feature request / question: file upload helpers


Slijkhuis:
What is the current recommended approach for file upload?

Let's say there's an endpoint `POST /photos` that allows you to upload a file using a `multipart-form` request. I think it would be amazing if you could do something like:

```
meta {
  name: Upload photo
  type: http
  seq: 1
}

post {
  url: {{baseUrl}}/photos
  body: multipartForm
}

script:pre-request {
  bru.setVar("image", readCollectionFile("example.jpg"));
}

body:multipart-form {
  title: my photo
  file: {{image}}
}
```

Where `readCollectionFile` (just an example name) would be a built-in function to read a file relative to root folder of the collection. It could simply wrap `fs.readFileSync`. 

Even better would be to be able to reference a file directly in the `body:multipart-form` block without requiring a script tag and setting a variable. But I'm not sure how that would look like since there can only be dictionary or text blocks right? Maybe the variable syntax could be used for a function call:

```
body:multipart-form {
  title: my photo
  file: {{readCollectionFile("example.jpg")}}
}
```

The problem with setting a variable is that the UI currently shows the variables as text when you hover over the eye-icon in the right-top corner. Having image data in variables doesn't really render nicely, you won't be able to see the full variable list.

The problem with allowing function calls within the variable syntax is that when a variable is set from the response of a request, it's a security issue if then the next request might interpret the value as a function call instead of just text. I currently (= after reading the docs, I haven't read the source yet) don't understand how Bruno distinguishes between the literal string `Greeting: {{hello}}` (expected result: `Greeting: {{hello}}`) and the variable substitution `Greeting: {{hello}}` (expected result: `Greeting: hi`). (where variable `hello` is equal to `hi`). Is there any documentation about that? Maybe it could help me come up with a good syntax to reference files.

Is there something for files already that I missed in the docs? If not I will try to find time to create a PR, but then I'd like some consensus on the preferred approach beforehand.

dot-bm:
Yes, it will be great feature

helloanoop:
> What is the current recommended approach for file upload?

Currently file upload support is not present. It'd be great to get a PR for this.

>  I haven't read the source yet) don't understand how Bruno distinguishes between the literal string Greeting: {{hello}} (expected result: Greeting: {{hello}}) and the variable substitution Greeting: {{hello}} (expected result: Greeting: hi). (where variable hello is equal to hi)

[...]

PR #258, opened 2023-09-30
Add initial multipart form file upload support

mkdillard:
This PR attempts to implement Multipart Form Data file uploads as discussed in #195

The code should handle file paths relative to the root of the current collection as well as absolute paths.

single file:
```
body:multipart-form {
  action: print
  file: @file(test.txt)
}
```

multiple files should be presented in a comma delimited list.
```
body:multipart-form {
  files: @files(test.txt,C:\Users\User\Documents\temp.txt)
  action: fax
}
```

Things still missing or potentially missing:

- I didn't see any test files explicitly for the surrounding code, I wasn't sure exactly where I should add test cases for this if any.
- I didn't add any this information to the docs, I wasn't sure where the updates should go, I saw several docs folders, but none of them looked like the docs on the site
- I didn't handle the UI updates to look nice, the value in the UI looks the same as it does in the .bru files
image

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -3,6 +3,8 @@ const https = require('https');
 const axios = require('axios');
 const Mustache = require('mustache');
 const FormData = require('form-data');
+const fs = require('fs');
+const path = require('path');
 const { ipcMain } = require('electron');
 const { forOwn, extend, each, get } = require('lodash');
 const { VarsRuntime, AssertRuntime, ScriptRuntime, TestRuntime } = require('@usebruno/js');
@@ -74,6 +76,29 @@ const getSize = (data) => {
   return 0;
 };
 
+const parseFormData = (data, rootCollectionPath) => {
+  const form = new FormData();
+  forOwn(data, (value, key) => {
+    if (value.toString().indexOf('@file') === 0) {
+      const stringValue = value.toString();
+      const filePaths = stringValue.substring(stringValue.indexOf('(') + 1, stringValue.indexOf(')')).split(',');
+      filePaths.forEach((filePath, index) => {
+        let trimmedFilePath = filePath.trim();
+        if (!path.isAbsolute(trimmedFilePath)) {
+          trimmedFilePath = path.join(rootCollectionPath, trimmedFilePath);
+        }
+        const file = fs.readFileSync(trimmedFilePath);
+        console.log(key);
+        console.log(file);
+        form.append(key, file, path.basename(trimmedFilePath));
+      });
+    } else {
+      form.append(key, value);
+    }
+  });
+  return form;
+};
+
 const registerNetworkIpc = (mainWindow) => {
   // handler for sending http request
   ipcMain.handle(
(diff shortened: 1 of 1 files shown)


jaktestowac:
Any updates on that?

kingzcheung:
Why hasn't such a good idea been merged？

helloanoop:
This will be a priority to get merged this week.

Current workaround is to use scripting: https://github.com/usebruno/bruno/discussions/385#discussioncomment-7196674

kym6464 (changes requested):
I can think of a few counterexamples that would cause the file path parsing to fail:

1. `some file (latest).mp4` — Parentheses in filenames are fairly common
2. `hello, world.txt` — Commas are totally valid as well 🤷‍♂️

I was able to fix 1 but 2 would require file paths to be surrounded by quotes (ie CSV rules) which, based on some discussions I've read, seems like this project is not fond of since it harms readability. So I didn't wanna adjust the syntax until we've settled on an approach. I would suggest this:

```
body:multipart-form {
  action: print
  file: @file(foo.txt)
  file: @file(bar.txt)
}
```

which avoids the need for CSV rules and parsing, and keeps it nice and readable. However, I'm not sure if the DSL supports this kind of syntax. @helloanoop thoughts?

@mkdillard RE "I wasn't sure exactly where I should add test cases for this if any" I think this is a good piece of logic to test:

```javascript
const assert = require("node:assert");

/**
 *
 * @param {string} value
 * @returns {?string[]}
 */
function parsePaths(value) {
  if (!(value && typeof value === "string")) {
    return;
  }

  if (value.slice(0, 6) === "@file(") {
    value = value.slice(6);
  } else if (value.slice(0, 7) === "@files(") {
    value = value.slice(7);
  } else {
    return;
  }

  if (value.at(-1) !== ")") {
    return;
  }

  const pathCsv = value.slice(0, -1).split(",");

  // TODO resolve to absolute paths, verify exists, etc

  return pathCsv;
}

const inputs = [
  "action: print",
  "@file(test.txt)",
  "@files(test.txt,C:\\Users\\User\\Documents\\temp.txt)",
  "@file(some file (latest).mp4)",
  "action: fax",
];

const expectations = [
  undefined,
  ["test.txt"],
  ["test.txt", "C:\\Users\\User\\Documents\\temp.txt"],
  ["some file (latest).mp4"],
  undefined,
];

inputs.forEach((input, i) => {
  const expected = expectations[i];
  const actual = parsePaths(input);
  assert.deepStrictEqual(actual, expected);
});
```

this-rafael:
Good morning, does anyone have any updates on this topic? Apparently, this valuable feature is already prepared, just waiting to be released. :cry:

gseroul:
I just discovered Bruno and did everything to import all my postman collection to Bruno. It is workijng well, fantastic software !  But I really miss the file upload, actually I have to keep Postman just for this. Script is not an option for us.
This file upload option is a must have to kick Postman out and free all of us ;-)

Poltuu:
don't make me create a postman account over this 🙏  😭 🙏

helloanoop:
Thank you for work on this @mkdillard ! 
Your work was an inspiration for the PR #1130 which has been merged.

File Uploads are available in Bruno `v1.8.0` 🥳

PR #259, opened 2023-09-30
fix(graphql-docs/build): rollup error thrown during build

sadn1ck:
- during the dts transformation, the css import was not recognized, hence marking it as external in the dts transform didn't throw the error

- "extract: true" in postcss plugin makes sure the CSS file gets extracted to the final bundle as well

image

---

I think the pre-commit caused stylistic changes 🤔 lmk if anything needs to be changed or added


helloanoop:
Nice! Merged!

Thanks @sadn1ck !

Issue #254, opened 2023-09-29
About Bruno Dialog is missing "App icon" image


Shoozza:
How to reproduce:
1. open bruno
2. in the menu click on help -> about bruno

Result:
* dialog opens but is missing app icon (see image below)

Expected:
* icon is visible

System:
* OS: Windows 8.1

![image](https://github.com/usebruno/bruno/assets/276143/e9d4546f-e667-42b7-b24f-7218eb909deb)

helloanoop:
This has been fixed in `v0.16.5`. Thanks to @mirkogolze !

![image](https://github.com/usebruno/bruno/assets/9350733/cc1e98a7-64c4-4de0-892d-61cacd47c31a)

Issue #231, opened 2023-09-28
[SUGGESTION] About Bruno dialog should allow copy version info


underhood:
When reporting a bug I noticed bruno about dialog is not allowing me to copy the bruno version info text. This forced me to type the versions manually. The dialog should allow to mark and copy the text or have a "copy version data to clipboard" button.

![Screenshot from 2023-09-28 12-55-51](https://github.com/usebruno/bruno/assets/6674623/c3a57bba-28f9-4aba-838e-e6211677e901)

#### OS
Fedora Workstation 38 AMD64
#### Bruno

appimage downloaded from usebruno.com
```
Bruno v0.16.1
electron: 21.1.1
chrome: 106.0.5249.103
node: 16.16.0
v8: 10.6.194.17-electron.0
```

helloanoop:
The icon has been fixed in `vo.16.5` thanks to @mirkogolze , and the text below the version number is copyable.
Ideally we also need to copy the version number.

Currently when you hover over the version number it becomes a pointer - upon click takes you to the website.
I think we don't need to redirect a user to website upon click, we need to configure it so that its just text and can be selected

![image](https://github.com/usebruno/bruno/assets/9350733/c4cc2a71-daf4-4656-abf6-8615b7c896fe)

mirkogolze:
I am able to select all version information.

I made the title and version information selectable. Only the description will not be selected.
With the correct mouse position you can select all from top to down and from down to top works too.
![image](https://github.com/usebruno/bruno/assets/66121941/e15e04a4-47b0-42de-9551-dcc109c5dc5c)

underhood:
Confirm icon is fixes in latest release. I can also select as per screenshot above. When I do CTRL+C with everything selected and paste here I get:
```
[...]

Issue #203, opened 2023-09-23
User should have access to About menu to find out what version they are currently using


jeffprinty:
Many apps will provide an About page in the help menu to allow users to know what version of the app is currently being used. Bruno should also have this too.

Example: As a Mac user I want to see if I have the latest Mac version of the app to know if the collections scrolling should be working. Currently there is no way to know what version of Bruno is running

Proposal: Use [electron-about-window](https://github.com/rhysd/electron-about-window) package to implement a basic About window in the Help menu

helloanoop:
This is cool !!

PR Merged. Thanks @jeffprinty !

helloanoop:
I have released this changes.

This works in local
![image](https://github.com/usebruno/bruno/assets/9350733/feae2191-2641-4482-a458-27b459bc60a8)

But is broken in the released build. Need to investigate.
![image](https://github.com/usebruno/bruno/assets/9350733/19653b39-ec70-4c85-9cfe-fc3531acbdf8)

mirkogolze:
you could try
```
icon_path: join(process.cwd(), '/resources/icons/png/256x256.png'), 
```

alefteris:
An issue from another user has also been opened for the broken app image in the about window #254. Since the about window is already implemented and released, I think the current issue can be resolved and track the icon issue in the new one?

Also a feature request for the about window has been opened at #231, requesting for the version text to be able to be copied.

helloanoop:
This is now working in latest release `v0.16.5`
![image](https://github.com/usebruno/bruno/assets/9350733/169cf7be-7cde-4aac-9ac6-a639f87671b9)

Thank you @mirkogolze and @jeffprinty !

PR #262, opened 2023-09-30
#203 #254 #231 add icon to asar content, change styling to allow copying the content

mirkogolze:
add icon to asar content, change styling to allow copying the content

The `resources` folder in bruno-electron is not packaged to the asar archive. So the image could not be loaded.


helloanoop:
Nice!

Thanks for taking care of this @mirkogolze !

Issue #257, opened 2023-09-29
Insomnia imports with duplicate request names will select only request to import


CrystalSpore:
In my Insomnia collection I had 1 request that was a POST, & then a second request that was a GET.

They had the same name due to one being for the deprecated service that we were moving away from, but some production systems still used it. They did have different endpoints however ( POST root.url/v1/endpoint vs GET root.url/v2/endpoint )

Since Bruno uses separate files per request, it makes sense that there is a collision, but having the import automatically append a number to the end for the number of duplicates would help make sure that everything gets imported without having this collision issue. We could then rename the request ourselves after it is imported

helloanoop:
@CrystalSpore This has been fixed by @Its-treason in `v0.16.5`
Can you check and confirm if this has resolved the issue?

CrystalSpore:
Will check this tomorrow morning! Thanks for the speedy fix!

CrystalSpore:
Works perfectly! Insomnia collection I exported had 64 endpoints/requests, & the import to Bruno had 64 endpoints/requests.

PR #264, opened 2023-10-01
fix(#257): Name collision during Insomnia collection import

Its-treason:
To avoid name collisions, my fix adds number suffixes to duplicate names.

grafik


helloanoop:
This is nice! Merged!

Thank you @Its-treason !

Issue #220, opened 2023-09-27
Can't change the order of the opened collections - new feature: make it possible to order (auto or manual) the collections


l0new4nder3r:
When importing or creating collections, their order in the left pane list remains the chronological order of their creations.

There is no way to sort them (alphabetical order, reversed / import date / reversed), or move them freely up or down in the list.

I (personally!) find this to be a bit annoying because I have quite a bunch of collections, named from microservices or apps I want to call, and I prefer having the list in alphabetical order to quickly find the right app name.

Task List
- [x] Sort Alphabetically upon clicking sort button
- [x] Reverse sort on clicking icon again
- [x] Go back to original import order when clicked again
- [x] Hide sort icon when there are no collections loaded

Basically 3 levels of sorting that happens everytime you click the icon
- Unsorted (based on import date) - use `IconArrowsSort`
- Alphabetical Sort - `IconSortAscendingLetters`
- Reverse Alphabetical Sort - `IconSortDescendingLetters`

We currently don't store the collection import timestamp. This will be needed to implement the functionality to go back to original import order. We can add a new property `collection.importedAt = new Date()` to store this info. I think a new property called as `collectionSortOrder` with values `['default', 'alphabetical', 'reverseAlphabetical']` will have to be maintained in the [appSlice](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/providers/ReduxStore/slices/app.js)

VipinVIP:
I would like to work on this issue. Can I kindly be assigned?

helloanoop:
Go ahead @VipinVIP !

helloanoop:
I recommend using a alphabetical sort icon to the right end of the `collection` label in this image.
On clicking of the icon, it can sort alphabetically in the redux store.

![image](https://github.com/usebruno/bruno/assets/9350733/7183d82e-869c-4506-a45e-6b9f18f216ef)

VipinVIP:
While opening the web version via `npm run dev:web` , i get the following error.

![image](https://github.com/usebruno/bruno/assets/58673683/5b0455ff-3d5a-44ab-9660-82af7f4667c7)

is it an error from my side or am i missing any step ?

helloanoop:
@VipinVIP the app is supposed to be used only in electron.

helloanoop:
@not-known-person I have merged your PR. Nice work!
Many Thanks!

I have also added the remaining list of things to be done (with some useful pointers) to complete this feature.

[...]

PR #272, opened 2023-10-01
added feat/-#220

not-known-person:
I have added a feature to sort the collections alphabetically as guided by @helloanoop in #220

Screenshot 2023-10-01 192800

@helloanoop , kindly suggest me , should I hide the icon when there is no collections available ?


helloanoop:
Merged!

Thanks @not-known-person !
I have updated the issue with the remaining list of things that are pending.

Issue #154, opened 2023-04-28
Request rearrange is not working properly in collection


sseenivasan89:
when i tried to move the PATCH request to top the DELETE request go and sit on top of the collection 
check the below video 

https://user-images.githubusercontent.com/44080743/235034926-11cce06c-3060-483f-8a28-79e7997f410d.mp4

lared:
Just to narrow it down a bit - it's definitely affecting all requests which are being dragged within the same folder.

Steps to reproduce:

1. Create a collection with empty requests A, B, C, D, E, in this order
2. Drag E over A, it correctly slots just underneath it
3. Drag D over A, the order right now is A, D, B, C, E, while expected is A, D, E, C, B

lared:
After doing some veeery basic debugging (I'm clueless in react, I just console.error the crap out of it), I noticed that the state we see at the entry of `moveCollectionItem` is **always** the same state as when the app is opened, seemingly as if redux store never gets updated. 

https://imgur.com/BIALKw5.png

That's weird for several reasons:

- I see reducers linked to file watches firing and updating sequence numbers
- sequence number change in the store cause the view to be updated (the order just after dragging is as we expect it)
- sequence number change is caused by `Collections` component re-rendering due to `useSelector` hook which reads `state.collections` updates

lared:
Ok, now I know what the bug is.

Whenever `seq` gets updated, we never update the order of items in the collection, but then when dragging, [we rely on that order being the same as `seq` to recalculate indexes](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/utils/collections/index.js#L147).

The fix would require relying on one or the other, but not both.

helloanoop:
> Drag D over A, the order right now is A, D, B, C, E, while expected is A, D, E, C, B (just as if the previous operation was forgotten)

Love the way you narrowed it down.

> I know close to nothing about react 

I disagree

> Whenever seq gets updated by the FS watch, we never update the order of items in the collection in redux, but then when dragging, [we rely on that order being the same as seq to recalculate indexes](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/utils/collections/index.js#L147).
The fix would require relying on one or the other, but not both. It seems like seq would make more sense since it's the golden source of the data plus we already sort stuff by seq when rendering.

[...]

PR #274, opened 2023-10-01
fix(#154): correct ordering during drag-and-drop

lared:
When dragging and dropping items, to delay changing sequence numbers until after all the resource-dependent logic has completed, we were relying on the order of children in redux store (which we then converted into new seq numbers).

This order of children was however **not updated** when sequence numbers changed (for example due to file watch changes). This resulted in a seemingly random drag-and-drop ordering, which in fact was linked to the initial order when the collection was  loaded.

This change sorts all the items by sequence number prior to reordering, so that those random jumps no longer happen. As this happens on a deep clone of the collection, no data gets hurt in the process.

disclaimer: As I have close to no idea about how any of the technologies involved work, please feel free to discard this and treat it more like a proposal for a fix.


helloanoop:
Fantastic 👏👏

Merged!

Thanks @lared !

PR #275, opened 2023-10-01
feat: client certificate support

phoval:
Add initial mutual TLS support with client certificate in global preferences

### DIFF
--- a/packages/bruno-app/src/components/CollectionSettings/ClientCertSettings/StyledWrapper.js
+++ b/packages/bruno-app/src/components/CollectionSettings/ClientCertSettings/StyledWrapper.js
@@ -1,0 +1,31 @@
+import styled from 'styled-components';
+
+const StyledWrapper = styled.div`
+  .settings-label {
+    width: 80px;
+  }
+
+  input {
+    width: 300px;
+  }
+
+  .textbox {
+    border: 1px solid #ccc;
+    padding: 0.15rem 0.45rem;
+    box-shadow: none;
+    border-radius: 0px;
+    outline: none;
+    box-shadow: none;
+    transition: border-color ease-in-out 0.1s;
+    border-radius: 3px;
+    background-color: ${(props) => props.theme.modal.input.bg};
+    border: 1px solid ${(props) => props.theme.modal.input.border};
+
+    &:focus {
+      border: solid 1px ${(props) => props.theme.modal.input.focusBorder} !important;
+      outline: none !important;
+    }
+  }
+`;
+
+export default StyledWrapper;
--- a/packages/bruno-app/src/components/CollectionSettings/ClientCertSettings/index.js
+++ b/packages/bruno-app/src/components/CollectionSettings/ClientCertSettings/index.js
@@ -1,0 +1,120 @@
+import React, { useEffect } from 'react';
+import { useFormik } from 'formik';
+import * as Yup from 'yup';
+
+import StyledWrapper from './StyledWrapper';
+
+const ClientCertSettings = ({ clientCertConfig, onUpdate, onRemove }) => {
+  const formik = useFormik({
+    initialValues: {
+      domain: '',
+      certFilePath: '',
+      keyFilePath: '',
+      passphrase: ''
+    },
+    validationSchema: Yup.object({
+      domain: Yup.string().required(),
+      certFilePath: Yup.string().required(),
+      keyFilePath: Yup.string().required(),
+      passphrase: Yup.string()
+    }),
+    onSubmit: (values) => {
+      onUpdate(values);
+    }
+  });
+
+  const getFile = (e) => {
+    formik.values[e.name] = e.files[0].path;
+  };
+
+  return (
+    <StyledWrapper>
+      <h1 className="font-semibold mt-4 mb-2">Current client certificates</h1>
+      <ul>
+        {!clientCertConfig.length
+          ? 'None'
+          : clientCertConfig.map((clientCert) => (
+              <li>
+                Domain: {clientCert.domain}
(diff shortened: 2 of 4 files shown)


helloanoop:
Apologies, I haven't gotten to this PR since last week. This will be a priority item on my list to get it merged this week.

adiesner:
@phoval thank you for taking care. I do miss that feature in Bruno too. 

But looking at your implementation it looks like you only support one client certificate and one domain. I think it is very common that you need to configure multiple client certificate for multiple domains. Just thinking about a preview and a prod environment. 

Do you have time to extend your implementation?

phoval:
Just moved the config in the collections settings. Made possible multiple client certificate.
UI still need to be polish.
Maybe encode the certificate/key file in base64 for easier git sync?

panda7789:
@phoval 👍  for your work.

Just a little error I found,
config must contain at least empty collection of clientCert:
image
When collection is not present, it fails on .push into undefined:
image

I think you didnt notice, because you already have it in config.

max-wittig:
Thanks for all the work here! This is really the last blocker to switch away from the buggy, data hungry Insomnia

mjhcorporate:
For reference: This is what the UI to set the certificate looks like (I like it!)
image

But: When I click "add", the values are not persisted. When I navigate away from the UI, save another request, and then come back, the UI is reset to all empty values.

panda7789:
@mjhcorporate Try to add into config this empty collection 
image
And than try save cert again. After click on add it should be listed on upper list, where you now have "None".

mjhcorporate:
Ahh, that was the error you mentioned earlier. Thank you, now it works. 

But: Saving the key in the config has two major downsides for us:
- (minor): The certificates expire and need to be refetched daily. I have a script that updates them on my harddrive. How would I trigger a re-import?
- (**major**): I am tracking my `bruno.json` in git. Under no circumstances must I upload the secret `key` to GitHub!

Proposal: Rename `cert` and `key` to `certPath` and `keyPath`, and store the path to the files. Then, on each request, read the file contents (no caching please). That way, the certs are guaranteed to be fresh. We lose a few ms for the file system access, but I think it's worth it in this case. This is also the path that Postman and Insomnia took.

helloanoop:
> I am tracking my bruno.json in git. Under no circumstances must I upload the secret key to GitHub!

@mjhcorporate  Checkout two options for managing secrets without having to commit credentials to git - https://docs.usebruno.com/secrets-management/overview.html

To interpolate env vars, you need to change
```javascript
httpsAgentRequestFields['passphrase'] = clientCert.passphrase;
```
to 
```javascript
httpsAgentRequestFields['passphrase'] = interpolateString(clientCert.passphrase, interpolationOptions)
```
You can checkout how we do this in proxy 

@phoval Can you share how the json schema will look like in `bruno.json`
I recommend to use the full name instead of shorthand.
```json
{
  "clientCertificates": [{
    "certFilePath": "assets/cert.pem",
    "keyFilePath": "assets/cert.key",
    "passphrase": "{{process.env.clientCertPassPhrase}}"
  }]
}
```

> But: When I click "add", the values are not persisted. When I navigate away from the UI, save another request, and then come back, the UI is reset to all empty values.

You need to implement the save functionality. Checkout how it's done in Proxy Settings Tab

Glad to see the progress !! @phoval @mjhcorporate Let me know as soon as you guys are in agreement. Then we can merge.

mjhcorporate (approved):
I retested, and it looks great! Thank you so much for your amazing work, @phoval !

Minor nitpicks:
1. There is no user-facing error when the filepath becomes invalid (file is deleted or moved). Ideally, the user would see an error message, just like the one they see when the cert-file is corrupted.
2. Absolute filepaths and relative filepaths work, but `~` does not.
3. `.env`-variables are not permitted for the paths. When entering `"certPath": "{{process.env.clientCertPath}}"`, I get `Error reading cert/key file Error: ENOENT: no such file or directory, open '{{process.env.clientCertPath}}'`
   - If this would be possible, then the nitpick above is solved, as every user can expand their `~` themselves, manually
5. I happen to have my cert in a directory that starts with a `.`, and to show that directory in the MacOS filepicker, I have to press `CMD + Shift + .`, every time I open the picker. It would be amazing if the UI remembered by `show-dot-files` choice.
6. The UI does not have enough space to show the chosen filenames or the "nothing selected" strings (see "Kein...ählt" and "car-...key" in the screenshot below) 
image
@helloanoop Everything good from my side! In my view, these points are minor and could be addressed by follow-up PRs some time in the future.

max-wittig:
@mjhcorporate Exactly. Let's get this in so people can use it and then the nitpicks can be addressed later.

voudin:
@helloanoop changed to long config name and fixed interpolate options
Should be good to be merged now

phoval:
@helloanoop changed the configuration name to the longer one and fixed the interpolation options.
```
"clientCertificates": [
    {
      "domain": "*.example.org",
      "certFilePath": "client.pem",
      "keyFilePath": "client.key",
      "passphrase": "test"
    }
  ]
```
It should be ready for merging now.

helloanoop:
Thanks @phoval for adding Client Certificates support in Bruno
I appreciate your patience and efforts towards incorporating the feedback from the community.

I added some polish https://github.com/usebruno/bruno/commit/333564f687b2e6f05dff90f5abe3184d850cd091
As well as updated the json structure to allow for enabling/disabling client certificate usage.

```json

  "clientCertificates": {
    "enabled": true,
    "certs": [
      {
        "domain": "localhost",
        "certFilePath": "/Users/anoop/bruno/bruno-testbench/certs/server.crt",
        "keyFilePath": "/Users/anoop/bruno/bruno-testbench/certs/server.key",
        "passphrase": ""
      }
    ]
  }
```

image

PR #282, opened 2023-10-02
feature/preview-response-html

Its-treason:
https://github.com/usebruno/bruno/assets/39559178/7ffe3afd-f557-4112-b1eb-77b14ebbfcf0

The tabs are build in a way which allows for more tabs to be added in the future. We could an extra Tab for the unformatted response.

Because of the Content-Security-Policy set in electron, images, CSS and other things are not loaded properly. With the settings updated, the Preview looks better:

```diff
    setContentSecurityPolicy(`
      default-src * 'unsafe-inline' 'unsafe-eval';
      script-src * 'unsafe-inline' 'unsafe-eval';
      connect-src * 'unsafe-inline';
-     base-uri 'none';
      form-action 'none';
-     img-src 'self' data:image/svg+xml;
    `);
```

https://github.com/usebruno/bruno/assets/39559178/494493e6-a218-4372-85ca-050ebca317e7

But I'm not sure what security problem might occur with the updated Content-Security-Policy, so I did not include them in this PR.

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/QueryResult/StyledWrapper.js
+++ b/packages/bruno-app/src/components/ResponsePane/QueryResult/StyledWrapper.js
@@ -3,14 +3,7 @@ import styled from 'styled-components';
 const StyledWrapper = styled.div`
   display: grid;
   grid-template-columns: 100%;
-  grid-template-rows: 50px calc(100% - 50px);
-
-  /* If there is only one element (the preview, not tabs) make it span over both grid rows */
-  > *:last-child:first-child {
-    grid-row: 1 / 3;
-    margin-top: 1.25rem;
-    height: calc(100% - 1.25rem);
-  }
+  grid-template-rows: 1.25rem calc(100% - 1.25rem);
 
   /* This is a hack to force Codemirror to use all available space */
   > div {
--- a/packages/bruno-app/src/components/ResponsePane/QueryResult/StyledWrapper.js
+++ b/packages/bruno-app/src/components/ResponsePane/QueryResult/StyledWrapper.js
@@ -1,9 +1,28 @@
 import styled from 'styled-components';
 
 const StyledWrapper = styled.div`
+  display: grid;
+  grid-template-columns: 100%;
+  grid-template-rows: 50px calc(100% - 50px);
+
+  /* If there is only one element (the preview, not tabs) make it span over both grid rows */
+  > *:last-child:first-child {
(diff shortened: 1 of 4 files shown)


helloanoop:
@Its-treason This is great! A couple of suggestions.

> Because of the Content-Security-Policy set in electron, images, CSS and other things are not loaded properly. With the settings updated, the Preview looks better:

You can remove this policy, if we block it, then as you said the images and css won't load 

Also, I have a suggestion, Can you update the tab design. Let's try to keep the views as symmetric as possible. 
In the pic below, we want to keep both the editors (left and right) as symmetric as possible
Here is my design proposal
image

Code FYR that I used to get this view in `QueryResult`
```javascript
const QueryResult = ({ item, collection, data, width, disableRunEventListener, headers }) => {
  // ... existing code
  const [view, setView] = useState('preview');

  return (
    <StyledWrapper className="px-3 w-full" style={{ maxWidth: width }}>
      <div className="h-full">
        <div className="flex items-center justify-end mb-1">
          <span
            className={`cursor-pointer mr-2 text-xs ${view === 'preview' ? 'text-yellow-600' : 'text-gray-400'}`}
            onClick={() => setView('preview')}
          >
            preview
          </span>
          <span
            className={`cursor-pointer text-xs ${view === 'raw' ? 'text-yellow-600' : 'text-gray-400'}`}
            onClick={() => setView('raw')}
          >
            raw
          </span>
        </div>
        <CodeEditor collection={collection} theme={storedTheme} onRun={onRun} value={value} mode={mode} readOnly />
      </div>
    </StyledWrapper>
  );
};

export default QueryResult;
```
The margin top for result would have to be removed [here](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/components/ResponsePane/index.js#L118)
something like
```html
      <section className={`flex flex-grow ${focusedTab.responsePaneTab !== 'response' ? 'mt-5' : ''}`}>
        {getTabPanel(focusedTab.responsePaneTab)}
      </section>
```
And the margin `mt-5` can be conditionally added in `QueryResult` depending on whether we have a `preview|raw` tabs or not

Its-treason:
I updated my PR with your suggested design and removed the CSP to show images etc correctly.

https://github.com/usebruno/bruno/assets/39559178/2a96ed91-1c10-484d-b196-1689885c2806

helloanoop:
This is looking great!

Merged. Thanks @Its-treason !!

PR #288, opened 2023-10-02
Fix for XML Formatting Options

Cibico99:
SafeParseXML - Option for collapseContent: true in QueryResult/index is not being passed to XML-Formatter.


helloanoop:
Merged.

Thanks @Cibico99 !

PR #293, opened 2023-10-02
fix: add package-lock.json

lucasew:
- Removed package-lock.json from gitignore

Other projects keep only one package manager lockfile, like if they use Yarn they ignore the lockfile from npm.

Right now git is ignoring all of them. You guys are using npm so I commented out the npm lock from .gitignore.

- npm install

Just to generate the lockfile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 node_modules
 yarn.lock
 pnpm-lock.yaml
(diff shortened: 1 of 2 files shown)


helloanoop:
@lucasew I have take care of this 
The app now has `package-lock.json` in the git repo.

PR #296, opened 2023-10-02
fix: Color in environment delete modal

Its-treason:
grafik


helloanoop:
Merged. Thanks @Its-treason !

Issue #291, opened 2023-10-02
[Feature] Double click to rename a request


thelazyoxymoron:

ilumin:
Should it render input in the request or render the modal to rename the request (right now when I click the rename in the option menu, it renders the modal to rename the request)

thelazyoxymoron:
Ideally, it should enable renaming in-place rather than a modal, but for the sake of consistency, a modal would also work.

helloanoop:
This was delivered by @ilumin and is now available in `v0.18.0`

PR #299, opened 2023-10-03
feat(291): add double click request to open rename modal

ilumin:
As discussed in #291 

- Add double click handler on `CollectionItem`
- On double click render rename modal


helloanoop:
Merged!

Thanks for taking care of this @ilumin !

Issue #270, opened 2023-10-01
Assert still tested even if deleted before unchecked


arnaduga:
When an assert is delete, the tests are still ran.

To reproduce:
- Create an assert and run 
<img width="1106" alt="image" src="https://github.com/usebruno/bruno/assets/56150873/34fd2182-f4db-4be4-ab48-368560641ad7">

- Delete the assert using the "bin" icon and run again
<img width="1106" alt="image" src="https://github.com/usebruno/bruno/assets/56150873/0be838f5-da65-403b-ba8a-4e143e054a10">

The assertions is failing the same.

To workaround:
- recreate an assert
- uncheck the newly created assert
- only then, delete the assert using the icon

Version: Bruno v0.16.5
OS: MacOS Sonoma 14.0

helloanoop:
This has been fixed by @Its-treason and is now available in `v0.18.0`

PR #308, opened 2023-10-03
fix(#270): Fix deleted assertions & tests shown in test tab

Its-treason:
After all tests or assertions were deleted, the store was not updated because of the length checks. Now the assertions/tests will always try to run, to ensure the data is always updated.

Before my fix:

https://github.com/usebruno/bruno/assets/39559178/0ad3e466-f7e3-4f3d-ae02-27ecd22c616e

With my fix:

https://github.com/usebruno/bruno/assets/39559178/90edded8-a23e-4d1b-a4fb-9b5611a2a00d


helloanoop:
Thanks for the PR @Its-treason !

Your video descriptions really helps a lot!

Merged!

Issue #314, opened 2023-10-03
Unable to delete Collection folder in GUI in case folder does not exist on filesystem


martinsefcik:
When folder is deleted on filesystem (outside Bruno) then it stays in Bruno and it is not possible to delete it due to the following error message:

```
Uncaught (in promise) Error: Error invoking remote method 'renderer:delete-item': Error: The directory does not exist
```

I guess this error should be ignored.

martinsefcik:
@lared I can reproduce it at least with versions 0.18.0, 0.17.0, 0.16.6 on Windows 11.

martinsefcik:
Steps to reproduce:
1. Create new subfolder in collection via some file commander (outside Bruno)
2. Folder is immediately showed Bruno UI
3. Delete this folder via file commander
4. Folder is still visible in Bruno UI
5. Try to Delete folder via Bruno UI -> nothing happen - no error popup, folder is not deleted, and the error message mention here in before is printed in Developer console.

martinsefcik:
I cannot anymore reproduce it. Probably fixed in #465

PR #321, opened 2023-10-03
fix(#314): tabbing order through tables (variables etc) was inconsistent or impossible

lared:
Two changes:

- the inline editor was made tab-friendly (it ignored tabs anyway, now it just treats them correctly)
- tabbing omits selection/deletion widgets as they seem much less likely to be used in practice

Now it's still not perfect for a few reasons:

- I'm not even sure what the user experience **should** be when taking into account sidebar
- if you add empty rows in headers and tab out to URL, the empty rows stay in place; if you start typing - they disappear

But a tiny bit less annoying nevertheless.

There is also an awful lot of copy-pastyism with those tables, might be work time consolidating at some point.

### DIFF
--- a/packages/bruno-app/src/components/RequestPane/Assertions/AssertionRow/index.js
+++ b/packages/bruno-app/src/components/RequestPane/Assertions/AssertionRow/index.js
@@ -197,10 +197,11 @@ const AssertionRow = ({
           <input
             type="checkbox"
             checked={assertion.enabled}
+            tabIndex="-1"
             className="mr-3 mousetrap"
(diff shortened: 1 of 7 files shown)


helloanoop:
Hey @lared ! Thanks for the PR!

> tabbing omits selection/deletion widgets as they seem much less likely to be used in practice

Agreed. I think its never gonna be used in a single line editor

> Now it's still not perfect for a few reasons:

I think we can live with them for now.

> There is also an awful lot of copy-pastyism with those tables, might be work time consolidating at some point.

Yup. Agreed.

I have merged the PR.

PR #327, opened 2023-10-04
decomment JSON before send

pove:
Easy way to remove comments before send request. In future, it could be better to use a package for this (https://www.npmjs.com/package/decomment)


helloanoop:
@pove Can you update the PR to use https://www.npmjs.com/package/decomment itself.

I am just worried if the regex breaks blocking people from using bruno until a fix arrives.

pove:
I am not a JavaScript developer, but I will try tomorrow

Issue #329, opened 2023-10-04
Basic Auth header not sent


nico-m-64:
Hi,
Just say that you made a very very great tool, hope this open source project will be 🚀 all around the solar system ! Thank you so much for your work.

It seems that the header "Authorization: Basic *****" is not sent when configuring the Auth tab with this type.

Have a nice day

l0new4nder3r:
It works fine for me with version 0.18.0: I just replaced a manually generated "Authorization" (Basic) header with the needed username and password in the "Auth" tab, and I can successfully access a restricted API in one of my apps.

Could you try again, or check if you do not have anything else interfering your test, @nico-m-64 ?

One other thing, though: when using "Generate code" on this new request, the "Authorization: Basic xxx" header is not included in the "curls". Should I create another issue, @helloanoop ?

helloanoop:
Yes, it worked for me too in my local tests. But we need to use headers. The current code uses axios inbuilt auth which is resulting in the header not being populated.

Here is the part we need to make changes: https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/ipc/network/prepare-request.js#L22-L33

I will send out a patch fix for this later tonight.

francocm:
I'm playing with Basic Auth on `v0.18.0` and observed that Basic Auth works fine if I put hardcoded values. However it breaks if I use variables.

# Example

## Works

* Username: `foo`
* Password: `foo`

## Does not work

* Username: `{{my-username}}`
* Password: `{{my-password}}`

lared:
@helloanoop should it be done there, or in dedicated step where we interpolate everything afterwards?

I don't think this is going to help interpolating generated code, as there the only piece of code that's actually being interpolated is the url.

helloanoop:
This has been fixed and released in `v0.19.0`
CLI support has also been released in `v0.11.0`

nico-m-64:
Thanks a lot for the quick fix ! Everything work perfectly 👍
I wish you the best for the future

PR #348, opened 2023-10-04
fix(#329): interpolate basic auth params

lared:
We were missing interpolation for basic auth. Bearer token is fine, it was just added as a header (and those are interpolated).


lared:
I see that CLI has the same code, will replicate it there too

helloanoop:
@lared I have take care of this. I decided to use header based approach so that the header shows up in response pane.
https://github.com/usebruno/bruno/commit/8d8949630469c2ec979ffac20c0e41d2c0aff564

PR #356, opened 2023-10-04
github star button color fix

therealrinku:
I don't know, if it was for me only but,
while on dark mode github star button was still looking white, so I made some changes to fix that.

checked on my device, works on both dark and light mode

before  changes ::
Screenshot from 2023-10-04 22-15-56

after changes ::
Screenshot from 2023-10-04 22-16-29


helloanoop:
Nice!

Thanks @therealrinku !

PR #368, opened 2023-10-04
fix(network): use axios interceptors to better reflect durations

luizfonseca:
Hello there 👋🏽 

### What has been done

- Create an `axios` instance that uses `interceptors.response` and `interceptors.request` to have a better/accurate request duration number
- Add missing dependencies. For example `typescript` and `prettier` were missing and they are needed by `npm run build:bruno-query` and `npm run build:graphql-docs` and `husky` respectively.
  - `pretty-quick` is an old package and depends on an older prettier version

In regards to the missing deps, let me know if you would rather have this discarded from the PR.

### Why

I recently started using Bruno after the Insomnia fiasco, and noticed that my API request durations weren't necessarily correct.

The issue being that using `Date.now()` before and after the request happens wouldn't yield correct duration results since axios has other timings steps. For example, a request in my server that takes 4ms would result in 12~14ms in the previous logic.

Now the duration hovers between 4~6ms, as we only run the timers exactly when a requests starts and it ends.

Important to note here that even using the `interceptors`, they won't provide an extremely accurate timing number since we don't have control of the axios internals to get the exact value (see issue here: https://github.com/axios/axios/issues/695). With more time we can maybe introduce an accurate timings lib.

### Before

Screenshot 2023-10-04 at 22 33 54

### After
Screenshot 2023-10-04 at 22 35 23

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "@playwright/test": "^1.27.1",
     "husky": "^8.0.3",
     "jest": "^29.2.0",
+    "prettier": "^2.0.2",
     "pretty-quick": "^3.1.3",
     "randomstring": "^1.2.2",
     "ts-jest": "^29.0.5",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "jest": "^29.2.0",
     "pretty-quick": "^3.1.3",
     "randomstring": "^1.2.2",
-    "ts-jest": "^29.0.5"
+    "ts-jest": "^29.0.5",
+    "typescript": "^5.2.2"
   },
   "scripts": {
     "dev:web": "npm run dev --workspace=packages/bruno-app",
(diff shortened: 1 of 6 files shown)


helloanoop:
This is great!

Thanks for the detailed writeup!

Merged!

helloanoop:
@luizfonseca For now, I have removed the `typescript` and `prettier` on the main branch.

> they are needed by npm run build:bruno-query and npm run build:graphql-docs and husky respectively.

this should not happen. You were using node 18 right ?

> pretty-quick is an old package and depends on an older prettier version

When you find some time, please raise a new PR for this

luizfonseca:
> this should not happen. You were using node 18 right ?
@helloanoop Oh you are right -- I was using node v20; node18 works fine without the deps. Thanks fore removing it 👍🏽 

> When you find some time, please raise a new PR for this
I'll be happy to, great work so far btw!

Issue #111, opened 2023-02-28
Upon creating a New Request it should open automatically


helloanoop:
Upon creating a New Request it should open automatically
Currently, the old tab remains active.

paulius-valiunas:
I would like to take this a step further and allow opening a "temporary" tab without adding it to a collection. Just like you can open a new tab/file in VSCode without necessarily choosing a save location yet.

sanjai0py:
Closing this!

As the [PR](https://github.com/usebruno/bruno/pull/373) adding the feature was already merged and released in [v0.20.0](https://github.com/usebruno/bruno/releases/tag/v0.20.0)

PR #373, opened 2023-10-05
feat(#111): Automaticly open a newly created request

Its-treason:
https://github.com/usebruno/bruno/assets/39559178/bd61036c-05ff-4e27-aaf2-21ba3bac1489

So this once implemented a little hacky, because of 2 problems:

I had some skill issues figuring out how Redux works, never worked with Redux.

The way creating files is implemented makes it a little hard to implement this feature properly. When creating the request, it is written to disk and then picked up by the file watcher and read into Bruno with a new UID. For a proper implementation of this feature, we might need to rewrite how file creating works, so that the new request is directly added to the Redux store and not by the file watcher with a new UID.

This now works by writing the last created request name into the store (`newRequestName`) and when the file watcher opens a new file it checks using this name if it was newly created.

### DIFF
--- a/packages/bruno-app/src/providers/App/useCollectionTreeSync.js
+++ b/packages/bruno-app/src/providers/App/useCollectionTreeSync.js
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import {
   collectionAddDirectoryEvent,
   collectionAddFileEvent,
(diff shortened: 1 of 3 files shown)


helloanoop:
Hey @Its-treason !

You always make beautiful video descriptions of the diffs!

> The way creating files is implemented makes it a little hard to implement this feature properly. When creating the request, it is written to disk and then picked up by the file watcher and read into Bruno with a new UID. 

This is by design. Think of the filesystem as the state and the application as view that reacts to any state change.
Most of the state updates are done by writing to the filesystem. The application is just a view that reacts to the state changes.

There are of course state updates (like status of the request, responses, opened tabs, test results, etc) that are not persisted to the filesystem. These are stored directly in the application.

The convention is 

// all filesystem updates are done here and the application reacts to it
`slices/collections/actions.js` 

// the rest of the state updates are done here
`slices/collections/index.js`

> So this once implemented a little hacky

Can you improve this. We have done a simlar thing for adding environments, where the new environment gets focused after creation.
We basically store a [lastAction](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js#L635-L642) and when we get a environment creation event, we check and see if the lastAction was a [environment creation](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js#L1041-L1064) and if so, we focus the newly created environment.

Its-treason:
Hey @helloanoop,

thank you for the explanation on the file system as a state and your feedback.

I updated my PR to use the existing lastAction in the store and remove the `newRequestName`.

helloanoop:
Thanks @Its-treason !

Merged.

I refactored the approach slightly  https://github.com/usebruno/bruno/commit/be8db1876a9eccc4e7f73d8604f1093ed109ff93 
I was worried, that adding `collections` and `tabs` in the ipc listeners useEffect might cause too many mount and remount ops
```
useEffect(() => {
  // ipc listeners
}, [isElectron, tabs, collections]);
```

PR #374, opened 2023-10-05
Changed Proxied Requests to Use https-proxy-agent/http-proxy-agent.

acostinescu:
Follow up PR to #252 which introduces HttpProxyAgent to replace the httpAgent in Axios and therefore properly handles HTTP requests through proxies and not just HTTPS requests.


helloanoop:
@acostinescu  Many thanks for working on this.

This is now available in `v0.21.0` Please test it once on your end.

Thanks!

PR #391, opened 2023-10-05
Show response time in milliseconds per request and total

dozed:
This PR shows the response time in milliseconds per request and total


dozed:
I added a PR to use axios interceptors: https://github.com/usebruno/bruno/pull/397

PR #393, opened 2023-10-05
Better BrowserWindow

zyrouge:
- Only setting `icon` property would set the icon of the window in Linux.
- Set `title` to `Bruno`.
- Enable `autoHideMenuBar` to prevent ugly menu bars in dark mode.

image

I was not able to test these changes due to `phantomjs` errors but I'm sure that these changes are correct.

### DIFF
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -36,6 +36,7 @@ app.on('ready', async () => {
       preload: path.join(__dirname, 'preload.js'),
       webviewTag: true
     },
+    title: 'Bruno',
     icon: path.join(__dirname, 'about/256x256.png'),
     autoHideMenuBar: true
   });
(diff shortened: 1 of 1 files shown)


zyrouge:
By the way, is opening multiple windows an intended behavior? The created window does not have the same theme as set in settings.

helloanoop:
Merged!

Thanks @zyrouge 

> Enable autoHideMenuBar to prevent ugly menu bars in dark mode.

This is nice. I hope people don't start looking for menubar and not know the "alt" key. We need to add this info somewhere in the app so that people don't get lost.

> By the way, is opening multiple windows an intended behavior? 

No, The expectation is that you have a single window. When you close and re-open, bruno currently remembers the last collections that were opened and loads them automatically. 

I haven't tested multiple windows actually. We might have some race conditions. Atleast for now, the recommendation would be to open a single window.

PR #395, opened 2023-10-05
Put constant in outer scope

dozed:
This PR fixes an issue with `start` constant not being in scope

There is still an issue with a failed request not being shown as failed (the yellow marked):

This should probably fixed in a separate PR?


helloanoop:
Thanks. Merged!

> There is still an issue with a failed request not being shown as failed (the yellow marked):

If I am not wrong, `Failed` implies that a test or an assertion failed right?
There can be 400 responses which can be a part of test that checks specifically for a 400.

dozed:
Ok, makes sense now :)

PR #397, opened 2023-10-05
Use axios interceptor to measure response time

dozed:
Measure response time with axios interceptor, similar to https://github.com/usebruno/bruno/pull/368

There is now some code duplication with `axios-instance.js` being twice in the project. Should we address this in this PR?

Note that I also modified `axios-instance.js` slightly, since `error.response` might be undefined:

```javascript
    (error) => {
      if (error.response) {
        const end = Date.now();
        const start = error.config.headers['request-start-time'];
        error.response.headers['request-duration'] = end - start;
      }
      return Promise.reject(error);
    }
```

### DIFF
--- a/packages/bruno-cli/src/utils/axios-instance.js
+++ b/packages/bruno-cli/src/utils/axios-instance.js
@@ -1,0 +1,40 @@
+const axios = require('axios');
+
+/**
+ * Function that configures axios with timing interceptors
+ * Important to note here that the timings are not completely accurate.
+ * @see https://github.com/axios/axios/issues/695
+ * @returns {import('axios').AxiosStatic}
+ */
+function makeAxiosInstance() {
+  /** @type {import('axios').AxiosStatic} */
+  const instance = axios.create();
+
+  instance.interceptors.request.use((config) => {
+    config.headers['request-start-time'] = Date.now();
+    return config;
+  });
+
+  instance.interceptors.response.use(
+    (response) => {
+      const end = Date.now();
+      const start = response.config.headers['request-start-time'];
+      response.headers['request-duration'] = end - start;
+      return response;
+    },
+    (error) => {
+      if (error.response) {
+        const end = Date.now();
+        const start = error.config.headers['request-start-time'];
+        error.response.headers['request-duration'] = end - start;
+      }
+      return Promise.reject(error);
+    }
+  );
+
+  return instance;
+}
+
+module.exports = {
+  makeAxiosInstance
+};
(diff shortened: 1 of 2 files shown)


luizfonseca:
➕   _If possible we should have just 1 of this util since the logic will be the same ^_

packages/bruno-cli/src/utils/axios-instance.js
```diff
@@ -0,0 +1,40 @@
...
+      const start = response.config.headers['request-start-time'];
+      response.headers['request-duration'] = end - start;
+      return response;
+    },
+    (error) => {
+      if (error.response) {
+        const end = Date.now();
+        const start = error.config.headers['request-start-time'];
+        error.response.headers['request-duration'] = end - start;
+      }
```
luizfonseca, line 30: 2cents: these 2 don't need to be inside the `if`, only the `error.response.headers['request-duration']` part.

It might be useful to be able to zero it if duration is not present etc
       

```
const end = Date.now();
const start = error.config.headers['request-start-time'];
```

dozed: Not sure I understand, what do you want to zero?

luizfonseca: because of the `if` we won't get `headers['request-duration']` if error is not a response, meaning it will be `undefined` 

in the line where you display it (`${responseTime} ms`) it needs to either be `0` or not be displayed for example

luizfonseca: You can probably solve it by always returning the duration or just

```js
// defaults to 0 if header is not present
responseTime = response.headers.get('request-duration') ?? 0
```

dozed: If `error.response` is `undefined` we don't need the `responseTime`, since there is a short-circuiting `return`: https://github.com/usebruno/bruno/blob/bb7d13d2d969feaafbe5c75299c570962db53b39/packages/bruno-cli/src/runner/run-single-request.js#L144-L165

In the other case the header is always present.

luizfonseca: ah gotcha 👍🏽 , then it's all good

dozed: Thanks for the feedback :+1:

helloanoop:
Merged.

Thanks @dozed  and @luizfonseca !

PR #413, opened 2023-10-06
request colors made consistent on the sidebar and tabs

therealrinku:
The colors of the request on the tab and the sidebar item were different, so I made some changes to fix that.

before::
before__

after ::
after___


helloanoop:
Nice!

Issue #345, opened 2023-10-04
Using dot (.) in variables names are recognized by the UI but not picked up when sending request


haneev:
Really great tool! Keep up the good work. 

However I found a small bug 🙃

Some variables/placeholders are not picked up by the engine (but are recognized by the UI).

I tested the following:
- {{auth_token}} ✅
- {{auth-token}} ✅
- {{auth.token}} ⚠️

It does not work globally, so it does not work for url, body or headers.

helloanoop:
> Some variables/placeholders are not picked up by the engine (but are recognized by the UI).

Yes, this needs to be fixed. Codemirror shouldn't be highlighting this variable. The `.` may lead to unpredictable behaviour since it may interfere with the [handlebars templating logic](https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/ipc/network/interpolate-vars.js#L31-L57)

lared:
@helloanoop is it just dots? Do we have some sort of a regex for valid variable names?

razzivofficial:
@helloanoop i tried to resolve this bug please review it if possible ! I am happy for any suggestions

helloanoop:
Hey @razzivofficial !

### Current and Expected Behaviour
Currently - env vars, request and other vars don't have a very strict validation for naming present. The expectation is that they should follow javascript's variable naming validation which is 

*Variable Names Must Start with a Letter, Underscore (_), or Dollar Sign ($)*

Unless there is a strong use-case (happy to change my mind if there is one) to name a var with other chars, bruno should use javascript's naming convention.

### Nested Vars
> Nested variables like {{auth.token}}

I have used the nested environments of insomnia in the past and like its benefits of have a base env and other env's extending the base env.

The reason why I went ahead with a simple key-val (like postman) was just because of its simplicity.

**Is there a plan to support nested environments ?**

Yes, In V2 - see [roadmap](https://github.com/usebruno/bruno/discussions/384)
Issue: https://github.com/usebruno/bruno/issues/292

### Changes need to nudge users that the name is invalid
To solve this, We need to show an error when people try to save an environment with an invalid name.
@razzivofficial Can you work on making these changes [here](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/components/Environments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js)

I am assuming we can make the table cell color red for invalid names, and disable the save button

[...]

Issue #292, opened 2023-10-02
Nested Environments?


drventure:
Just learned about Bruno and really liking it so far.

I'm coming from Insomnia and one of their fairly nice features is nested environments, where you can have a base environment and 1 or more additional environments nested beneath it that include (or override) variable values in the base environment.

I couldn't find any mention of that in the docs, so if it's not there now, I'd like to suggest it as a possibility.

Thanks!

danielraab:
Would like this feature too. It is very helpful if you have a lot of environments.

helloanoop:
I have used the nested environments of insomnia in the past and like its benefits of have a base env and other env's extending the base env.

The reason why I went ahead with a simple key-val (like postman) was just because of its simplicity.

**Is there a plan to support nested environments ?**

Yes, In V2 - see [roadmap](https://github.com/usebruno/bruno/discussions/384)

I imagine allowing a configuration in `bruno.json` that allows people to choose the environment type.
something like 
```json
{
  "version": "1",
  "name": "bruno-testbench",
  "type": "collection",
  "nestedEnvironments": true
}

drventure:
Definitely this. Insomnia's environments are a pretty simply structure. Take the base, then iteratively apply each level of nesting to the base, till you get to the selected environment.

There's also an additional slight twist to this.

Each FOLDER that you create can also have environment overrides (that are applied lastly).

This allows you to group those requests that have similar environment requirements together and override base env settings for that specific group.

bbbco:
This is one of the primary hangups preventing me from migrating completely to Bruno. I'm still using Insomnia ... I reverted 2023.4.0 because of issues experienced in 2023.5.x and never did upgrade past that when I saw all the headaches with forcing users to sync with the cloud.

Faqual:
Likewise, nested env and nested variables are the main things making me hesitate to migrate immediately

benyamynbrkyc:
Nesting environments and variables and having these configured by workspace would be fantastic. Would switch immediately

minamz:
Yes please

lauhon:
Would also love this!

An easy alternative for Bruno could be to just enable referencing environment files within another.
If it is a good first issue I'm happy to contribute :)

drventure:
Was playing around a little more with Bruno looking for a way to resolve nested environments.

One thing I noticed is that variables from .env files don't seem to ever be "resolved".

What I mean by that is say you have a .env file like this

var1=value1
var2=custom{{var1}}

[...]

PR #414, opened 2023-10-06
insomnia-importer: import environments

petoc:
Import environments.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,6 +37,7 @@
     },
     "node_modules/@alloc/quick-lru": {
       "version": "5.2.0",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=10"
@@ -47,6 +48,7 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.2.1",
+      "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@jridgewell/gen-mapping": "^0.3.0",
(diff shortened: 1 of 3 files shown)


helloanoop:
@petoc 

Thank you for the PR!

Just a quick question, Have we taken care of converting insomnias nested variables and replace `.` with `_`

As in `auth.token -> auth_token`
The expected behaviour is that env vars should follow javascript's variable naming validation which is

*Variable Names Must Start with a Letter, Underscore (_), or Dollar Sign ($)*

Also See https://github.com/usebruno/bruno/issues/345

petoc:
Sorry about that I was not aware of that, thanks for pointing that out. Updated PR to address this.
- environment variables (objects/arrays) should be flattened
- variables in body/query/auth/headers are normalized too

I suppose there will be some issues with variables due to nested environments in Insomina (i.e. one variable from Base Environment and other from Sub Environment in one request), but it should be resolved in Bruno v2 (#292).

petoc:
@helloanoop if there is anything else you would like to address, please let me know.

I noticed that there is already issue #472 which should be solved by this PR.

F0rce:
LGTM 🎉

ItsJustRuby:
@petoc Could you please resolve the conflicts on this branch?
I am looking forward to this being merged 😊.

petoc:
Updated, but it is up to @helloanoop if it will be merged.

Issue #428, opened 2023-10-06
[Bug] Requests order in a folder is reset when any request is dragged in or out of that folder


lared:
Steps to reproduce:

1. Create a new collection
2. Create a folder
3. Inside that folder, create requests `D`, `C`, `B`, `A` - exactly in that order
4. Drag them so that they are now in order `A`, `B`, `C`, `D`
5. Outside that folder, create request `E`
6. Drag request `E` into the folder

Expected: requests are in order  `A`, `B`, `C`, `D`, `E`
Actual: requests are in order  `D`, `C`, `B`, `A`, `E`

I'll try to troubleshoot and fix. Same thing happens when we drag things out, but only to the root of the collection.

helloanoop:
@lared This has been fixed in `v0.21.0`

Please test this in the latest release and close if its working as expected.

lared:
Yep it's working as expected, thanks!

PR #430, opened 2023-10-06
fix(#428): moving requests into directories or into root of collection broke ordering

lared:
Previously I touched some code for moving requests between folders in #154, those two additional fixes should (hopefully) fully address the situation.


helloanoop:
@lared Dont worry abt Test failure. Its because your branch does not have package-lock.json that is now present in main branch.

I will review this and get it merged.

helloanoop:
Merged.

Thanks @lared !

PR #433, opened 2023-10-06
Added a setting that allows you to change the font of the code

qweme32:
Preview
electron_Msz4JnnuKO

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/StyledWrapper.js
+++ b/packages/bruno-app/src/components/CodeEditor/StyledWrapper.js
@@ -4,6 +4,7 @@ const StyledWrapper = styled.div`
   div.CodeMirror {
     background: ${(props) => props.theme.codemirror.bg};
     border: solid 1px ${(props) => props.theme.codemirror.border};
+    font-family: ${(props) => props.font ? props.font : "default"};
   }
 
   .CodeMirror-overlayscroll-horizontal div,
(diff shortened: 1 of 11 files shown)


qweme32:
I may move this setting to the "Theme" section, that seems more logical to me, what do you think @helloanoop ?

helloanoop:
Nice.

@qweme32 Is it possible to load the fonts from the system and display them as a dropdown to choose from ?

qweme32:
> Nice.
> 
> @qweme32 Is it possible to load the fonts from the system and display them as a dropdown to choose from ?

image
@helloanoop Yes, I found a function that returns all local fonts, but there is one thing - there are too many of them (300+ in my case), you need to do a full search on them. Also the user will use only monospaced fonts in the code. Of course I can build such a form, but is it worth it, I think it's better to leave just the input and not to load ui.

qweme32:
@helloanoop any comments?

helloanoop:
Sure @qweme32 

I have one comment post that we can merge this. Can you update the input box styling to match the theme, You can take a look at the input boxes in `New Request` / `Rename Request` components ?

qweme32:
@helloanoop 

Dark
image

Light
image

helloanoop:
@qweme32 Thank you for your patience. I know you would like to get this merged. I have shipped key bug fixes today. Will visit this tomorrow.

qweme32:
@helloanoop it's been 3 days

newtoallofthis123 (approved):
lgtm

dw-0:
Having to type in a font name seems like bad UX to me. 

> Also the user will use only monospaced fonts in the code.

I don't think that all users want monospaced fonts if they get the chance to change the font.

Is having 300+ fonts installed a regular thing?

BoscoDomingo (approved):
This looks good, can we get it merged? The default font is kinda ugly!

PR #434, opened 2023-10-06
fix: add missing CSS for GraphQL docs explorer

oscarpas:
The GraphQL docs explorer doesn't have any CSS styling which makes very difficult to use, see image:
Skärmavbild 2023-10-06 kl  17 28 32

I'm new to Bruno so not sure if this has worked before.

This PR simply imports the necessary CSS file in the `_app.js` file. Not very familiar with the code base so let me know if there's a better way to do this.

After image:

Skärmavbild 2023-10-06 kl  17 50 09


helloanoop:
Merged!

This is working before, I guess this broken when we fixed a rollup related error in the `graphql-docs` package.

Thanks for taking care of this @oscarpas !

Issue #419, opened 2023-10-06
fix: generate code breaks app if the url is empty


cvmocanu:
If the URL is empty, choosing "Generate code" will show a blank page with "index.html" as the title.
The only way to get back to the collection is to close Bruno and open it again.

HondaChan14:
Hello may I be assigned this issue? 😊

cvmocanu:
I'm sure the author wouldn't mind a PR :)

not-known-person:
Hi @HondaChan14, I'm actively addressing this issue and it's about to be resolved. Feel free to focus on other matters. Thank you for showing your interest.

helloanoop:
This has been fixed by @not-known-person and is now available in `v0.21.0`

PR #436, opened 2023-10-06
fix-/#419

not-known-person:
I fixed white screen issue on clicking generate code on being url empty , 
1. Added toast error showing " URL is required " to generate a code ,
2. Added optional chaining while accessing appropriate properties,
3. Removed undefined variable "str" which was the root cause of #419 ,
4. Checked drafted request when not getting the url from the request object


helloanoop:
Nice! Merged!

Thanks for taking care of this @not-known-person !

Issue #354, opened 2023-10-04
[Feature] Allow configuring headers/authorization for GraphQL schema introspection query


grubersjoe:
Hey. First of all, great app, thanks for sharing!

My issue is that the GraphQL schema introspection is failing, because it requires authentication with a bearer token and currently there seems no way to provide the credentials:

https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/ipc/network/prepare-gql-introspection-request.js

I'm fetching the token in another request and store it in a collection variable then like this:

```
script:post-response {
  bru.setVar('accessToken', res.body.access_token)
}
```

So ideally there would be some setting where you can configure authorization and/or the headers for the introspection query including support for variables. If you like I can also give this a try and open up a PR :).

grubersjoe:
I'm currently working on it and I think it goes well. I'll open up a PR when I'm ready. There are also surely things that needs to be discussed like the data structure or how the UI should look like etc. Currently I've added another tab to the GraphQL request pane to configure headers for the introspection query:

![image](https://github.com/usebruno/bruno/assets/3109364/c411ff1d-3067-4998-9210-247f136389ae)

grubersjoe:
I've opened up a PR that adds the `Auth` tab to GraphQL requests to solve this.

helloanoop:
Hey @grubersjoe 

Many Thanks!
I have merged the PR. There are still some enhancements that need to be implemented.
It'd be great if you could raise a new PR when you find some time.

The [prepare-gql-introspection-request.](https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/ipc/network/prepare-gql-introspection-request.js) still uses mustache js which we no longer use.

We have started using handlebars at all [other places](https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/ipc/network/interpolate-vars.js#L41-L60)

Changes needed
1. Update to use handlebars
2. Interpolate env variables

Bruno allows pulling dotenv varaibles at the root of you collection and accessing them via `{{process.env.secret}}`
See [docs](https://docs.usebruno.com/secrets-management/dotenv-file.html)

The current interpolation does not consider the interpolation of env vars.

oscarpas:
This is a great addition. If you're still working on this @grubersjoe I have another use case that would be great if it's covered:

I need to set custom headers to authenticate the schema request. So either add a "custom header" option to the dropdown on the auth tab. Or use the headers from the headers tab when making the schema req.

[...]

PR #441, opened 2023-10-06
Improve GQL introspection request

grubersjoe:
This fixes the things mentioned here: https://github.com/usebruno/bruno/issues/354#issuecomment-1749094454

> 1. Update to use handlebars
> 2. Interpolate env variables

Additionally, I added one more commit to add the set headers to the introspection request as well as asked here: https://github.com/usebruno/bruno/issues/354#issuecomment-1750191137

One more side note about handlebars: it's only used in two more places now and it should be pretty easy to get rid of this dependency completely.


grubersjoe:
Seems like `package-lock.json` is not up-to-date right now. Added another commit for that.

helloanoop:
Merged. Thanks for taking care of this @grubersjoe

PR #443, opened 2023-10-07
Building electron script in js to support different OS

survivant:
I created a node js script instead of a bash script.  I'm on Windows and running the script in cygwin or wsl is something painful.  Instead  this node js script will works on any OS.

I also generate the distribution based on the OS.  By default it was MAC.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
     "jest": "^29.2.0",
     "pretty-quick": "^3.1.3",
     "randomstring": "^1.2.2",
-    "ts-jest": "^29.0.5"
+    "ts-jest": "^29.0.5",
+    "fs-extra": "^11.1.1"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


helloanoop:
Nice work @survivant !

Need to get a release out tomorrow. Will get this merged after that.

survivant:
please, can you add this label at the end : hacktoberfest-accepted   (it's just for fun.. there are no t-shirt this year :( )

survivant:
there are my first PR for a React project, so maybe there will need a little review, just to be sure that they respect React good pratice.

helloanoop:
Something seems off here.

This got reverted: https://github.com/usebruno/bruno/commit/5ef0de2e4d2e8f4e42e667ae5b7af0e3bebb2754

@survivant if you don't mind, please raise a new PR

PR #447, opened 2023-10-07
minor styling fixes

therealrinku:
Minor styling fixes
- Increased checkbox and radio sizes, I think they were too small.
- Added background color on the modal.


helloanoop:
Neat! Thanks @therealrinku !

Issue #451, opened 2023-10-07
Electron app doesn't store window state/size


jarne:
Surely a minor issue, but implementing it would improve user experience.

**Current state:**

Window state (position, size, etc.) are not kept when closing and re-opening the application

**Way to reproduce:**

1. Resize or move application window

<img width="1102" alt="Bildschirmfoto 2023-10-07 um 12 29 42" src="https://github.com/usebruno/bruno/assets/15860314/79055d20-e7bc-4c5d-b396-bcb2aae08f49">

2. Close app
3. Open again (starts in default size/position)

<img width="1210" alt="Bildschirmfoto 2023-10-07 um 12 29 25" src="https://github.com/usebruno/bruno/assets/15860314/37538a62-44a1-4434-a592-185ae565bd58">

**Should be:**

Store the window state, persist and load when opening the application.

Likely I'll submit a PR for this soon.

helloanoop:
Reopening: https://github.com/usebruno/bruno/pull/452#issuecomment-1755101199

jarne:
completed with #536

PR #452, opened 2023-10-07
Persist window state (size, position)

jarne:
This PR implements issue #451 and adds code to persist the window state of the Electron application, including its size and position so it's starting at the same position and size when opening the application as it was closed. I also added some checks to reset both position or size to default if it's not inside the screen border's (e.g. when switching to a smaller screen).


martinsefcik:
@jarne this implementation is not working correctly in all cases. At least on Windows when window is maximezed then `x` and `y` values are stored with negative values and then during startup reset to `undefined`.

PR #457, opened 2023-10-07
Fix 'between' assertion

Sl-Alex:
This is a small fix for the 'between' runtime assert.

Without this fix, the assertion fails with the message "value is not defined"


helloanoop:

Issue #440, opened 2023-10-06
It's difficult to find how to open web console in Bruno v0.20.0


lared:
Affected version: Bruno v0.20.0, electron

Version 0.18.0 (and 19):

![](https://imgur.com/YX81qmH.png)

Version 0.20.0:

![](https://imgur.com/GEv5Vug.png)

It's *very* unintuitive for me that I need to press alt to see the menu, especially since it's the only place from which I can access developer tools to debug my poorly written script (if I don't remember the keyboard shortcut anyway, which I dont)

lared:
Actually I see this is related to https://github.com/usebruno/bruno/pull/393 - it is **very** unintuitive for me to figure out that I need to press alt to see the menu. I think some other users might feel the same.

helloanoop:
> It's very unintuitive for me that I need to press alt to see

100%

I expected this.
This got introduced in this PR:https://github.com/usebruno/bruno/pull/393#issuecomment-1749117914

> to debug my poorly written script

You're being too modest again :P 
If you are on linkedin, would love to connect with you. I am [here](https://www.linkedin.com/in/anoop-m-d-868099100/)

Here are my thoughts. 
1. Bring this back for now in `v0.21.0` that I plan to ship in an hour.

2. Work on a new PR that allows a user to hide the menubar via app preferences

3. Add another option called Devtools here

![image](https://github.com/usebruno/bruno/assets/9350733/b2137c91-c7cd-4d72-84ed-69f922ad6773)

What do you think?

lared:
I think this is a great plan, thank you!

helloanoop:
> Bring this back for now in v0.21.0 that I plan to ship in an hour.

Shipped in v0.21.0

Remaining
2. Work on a new PR that allows a user to hide the menubar via app preferences
3. Add another option called Devtools

martin-braun:
I'm very sad to figure out that this is behind a paywall, I was eventually getting the Golden Version, but I have trouble to recommend it. DevTools are too essential! -.-

Its-treason:
> I'm very sad to figure out that this is behind a paywall, I was eventually getting the Golden Version, but I have trouble to recommend it. DevTools are too essential! -.-

I think you mean the Browser DevTools, they are not locked behind the golden edition. You can Open them with `Ctrl` + `Alt` + `I` or by clicking on the 3 dots in the top left corner.

The DevTools that are locked behind the Golden Edition include things like: A base64 encoder, JSON to YML.

---

[...]

PR #459, opened 2023-10-07
fix(#440) open dev tools from menu

lared:
Addresses issue in #440 where a common operation, opening dev tools (ex. to debug scripts) was only available from menu, which by default is hidden, which generated some friction in user experience.


helloanoop:
Nice! Thanks @lared !

Issue #265, opened 2023-10-01
folder showing even after deleting 


not-known-person:
folder inside collection doesn't get remove from the list even after deleting , 
![Screenshot 2023-10-01 145005](https://github.com/usebruno/bruno/assets/134089980/77b2971a-a64d-447b-88aa-e254959213af)
I have deleted the Demo folder in the Demo12389 , but it still showing  ,
but restarting the application fixes the issue .

helloanoop:
Hey @not-known-person Thanks for taking caring of this!

The phantom folder issue is only on Windows. This works perfectly fine in mac and linux. add this [comm
I've just added this [commit](https://github.com/usebruno/bruno/commit/fcc12fb089472716328054665d767d7b0ae48e04) so that this fix is applied only on windows.

On Mac and Linux this might end up with multiple redux updates (like not harmful), but still choosing not to do it since rename and delete feature is pretty stable on mac and linux.

Also, another debugging tip. When you bring up the app, you can open developer tools and turn on debugging by executing `window.__IS_DEV__ = true`

On my mac, as you see below in the screenshots I have 3 IPC events. `addDir` when I added the test directory and upon rename - two subsequent events `unlinkDir` and `addDir` which is basically a rename. I am suspecting you are not seeing this in windows. 

![image](https://github.com/usebruno/bruno/assets/9350733/d6132be2-d553-462c-b0aa-22fd5bad9675)

helloanoop:
This has been released in `v0.16.5`
You can download the latest version from the website.

Thanks @not-known-person 

Closing this issue. Please re-open if the issues still persists.

not-known-person:
Thank you so much for your kind words! It's been a pleasure contributing to the project, and I'm glad I could be of help. I look forward to more opportunities to work together and make the project even better.

Issue #251, opened 2023-09-29
renaming folder under collection ends up with a phantom folder


mreiland:
Repro:

1. create a collection named 'test'
2. create a folder named 'inner folder' inside the test collection
3. create a request named 'google' inside the folder named 'inner folder'
4. select the 'elipses' menu next to the folder and select rename
5. rename to 'inner folder (renamed)
6. observe there are now two folders, 1 named 'inner folder' with nothing in it and 1 named 'inner folder (renamed)' with the 'google' request in it.
7. you will not be able to interact with the 'inner folder' in any way.
8. close and restart bruno and the 'inner folder' goes away.

see attached screenshot for evidence.

![bruno-bug](https://github.com/usebruno/bruno/assets/737163/149dd573-5a22-4f8a-9d06-73ef3bd17906)

helloanoop:
Hi @mreiland 

The rename works fine on my machine. I am using Mac x64
Can you let me know which OS you are using ?

not-known-person:
Hey , I am using windows 11 x64 , Same issue is with me

helloanoop:
@mreiland This has been fixed by @not-known-person  in `v0.16.5`
You can download the latest version from the website.

Closing this issue. Please re-open if the issues still persists.

mreiland:
@helloanoop 

Apologies for the delay, I'm on Windows 10 and can confirm this is fixed in the latest version of bruno.

Thank you for such a quick response!

Scotsoo:
I was looking at this last weekend and where I think this fix does indeed fix the issue, I don't think it's the real fix. 
I think the real problem is due to `path.sep` returning the unix `/` rather than the windows `\\` as it's running in a browser environment path.sep gets a bit confused. Meaning that the context doesn't update using the natural triggers. 

See example, this is based on `v0.16.5` on a windows machine.
![electron_Fqr7Km1Hms](https://github.com/usebruno/bruno/assets/6567996/76148c0b-8d77-4891-bd58-5571e8f73a7c)

PR raised: https://github.com/usebruno/bruno/pull/465

Scotsoo:
Yep, that's correct. I'm not seeing path.sep as \\ on windows as it's something to do with the runtime actually running as a browser.
My gif above is recorded on a windows machine

helloanoop:
Closing this issue as this has been fixed in the latest version.

PR #465, opened 2023-10-08
Bugfix/fixes 251

Scotsoo:
Fixes underlying issue with using path.sep in windows: https://github.com/usebruno/bruno/issues/251


packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -24,8 +23,9 @@ import {
 } from 'utils/collections';
 import { parseQueryParams, stringifyQueryParams } from 'utils/url';
 import { getSubdirectoriesFromRoot, getDirectoryName } from 'utils/common/platform';
+import os from 'os';
 
-const PATH_SEPARATOR = path.sep;
+const PATH_SEPARATOR = /Windows/i.test(os.release()) ? '\\' : '/';
```
martinsefcik: You can use `isWindows` function from `utils/common/platform`

Scotsoo: Done

helloanoop:
@Scotsoo Can you reply to this on the main issue ? https://github.com/usebruno/bruno/issues/251#issuecomment-1753674309

helloanoop:
Need some help here.

Can someone test this on windows and confirm?
cc @Its-treason ?

This was the initial fix (https://github.com/usebruno/bruno/commit/fcc12fb089472716328054665d767d7b0ae48e04) for the phantom folder issue that occurred while renaming

The rename perfectly works in linux and mac. The problem is only on windows.
If this PR fixes the issue, then I would remove the fix we did here https://github.com/usebruno/bruno/commit/fcc12fb089472716328054665d767d7b0ae48e04 in this PR

Scotsoo:
> @Scotsoo Can you reply to this on the main issue ? [#251 (comment)](https://github.com/usebruno/bruno/issues/251#issuecomment-1753674309)

Apologies, replied

Its-treason:
So have been testing around a bit and the Path separator is the underlying issue. So when this fix is merged, [`fcc12fb`](https://github.com/usebruno/bruno/commit/fcc12fb089472716328054665d767d7b0ae48e04) can be reverted.

But before we merge this, I would suggest moving the `PATH_SEPERATOR` to `src/utils/common/platform.js` so it is globally in one place. And then remove other `PATH_SEPERATOR` constants in other places like `src/providers/ReduxStore/slices/collections/index.js`.

Scotsoo:
Done! 

After reverted changeset:
electron_nwb4nUSc5B

Scotsoo:
Also deletion works so fixes #265 too
electron_varvRbPR78

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -295,19 +289,10 @@ export const renameItem = (newName, itemUid, collectionUid) => (dispatch, getSta
...
-        // This looks like an issue in our watcher library chokidar
-        // GH: https://github.com/usebruno/bruno/issues/251
-        if (isWindowsOS()) {
-          dispatch(_renameItem({ newName, itemUid, collectionUid }));
-        }
-        resolve();
-      })
-      .catch(reject);
+    ipcRenderer.invoke('renderer:rename-item', item.pathname, newPathname, newName).then(() => {
+      dispatch(_renameItem({ newName, itemUid, collectionUid }))
```
Its-treason: I think the `_renameItem` call can be removed, because this was basically the original fix.

Scotsoo: Removed

helloanoop:
Merged. Thanks @Scotsoo !

PR #466, opened 2023-10-08
feat: added active query, headers, vars and asserts count

therealrinku:
Added active params, headers, vars and assertions count on the request options tab

image


martinsefcik:
@therealrinku Wouldn't it be good to have it also on Vars tab ?

therealrinku:
@martinsefcik 
that makes sense, I added it as well.
thanks!

helloanoop:
Thank you @therealrinku and @martinsefcik !

Merged!

majkinetor:
It would be good to have some kind of marker such as a dot for other things, like non empty Body, Script, Doc...

PR #471, opened 2023-10-09
showing error response in response tab instead of alert message

therealrinku:
- Showing error message in response panel instead of just showing alert message.
- Decreased Loading Overlay height because it was causing to show vertical scrollbar.

before:
Screenshot from 2023-10-09 12-26-43

after:
Screenshot from 2023-10-09 12-27-16

loading overlay was causing the vertical scrollbar to appear:
image


helloanoop:
@therealrinku I have pushed some fixes for the overlay, Can you sync your fork with the latest main and confirm if the scrollbar issue is fixed.

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -117,22 +117,23 @@ export const sendRequest = (item, collectionUid) => (dispatch, getState) => {
...
-            response: null
+            response: errorResponse
           })
         );
-
-        if (err && err.message === "Error invoking remote method 'send-http-request': Error: Request cancelled") {
-          console.log('>> request cancelled');
-          return;
-        }
-
```
helloanoop, line 155: We need to keep above lines and should not show any error if the request was manually cancelled using the cancel button.

packages/bruno-app/src/components/ResponsePane/index.js
```diff
@@ -116,8 +116,13 @@ const ResponsePane = ({ rightPaneWidth, item, collection }) => {
...
       </div>
       <section className={`flex flex-grow ${focusedTab.responsePaneTab === 'response' ? '' : 'mt-4'}`}>
-        {isLoading ? <Overlay item={item} collection={collection} /> : null}
-        {getTabPanel(focusedTab.responsePaneTab)}
+        {isLoading && <Overlay item={item} collection={collection} />}
+
+        {response.isError ? (
+          <span className="mt-5 px-5 text-red-500">{response.error}</span>
+        ) : (
+          getTabPanel(focusedTab.responsePaneTab)
```
helloanoop: Instead of always displaying the error, Can we pass this error to the `QueryResult` component and display it there.

```html
<QueryResult
  item={item}
  collection={collection}
  width={rightPaneWidth}
  data={response.data}
  error={response.error}
  headers={response.headers}
/>
```

That way, the user can switch to the timeline tab and see the request that was sent.

helloanoop (changes requested):

therealrinku:
@helloanoop thanks for the feedbacks, 
I have made the changes you requested. you can check it out when you have some time.

helloanoop:
Thanks @therealrinku !
Merged!

PR #475, opened 2023-10-09
Fix assertion value modification

Sl-Alex:
This PR fixes the following bug in GUI:

> Assertion operator falls back to "eq" when user starts changing assertion value.
> 
> Steps to reproduce:
> 1. Create any assertion that requires value, except "eq" (for example "neq")
> 2. Start changing the value
>
> Expected: Operator doesn't change
> Actual: Operator falls back to "eq"


helloanoop:
Thanks for taking care of this @Sl-Alex !!

Merged.

Issue #474, opened 2023-10-09
[Bug] GraphQL query validation not working


grubersjoe:
Hey. When writing GraphQL I would expect to see errors like referencing non-existing fields in my query. For example here it should show a red squiggly underline for `unknownField` since this field does not exist for a `Machine`:

![image](https://github.com/usebruno/bruno/assets/3109364/394d64ef-6d73-434b-a58b-ce3c40dc6d72)

I've looked a bit at the [`<QueryEditor />`](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/components/RequestPane/QueryEditor/index.js) component and it seems like this should be handled by Codemirror but errors are not shown. Not sure what's wrong. Does anyone have an idea?

Edit: found the issue, there's a PR now

grubersjoe:
Seems like Codemirror initialization is not correct. If I add the respective `codemirror-graphql` imports in the QueryEditor file it works:

![image](https://github.com/usebruno/bruno/assets/3109364/34417432-2f25-43fb-ba3a-5b6977e8e94b)

grubersjoe:
Fixed by #480

helloanoop:
Thanks for working on this @grubersjoe 
This is now available in `0.22.1`

PR #480, opened 2023-10-09
Fix GraphQL validation

grubersjoe:
Hey,

there have been two issues with GraphQL validation (issue https://github.com/usebruno/bruno/issues/474):

* the `codemirror/addon/lint/lint` module was not imported, consequently the GraphQL linting/validation was not enabled
* the content security policy did not allow to load base64 inline images, required for the red squiggly underlines and the little icon in the popover

image

I have also allowed myself to make the CSP stricter and successfully tested it in the development and production builds. Now the `unsafe-*` policies will only be used during development where they are needed.

### DIFF
--- a/packages/bruno-app/src/pageComponents/Index/index.js
+++ b/packages/bruno-app/src/pageComponents/Index/index.js
@@ -14,24 +14,25 @@ const SERVER_RENDERED = typeof navigator === 'undefined' || global['PREVENT_CODE
 if (!SERVER_RENDERED) {
   require('codemirror/mode/javascript/javascript');
   require('codemirror/mode/xml/xml');
-  require('codemirror/addon/scroll/simplescrollbars');
+  require('codemirror/addon/comment/comment');
+  require('codemirror/addon/dialog/dialog');
+  require('codemirror/addon/edit/closebrackets');
   require('codemirror/addon/edit/matchbrackets');
   require('codemirror/addon/fold/brace-fold');
   require('codemirror/addon/fold/foldgutter');
-  require('codemirror/addon/mode/overlay');
   require('codemirror/addon/hint/show-hint');
-  require('codemirror/keymap/sublime');
-  require('codemirror/addon/comment/comment');
-  require('codemirror/addon/edit/closebrackets');
+  require('codemirror/addon/lint/lint');
+  require('codemirror/addon/mode/overlay');
+  require('codemirror/addon/scroll/simplescrollbars');
+  require('codemirror/addon/search/jump-to-line');
   require('codemirror/addon/search/search');
   require('codemirror/addon/search/searchcursor');
-  require('codemirror/addon/search/jump-to-line');
-  require('codemirror/addon/dialog/dialog');
+  require('codemirror/keymap/sublime');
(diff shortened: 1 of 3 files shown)


packages/bruno-app/src/pageComponents/Index/index.js
```diff
@@ -14,24 +14,25 @@ const SERVER_RENDERED = typeof navigator === 'undefined' || global['PREVENT_CODE
...
+  require('codemirror/addon/edit/closebrackets');
   require('codemirror/addon/edit/matchbrackets');
   require('codemirror/addon/fold/brace-fold');
   require('codemirror/addon/fold/foldgutter');
-  require('codemirror/addon/mode/overlay');
   require('codemirror/addon/hint/show-hint');
-  require('codemirror/keymap/sublime');
-  require('codemirror/addon/comment/comment');
-  require('codemirror/addon/edit/closebrackets');
+  require('codemirror/addon/lint/lint');
```
grubersjoe, line 24: This is the added line. Sorry for the noise, I sorted all the imports...

packages/bruno-electron/src/index.js
```diff
@@ -13,12 +13,16 @@ const { loadWindowState, saveWindowState } = require('./utils/window');
...
-	script-src * 'unsafe-inline' 'unsafe-eval';
-	connect-src * 'unsafe-inline';
-	form-action 'none';
-`);
+const contentSecurityPolicy = [
+  isDev ? "default-src 'self' 'unsafe-inline' 'unsafe-eval'" : "default-src 'self'",
+  "connect-src 'self' https://api.github.com/repos/usebruno/bruno",
+  "font-src 'self' https://fonts.gstatic.com",
+  "form-action 'none'",
+  "img-src 'self' blob: data:",
```
grubersjoe, line 21: This is the relevant line for inline image loading

helloanoop:
Nice!

@grubersjoe some comments.

```javascript
isDev ? "default-src 'self' 'unsafe-inline' 'unsafe-eval'" : "default-src 'self'",
```
Would the policy prevent users from loading html data.
image

@Its-treason had added support for html preview in https://github.com/usebruno/bruno/pull/282
@Its-treason do you see this breaking preview

grubersjoe:
Ah good point with the HTML preview. Maybe we want to overthink the `data:text/html` approach? Non ASCII characters like German umlauts (ä, ü, etc) are also rendered incorrectly this way. Instead one could do this and you would not be required to inject the base URL:

image

However, the page will be reload every time you navigate to the preview tab. Perhaps that is not inherently bad.

grubersjoe:
It's a bit weird. On the wikipedia page I also don't have the issue, but here for google.de I do see the broken characters (on the `main branch):

image

For some reason the timeline also shows this:

> < content-type: text/html; charset=ISO-8859-1

However if I check out this header in `curl` the charset in the response header is `utf-8`...

---

Anyway, back to this PR: HTML for pages like usebruno.com still render fine with these changes.

helloanoop:
Thanks @grubersjoe @Its-treason 

Merging this!

Issue #473, opened 2023-10-09
Impossible to use built-in AJV library 


arnaduga:
In a request script or test, impossible to import the AJV library:

`const Ajv = require("ajv")` 
leads to an error message 
`Error invoking remote method 'send-http-request': VMError: cannot find module 'ajv'`

lared:
I'll take it

arnaduga:
Seems to be good in UI version 0.24+
Thanks

However I opened the #612 for a similar problem with the bru-cli

PR #488, opened 2023-10-09
fix(#473) add Ajv to the runtime

lared:
Simply wasn't added to the runtime.


helloanoop:
That was fast!

Merged!

PR #490, opened 2023-10-09
fix: remove unwanted scrollbar

snippetkid:
fixes https://github.com/usebruno/bruno/issues/491


snippetkid:
After applying the fix

https://github.com/usebruno/bruno/assets/16718733/12a114f2-0117-405b-b806-a3fca02f7fc1

PR #495, opened 2023-10-09
Fix #437 (start/endWith assertions)

### LINKED ISSUE #437: The "startWith" and "endWith" assertions do not work
image

assert {
  res.status: 200
  res.body.data.url: startsWith "https://gateway.test.com"
}

https://github.com/usebruno/bruno/blob/6db36513c50a7af2d4c58b35d017db81ff0ca43b/packages/bruno-js/src/runtime/assert-runtime.js#L265

I don't see those assertions in the chai library but in the chai-string library: https://www.chaijs.com/plugins/chai-string/

Also, apparently the syntax should be : expect('abcdef').to.startsWith('abc');

Sl-Alex:
This PR adds and enables "chai-string" plugin, making "startWith"/"endWith" assertions work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5600,6 +5600,14 @@
         "node": ">=4"
       }
     },
+    "node_modules/chai-string": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/chai-string/-/chai-string-1.5.0.tgz",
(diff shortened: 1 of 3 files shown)


Sl-Alex:
Should I also push my auto-generated ```package-lock.json``` to make unit tests work?
I'm not sure if it is really needed

grubersjoe:
@Sl-Alex yes, you should also commit the updated `package-lock.json` whenever dependencies change. More background here: https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json

Sl-Alex:
@grubersjoe Thank you, updated it

helloanoop:
Nice.

PR #504, opened 2023-10-10
feat(#354): Use collection level auth if available for introspection request

grubersjoe:
With https://github.com/usebruno/bruno/commit/1ce8d707f1b4cc461235333174b3bdcb290c9c40 merged I think it makes sense to reuse the same logic for the introspection request. Use the collection level authentication if possible, else use request level authentication.


helloanoop:
Nice!

Thanks for taking care of this @grubersjoe !

PR #506, opened 2023-10-10
fix(#505): fixed displaying shortcuts based on OS

### LINKED ISSUE #505: [Bug] Shortcuts should be displayed based on OS type
The following shortcuts on request tab should be displayed based on operating system. So e.g. `Cmd + Enter` on Mac and `Ctrl + Enter` on other OS type.

image

martinsefcik:
Fixes #505


helloanoop:
Nice!

PR #512, opened 2023-10-10
fix: add request and folder modal validation fixes

therealrinku:
- Fixed Name Validation in Add Request and Add Folder Modals, previously it wasn't trimming the string
- Fix validation message on Add folder modal


helloanoop:

Issue #485, opened 2023-10-09
Middle click on tab to close the tab


kpau:
Hello, the Bruno app is really great. Thanks!

I have one small request: Can you add support for closing the tab with mouse middle click?

Thanush19:
can u describe the issue with some more more clearence, like send adequete screenchots of the issue @kpau ....

kpau:
Currently we can close the tabs in the GUI client with the "X" button or with CTRL+W.
It would be useful if we could close the tabs when clicking them with the middle button of the mouse. The same way we can close the tabs in a web browser or VS Code.

![image](https://github.com/usebruno/bruno/assets/1583751/a86484b6-76c0-4690-b048-6fe67b1fff6b)

tibormichel1:
Hello can i work on this issue ?

tibormichel1:
Hello can i work on this issue ?

Sahilm416:
@kpau will you please explain where the middle click event should happen , anywhere in main content section or just by pressing " X " on the tab ??

jakoss:
Anywhere on the tab, it's pretty much default for any tab application these days

kpau:
As @jakoss said, the tab should be closed by clicking anywhere on it with the middle button.
You should be able to close any tab in that way, not only the active one.
This behavior can be seen in any web browser and in VS code (probably most other apps with tabs too).
Thanks for working on it 🙂

ravermeister:
Any news on this? Isn't it a quick win?

shivamlikhar29:
Is this issue fixed ?

ravermeister:
No

Oxymoron290:
This conversation has seen various inputs and attempts at clarification, but it seems we've yet to converge on a solution that has been merged into the main codebase.

The request is to enable users to close tabs within the Bruno app by clicking anywhere on the tab with the middle mouse button, mirroring the functionality found in most web browsers and applications with tabbed interfaces, such as Visual Studio Code. This feature is not just a convenience but has become an expected behavior for users interacting with tabbed interfaces.

The middle mouse button's use for closing tabs has evolved into an emergent convention, as noted by Jeff Atwood in his [March 2008 blog post](https://blog.codinghorror.com/what-should-the-middle-mouse-button-mean/), highlighting its adoption in tabbed user interfaces. While acknowledging that there isn't a universal standard, the consistency in user experience across major platforms and applications suggests a strong user expectation for this behavior.

[...]

PR #514, opened 2023-10-10
Feat (#485) : added feature to close the active tab with middle mouse click

Sahilm416:
-> feat #485 
-> Added the functionality to close the active tab with middle mouse click
-> Middle click functionality is added inside the whole main section

### DIFF
--- a/packages/bruno-app/src/components/RequestTabPanel/index.js
+++ b/packages/bruno-app/src/components/RequestTabPanel/index.js
@@ -16,9 +16,11 @@ import RunnerResults from 'components/RunnerResults';
 import VariablesEditor from 'components/VariablesEditor';
 import CollectionSettings from 'components/CollectionSettings';
 import { DocExplorer } from '@usebruno/graphql-docs';
-import { handleCloseClick } from 'components/RequestTabs/RequestTab/index';
 import StyledWrapper from './StyledWrapper';
-
+import { closeTabs } from 'providers/ReduxStore/slices/tabs';
+import ConfirmRequestClose from 'components/RequestTabs/RequestTab/ConfirmRequestClose/index';
+import { saveRequest } from 'providers/ReduxStore/slices/collections/actions';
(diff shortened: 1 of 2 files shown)


packages/bruno-app/src/components/RequestTabPanel/index.js
```diff
@@ -150,7 +155,7 @@ const RequestTabPanel = () => {
       <div className="pt-4 pb-3 px-4">
         <QueryUrl item={item} collection={collection} handleRun={handleRun} />
       </div>
-      <section className="main flex flex-grow pb-4 relative">
+      <section className="main flex flex-grow pb-4 relative" onMouseDown={(e) => MiddleClickClose(e)}>
```
helloanoop: Please rename `MiddleClickClose` as `handleMouseDown`

You can add a comment in the callback like this.
You also don't have to write an explicit return.
 
```javascript
const handleMouseDown = (e) => {
  // middle click
  if (e.button === 1) {
    handleCloseClick(e, focusedTab, dispatch);
  }
};
```

packages/bruno-app/src/components/RequestTabPanel/index.js
```diff
@@ -16,7 +16,7 @@ import RunnerResults from 'components/RunnerResults';
 import VariablesEditor from 'components/VariablesEditor';
 import CollectionSettings from 'components/CollectionSettings';
 import { DocExplorer } from '@usebruno/graphql-docs';
-
+import { handleCloseClick } from 'components/RequestTabs/RequestTab/index';
```
helloanoop: Let'd not wire components like this.
You need to use a redux action.

helloanoop (changes requested):

helloanoop:
Also @Sahilm416 We have a functionality where - if a user is editing something, upon clicking in close we show an alert modal requesting to either save or cancel the draft.

The middle click implementation should ensure that this flow is not broken.

Sahilm416:
Thanks for the review !

Issue #500, opened 2023-10-10
Copy URL with query string does not parse


arnaduga:
As part of the migration from Postman to Bruno, I copy/paste URL of my queries.

However, when I paste an URL with query string, like `https://httpbin.org/anything?q=test`, it does not parse and does not create an entry the query tab.

As soon as I edit the URL field or the query value both (url and query list) are correct.

helloanoop:
@arnaduga This as been fixed by @Its-treason in `v0.23.0`

premeaswaran:
@helloanoop @Its-treason Can this issue be closed if it is implemented already?

PR #515, opened 2023-10-10
fix(#500): Parse query url on request creation

Its-treason:
https://github.com/usebruno/bruno/assets/39559178/6f5ca073-d8f4-40d4-aec5-64eefd41b632


helloanoop:

Thanks for the video diff :)

PR #516, opened 2023-10-10
Add support for SPARQL bodies

dozed:
This PR adds support for [SPARQL](https://en.wikipedia.org/wiki/SPARQL) bodies:

SPARQL bodies are integrated into bruno-lang v2, but not in v1, is this ok or should v1 also be supported?

Example query:

```sparql
## All predicates, ordered by number of subjects
SELECT ?predicate (COUNT(DISTINCT ?subject) AS ?count) WHERE {
  ?subject ?predicate ?object .
}
GROUP BY ?predicate
ORDER BY DESC(?count)
LIMIT 10
```


helloanoop:
Fantastic !!

Thank you @dozed for adding this in Bruno!

PR #518, opened 2023-10-10
Fixed error handling

dozed:
When you try to do a request to an invalid URL, e.g. `https://`, then the following error was shown:

With this fix the actual error is shown:


helloanoop:
Thanks for taking care of this @dozed !

Issue #359, opened 2023-10-04
[Bug] Preview for PNG files results in white screen/crash


bholm:
When making a request that returns a PNG, the application freezes at a white screen and I need to restart it.

Steps to reproduce:
- To reproduce this issue, you can simply try sending a request to `https://github.com/usebruno/bruno/blob/main/assets/images/landing-2.png?raw=true`.

Expected result:
- To see the image in the preview pane.

helloanoop:
Reopening: https://github.com/usebruno/bruno/pull/519#issuecomment-1756958177

PR #519, opened 2023-10-10
feat: shows image if url return a image

Nikhil569:
fixes: #359

Below is a preview after the change:
fix

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/QueryResult/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/QueryResult/index.js
@@ -45,6 +45,10 @@ const QueryResult = ({ item, collection, data, width, disableRunEventListener, h
       return safeStringifyJSON(data);
     }
 
+    if (mode.includes('image')) {
+      return item.requestSent.url;
+    }
+
     // final fallback
     if (typeof data === 'string') {
       return data;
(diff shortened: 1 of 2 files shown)


helloanoop:
Awesome !!

Thanks @Nikhil569 for taking care of this!

martinsefcik:
@helloanoop 
If I understood these changes correctly new additional request is done to show image. But do we want to do 2 requests in case response content is image ? We have already image data in response body, so we should construct data URL for it and show image based on this data URL and not to do addiational call.

This two request approach is also not good because the second request to download and show image is always GET so it ignores configured request method, also another configuration of original request (like headers, proxy, ...) is ignored. What if endpoint is returning image only if I use `Accept: image/png` header and otherwise it returns json with some image metadata. In such case original request returns png but this second request in preview tab downloads json. Also what if image is dynamically returned based on time or based on request counts, I can retrieve one image in initial call and totally different image in second call. So we display something else as was the response in initial call.

helloanoop:
@martinsefcik I agree with you. We should improve this to render the image that came back instead of making a new call. Reopening the ticket.

martinsefcik:
@helloanoop 
Question is how to get raw response data from axios response? Is it even possible if axios `resposeType` request parameter is set to json (default value) and `responseEncoding` is utf8 (default value)? Because if response data are binary then data in axios `response.data` are a bit bigger and it does not correspond to the binary data retuned by endpoint. I think it is happening because of these two parameters and because some transformation to utf8 is happening there and the data are somehow incorrectly stored in `response.data`.  I think that axios instance used by Bruno should use `responseType = stream` or `arraybuffer` or something general and then it should be manually converted to json object only in cases when for example response content type is json otherwise `response.data` object should contain binary data without any transformation.

I see that @DivyMohan14 faced the same problem here in PR #528 and he had to change `responseType` to `arraybuffer` and I was also thinking how to implement feature which allows to save response data as file, but for example for PDF I was not able to achieve that when `responseType` and `responseEncoding` are set to default values and PDF was somehow malformed.

Issue #497, opened 2023-10-10
Response size always seems to be 0B


therealrinku:
![image](https://github.com/usebruno/bruno/assets/76877078/df23d8d3-2529-41c9-9edd-19f55bf8ed8b)

snippetkid:
I'd like to work on this, @helloanoop can you please assign this to me?

PR #520, opened 2023-10-10
Show response size even if content-length header isn't present

snippetkid:
fixes https://github.com/usebruno/bruno/issues/497

image


helloanoop:
Nice!

Thanks for taking care of this @snippetkid !

PR #526, opened 2023-10-11
Add "Prettify GraphQL" button

j0k3r:
The goal is to achieve the same behavior from Insomnia which allow to format a GraphQL query using prettier ([see](https://github.com/Kong/insomnia/blob/264177b56fcf0a663bebfc8325eb95adb2a2f70e/packages/insomnia/src/ui/components/editors/body/graph-ql-editor.tsx#L260-L266)).

I moved the `prettier` deps from `devDependencies` to `dependencies` because it's now used within the application. I was forced to import `prettier/standalone` & `prettier/parser-graphql` ([as it is explained here](https://prettier.io/docs/en/browser.html)) otherwise I got that error:

> Couldn't resolve parser "graphql". Parsers must be explicitly added to the standalone bundle.

Things I'm not sure:
1. I used `this.editor.setValue()` to update the GraphQL query, is there a better way or it's ok?
2. The button is _badly rendered_ for now (I'm really not sure about all the CSS classes). I tried to put it in the same place as Insomnia (right below the code editor) but I failed. So it is defined above the editor. I think we can find a better place.
   
   image

In case of error because the GraphQL query is badly formatted, I raise a toast error:
image


helloanoop:
@j0k3r looks good to me, Can you resolve the package-lock conflicts?

I don't think you could hand resolve them, unless there is a better way - you basically sync your fork with the main, and delete package-lock.json and recreate it.

j0k3r:
@helloanoop conflicts resolved. Looks like moving `prettier` doesn't generate change in `package-lock.json` anymore 🤷‍♂️

mvmn:
Why not use the same prettier to also prettify JSON request body?

j0k3r:
@helloanoop any chance to get this merged?

grubersjoe:
> It's already done here: #604

I think the linked PR only handles REST requests and this PR is about formatting GraphQL queries which is not supported yet. I would also appreciate this feature. The added logic in the other PR can also surely be reused.

tkstang:
@j0k3r looks like there are conflicts again

j0k3r:
Well, I can handle the rebase.

j0k3r:
But looks like some upstream tests are failing..

helloanoop:
Merged.
Appreciate the patience on this PR. Thank you @j0k3r !

Issue #507, opened 2023-10-10
[Feature Request] Add support for Console Logs with DEBUG level in scripts


martinsefcik:
As part of #190 (Console Logs in scripts should be written into developer console) there was added possibility to log messages from scripts, but DEBUG level (`console.debug()`) was not added. I'm not sure it was intentionally or not, but it would fine to have this level also supported.

Currently request with `console.debug()` in script fails:
![image](https://github.com/usebruno/bruno/assets/1073299/9946732e-0b49-41c6-9fbd-c29a577d5b3e)

If you agree with adding it I can prepare PR.

martinsefcik:
It was implemented in #527

PR #527, opened 2023-10-11
Added support for debug level for console logging in scripts

martinsefcik:
Implements #507


helloanoop:

PR #531, opened 2023-10-11
Fixed application crash on renaming request names by changing cases

Kavinkumar07:
Fixed the issue `Rename dialog ignores capitalisation changes`.


Its-treason:
Hey @Kavinkumar07, thank you for taking your time to look into this issue!

I noticed that, environments also have the same problem as requests when renaming the casing. I think the fix should be the same, could you also look into that? If you don't have time or there's another issue with environments, you can just skip this.

Kavinkumar07:
Hi Its-treason, 
I have made the requested changes. Please review it.

Its-treason:
Hey @Kavinkumar07,

so I see that you updated the check everywhere, but I don't think this is necassary, because the `fs.existSync` is only a problem when renaming i think.

I written a review comment yesterday, but this doesn't seem to be visible. Heres what i wrote:

I think it's better to put this function in `src/util/filesystem.js`.

Also check if the try / catch is really needed. If it is needed, you can remove the `(error)` from the catch, because error is not used.

Second Comment:

After testing this is noticed that now, requests with a different casing can override each other


https://github.com/usebruno/bruno/assets/39559178/2805ff4e-ae63-4880-9255-95203ced411e



Maybe this could be a better fix:

```js
function fileExistsWithCase(newFilePath, oldFilePath) {
  const newFileName = path.basename(newFilePath);
  const oldFileName = path.basename(oldFilePath);
  // Case only changed, allow renaming / overriding. Fix for case-insensitive filesystems
  if (newFileName.toLowerCase() === oldFileName.toLowerCase()) {
    return false;
  }

  return fs.existsSync(newFilePath);
}
```

Kavinkumar07:
Fixed this by following the solution proposed. Please review it

Its-treason:
Hey @Kavinkumar07, I reviewed and tested your changes. Everything seems to work as expected now, good work!

But I have one little thing: I think there was a misunderstanding with the path I wanted you to move your new function. I wanted it to be moved here: https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/utils/filesystem.js. But this is my mistake, I should have linked the correct path in my previous comment.

When this is changed, we can merge your PR! Nice job!

Kavinkumar07:
Changed the file to appropriate path. Thanks. 😊

martinsefcik:
Shouldn't it be handled differently per OS type (case sensitive vs. case insensitive) ? 
For example on Linux i can have two requests with file names `a.bru` and `A.bru` and If I try to rename `a.bru` to `A.bru` with current PR implementation then I override existing request, because `fileExistsWithCase` returns `false`. Or am I wrong ?

Also method name `fileExistsWithCase` is not good from clean and easy code readable perspective. It's confusing why it has two arguments. I would not expect that from method which is checking if file exists or not. And also it is confusing that it returns `false` in case file exists but has different casing. I wouldn't expect it from method witch name starts with `fileExists`

So I probably It could be implemented something like this (not tested);
```js
      const changingCasingOnlyOnCaseInsensitiveOS = (isWindows || isMac) && oldPath.toLowerCase() === newPath.toLowerCase();
      if (fs.existsSync(newPath) && !changingCasingOnlyOnCaseInsensitiveOS) {
        throw new Error(`path: ${oldPath} already exists`);
      }
```

or even better, as Mac can be running in both modes (case sensitive or case insensitive):
```js
      const changingCasingOnlyOnCaseInsensitiveOS = (isWindows || isMacInCaseInsensitiveMode) && oldPath.toLowerCase() === newPath.toLowerCase();
      if (fs.existsSync(newPath) && !changingCasingOnlyOnCaseInsensitiveOS) {
        throw new Error(`path: ${oldPath} already exists`);
      }
```
but frankly speaking I have no idea yet how to implement `isMacInCaseInsensitiveMode` function and check in which configuration Mac is running.

And also name of this PR a bit confusing, because I guess application is not crashing during renaming. It just do nothing.

martinsefcik:
Also it looks like there is very closely related another bug. If I rename request's bru file for example from `Update request.bru` to `Update.bru` and I don't touch `meta.name` in bru file, so it stays `Update request` then If I want to rename it in Bruno app to `Update` it is not allowed because File already exists error in console.

So probably this condition when to show this error should be extended by `oldPath !== newPath`

Zomzog:
Hi @Its-treason 
I've seen your comment on #1171 for the args that it is a breaking change for Linux, I can understand it. But in another hand, one of the main benefits I see with Bruno is to be able to share collections with teammates and we have both Linux, Mac, and Windows users. 
Is it more important to use the full naming capacity of one OS or have (in my opinion) a smoother user experience?

Nahuel92:
Hi, I have a silly question. Is there anything preventing the UI from grabbing the request's name from the metadata instead of the filename?

Its-treason:
Ater some testing i came up with this function:

```js
const canRenameFile = (newFilePath, oldFilePath) => {
  const newFileExists = fs.existsSync(newFilePath);
  if (!newFileExists) {
    // File does not exists, so wen can safely rename
    return true;
  }

  // Try to resolve both paths, if they reference the same file. We are on a case-insentive filesystem like NTFS and are trying to rename a file
  return fs.statSync(newFilePath).ino === fs.statSync(oldFilePath).ino;
};
```
It tries to check if both reference the same file but with different casing, by using `stat` on both files and then comparing the Inode number.

Might need some testing on different use cases, e.g. Mac, Network Drive etc.

helloanoop:
This issue has been fixed by in PR #3171 and will be released in the upcoming `v1.32.0` release next week.
Closing this PR.

PR #532, opened 2023-10-11
Update GitHub workflows

j0k3r:
- jump to `actions/checkout` v4 (latest version)
- retrieve node version from NVM instead of hard-coded
- add a new job to run prettier (in case people skip pre-commit hook)


helloanoop:
Thanks @j0k3r !

Merged!

PR #536, opened 2023-10-11
Fix: save window maximized state

jarne:
Fix for https://github.com/usebruno/bruno/pull/452#issuecomment-1755101199: save if the window is maximized in the window state and maximize on start if it was before.

### DIFF
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -67,8 +67,14 @@ app.on('ready', async () => {
   mainWindow.loadURL(url);
   watcher = new Watcher();
 
-  mainWindow.on('resize', () => saveBounds(mainWindow));
-  mainWindow.on('move', () => saveBounds(mainWindow));
+  const handleBoundsChange = () => {
+    if (!mainWindow.isMaximized()) {
+      saveBounds(mainWindow);
+    }
+  };
+
+  mainWindow.on('resize', handleBoundsChange);
+  mainWindow.on('move', handleBoundsChange);
 
   mainWindow.on('maximize', () => saveMaximized(true));
   mainWindow.on('unmaximize', () => saveMaximized(false));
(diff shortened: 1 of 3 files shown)


jarne:
@martinsefcik it's working for me now on Ubuntu/macOS (but those ones also didn't save negative positions for me), I wasn't able to test on Windows as I don't have a Windows environment here. If you have some time, it'd be great if you could confirm the bug fixed.

Its-treason:
Just tested it on windows, works there too!

packages/bruno-electron/src/index.js
```diff
@@ -63,8 +67,11 @@ app.on('ready', async () => {
   mainWindow.loadURL(url);
   watcher = new Watcher();
 
-  mainWindow.on('resize', () => saveWindowState(mainWindow));
-  mainWindow.on('move', () => saveWindowState(mainWindow));
+  mainWindow.on('resize', () => saveBounds(mainWindow));
+  mainWindow.on('move', () => saveBounds(mainWindow));
```
martinsefcik: I would do small enhancement here
```js
  mainWindow.on('resize', () => {
    if (!mainWindow.isMaximized()) { 
      saveBounds(mainWindow)
    }
  });
  mainWindow.on('move', () => {
    if (!mainWindow.isMaximized()) {
      saveBounds(mainWindow)
    }
  });
```
(probably some code cleanup would be fine to minimize duplicity in my snippet)

So window bounds would be saved only in case window is not maximized. And why?
Imagine case that you resize and move window to x=0, y=0, width=100, height=100. This bounds are saved. Then you maximize window. Maximation flag is saved, but also bounds are replaced with some values which are not important when window is maximized. Then you close app. After starting app again I would expect it starts as maximized, but in case I unmaximize it then bounds will be restored to x=0, y=0, width=100, height=100. If you apply my change then It will happen. If not then after unmaximize action window position is restored to default value because it has negative x, y values on Windows and on Linux and Mac (in case there are no negative values after maximize action) window position and size I guess would be very similar if not the same as when window is maximized.

jarne: Thanks for your detailed explanation, makes absolutely sense, I'll change the implementation to work this way.

Issue #533, opened 2023-10-11
[Bug] HTTP Method name and request name collidates in sidebar


martinsefcik:
Longer method names like DELETE, OPTIONS collides in sidebar with request name or cause that request name is indented more then it should be. This behavior depends on sidebar width.

Probably It would be better for this longer method names to use just shortcuts like DEL, OPT the same way as it is in other popular clients as Postman, ...

![image](https://github.com/usebruno/bruno/assets/1073299/8bdd2e0e-30d5-4e4d-9374-4659be966fbd)
![image](https://github.com/usebruno/bruno/assets/1073299/74382c52-2522-4918-be69-b1d6903195fa)

theavitw:
@martinsefcik can I work on this?

martinsefcik:
@theavitw there is already #538

PR #538, opened 2023-10-11
Shorten method name in sidebar

Nikhil569:
fixes: #533 

If a method length in greater than 4, sidebar shows only starting 3 character

fix 533(2)

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/RequestMethod/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/RequestMethod/index.js
@@ -23,7 +23,9 @@ const RequestMethod = ({ item }) => {
   return (
     <StyledWrapper>
       <div className={getClassname(item.request.method)}>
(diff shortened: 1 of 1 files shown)


joao-silva1007:
Wouldn't it be more user friendly if we just scaled the text down a bit and aligned it to the right? This way the request names are aligned and the user still sees the full method name

martinsefcik:
@Nikhil569 In your solution I would replace greater than 4 chars condition with greater than 5 to have `PATCH` displayed as full name, because `PAT` shortcut is not very good, but I also like @joao-silva1007 proposal with smaller text for request method as for me it is less important information as request name

Nikhil569:
@martinsefcik I think it is possible to showcase the full method name without scaling down the request method text size(i think it looks weird if text size are not same), but the request name would show only starting 8 characters(if `OPTIONS` method is selected).

What are your thoughts about this?

fix 533(3)

Nikhil569:
I am closing this to create a new pull request, containing a better fix for #533

Issue #529, opened 2023-10-11
Shortcut Ctrl+W close the window, not the tab


arnaduga:
When pressing the Ctrl+W, the whole Bruno application is closed, without saving the current request edition.

Expected behavior: whatever the object closed (window or tab), it should be offered to save current edit.

Ideally (but maybe larger discussion), shortcut Ctrl+M should close tab only

martinsefcik:
For me it is happening only if focus is in Search edit box located in sidebar, when focus is somewhere else this shortcut does not close whole application.  Is it happening the same way for you @arnaduga ?

snippetkid:
Not only the sidebar search, whenever **any** text field is in focus and you press <kbd>Ctrl</kbd> + <kbd>W</kbd> the whole application gets closed. I've tried with the following text fields

- Sidebar Search
- Request URL
- Rename Request

helloanoop:
Closing this issue as it has already been resolved in Bruno. The `Ctrl+W` shortcut now closes the tab

PR #540, opened 2023-10-11
fix(#529): Fix Ctrl+W closes Bruno

Its-treason:
The default shortcut to close in the menu is Ctrl+W, I changed it to Ctrl+Shift+Q because Firefox uses this shortcut for closing

### DIFF
--- a/packages/bruno-electron/src/app/menu-template.js
+++ b/packages/bruno-electron/src/app/menu-template.js
@@ -44,7 +44,7 @@ const template = [
   },
   {
     role: 'window',
-    submenu: [{ role: 'minimize' }, { role: 'close' }]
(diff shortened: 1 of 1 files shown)


martinsefcik:
Question is what is the purpose for this `Window` submenu with `Minimize` and `Close` actions if these actions are in any OS type available in application title bar. For me this submenu is useless.

Its-treason:
> Question is what is the purpose for this `Window` submenu with `Minimize` and `Close` actions if these actions are in any OS type available in application title bar. For me this submenu is useless.

I think most of the options in the menu are Electron defaults and not very useful. But some users probably expect them to be there, because other Electron applications also have these options in the menu. Having them there doesn't hurt.

vkopanja:
Can't wait for this to be merged, so annoying that I close Bruno every time I press `Cmd + W`. Muscle memory is a bit*h 😄

PR #548, opened 2023-10-12
Feat (#485) : added feature to close the active tab with middle mouse click

Sahilm416:
-> **feat #485** 
-> Added the functionality to close the active tab with middle mouse click
-> Middle click functionality is added inside the whole main section

### DIFF
--- a/packages/bruno-app/src/components/RequestTabPanel/index.js
+++ b/packages/bruno-app/src/components/RequestTabPanel/index.js
@@ -16,6 +16,10 @@ import RunnerResults from 'components/RunnerResults';
 import VariablesEditor from 'components/VariablesEditor';
 import CollectionSettings from 'components/CollectionSettings';
 import { DocExplorer } from '@usebruno/graphql-docs';
+import { closeTabs } from 'providers/ReduxStore/slices/tabs';
+import ConfirmRequestClose from 'components/RequestTabs/RequestTab/ConfirmRequestClose/index';
+import { saveRequest } from 'providers/ReduxStore/slices/collections/actions';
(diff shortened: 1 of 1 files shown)


packages/bruno-app/src/components/RequestTabPanel/index.js
```diff
@@ -147,10 +167,43 @@ const RequestTabPanel = () => {
...
+                console.log('err', err);
+              });
+          }}
+        />
+      )}
       <div className="pt-4 pb-3 px-4">
         <QueryUrl item={item} collection={collection} handleRun={handleRun} />
       </div>
-      <section className="main flex flex-grow pb-4 relative">
+      <section className="main flex flex-grow pb-4 relative" onMouseDown={(e) => handleMouseDown(e)}>
```
martinsefcik, line 206: I tried the following applications (Firefox, Edge, Chrome, Posman) and all of them are closing tabs on middle click up event and not down.

Sahilm416:
@martinsefcik  so i should trigger it onMouseUp right ?

Sahilm416:
@martinsefcik will you please elaborate, we just need to add this functionality to everywhere in Main section or just on " X " button to close with OnMouseUp ??

jakoss:
@Sahilm416 I'm pretty sure what he mean is to add middle click on the tab itself, not just the X button but whole tab. It's pretty much the default behavior for any app that has tabs now

martinsefcik:
@Sahilm416 Yes, the whole tab, not X button. It would make no sense. I would just use left click on x button and not middle click in case it would be only on X button.

PR #549, opened 2023-10-12
fix(bruno-cli): add missing variable to `prepareRequest`

bacebu4:
Due to missing variable while calling `prepareRequest()` auth from `collection.bru` was not added to the request during `bru run`. 

Is a reproduction of the bug needed?


helloanoop:
Thanks for taking care of this @bacebu4 !

PR #570, opened 2023-10-13
Fixes colliding method and request name in sidebar

### LINKED ISSUE #533: [Bug] HTTP Method name and request name collidates in sidebar

Nikhil569:
fix #533 

Added `flex-shrink:0` property in the so method name div wont shrink even if size less.

fix 533(3)


helloanoop:
Clever, Neat and Simple ❤️ !!

Thanks @Nikhil569 !

martinsefcik:
Simple solution, but do we want to have request name for two requests on the same tree level starting at different X position ?

helloanoop:
Good point @martinsefcik 
Created an issue to track this https://github.com/usebruno/bruno/issues/584

PR #573, opened 2023-10-13
assert operator dropdown bgcolor fix

therealrinku:
In dark mode assert operator dropdown background color was not changing to dark
before:
before

after:
after


helloanoop:
Nice!

PR #575, opened 2023-10-13
fix: switching tabs resets timer

### LINKED ISSUE #186: Switching between tabs while a request is running resets the timer
If while a request is running I switch to a different opened tab and then switch back to the tab with the running request, the timer above the "CANCEL REQUEST" button resets and starts again. The request doesn't appear to be resent and the final request runtime is correct so it appears to be limited to this one UI element.

panda7789:
Fixes #186


helloanoop:
That was neat !!

Many Thanks @panda7789 !!

PR #576, opened 2023-10-13
feat: add button to clear collection search

dw-0:
This PR will add a small clear button in the collection search field to quickly delete the input. Previously it was required that the input field was either cleared character for character or by selecting the input and hit backspace. This is a small QoL feature.

23-10-13_17-33-25_electron


helloanoop:
This is cool !

Thanks for implementing this @dw-0 !

Issue #142, opened 2023-03-29
Auto Save Changes to request


helloanoop:
Currently Bruno does not auto save changes made to the request to the file unless user explicitly saves it by pressing Ctrl/Cmd+S

Some of our users prefer auto-save functionality (Similar to IntelliJ editor)
I am thinking of supporting this via a UI toggle in app settings.

stratosgear:
Well, eventually I "figured it" out that there must be some way to save the edits in the requests tab when i was desperately trying to get the changes persisted.  I was looking for a File/Save option or something, until eventually tried a CTRL+S randomly in the blind.  That was a very confusing experience.

To be more precise Bruno 0.19.0 still does not persist URL changes to requests....

gjrivero:
There should be some label indicating how to save the request. I had to search the forum to find out that the command was Ctrl-S

helloanoop:
We have heard this from multiple people. And I understand this is frustrating to lose data that you though you had saved.

We can make auto-save the default and provide a toggle in the app preferences to turn auto toggle off.

helloanoop:
Another option is to provide a "Save" button for now next to the send icon

![image](https://github.com/usebruno/bruno/assets/9350733/cf2091fa-c433-4dd4-acab-2a7c6176dc60)

Can someone in the community help with this?

srikary12:
@helloanoop Do we need toggle to auto save tests?

helloanoop:
@srikary12 I published a patch in `v0.21.1` that introduces several nudges to indicate that they need to save the request.
We also show a popup when a user attempts to close a request without saving.

> Do we need toggle to auto save tests?

Our current implementation basically saves the request draft inside the local redux store. It only gets written to the filesystem, when we save `Cmd+s/Ctrl+S` or the newly introduced save button in the url bar.

![image](https://github.com/usebruno/bruno/assets/9350733/e275bef2-9eb0-4dff-a9a4-1d232c7c487b)

We definitely want to support Auto-Save Toggle for users who prefer auto save - This is will be taken in v2.
Platform stability and other features in [roadmap](https://github.com/usebruno/bruno/discussions/384) are priority for v1

srikary12:
@helloanoop I've started working on this, Can expect a PR before weekend.

srikary12:
@helloanoop raised a PR. Please let me know if something needs to be improved :).

MSGoodman:
[...]

PR #582, opened 2023-10-13
feature/autoSave

srikary12:
Auto Save functionality with default 15sec interval. 

Link to discussion: https://github.com/usebruno/bruno/issues/142.


srikary12:
ipcRenderer.invoke('renderer:set-preferences', preferences) is throwing undefined error on playwright but is working fine on manual testing, Is this something expected?

Its-treason:
Quick idea: Would it be possible to listen to the user focusing another application and then auto saving? Maybe with https://stackoverflow.com/a/12486475? I feel like this would be more intuitive than the 15 seconds.

helloanoop:
Looks promising!
We will have to refine this a bit, but a good start.

@srikary12 Fully onboard with supporting this, but I would take time to gather feedback and refine the implementation before we merge this. Just saying this early so that you are not frustrated on why the PR is not getting merged :)
Some ideas from my side.

In my past, I had built an a product (that colossally failed) which was a remix of Roam Research and Github.
Like a Github for Social Notetaking

There is this functionality where a small circle on the right side displays the state.
orange - changes in collection yet to be saved
green - all synced.

The sync happens every 15 seconds (which can be made configurable)
If a user wants to immediately sync, they can hover over the circle icon and a popover will show them the number of changed files, with a button to sync immediately.

In the image below, you can see the green dot in the right top

image

srikary12:
@Its-treason I'll try to get window listener done.

@helloanoop I'll add an option to update auto save time with saving when the window is closed. I'm thinking of adding colors and hover to the current save icon.

helloanoop:
@srikary12 

> I'll add an option to update auto save time

This part can wait for a second PR. For now you can keep timeout set to 15s. Let's try to keep the PR simple, I just want to get the core part right at the first go with a clear direction on what's needed in future PRs.

douhaolee:
> @srikary12
> 
> > I'll add an option to update auto save time
> 
> This part can wait for a second PR. For now you can keep timeout set to 15s. Let's try to keep the PR simple, I just want to get the core part right at the first go with a clear direction on what's needed in future PRs.

is this configurable? thanks

srikary12:
> > @srikary12
> > 
> > > I'll add an option to update auto save time
> > 
> > This part can wait for a second PR. For now you can keep timeout set to 15s. Let's try to keep the PR simple, I just want to get the core part right at the first go with a clear direction on what's needed in future PRs.
> 
> is this configurable? thanks


Not in this PR. But would make it configurable in the next iteration.

jenswegar:
Another suggestion on solving auto-save could be to listen for changes in the documents and then some X seconds after the last change do the save? That way you don't have to run a timer that saves "just in case" every 15 seconds. Though of course it in turn would require a timer that delays the saving (if the delay is necessary).

whalemare:
Can we merge it? Great feature

NoahKr:
Any chance of this being merged? Would really love to have this feature included.

uilton-oliveira:
it would be nice to be optional, i would not like to rollback it on git every time i change some parameters to make a request..

kylehz:
need auto save!

srikary12:
I can fix it, I see that it's not merged yet.

sid-bruno:
PR is being merged into an internal version managed by @pooja-bruno, you can track the updates on the internal PR as we move forward to update the code from this PR to the current state of bruno.

Update: 

PR: https://github.com/usebruno/bruno/pull/6211

PR #587, opened 2023-10-14
Fixes: #552: Runs Post response vars and scripts on HTTP error.

### LINKED ISSUE #552: Body is not read in case of a 403
Using a post response script
`bru.setEnvVar("mfa-token", res.body.mfa_token);`

This does not work when the HTTP response status is a 403
But it does work when the HTTP response status is a 200

Could it be that Bruno does not read the body in case the call was not successful?

rayoz12:
Fixes #552: Runs Post response vars and scripts on HTTP error.

This normalises behaviour between:
- Assertions
- Scripts
- Tests

Let me know if you need anything to be changed or if this isn't desired.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -435,6 +435,56 @@ const registerNetworkIpc = (mainWindow) => {
           });
         }
 
+        // run post-response vars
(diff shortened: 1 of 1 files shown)


helloanoop:
Thanks for taking care of this @rayoz12 !!

PR #588, opened 2023-10-14
Shorten method name for OPTIONS and DELETE

### LINKED ISSUE #584: Http Method Shorthand in UI
Since http methods can have varying lengths, this can cause the UI to not aligned

image

### Proposal:
Use shorthand representation for methods that are longer then 4 characters

| Method | Display | Comment |
| --- | --- | --- |
| `DELETE` | `DEL` | This seems to be a reasonable shorthand. |
| `OPTIONS` | `OPT` | This seems to be a reasonable shorthand. |
| `PATCH` | `PATCH` | This one is hard, better to leave as is (unless you have a better idea) |

See:
https://github.com/usebruno/bruno/pull/570#issuecomment-1762258492 https://github.com/usebruno/bruno/pull/570#issuecomment-1762258492

Nikhil569:
fixes #584 

The fix shorten the method name to only starting 3 characters which have length greater than 5

Below is the screenshot for how it looks

fix584

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/RequestMethod/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/RequestMethod/index.js
@@ -23,7 +23,9 @@ const RequestMethod = ({ item }) => {
   return (
     <StyledWrapper>
       <div className={getClassname(item.request.method)}>
-        <span className="uppercase">{item.request.method}</span>
+        <span className="uppercase">
+          {item.request.method.length > 5 ? item.request.method.substring(0, 3) : item.request.method}
+        </span>
       </div>
     </StyledWrapper>
   );


helloanoop:
@Nikhil569 

```html
<span className="uppercase">
  {item.request.method.length > 5 ? item.request.method.substring(0, 3) : item.request.method}
</span>
```

I know this fixes the issue, but imo the code does not imply what is being done.

Can you update the PR with something like this

```javascript
const shortenMethod = (method) => {
  method = method.toLowerCase();

  switch(method) {
    case 'delete':
      return 'DEL';
    case 'options':
      return 'OPT';
    
    default:
      return method;
  }
}
```

In a large scale project like Bruno, readability of code trumps over number of lines needed to achieve the task.

helloanoop:
Sorry for the long wait on this.
Merged! Thank you @Nikhil569 !

Issue #193, opened 2023-09-14
Support Environment import and export


ChristophSchwenk:
Currently it't not possible to export environments and share them with the team. Also an import option of existing postman environments would be very helpful.

helloanoop:
Hi @ChristophSchwenk 

In Bruno, When you export a collection as a file, the environments are also exported in the same file.

> Also an import option of existing postman environments would be very helpful.

Yes, We will support this soon.

cardonator:
@helloanoop has any work been done yet on importing Postman environments?

helloanoop:
Not yet. This should be rather straight forward parsing of a simple key value pair.

@therealrinku please see if you could pick this up when you have some time.
UX wise, I'd imagine a "Import button" just below the "+ Create" button in Bruno's Environment Config UI

premeaswaran:
@helloanoop I am working on this currently. Will raise the PR asap unless someone else beats me up on it. Just need some help/inputs in testing out the changes in my local.

helloanoop:
@premeaswaran Send a across the PR once it's ready. I will review it.

nubunto:
was going to do it but I can wait!

premeaswaran:
@helloanoop Please let me know if any changes required. Newbie here for js and react. So any kind of inputs welcomed!

zachradtka:
I was just trying to export environment variables from bruno and currently I can export my collections but not my environments. I am on v1.11.0.

I might be missing something, but this issue is asking for support of import and export of environments and this was closed without support export.

vazome:
@zachradtka yeah, I don't care much about postman migrations. I have a clean start, but it would be cool to share my environment to collaborate on.

Look into this one: https://github.com/usebruno/bruno/issues/2017

iutale-lilly:
still not able to export env vars

anusree-bruno:
> still not able to export env vars

As mentioned [here](https://github.com/usebruno/bruno/issues/193#issuecomment-1726356899), environment variables are exported along with Bruno collections.

Are you trying to export them individually? Could you please share a bit more about your use case? For importing and exporting Global Environments specifically, we are tracking that separately in [this issue](https://github.com/usebruno/bruno/issues/3529).

iutale-lilly:
i've meant we are not able to export them to a postman collection.
we need to share the collections and envs with colleagues that do not use Bruno (for now), and it would be nice to be able to export also the envs

minamz:
[...]

PR #589, opened 2023-10-14
Feature: Support Import postman environment - Fixes issue #193

premeaswaran:
-> Fixes https://github.com/usebruno/bruno/issues/193
-> Feature to support Importing Postman environment


helloanoop:
Thanks @premeaswaran for building this out!

Just an FYI, I added some polishing as well as refactoring - https://github.com/usebruno/bruno/commit/1f8c4431e06eaccd1b85462995eb8869c790d0dc

cardonator:
@premeaswaran thanks for this, it's great! Can I make one recommendation? I have several Postman collections I've exported with "bad" characters in the name (such as `/`) and the importer fails to render the files to disk because it can't find them as they are created in subdirectories. It seems like the importer should "clean up" the filename before trying to write it out to prevent any bad characters.

Issue #568, opened 2023-10-13
[Feature request]: no proxy setup


qck4fun:
Could you please add support for no proxy addresses? Basically option to exclude certain addresses from proxying such as other tools have eg. screenshot from Insomnia below.

![image](https://github.com/usebruno/bruno/assets/7276555/be4531a5-8aad-46ab-96ad-b40d75db965e)

Thanks for you effort & feedback! :pray:

mirkogolze:
I'am working on it. Its finished in the next days

qck4fun:
> I'am working on it. Its finished in the next days

Thank you very much!

martinsefcik:
@qck4fun I guess you should leave this issue open until it is done

qck4fun:
Okay, let's wait until it's released and close it then

qck4fun:
Released, closing this. Thank you! :pray:

Issue #224, opened 2023-09-27
Proxy Support


helloanoop:
User should be able to use proxy at the collection level and app level

Tasks

- [x] Store Collection Proxy Settings in `bruno.json`
- [x] Electron layer updates to pickup apply proxy settings
- [x] Bump axios at electron layer to 1.5 
- [x] UI support for configuring proxy settings
- [x] CLI support for using proxy settings
- [x] Bump axios at cli layer to 1.5 
- [x] Use proxy config in collection runner in UI
- [x] socks5 proxy support
- [x] Support for proxy bypassing by hostname
- [x] Proxy settings at app level.
- [x] Use OS proxy configuration if found https://github.com/usebruno/bruno/issues/224#issuecomment-1753076833

Below is the axios config to use proxy
```
proxy: {
    protocol: 'https',
    host: '127.0.0.1',
    // hostname: '127.0.0.1' // Takes precedence over 'host' if both are defined
    port: 9000,
    auth: {
      username: 'mikeymike',
      password: 'rapunz3l'
    }
  }
```

mirkogolze:
The proxy must only be enabled for the request, if the request URL does not match the proxy exceptions

mirkogolze:
Hi @helloanoop,
I am not sure if proxy support is perfectly placed on collection level. Someone needs another proxy than another developer. So it is in my current project. 
In my eyes it should be a global configuration in Bruno preferences. So everybody can enable it for its personal situation. As it is in Postman.
And the usage of the CLI will mostly be in CI/CD. Here you have different proxy configuration. So the global env Variables `http_proxy`, `https_proxy` and `no_proxy` should be enough. Or I need any other configuration for runtime environment.
And the third thing is. We are currently using `axios` in our custom scripts to run the OIDC authentication request to get a token. So I would need to have access to the proxy configuration inside global object `bru`.

Your are too fast with implementing.

helloanoop:
Hey @mirkogolze !

> I am not sure if proxy support is perfectly placed on collection level. Someone needs another proxy than another developer. So it is in my current project.
In my eyes it should be a global configuration in Bruno preferences. So everybody can enable it for its personal situation. As it is in Postman.

I feel we should support it at both levels.
If a proxy config is defined at collection level, then the runtime uses collection level proxy, else it fallbacks to global proxy config.
This has been asked by community to be supported in Postman - https://github.com/postmanlabs/postman-app-support/issues/4739
What do you think?

[...]

PR #596, opened 2023-10-15
proxy settings on global and collection level

mirkogolze:
#568 #224
proxy settings on global and collection level

@helloanoop: Please make a tool tip information for the config property.
image


mirkogolze:
Hi @helloanoop,

pull request: https://github.com/usebruno/bruno/pull/617 should be included in my PR.

helloanoop:
Hey Mirko!

This is great! Thanks for taking so much time to work on this to support App Level Proxy Settings, Adding validation and also refactoring some code in the electron ipc network, and many other touches❤️

There are some things I don't concur on the implementation (specially around preferences store). It will likely require a lot of back and forth to discuss this. And this being a long PR, it'll need to be synced many times as other PR's get merged.

So, I am going to merge this now, and make some updates on main. You can raise a separate PR post that if you feel there was something I missed.

Thanks again for getting this done!

mirkogolze:
I think you mean the preferences object in the store. I prefer to have the logic more separated. 
And the usage of the configuration only needs to call the function. E.g. `preferences.isTlsVerification`.
I don't prefer to work on property level on the last level. For a store it is quit easy. 
I you have more complex logic, you only have to change the `isTlsVerification` and not the many places where it is used.
But here you are the boss.

mirkogolze:
Currently I think the proxy configuration is not enabled for `fetch-gql-schema`. I think all the request preparation is necessary too.

helloanoop:
@mirkogolze 

When it comes to state management, there are three ways in which Bruno does this
1. For requests (.bru files) and bruno.json. The state gets written into the file, the watcher(main) picks it up and emits an event, the renderer listens and update the store

2. For many other things such as responses, opened tabs, current environment, ongoing collection runs etc - the state is maintained in redux itself (This state vanishes when the app closes)

3. For Preferences, the way it is implemented is that after you click on "Save" if the ipc `save-preferences` call is successful, we update it to the redux store as you see [here](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/providers/ReduxStore/slices/app.js#L62-L73) I don't see a need for the preferences file watcher. The preferences file is a bruno system level config that is expected to be changed only via the app only. A user should not be hand-editing this file.

Do you have a reason why we need the preferences watcher over the current way (which already works for storing TLS Validation and Font Settings) ?

helloanoop:
> I don't prefer to work on property level on the last level. For a store it is quit easy.
I you have more complex logic, you only have to change the isTlsVerification and not the many places where it is used.

@mirkogolze I agree with you here. It makes sense to have this in a common place. I like the approach you went with.

mirkogolze:
> Do you have a reason why we need the preferences watcher over the current way (which already works for storing TLS Validation and Font Settings) ?

For me it was the intention to have the change, when the file is changed external. Yes, of course its only for one per mill of users.

And the all watcher logic is completely on collection level. I thought it could be nice to have a global watcher level.

helloanoop:
> For me it was the intention to have the change, when the file is changed external. Yes, of course its only for one per mill of users.

We can revisit this at some point. I still feel we should support this as users might try to mess with the preferences file and cause the app to break.

> And the all watcher logic is completely on collection level. I thought it could be nice to have a global watcher level.

I felt this too. The watcher is currently collection level stuff only.

helloanoop:
@mirkogolze 

Overall Fantastic PR !!

I added some polish https://github.com/usebruno/bruno/commit/d809a58deb6bc1d5ea7bf3dcd4d6bd55307f976a

Changes mainly include
- Renaming `noProxy` to `bypassProxy` - Although noProxy is elegant and shorter, I felt that its better to use the same wording as we represent it in the UI and is more verbose about what the property does
- Changing `preferences` to `preferencesUtil` - The property `preferences` feels more like a set of data. I felt `preferencesUtil` is a better choice as it feels more like a set of functions

Some other points
- Totally appreciate and onboard with your efforts to refactor `configureRequest`. The file was slightly getting out of control.
- `preferencesUtil` was a really good idea and helps use a neat api to get preferences

I plan to ship a release later in the day. Let me know if there is anything that you feel we need to discuss.

Issue #597, opened 2023-10-16
Bug: Request timeout will accept a negative value


Ross-Gargett:
This is pretty minor, but it causes a JavaScript error if you enter a negative timeout. 

I will make the code change to fix this shortly.

<img width="1037" alt="image" src="https://github.com/usebruno/bruno/assets/73141350/aae7359c-93b6-4f51-9e9b-a6a230ecdbcb">

PR #598, opened 2023-10-16
Fixes Bug #597 : Timeout field validation

Ross-Gargett:
# Description
[Issue #597 ](https://github.com/usebruno/bruno/issues/597)
This is a fix for a small bug in the request timeout preference text field. The field allowed any character, and allowed you to save negative values. 

- Prevented any character outside of 0-9 from being accepted
- I tied the fields value property to the `timeout` stateful value. This means that character we are not accepting into state, will also not be displayed in the field
- Made the field slightly wider, as it felt slightly cramped to me

Just let me know if you want any changes.

# Contribution Checklist:


mirkogolze:
I think this solution is not the best way. In bruno the YUP-Framework ist used for validation. Can you have a look at the different . A more complex example is here.
https://github.com/mirkogolze/bruno/blob/feature/proxy-global-and-collection/packages/bruno-app/src/components/CollectionSettings/ProxySettings/index.js

Ross-Gargett:
image
image

Thanks for the feedback @mirkogolze> I updated the form to use Formik and Yup. Let me know if you would like any additional changes.

Issue #143, opened 2023-03-30
Collection Runner Improvements


helloanoop:
1.  Collection runner Stop/Close button should appear immediately to stop unintended runs
2. The Runner should open in a tab like a request, so you can navigate away from it and then back to it.

Source: Cibico99 on Discord

mj-h:
Current state: The runner opens in a new tab, but there is still no stop button.

When closing the tab, the runner continues in the background, and each time a request is fired, any open context menus close again.

PR #619, opened 2023-10-16
feat: bru.setNextRequest()

### LINKED ISSUE #534: [Feature Request] Influence the Collection run order
It would be great to have something like

```js
// set request by id
bru.setNextRequest(12)
// alternative: set request by name?
bru.setNextRequest("GET BookingDetails")
```

That way, we could do branches and loops for error-handling and async workflows.

mj-h:
# Description
This PR adds the method `bru.setNextRequest(requestName)`. It can be used in pre- and postrequest scripts (last write wins). It works in the collection-runner and in the CLI. 

Closes #534. 

# Contribution Checklist:

---
**Usage**::
image

---
**Result in the UI Collection Runner**
image

---
**Result in the CLI**
image


jhoppenstedt (approved):

helloanoop:
@mj-h Some questions

1. Can you share a real world scenario where the `setNextRequest()` api become useful.
2. Should we consider supported full paths like `setNextRequest("folder/get users")`

mj-h:
Sure!
1. Typical real-world scenarios for `setNextRequest()` are:
    - Async service with polling pattern: For these services, you send "start-process" / "get-status" / "verify-final-result" requests. The "get-status" request needs to be repeated until the status changes, or until a maximum time / number of requests has been reached. We use this pattern a lot internally, and it's a common pattern for loading results of of slow processes into the UI, asynchronously. 
    - Service-creation with cleanup: Often, you find the "create-service" / "test-service" / "destroy-service" pattern. When one of the tests fail, you often want to jump straight to "destroy-service".
2. That's an interesting idea. I'm definitely open to it, and it would reduce the chance of getting annoying name-collisions -- but it increases the chance for breaking things, when someone renames a folder without checking the code.  Tomorrow, I'll ask a few colleagueswhom I know to be using `.setNextRequest` heavily. I would tend to make the foldername optional -- if it's there, it must be matched, if it's not there, it's not part of the matching-logic.

 I truly appreciate your thoughts here. The `.setNextRequest()`-feature keeps me up at night, and I'm not sure I am doing humanity a favor when implementing it. The worst Postman collections I had to maintain were collections that made heavy use of `.setNextRequest` and variables, which are really just `GOTO`s and global variables. Not fun. But on the other hand, Postman was the only test-tool that supported async APIs.

mj-h:
Two thoughts on how `.setNextRequest()` could be made more palatable:
1. [Similar to `cmd + enter`](https://github.com/usebruno/bruno/blob/f08871d491df711588f2eb79227d12951d20b2f4/packages/bruno-app/src/providers/Hotkeys/index.js#L68), we could have a keyboard shortcut `cmd + shift + enter` that sends a request and goes to the "next" request, just like the collection runner would. That way you could step through a collection in a way that Postman never could. This is the main debugging feature I missed in Postman. (And for people who don't use `.setNextRequest`, it's still a convenience feature that allows them to run collections top-to-bottom more quickly).
2. [Like hurl is discussing it](https://github.com/Orange-OpenSource/hurl/issues/1862#issuecomment-1701045623), we could ave `skip` and `repeat` blocks in the bru file. Something like:
   ```
   skip-if {
     someOtherVar != "some-value"
   }
   
   repeat-if {
     someVar == "not-ready" 
   }
   ```
   Then the skip / repeat information could be placed more prominently in the UI (like the assertions), and we could do safe renaming of requests and folder. But the bru language would become bigger and bigger...

bdentino:
Chiming in with another real-world use case I would love to see supported before I can fully migrate my workloads from postman:

A common pattern in our automation tests & cache-warming flows are to make a request to one endpoint which provides a list of available resources, and then issue a different request for each resource returned. So for example I will have a postman collection with 2 requests: `GET /foos` and `GET /foos/:fooid`. The first request is expected to return a list of `n` fooids, and I then I'll issue `GET /foos/:fooid` `n` times, while updating the value of `fooid` via post-request scripts during the run.

I agree with @mj-h that `setNextRequest` feels a bit "unclean" from a design pov. Not just because of the GOTO/globalish nature of it, but also because it forces me to modify collection variables during a run in order to achieve flows like this and it feels wrong to have the collection end up in a different state after any individual run. 

Another approach which would enable my use-case, and presumably others, would be to inject a function you can call from scripting contexts to 'queue' additional requests. I suppose it would only make sense in the context of a collection run as opposed to executing a single request from the ui (then again, so does getNextRequest). But I'm thinking it could look something like this: 

```js
// post-request script

// run is a new global variable, alongside `req`, `res`, `bru`, etc...
// requests is a list of requests defined in the running collection
// queue is a list of pending requests waiting to run
const { requests, queue } = run

// Find the request you want to queue. Solves 'globalness' problem 
// in that you can only queue requests which you know exist. Does 
// limit you to requests within the collection, though so no reaching 
// out to other collections/folders (personally i think that's a good 
// thing for maintainability/cohesion but could be convinced 
// otherwise).
const newRequest = requests.find((r) => r.name === 'Get Foo')

// Create an instance of this request that will execute with specific
// variables that would supersede any pre-defined collection variables
// (solves the problem of having to modify collection state during run)
const toRun = newRequest.withVars({ ...overrideVars })

// Run new request last (after other queued requests)
queue.push(toRun)

// Run new request next (before other queued requests)
queue.unshift(toRun)
```

What do you think?

Timwun (approved):

mj-h:
I talked to colleagues, and I checked usage on various GitHubs. My findings:
- on public GitHub, only 2% to 3% of collections use `postman.setNextRequest`. 
- on our Enterprise GitHubs (yes, several GitHubs, because $enterprise...), about 20% to 25% of the Postman collections use `postman.setNextRequest` (!)
  - half of those usages are `postman.setNextRequest(null)` or `postman.setNextRequest()` to abort the collection run. Most times, the author actually wants to exit with error. On the command line, they could have used `newman run --bail`, but the UI collection runner has no such feature (afaik). 
    - **Idea**: "stop-on-error" should be a collection-level property that is respected by the CLI and the UI collection runner.
  - a third of usages are `postman.setNextRequest(request.name)` for a `do-while` loop, mostly to check the results of async APIs. It is occasionally used as generic retry mechanism.
  - the remaining usages seem to be a hodgepodge of "skip next request if previous response is foo" and "goto error-handling-part".
  - one clear misuse I saw was an author who painstakingly used `postman.setNextRequest("foo")` on every request to jump to the next request, because he did not know that that's the default behavior of the collection runners. But that was an outlier and should not influence our design.
- no user I spoke to had a clear preference of `.setNextRequest` vs `skip`+`repeat`.

@bdentino: I did not encounter your scenario in our collections, but I did not check every collection in detail. In any case, I would say it's a rare (but still valid!) usage. Your proposal, however, boils down to a `goto X then goto Y then goto Z` statement, which is powerful, and which makes your usecase easier, but which is harder to use and harder to reason about than a simple `goto`.

mj-h:
My proposal going forward:
- let's include `bru.setNextRequest()` in the v1 release, to enable more users to switch to bruno. Use similar semantics as in postman. Clearly mark it as "experimental / not recommended" in the documentation.
- let's think about better control-flow logic for the v2 release. I think a strong case can be made for `skip` and `repeat`, but making it ergonomic takes some more thinking and tinkering.  
- possibly mark `bru.setNextRequest()` as deprecated in the future.

mj-h:
Merged main and confirmed that it still works in UI and CLI. To be discussed: How to handle the "request name not found" case. Postman exited early here; with exit code 0, I believe. For easier compatibility, we could do the same.

bdentino:
@mj-h I found some UI problems in my own testing and created a PR with fixes here: https://github.com/mj-h/bruno/pull/1

mj-h:
While testing this further, I noticed that there is no way to stop the collection runner once it started. But that's a separate issue (https://github.com/usebruno/bruno/issues/143).

helloanoop:
@mj-h Thank you for leading the discussion around this.

I'd like to spend some more time on this. Been busy with many other things. I will put some dedicated time towards this issue the coming week.

PS: See [v1.0.0 discussion](https://github.com/usebruno/bruno/issues/643)

mj-h:
Sure, no problem. In the meantime, I'll add two missing behaviors:
- `setNextRequest(null)` and `setNextRequest(undefined)` should break the the execution loop (no error)
- there should be a maximum number of loop iterations to prevent truly infinite loops. Other popular products use `N=10000`. I think that's a reasonable choice.

fpassaniti:
For my usages, this feature is a mandatory prerequisite to move from Postman to Bruno, as it's a pillar to all my collections. 
I need to dynamically, depending on file inputs, or my API results, to switch from a request to another, or even loop over the same request several time in my collection. 
I intensively use postman.setNextRequest to chain, loop, or even stop my collection dynamically. 
I look forward to see this function released, as for now I'll be force to put on hold our migration to Bruno without it. 
Thanks

packages/bruno-cli/src/commands/run.js
```diff
@@ -369,6 +372,28 @@ const handler = async function (argv) {
       );
 
       results.push(result);
+
+      // determine next request
+      const nextRequestName = result?.nextRequestName;
+      if (nextRequestName !== undefined) {
+        nJumps++;
+        if (nJumps > 10000) {
```
fpassaniti, line 380: Shouldn't it be a parameter in bruno.json (the max number of jumps allowed) ? just in case

mj-h: I thought about it -- but there are already a ton of config options... Do you have a use-case that needs it? If so, we can add it. If not, I would say YAGNI.

fpassaniti: YAGNI it is then ;)

CoenraadS:
Just would like to share how postman is used at my $dayjob

Similar to what bdentino already described

- GET /api/foo/{{id}}
  - If ID is not defined, then setNextRequest
    - GET /api/foo/latest => Set {{id}} var
    - If Auth is not refined, then setNextRequest
      - POST /api/login => Set auth variables

mj-h:
@helloanoop : Any news on if and/or when this can be included? I'm under pressure to migrate our API-tests away from newman, and I can stall my org for maybe one or two more weeks...

helloanoop:
Thank you @mj-h for your work on this !
Everyone, Thank you for your patience.

helloanoop:
Its live  🎊 

GUI - v1.4.0
CLI - v1.2.0

CoenraadS:
I just tried it, but I couldn't get it to work?

image

I tried

`bru.setNextRequest("Hacker News/GetById");`

`bru.setNextRequest("GetById");`

In `GetById` I can in the pre-request script a console.log("Hello World") but I never see it,

mlestoquoy:
First, thanks for this feature. I managed to make it work from bruno, but not from CLI:
image
works with bruno 1.4.0 (Run):
image
and not with bru 1.2.0 (bru.setNextRequest is not a function):
image

Did I miss something ?

mlestoquoy:
> @mlestoquoy I just released a patch fix for this in `bru cli v1.2.1`
> 
> Can you install it and check if this issue is resolved ?

Yes it works !
Thanks @helloanoop 
image

Issue #609, opened 2023-10-16
[feature] Use raw response data instead of the default output from axios


Sl-Alex:
As a user, I'd like to have access to the original response data.
The current implementation relies on the default axios behavior, that does two things with the response data:
1. Converts it to string, which breaks binary responses (e.g. images)
2. Tries to convert it to JSON, which removes the original formatting (spaces, tabs and so on)

These implicit conversions change the data size and the content, so the size displayed in Bruno, is correct only for text (not even JSON) responses.
I'm working with embedded HW, so knowing the exact response size is sometimes a must.

Using raw data also seems to be the only way to solve the problem with images (and, probably, with other binary file types as well). As @Its-treason commented in #593:
> currently, Bruno does not render the original response body as an Image. Instead, it tries to load the image using a separate GET request to the requested URL.

If we had access to the raw response data, it would be much easier to implement such a rendering and show a real response size

Issue #593, opened 2023-10-15
Is bruno unable to render images in response?


IamLizu:
My Bruno version is v0.24.0

I am getting this in response,
![image](https://github.com/usebruno/bruno/assets/26184316/a1264568-4ec8-4630-81af-687a86371320)

hasan-zaag:
For another response, I am getting
![image](https://github.com/usebruno/bruno/assets/136310205/e8fb6933-987e-4b11-ba63-a5e017a2caa0)

Its-treason:
Hey @IamLizu,

currently, Bruno does not render the original response body as an Image. Instead, it tries to load the image using a separate GET request to the requested URL.

We are definitely going to update the preview to render the original Image response! Thank you for opening an issue.

IamLizu:
Thank you, I am closing this now as it already planned.

helloanoop:
@IamLizu let's keep this open. Will close this once this is delivered.

sanjai0py:
Closing this as the PR implementing feature #622 has already been merged and delivered in [v0.26.0](https://github.com/usebruno/bruno/releases/tag/v0.26.0).

Issue #539, opened 2023-10-11
[FEATURE] Binary file download


majkinetor:
It should be possible to save a binary file locally when `content-disposition` header is present

<details><summary>details</summary> 

![image](https://github.com/usebruno/bruno/assets/85767/bb0a5784-2a0c-49a5-9299-9cf95b8f6c60)

![image](https://github.com/usebruno/bruno/assets/85767/3ef9af7b-a844-4bfd-a564-d7728b54fce2)

</details>

martinsefcik:
I think it should be possible to store response body to file for any type of response and even if `content-disposition` header is not present (it can be just hint for filename during saving if such header present).

majkinetor:
Indeed. However, its totally needed for binary files because you just cant see response at all. 

I generated shell code for my case to be able to see it until we eventually have this:

```
script:post-response {
  var path = require("path");
  const HOST = bru.getEnvVar("HOST");
  bearer = req.getHeader("Authorization");
  let cd = res.getHeader('content-disposition');

  var match = cd.match(/(filename=")(.*)";.+$/);
  var filename = match && match[2] || 'default-filename.out';
  var filename = path.basename(filename)
  let url = req.getUrl().replace("{{HOST}}", HOST)
  let method = req.getMethod();
  console.log(`rm "${filename}"; curl.exe ${url} --request ${method} --header "Authorization:  ${bearer}" -O -J; start ${filename}`);
}
```

I use CTRL + I to copy/paste it into shell

martinsefcik:
@helloanoop I guess this issue was implemented as part of PR #766 and can be closed.

helloanoop:
Closing this issue. @majkinetor please reopen if you encounter any issues with the functionality.

majkinetor:
The only issue I found is that larger PDF files make bruno unusable (e.g. 500 pages). It takes minutes to render during which time bruno is frozen. The same PDF opens instantly in browser.

martinsefcik:
> The only issue I found is that larger PDF files make bruno unusable (e.g. 500 pages). It takes minutes to render during which time bruno is frozen. The same PDF opens instantly in browser.

But I guess it is caused by pdf file preview feature implemented in PR #982 implemented by @n00o (and not by binary file download feature implemented in PR #766) as it consumes much more resources in case of big pdf files and renders big HTML content.

[...]

Issue #226, opened 2023-09-27
Bruno Runtime Refactor


helloanoop:
A lot of code is currently shared across `bruno-cli` and `bruno-electron`

This issue is for consolidating this logic inside a `bruno-runtime` package

helloanoop:
We will do this after our `v1` launch.

The current approach has been in development for a long time.
At our current stage stability is more important than the slowness in dev introduced code duplication.

Once we finish `v1`, then we will have enough time on our hands to first setup a strong test suite and then proceed with the refactoring.

This test suite should give us the confidence to refactor with peace of mind.

cc @mirkogolze

berlingoqc:
Do you plan on starting this soon ? I could take a look because i want to added feature to the cli but the code is in bruno-app but could be shared.

For exemple i want to add export via the cli without copy pasting the same code

helloanoop:
Hi @berlingoqc, Given our current PR and issue backlog, it looks like we won’t be able to tackle this within the next three months at least.

berlingoqc:
@helloanoop I have time, i want to upgrade the cli tooling and i wish we could refactor it before.

if you want to shared me what your vision on it. I can tackle it

helloanoop:
This activity is already in progress. We've started migrating things to the bruno-common and bruno-requests packages.
I don't see a need to keep tracking this as an open issue.

Closing the issue.

PR #622, opened 2023-10-16
 feat(#593): Show real response in image preview

Its-treason:
Fixes: #609, #593 & #519

Now the original body is encoded as Base64 and send to the browser. This allows us to render the original Image response as a preview.

I also refactored the `QueryResult`-component into to separate components, this makes extending it with more preview modes easier.

This PR does change quite a lot, so I would appreciate anyone testing it locally to check if there are any issues.

---

Now that we have the original response, we can start working on other issues, like: #539 and implement other preview modes.


packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -296,6 +296,14 @@ const registerNetworkIpc = (mainWindow) => {
       /** @type {import('axios').AxiosResponse} */
       const response = await axiosInstance(request);
 
+      const dataBuffer = Buffer.from(response.data);
+      // Overwrite the original data for backwards compatability
+      response.data = dataBuffer.toString('utf-8');
```
Sl-Alex: What about using charset from "Content-Type" with a fallback to UTF-8? Something like this:
```js
      const content_type = response.headers['content-type'];
      // RegExp magic from https://stackoverflow.com/a/33192813
      let charset = new RegExp('/charset=([^()<>@,;:\"/[\]?.=\s]*)/i').exec(content_type);
      if (charset == null) {
        charset = "utf-8"
```

martinsefcik:
@Its-treason Great!

One small note. Can we use raw response body size now for displaying in UI in case `content-length` header is missing, because I think the fix implemented in #520 was I would say just temporary workaround which does not show real size of transferred body.

Sl-Alex:
@Its-treason I like the way you did it. I tried to do the same before raising #609, and it even worked, but your solution is definitely better.
I have only one general comment: do we need more or less the same for CLI?

Its-treason:
Hey everyone, thank you for feedback! Really appreciate it!

I update my PR: I noticed that runners were broken, because I need to convert `response.data` from the buffer to a string/object.

> What about using charset from "Content-Type" with a fallback to UTF-8? Something like this:
> 
> ```js
>       const content_type = response.headers['content-type'];
>       // RegExp magic from https://stackoverflow.com/a/33192813
>       let charset = new RegExp('/charset=([^()<>@,;:\"/[\]?.=\s]*)/i').exec(content_type);
>       if (charset == null) {
>         charset = "utf-8"
> ```

That sounds like a very good idea, I implemented it!

> @Its-treason Great!
> 
> One small note. Can we use raw response body size now for displaying in UI in case `content-length` header is missing, because I think the fix implemented in #520 was I would say just temporary workaround which does not show real size of transferred body.
> 
> And maybe also in cases if `content-length` is present I would prefer raw body size to be displayed.

Also didn't think of this at first, also implemented it!

> @Its-treason I like the way you did it. I tried to do the same before raising #609, and it even worked, but your solution is definitely better. I have only one general comment: do we need more or less the same for CLI?

Yes, we could also add this in CLI, but It's currently not needed because: CLI has its own code for request preparation and the buffer is not accessible to tests and assertions.

We should probably change this with: https://github.com/usebruno/bruno/issues/226

helloanoop:
@Its-treason I am going to review this today and get it merged.

Is this PR `Good to Merge` from your side ?

Its-treason:
> @Its-treason I am going to review this today and get it merged.
> 
> Is this PR `Good to Merge` from your side ?

Sorry for the late reply, I was still testing this during work with a “production” environment. Everything was working fine, so It's ready to merge!

PR #644, opened 2023-10-17
Fixes #637: autofill folder name as collection name

### LINKED ISSUE #637: [User Experience] Use collection name as Folder name by default
While trying to create/import a new collection, more often than not, users prefer to have the collection name and folder name as same (I don't see a reason why not). 

I would prefer to have it the below way as an end user. Collection name is filled in the Folder name by default (as and when the user types in) and the user can change it only if needed. It would reduce some keyboard work for the user.

Screenshot 2023-10-17 at 3 50 01 PM

premeaswaran:
# Description
- Autofill folder name to be the same as collection name as and when the user types in the collection name.
- In order to improve user experience further, once the user consciously updates the folder name, and then comes back to update the collection name, this time around, folder name is not updated automatically.
- Fixes issue https://github.com/usebruno/bruno/issues/637

# Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
+++ b/packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
@@ -76,7 +76,14 @@ const CreateCollection = ({ onClose }) => {
             name="collectionName"
             ref={inputRef}
             className="block textbox mt-2 w-full"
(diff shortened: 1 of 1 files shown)


helloanoop:
Nice!

Thanks @premeaswaran !!

PR #647, opened 2023-10-17
Fix #646: show exact response size in bytes on mouse hover

### LINKED ISSUE #646: [UX] Display the exact response size in bytes on mouse hover
It would be nice to display the exact response size in ResponseSize component:

image

Sl-Alex:
This PR just adds a title to the ResponseSize component, showing the size in bytes, in addition to human-readable KB and MB.


helloanoop:

Issue #663, opened 2023-10-18
Environments should support the "meta" grammar


cardonator:
Collections have a tag for meta information which allows the collection to have a name that's different from the filename stored to disk. Environments should support this same behavior.

anusree-bruno:
We've made updates to environments, and Bruno now supports importing environments that contain special characters.

PR #349, opened 2023-10-04
fix: Fix name validation consistency

Its-treason:
With this PR I want to fix validation problems when creating folder, requests or environments. All names now use a validation regex:
```regex
/^(?!CON|PRN|AUX|NUL|COM\d|LPT\d|^ |^\-)[\w\-\. \(\)\[\]]+[^\. ]$/
```
This regex ensures various rules when naming files on the file system, so users don't run into weird problems:

- Only allowing certain characters: `a-z`, `A-Z`, `0-9`, `_`, `-`, `.`, `(`, `)`, `[`, `]` and ` ` (Space) to ensure file name compatibility between operating and file systems
  - Everything from [Portable Filename Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282) and `(`, `)`, `[`, `]`, ` `
- Disallowing special cases, e.g., Windows file names must not end with a `.`
- Disallowing Windows Device names e.g., CON, NUL

You can test this regex on regexr: https://regexr.com/7l552

For reference for all limitations, I used:
- https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names
- https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words

I also removed the `is-valid-path`/`isValidPathname` check from the collection creation as this check was very strict and would not pass with names containing common special characters and a failed checked would only fail with an “An error occurred” toast. This should resolve #174 

---

For the future we should make it, so all characters are allowed in the Request/Folder name, but the special characters are only saved in the .bru file meta and the file name is automatically stripped from special characters. This should then resolve: #189. But will take some time to fix as it is more complicated.

PR #665, opened 2023-10-18
sanitize the imported environment filename before writing it to disk

cardonator:
# Description

Currently if the environment name has a bad character (like /) in it then it fails to write to the filesystem. I'm proposing this however I don't think this is functional to merge because of #663 right now. The Environment name on the UI ends up being the filename and it really should be a name stored in the environment file and not just the filename.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/collection.js
+++ b/packages/bruno-electron/src/ipc/collection.js
@@ -141,7 +141,8 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
         await createDirectory(envDirPath);
       }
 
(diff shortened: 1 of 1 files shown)


Its-treason:
Hey @cardonator,

In #349 i already implemented a meta tag/grammer in the environment files to save the name. With this you can use any character in the environment name. The PR is planned to be reviewed and merged before the v1 release, so it should be shipped shortly!

cardonator:
@Its-treason sweet, thanks. You're right that the proposed change will make this change obsolete. I'll take a look at and comment on that other PR.

PR #668, opened 2023-10-18
build: update Node.js to v18.18.2

wheinze:
# Description

Updates Node.js to latest v18.

### Contribution Checklist:


.nvmrc
```diff
@@ -1 +1 @@
-v18.13.0
\ No newline at end of file
+v18.18.2
```
j0k3r: I think we can directly remove minor version
```suggestion
v18
```

wheinze: Me personally, I like to set up environments as much predictable as possible in regards of [reproducibility](https://reproducible-builds.org/), that's why I would vote to define the exact version here.

But I would also be fine with @j0k3r's approach if this is the preferred way.

helloanoop:
> Me personally, I like to set up environments as much predictable as possible in regards of [reproducibility](https://reproducible-builds.org/), that's why I would vote to define the exact version here.
But I would also be fine with @j0k3r's approach if this is the preferred way.

I feel v18 should work just fine.
Main reasoning is that contributors normally use `nvm use` and will be forced to install a new version for this.

On reproducibility of bugs, we can move towards a exact versioning when we see cases of people encountering different behaviours on the same major version of node.

wheinze:
> > Me personally, I like to set up environments as much predictable as possible in regards of [reproducibility](https://reproducible-builds.org/), that's why I would vote to define the exact version here.
> > But I would also be fine with @j0k3r's approach if this is the preferred way.
> 
> I feel v18 should work just fine. Main reasoning is that contributors normally use `nvm use` and will be forced to install a new version for this.
> 
> On reproducibility of bugs, we can move towards a exact versioning when we see cases of people encountering different behaviours on the same major version of node.

I'm fine with only requiring the major version to be aligned to simplify local development, but would [propose](https://github.com/usebruno/bruno/pull/667/commits/2b8c6468548eea2d5a9b4ce19f4ff79ff78c1ba4) to run CI against latest patches. 

The more interesting question as of today would be, if we think to update to Node v20 any time soon as it got recent LTS version now :) (Would be willing to file a PR for this one)

Issue #704, opened 2023-10-20
[Feature Request] Run pre-request and post-response actions when fetching GQL schema


cwilper:
# Why

In my environment, I need to digitally sign all request payloads, including GQL introspections. Since it's a very custom authentication scheme that is not in wide use, I thought it would be good to use a pre-request script for this. But I discovered that GQL schema fetching doesn't invoke pre-request scripts.

Although my use case doesn't call for it, I know Bruno has pre-request *and* post-response vars and scripts, so it seems like it would be good to also run those, and treat these requests just like any other request. Can anyone think of downsides to doing that?

# Related

Postman's GQL schema auto-fetch does not currently support running Pre-request scripts either, but people clearly want it:
*  https://github.com/postmanlabs/postman-app-support/issues/10957

MatisseHack:
I'm in the same boat! I handle authentication with a pre-request script, which isn't called when loading a GraphQL schema through introspection.

jwsloan:
I agree that this does not seem to work.

Jackevansevo:
I'm pretty sure there's been a regression since https://github.com/usebruno/bruno/pull/705 got merged.

I'm on v1.16.1 and currently affected by this.

helloanoop:
I have merged the PR for the fix.
This is set to go out in the upcoming `v1.20.0` release

PR #705, opened 2023-10-20
feat: run prerequest/postresponse actions when fetching gql schema

cwilper:
This PR addresses the following feature request:
* https://github.com/usebruno/bruno/issues/704

It automatically runs the usual pre-request and post-response code (vars and scripts) when fetching the GQL schema. It also tries to consolidate some of the relevant code to make re-use more manageable.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -514,8 +514,25 @@ const registerNetworkIpc = (mainWindow) => {
         });
       }
 
-      const processEnvVars = getProcessEnvVars(collection.uid);
-      interpolateVars(preparedRequest, envVars, collection.collectionVariables, processEnvVars);
+      const requestUid = uuid();
+      const collectionPath = collection.pathname;
+      const collectionUid = collection.uid;
+      const collectionVariables = collection.collectionVariables;
+      const processEnvVars = getProcessEnvVars(collectionUid);
+      const brunoConfig = getBrunoConfig(collection.uid);
+      const scriptingConfig = get(brunoConfig, 'scripts', {});
+
+      await runPreRequest(
+        request,
+        requestUid,
+        envVars,
+        collectionPath,
+        collectionRoot,
+        collectionUid,
+        collectionVariables,
+        processEnvVars,
+        scriptingConfig
+      );
 
       const axiosInstance = await configureRequest(
         collection.uid,
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -189,21 +189,87 @@ const parseDataFromResponse = (response) => {
 };
 
 const registerNetworkIpc = (mainWindow) => {
+  const onConsoleLog = (type, args) => {
+    console[type](...args);
+
+    mainWindow.webContents.send('main:console-log', {
+      type,
+      args
+    });
+  };
+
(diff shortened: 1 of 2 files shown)


helloanoop:
Thanks for taking care of this @cwilper !

> It also tries to consolidate some of the relevant code to make re-use more manageable.

Makes sense. This should help us move this logic seamlessly to a separate package (post v1 release) to be shared across electron and cli.

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -502,8 +540,25 @@ const registerNetworkIpc = (mainWindow) => {
         });
       }
 
-      const processEnvVars = getProcessEnvVars(collection.uid);
-      interpolateVars(preparedRequest, envVars, collection.collectionVariables, processEnvVars);
```
mheidinger, line 506: This unfortunately breaks interpolation for the introspection query.
The `runPreRequest` and with that the `interpolateVars` is executed on `request`, but actually executed is `preparedRequest` from https://github.com/usebruno/bruno/pull/705/files#diff-eaeb1da31271859ebecb449a3291ab31d513dbee71aff0df896001827f94cc90R533 in https://github.com/usebruno/bruno/pull/705/files#diff-eaeb1da31271859ebecb449a3291ab31d513dbee71aff0df896001827f94cc90R570.

Not sure how to properly fix this best 🤔

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -514,6 +569,19 @@ const registerNetworkIpc = (mainWindow) => {
       );
       const response = await axiosInstance(preparedRequest);
 
+      await runPostResponse(
```
mheidinger, line 572: On first thought I am not sure if running post request stuff makes sense.
At least my expectation is that post request scripts are written to run on the actual response of the request (i.e. extract data from the response, run assertions on the response). If now suddenly the response is the introspection data, I'd expect that those scripts and assertions will definitively fail or result in unexpected behaviour.

packages/bruno-electron/src/ipc/network/index.js
mheidinger: Tagging @helloanoop directly, as this should be fixed before the 1.0 release

helloanoop: @mheidinger 

I will spend some time next week, This file has gotten really hard to maintain. We have `prepareRequest` and `configureRequest`. Then we have configuring proxy, and the entire request lifecycle along with nuances around running single request, running a folder (collection run), and fetching gql schema.

Need to look at this more holistically. Ideally we should have just 1 separate file with a single`runRequest` method with a good configurability (ex: disable running assertions/tests etc)

For now, to solve this issue, can you update the `fetch-gql-schema` [ipc handler](https://github.com/usebruno/bruno/blob/main/packages/bruno-electron/src/ipc/network/index.js#L551-L626) and duplicate the code inside it needed to run request lifecycle stuff (pre request vars and script, post response vars and script)

@cwilper @mheidinger @grubersjoe If anyone of you have bandwidth, please pick this up.

Its ok to duplicate code.
We'll get to the ideal DRY version post release 1.0.0

mheidinger: Created https://github.com/usebruno/bruno/pull/809 to just add back in that one `interpolateVars` call

PR #707, opened 2023-10-20
Add most common request url as initial in new req

thehorse2000:
# Description
This change sets the **initial request URL** field to the most common URL used in the collection.
This change is made to facilitate creating new requests in a collection Since most requests in the same collection share the same URL (especially GraphQL requests).
However, I didn't limit this feature to graphQL requests only, I believe it can also be helpful in restful request creation as well, because the user has to change only the last part of the URL(endpoint) and won't have to write the URL from scratch.
It is a small change but I believe it is very important for User Experience.

I will be waiting for Code Review, If there are any recommended changes I will be happy to apply them.

### Contribution Checklist:

https://github.com/usebruno/bruno/assets/7264374/1df36702-d595-4d13-9e05-b12513ca4604

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
+++ b/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
@@ -11,16 +11,18 @@ import { addTab } from 'providers/ReduxStore/slices/tabs';
 import HttpMethodSelector from 'components/RequestPane/QueryUrl/HttpMethodSelector';
 import { getDefaultRequestPaneTab } from 'utils/collections';
 import StyledWrapper from './StyledWrapper';
+import { getMostCommonUrlFromCollection } from 'utils/collections/index';
 
 const NewRequest = ({ collection, item, isEphemeral, onClose }) => {
   const dispatch = useDispatch();
   const inputRef = useRef();
+  const commonUrl = getMostCommonUrlFromCollection(collection);
   const formik = useFormik({
     enableReinitialize: true,
     initialValues: {
       requestName: '',
       requestType: 'http-request',
-      requestUrl: '',
+      requestUrl: commonUrl,
       requestMethod: 'GET'
     },
     validationSchema: Yup.object({
--- a/packages/bruno-app/src/utils/collections/index.js
+++ b/packages/bruno-app/src/utils/collections/index.js
@@ -8,6 +8,14 @@ import filter from 'lodash/filter';
 import sortBy from 'lodash/sortBy';
 import isEqual from 'lodash/isEqual';
 import cloneDeep from 'lodash/cloneDeep';
+import head from 'lodash/head';
+import flow from 'lodash/flow';
+import countBy from 'lodash/countBy';
+import maxBy from 'lodash/maxBy';
+import last from 'lodash/last';
+import partialRight from 'lodash/partialRight';
+import entries from 'lodash/entries';
+
 import { uuid } from 'utils/common';
 import path from 'path';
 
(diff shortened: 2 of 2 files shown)


dw-0:
What if you have a lot of different requests which are different from each other. Let's say you have two endpoints, `http:127.0.0.1/api/cities/spain` and `http:127.0.0.1/api/cities/germany`. That would make `http:127.0.0.1/api/cities/` as the initial request URL, right? Now let's say you happen to not have any `cities` endpoints anymore and also happen to not have a higher quantity of another form of request which qualifies as a new initial request URL. Would you be stuck with `http:127.0.0.1/api/cities/` as the default request URL? If yes, wouldn't that be even annoying? In your showcase you would have two other approaches that come to my mind:
1) set the most used URLs as an environment variable and use the variable name
2) clone the request which has already the most fitting URL

Otherwise it would probably be even better if you have the option to define a specific default URL as kind of a collection setting or property. But not set one without the option to disable that behavior.

thehorse2000:
Hey,
The current implementation scans the collection URLs and sets the most repeated URL as the initial for a new request.
This is great for graphQL requests, But for REST, I do agree that it is not the best implementation.
**So I have two options in mind for this:**
1. The perfect implementation is to set a property for each collection for the Base URL for this collection.
But I don't think there is a way to set properties for collections, right?
2. Or just do this current implementation only when the user selects graphQL as the type of request.
Waiting for your opinion on this,
Thanks

dw-0:
Im not too familiar with working with graphQL, so take my comments with a grain of salt.

To 1: Yes, that would require a general implementation for customizable collection properties first. But that could be possible to implement. Each collection has a top level JSON where its version, name and type is stored. You could store its properties in there as well? Of couse, that would result in a much bigger workload for this PR. But i think that would lay a great foundation to have customizable collection properties and maybe even subfolder specific properties (e.g. subfolder level env-variables?) in the future.

To 2: I guess other people have to chime in on that part, which regularly use graphQL. Im not sure if that's a feature that is always wanted by everyone. I would assume, at least a settings options should be delivered to disable this behavior if it's not wanted.

thehorse2000:
I do agree
I like the sound of creating properties for Collections and folders. In which we can set a Base URL for any new request made in this collection/folder.
I would love to work on that.
How about i create a new PR for impelmenting the properties feature and another PR after for creating the BASE URL functionality?
This way PRs will be small as possible

dw-0:
> How about i create a new PR for impelmenting the properties feature and another PR after for creating the BASE URL functionality?

I tend to agree, but don't you need a first custom property to implement such feature? That could be the feature for setting a collection/folder wide base url? :) Maybe @helloanoop can steer you in the right direction here and decide what would he prefer?

thehorse2000:
Yes that is what i meant
We need to first create setting custom properties for collections and folders.
I would love to hear @helloanoop opinion on this idea and the possibility of its implementation.

thehorse2000:
Hey as promised, I created a new PR for the properties feature for collections
I think it is worth checking it out.
https://github.com/usebruno/bruno/pull/718

helloanoop:
Fully onboard with supporting this.
Thanks @thehorse2000 and @dw-0 
Going to close this PR as this is being tracked in https://github.com/usebruno/bruno/pull/718

helloanoop:
Nice work @thehorse2000 👏 ! I have merged your PR #718

Many thanks again @dw-0 for your comments and feedback, it helped us implement this functionality in an elegant way 🙌

PR #718, opened 2023-10-21
Added properties to collections

thehorse2000:
# Description

This change introduces storing properties for collections.
Currently, two properties are stored in collections:
1. BASE URL which will be the initial default value for any new request within that collection
2. Default Request Type also sets initially the request type for new requests.

This is very helpful to avoid repetition of entering the same URL over and over again.
This also opens up more space for future customizations for the collections.

#### 
The collection properties are stored in the bruno.json file of the collection
### Contribution Checklist:

For previous comments of this PR. See https://github.com/usebruno/bruno/pull/707

image

https://github.com/usebruno/bruno/assets/7264374/2cf9516a-36e2-428a-a0c0-522558bb7da2


helloanoop:
@thehorse2000 Nice!

Fully onboard with supporting this. I

I'd like to recommend a few changes. 
The `properties` modal was kind of inspired by how we have properties of a file or folder in any OS.
I recommend putting this inside the collection settings and calling the Tab `Presets` 
The config can be stored inside `bruno.json` under the key `presets`
```json
{
  "presets": {
     "defaultUrl": "",
     "defaultRequestType": "",
  }
}
```

image

thehorse2000:
Hmm, I understand your view. But don't you think it would be more noticeable if we insert a new tab item for collections named Presets and introduce a new modal for it?
It will contain those two fields, And more hopefully in the future.
My concern is that it might not be noticeable at all if we set it in collection settings. Because I noticed that we can only access the collection settings from the request page. I didn't even notice it before you mentioned it 😢  

What do you think?
image

thehorse2000:
Hey, Thanks for the info. I actually didn't know there was a collections settings area 😿 
I think that is because we can only access it from requests.
I believe it would be better if we could access these settings from the menu of the collection (like properties)
But maybe that is for another PR 😄

thehorse2000:
### New Changes
- Changed "properties" to "presets" in bruno.json file
- Added Presets tab to collection settings
- Returned collection properties modal to its previous state.

image

https://github.com/usebruno/bruno/assets/7264374/39dd48e7-510a-4988-b662-42661eaff65a

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -94,6 +94,28 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
     }
   });
 
+  // update collection properties
```
helloanoop: @thehorse2000 You don't need a new electron event. There is already one which handles the updates for Bruno Config file.

https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/components/CollectionSettings/index.js#L34-L42

Can you update the code to use that?

packages/bruno-app/src/components/CollectionSettings/index.js
```diff
@@ -114,6 +118,9 @@ const CollectionSettings = ({ collection }) => {
         <div className={getTabClassname('headers')} role="tab" onClick={() => setTab('headers')}>
           Headers
         </div>
+        <div className={getTabClassname('presets')} role="tab" onClick={() => setTab('presets')}>
```
helloanoop: Lets place this tab after `Tests` and before `Proxy`

helloanoop (changes requested):
@thehorse2000 Nice work. I have a couple more comments. Once these addressed, I feel this should be good to merge.

martinsefcik:
@thehorse2000 @helloanoop Probably this Settings tab for collection could be opened by clicking on collection name in sidebar, the same way how it is working in Postman. It could be then also used to switch quickly environment, because now if I want to run whole collection or folder I have to unfold many subfolders find first request in collection open it and then I can switch environment which is not very user friendly.

Also this Collection properties in modal and Collection settings in tab could be merged to Collection tab. It doeas not make sense to me to have two places in app for similar things.

helloanoop:
> Probably this Settings tab for collection could be opened by clicking on collection name

Agreed.

> Also this Collection properties in modal and Collection settings in tab could be merged to Collection tab. It doeas not make sense to me to have two places in app for similar things.

Agreed. Thinking of a new Tab called `Info` which will display the same stuff that we display in `Properties`

thehorse2000:
@helloanoop Resolved the required changes.
I do also agree with you and @martinsefcik opinions on making the collections settings page more accessible.
So I will work on it, if you guys don't mind.
Will create a new PR for it and mention you

helloanoop (approved):

helloanoop:
Nice work @thehorse2000 !

PR Merged.

helloanoop:
> Probably this Settings tab for collection could be opened by clicking on collection name in sidebar

@martinsefcik This is done.

> Also this Collection properties in modal and Collection settings in tab could be merged to Collection tab. It doeas not make sense to me to have two places in app for similar things.

This is also done. See #989
I also removed `Properties` in the collection dropdown and added `Settings`

>  It could be then also used to switch quickly environment, because now if I want to run whole collection or folder I have to unfold many subfolders find first request in collection open it and then I can switch environment which is not very user friendly.

@martinsefcik I didn't understand. Can you explain this?
You should be able to select any request and switch the env on the top-right corner right ?

martinsefcik:
> > It could be then also used to switch quickly environment, because now if I want to run whole collection or folder I have to unfold many subfolders find first request in collection open it and then I can switch environment which is not very user friendly.
> 
> @martinsefcik I didn't understand. Can you explain this? You should be able to select any request and switch the env on the top-right corner right ?

@helloanoop If you have for example the following collection structure:

```
- folder 1
  - folder 1.1
    - folder 1.1.1
      - folder 1.1.1.1
        - request 1
        - request 2
        - ...
- folder 2
  - folder 2.1
    - folder 2.1.1
      - folder 2.1.1.1
        - request a
        - request b
        - ...
```
then you have to unfold first 4 folders to get inside folder with request, then open request and the switch environment.

So I meant this that it isn't/wasn't very user friendly.

garposmack:
I loved this new feature.
It's possible to go one step further, and have folders inside a collection to have the same settings and properties as a collection?

Issue #362, opened 2023-10-04
[Bug] Disabling SSL doesn't seem to work


deanveloper:
At the company I work for, we use self-signed certificates for our development environment. In Postman, we need to add our own CA, which is done (in postman) via Settings > Certificates > CA Certificates > PEM File.

It doesn't appear that bruno supports this, although there is an option to disable SSL Certificate Authentication. But when I do that, I get the error `Error invoking remote method 'send-http-request': Error: Key is required`.

lared:
Can you replicate it with any of those domains?

https://badssl.com/

TbirdDuncan:
Resolved it. There's no documentation on this except for https://github.com/usebruno/bruno/issues/168. You go to the bottom left of the app and theres a tiny gear cog with the options to disable ssl

deanveloper:
> Can you replicate it with any of those domains?
> 
> https://badssl.com/

Unfortunately not. Not sure if it's at all helpful, but the error that I get is `Error invoking remote method 'send-http-request': Error: unable to verify the first certificate`

Lotti:
@deanveloper you just have to do a get call to https://self-signed.badssl.com/ to reproduce the problem.

Disabling SSL self-signed check doesn't work (I used the cog in the bottom left corner)

helloanoop:
@Lotti Are you on the latest version `v0.22.0` ?

This is working on my system. MacOS Moneterey Intel x64

![image](https://github.com/usebruno/bruno/assets/9350733/b1bc8d4a-fbe1-4b6f-b503-2cc8cf0253c8)

Lotti:
Thank you for your fast reply.
I'm running in windows 10 with v0.22.0 (installed this morning to replace postman)

helloanoop:
Can you join our [discord community](https://discord.com/invite/KgcZUncpjq). I'll ask our windows user there to check this if they are experiencing the same.

deanveloper:
Can the findings be posted here? It'd be nice to have a solution posted in an indexable/archivable location

Lotti:
We find out that the bug exists only when using a proxy

lllama:
Echoing what @Lotti says: if I enable a proxy (Burp in this case) then I get the cert error.

cwilper:
I got the same issue when using a proxy that self-signs certs in proxied HTTPs traffic, even after checking "SSL Certificate Verification" is un-checked. This is still happening as of `v0.25.0`.

**Workaround:**

[...]

PR #725, opened 2023-10-22
fix: respect rejectUnauthorized and ca opts when proxying https

cwilper:
This PR fixes https://github.com/usebruno/bruno/issues/362

### Problem

When Bruno would make SSL/TLS connections through a proxy, it would not respect connection options like `rejectUnauthorized` or `ca`. This was making it difficult to use when going through a self-signing MiTM proxy, or using a custom certificate authority.

### Cause

The issue was caused by an upstream bug in [https-proxy-agent](https://github.com/TooTallNate/proxy-agents/tree/main/packages/https-proxy-agent), where it doesn't permit options  to be passed down when upgrading sockets to TLS:

* https://github.com/TooTallNate/proxy-agents/issues/194

The same underlying issue has been reported against multiple versions of that library over the years. As you can see in the comments, people have either switched to other libraries, or patched it to pass down such options:

* https://github.com/TooTallNate/proxy-agents/issues/89
* https://github.com/TooTallNate/proxy-agents/issues/121

### Solution

The solution employed here is to patch the class, similar to [what others have done](https://github.com/pnpm/components/blob/02d17cc303c94dc7d85458fd233303edcc452c8b/network/proxy-agent/proxy-agent.ts#L146-L157).

When/if the issue is fixed in `proxy-agents`, that should be used instead.

### Note to Reviewers:

* I put this in the existing `proxy-util.js` file, which is currently duplicated for CLI and Electron. I didn't see a good way to just have one copy of the code. Maybe the JS package would make sense eventually -- there seems to be lots of opportunity for consolidating common code there. 
* I was able to test successfully using the UI, but not the CLI -- I think there may be a separate bug in CLI currently where the Collection-configured proxy is not actually attempted to be used -- the code seems to look for `proxy.enabled: true` but it's saved as `proxy.use: true` in my `bruno.json` file in the collection dir. I didn't try to address that as part of the PR, as it seems a separate topic.

### Contribution Checklist:

- [x] **Linked issue to the pull request.**

### DIFF
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -12,11 +12,10 @@ const { ScriptRuntime, TestRuntime, VarsRuntime, AssertRuntime } = require('@use
 const { stripExtension } = require('../utils/filesystem');
 const { getOptions } = require('../utils/bru');
 const https = require('https');
-const { HttpsProxyAgent } = require('https-proxy-agent');
 const { HttpProxyAgent } = require('http-proxy-agent');
 const { SocksProxyAgent } = require('socks-proxy-agent');
 const { makeAxiosInstance } = require('../utils/axios-instance');
-const { shouldUseProxy } = require('../utils/proxy-util');
+const { shouldUseProxy, PatchedHttpsProxyAgent } = require('../utils/proxy-util');
 
 const runSingleRequest = async function (
   filename,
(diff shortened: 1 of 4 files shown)


helloanoop:
@cwilper You're awesome !!!

> there seems to be lots of opportunity for consolidating common code there.

Yup, post out v1 launch, we will consolidate this.

> I was able to test successfully using the UI, but not the CLI -- I think there may be a separate bug in CLI currently where the Collection-configured proxy is not actually attempted to be used -- the code seems to look for proxy.enabled: true

Yup, this discrepancy needs to be fixed. We will be using `enabled` in both places.

Thanks again @cwilper !

Merged!

PR #733, opened 2023-10-23
[bugfix] Put scripts in comments in Postman import

survivant:
# Description

Bug fix to put scripts in comments in postman import

### Contribution Checklist:


helloanoop:
This approach would break if the script already has /* */ comments

The safest way is to iterate over each line and just add // in the beginning of each line 

Or best of both worlds.
Check if the script has. /* match.
If match found, then // on each line
Else just wrap the whole script like you already have done using /* */

I am fine with either approach

survivant:
Ok, I will do that tomorrow, I just wanted to do a quick fix, but I agree
with you.

Le lun. 23 oct. 2023, 07 h 25, Anoop M D ***@***.***> a
écrit :

helloanoop:
Sounds great @survivant !

Merged !!

PR #755, opened 2023-10-23
feat: Improve respone & request pane styling

Its-treason:
# Description

Adds "PaneContent"-Component that wraps all response / request panes to ensure that overflow will never cause the page to scroll & styling of all tabs to look the same.

- Updated body select to be inside pane header
- Updated auth select to use pane head styling
- Update Docs Edit button style to match pane head style
- Show Docs editor if docs are empty

Currently:

https://github.com/usebruno/bruno/assets/39559178/81621f58-0459-4fd0-8a0f-4e13b50d702b

Updated:

https://github.com/usebruno/bruno/assets/39559178/493d0311-041a-473e-a9f6-bfb0a75ee102


martinsefcik:
Great work!

helloanoop:
Awesome @Its-treason !

There are lots of files, I need some time to review this. A bit held up prepping for the upcoming v1 release.
This will be a priority item to be merged first thing next week.

martinsefcik:
Similar tunning would be needed also for runner tab. It is needed to have two scrollbars for left part of the tab (with executed request list) and for the right part of the tab (specific request response preview), because currently both parts (left and right) are scrolled with only one scrollbar and If I run many requests in runner and I want to click on requests which are at the end of list to see the response preview for such request I have scroll top to see the response and then If I want to click on another request at the end of the list then I have to again scroll down to click on it and scroll top to see response preview, etc. ... It is quite annoying

helloanoop:
Will get to this PR once I get the Digest and OAuth 2.0 completed.

Its-treason:
I think this is MR is too big and i have don't want to fix all the conflicts. Also fixing all issues like in #1058 might be better

sanjai0py:
Hey @Its-treason, love your work! Need help with merge conflicts?

Its-treason:
No, I think this is a too complicated solution for the style problem. Implementing the fix from #1058 on the other panels / tabs would be better.

Issue #740, opened 2023-10-23
Local development setup not working if port 3000 is used


Oryss:
**Just letting people know about this issue, if you're trying to setup the local development and run into this issue.
This was fixed by shutting down the process which was using port 3000 and restarting dev:web**

I have followed the development.md file to setup the project but the app does not load and display a white error page : 

![image](https://github.com/usebruno/bruno/assets/43440732/f3be7dc6-6a3a-4758-8146-e8029e7089c7)

When using npm run dev:web, the app loads but I cannot select the location of the collection 

![image](https://github.com/usebruno/bruno/assets/43440732/d7955bc9-0180-4245-be5f-fc5e3eb7ac76)

I'm on Ubuntu 22.04.3 LTS

helloanoop:
During Development, Bruno cannot load directly inside a browser. It only loads inside of Electron 

You need to run
`npm run dev:electron` in a second terminal

Its-treason:
We could look into forcing Next to use port 3000 and abort if it's not available. This should help avoid confusion when the port is unavailable.

Oryss:
> During Development, Bruno cannot load directly inside a browser. It only loads inside of Electron
> 
> You need to run `npm run dev:electron` in a second terminal

Ok ! But the Electron window crashes (second screenshot)

I fixed it

helloanoop:
> We could look into forcing Next to use port 3000 and abort if it's not available. This should help avoid confusion when the port is unavailable.

Yes. 

Also, I would recommend showing a popup saying that Bruno can only be loaded on electron while a user tries load the app directly on browser.

PR #768, opened 2023-10-24
feat(#740): Add hints for local bruno setup

Its-treason:
Fixes: #740

- Custom error when opening bruno in the browser

grafik

- Custom error when starting electron without the web server running

grafik

- Force Next dev server to use port 3000
  - Added `-p 3000` to next dev command


helloanoop:
This is cool! You just saved a lot of future github issues :P

Thanks @Its-treason !

Issue #653, opened 2023-10-18
Relative client certificate paths do not work


tibeer:
relates #275

Even though mentioned in the PR that relative certificate path should work, I cannot get them to work.
I placed them directly within my collection folder next to my bruno.json which looks like this:

```json
{
  "version": "1",
  "name": "tibeer",
  "type": "collection",
  "clientCertificates": {
    "enabled": true,
    "certs": [
      {
        "domain": "api.tibeer.de",
        "certFilePath": "prod.cer",
        "keyFilePath": "prod.key",
        "passphrase": ""
      },
      {
        "domain": "api-test.tibeer.de",
        "certFilePath": "test.cer",
        "keyFilePath": "test.key",
        "passphrase": ""
      }
    ]
  }
}
```

Might also be an user error.

mj-h:
Interesting: They work in the CLI, but not in the UI. I'm having a look.

tibeer:
Aweseome job @jwetzell! Closing this :)

PR #774, opened 2023-10-24
fix cert/key loading in electron for relative path

jwetzell:
- add `collectionPath` to `configureRequest` in electron
- use `collectionPath` to resolve relative paths to cert/key if the configured path is not absolute

addresses issue #653


helloanoop:
Nice! Thanks for taking care of this @jwetzell !

Merged!

PR #819, opened 2023-10-29
fix: trimming variable path

### LINKED ISSUE #936: [Bug] Variable not showing if it has a leading or trailing white space
when having spaces in the left or right of a variable inside the `{{ }}` syntax bruno does not find the variable that is defined in the environment, example:
image

giuliano-macedo:
fixes #936
# Description

Fixes a bug where when having spaces in the left or right of a variable inside the `{{ }}` syntax bruno would not find the variable:

before the fix:
image

after the fix:
image

I would happily add tests for this if it is needed, but I need some help :>)

### DIFF
--- a/packages/bruno-app/src/utils/common/codemirror.js
+++ b/packages/bruno-app/src/utils/common/codemirror.js
@@ -9,7 +9,7 @@ if (!SERVER_RENDERED) {
 }
 
 const pathFoundInVariables = (path, obj) => {
-  const value = get(obj, path);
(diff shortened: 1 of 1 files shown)


helloanoop:
Hey @giuliano-macedo 
Sorry for being late to the PR review
Thanks for taking time to work on this PR. For now, we'd like to keep the interpolation logic strict.
We may consider this in the future, if there is a wider interest from the community or a strong reason on why the variable must be trimmed.

PR #832, opened 2023-10-30
fix (#680): fixed non-string variables to be displayed correctly in code editor

### LINKED ISSUE #680: [Bug] Collection variable is defined and has value but still displayed as red in request address or query params tab
Sometimes it happens that collection variable is defined and has any value set (I can see it on Variables tab), but it is still displayed with red color in address bar or in Query params table on request's tab. Also it is not possible to see value on tooltip during moving cursor over it. If I run request then this variable is correctly send with proper value.

image
image
image

It's happening somehow randomly and I don't know how to reproduce it.

Bruno version: 0.25.0

martinsefcik:
# Description

Fixes #680.

Non-string collection variables are displayed in code editor as red without tooltip with variable value. So this PR fixes this bug and they are now correctly displayed as green also in cases value is null. They are red only in case they are undefined.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/utils/common/codemirror.js
+++ b/packages/bruno-app/src/utils/common/codemirror.js
@@ -10,7 +10,7 @@ if (!SERVER_RENDERED) {
 
 const pathFoundInVariables = (path, obj) => {
   const value = get(obj, path);
-  return isString(value);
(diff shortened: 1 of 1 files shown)


helloanoop:

PR #833, opened 2023-10-31
fix (#402): response/body cursor offset on resize

### LINKED ISSUE #402: [Bug] Adjusting divider between response/body will break body editing
Sometimes the body is complex and requires adjusting the divider to see especially long text in 1 line. The problem is that it will mess up the place that the text cursor shows vs where it is actually editing. The only way to really fix it is to exit the tab and come back which will reset the divider back to normal. It's more noticeable if the divider is changed drastically. This is a small issue but it can be annoying.

It appears to fix itself if you edit 1 character but that 1 character is not where the text cursor shows it to be.

n00o:
Now response/body will be refresh each time it gets resized so the cursor always stays in the right place.

# Description
This will fix #402 and make it easier to adjust the response/body window size to see them better while not losing the cursor. It might not be very efficient if the response/body is very large (haven't noticed any issues yet).

### before
https://github.com/usebruno/bruno/assets/115114979/49d53071-7a27-45d2-ad39-52754efb2508

### after
https://github.com/usebruno/bruno/assets/115114979/7c07201e-5f8a-4c8a-b0ad-f8f45d7daac3

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -146,6 +146,9 @@ export default class CodeEditor extends React.Component {
   }
 
   render() {
+    if (this.editor) {
+      this.editor.refresh()
+    }
     return (
       <StyledWrapper
         className="h-full w-full"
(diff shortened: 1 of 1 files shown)


helloanoop:
Nice! Thanks for taking care of this @n00o !

Issue #837, opened 2023-10-31
JSON in the ResponsePane / QueryResult is not indented for broken APIs


Joschasa:
Some APIs return JSON payload using the Content-Type "application/html" header.

This is parsed as JSON by Bruno, but without added indentation.

Before:
![image](https://github.com/usebruno/bruno/assets/213920/4ff3206f-c668-443b-9f7d-00911353ad79)
(Headers:)
![image](https://github.com/usebruno/bruno/assets/213920/7d926f20-b1fa-4432-9ff1-ac075ef32da3)

If the JSON is parsable it should (in my opinion) be indented as well. This makes working with those broken APIs a lot easier.

sanjai0py:
Closing this issue as your [PR](https://github.com/usebruno/bruno/pull/838) implementing the feature has already been merged and released in [v1.5.0](https://github.com/usebruno/bruno/releases/tag/v1.5.0).

Thanks for taking the time to report this suggestion and also a big thanks for the fix PR!

PR #838, opened 2023-10-31
Feature/always indent json in queryresult

Joschasa:
# Description

An unused condition was removed and the QueryPreview window should always indent the JSON payload.
Even for APIs with a Content-Type info of "application/html".

Before:
image
(Headers:)
image

After:
image

However the resulting JSON is not highlighted / foldable, as the CodeEditor still treats this as HTML.

We could leave this as is, so that one might recognize the API returns the wrong Content-Type.

### Contribution Checklist:

- [x] **Create an issue and link to the pull request.** https://github.com/usebruno/bruno/issues/837


Its-treason:
Hey, @Joschasa, I think it would be better if we instead edit the `getCodeMirrorModeBasedOnContentType` function to return the correct mode instead of changing just the formatting. This way JSON will also be highlighted correctly.

I think you can just pass the body to `getCodeMirrorModeBasedOnContentType` and check if it's an object or something and then return the `application/ld+json` mode.

Its-treason:
> Hi @Its-treason, is this what you had in mind?

Yes, and you should be able to revert the changes to the `formatResponse` function.

Looks good, nice work!

Its-treason (approved):

Joschasa:
Any update on this?

Issue #338, opened 2023-10-04
Convert curl commands to request.


sanskar-mk2:
In Insomnia, whenever you post a curl request to the url bar, it autofills the headers/url/parameters/creds accordingly to that curl request. It is very important feature since curl is very universal to http requests. You can copy curl directly from browser.

- [x] Import Curl when creating a new request
- [x] Switch to curl mode automtically when a curl command is pasted in NewRequest modal
- [ ] Allow importing of curl when a curl command is pasted in an already opened tab

cvmocanu:
+1

In addition, an "Import from curl" option on the directory menu (next to "New request") should be added, to make it obvious that Bruno has this feature.

davidsr2r:
If you can't export to curl, that would also be great to have added. Apologies if it's already a feature, I couldn't see any obvious place in the docs that point that out & GOOG simply pointed me here.

Olian04:
@davidsr2r You can export a request as a curl command from the "generate code" menu.
<img width="248" alt="Screenshot 2023-10-13 at 09 01 56" src="https://github.com/usebruno/bruno/assets/5493406/54aa72af-0a50-41af-ac9a-aa404a5e831f">
<img width="848" alt="Screenshot 2023-10-13 at 09 02 11" src="https://github.com/usebruno/bruno/assets/5493406/366e3e38-c4a1-40a1-b993-101d82aceb5f">

snippetkid:
@sanskar-mk2 @helloanoop @cvmocanu I'll start working on this issue.

snippetkid:
A quick update, I'm through with the changes for this and is currently working on a bug while updating headers from the curl command.

helloanoop:
Would love to see a teaser video :)

snippetkid:
https://github.com/usebruno/bruno/assets/16718733/2f1caea9-ffee-4e29-998a-f96d8017109c

@helloanoop Here is what it looks like now. Please pass on review comments.

helloanoop:
Curl Import functionality is now available in `1.1.0` 🎉

girishkumar9:
@helloanoop it would be good to also have a curl script generated on an existing request. Postman has this feature and it is very handy to show the output of an API call

cvmocanu:
@girishkumar9 that is already implemented: hover your mouse cursor over a request in the left-side tree, click the 3 dots, Generate Code, Shell-Curl.

tbkahuna48:
[...]

PR #840, opened 2023-10-31
Add new request from cURL command

snippetkid:
# Description

This PR add the capability to create an HTTP request from a valid cURL command (fixes https://github.com/usebruno/bruno/issues/338)

### Contribution Checklist:

### Changes:
- Added **From Curl** option in the New Request modal.
- Added `parse-curl` as a dependency and new utility method `getRequestFromCurlCommand` to parse cURL commands

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12178,6 +12178,14 @@
         "xml2js": "^0.4.5"
       }
     },
+    "node_modules/parse-curl": {
+      "version": "0.2.6",
+      "resolved": "https://registry.npmjs.org/parse-curl/-/parse-curl-0.2.6.tgz",
+      "integrity": "sha512-ENhXeIxG4A6wFvYSU87b0o3Tp6U+Wup069GYhXCn0ZP/E7evfvomTSM/MeEOs3QTyuye+u2r7fFRj6nX0q9kQA==",
+      "dependencies": {
+        "shellwords": "^0.1.0"
+      }
+    },
     "node_modules/parse-headers": {
       "version": "2.0.5",
       "dev": true,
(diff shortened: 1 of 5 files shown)


snippetkid:
https://github.com/usebruno/bruno/assets/16718733/890699ab-c5fc-4ebe-b495-bc68ce0d37a3

Olian04:
@snippetkid how does this handle the  [-k, --insecure](https://curl.se/docs/manpage.html#-k) flag? Seeing as bruno currently doesn't support disabling TLS/SSL certificate verification on an individual request basis.

snippetkid:
@Olian04 I think your question answers itself. Since bruno doesn't support insecure requests, so does the PR. Neither do I see `-k, --insecure` in Insomnia's supported args (which you can take a look [here](https://github.com/Kong/insomnia/blob/e9e1d27add7490b32c33ca1e83b93e0f7f2f54e6/packages/insomnia/src/utils/importers/importers/curl.ts#L12)) so my question would be do we really need that?

Apologies but I fail to understand what is improper in this PR. I believe this PR might be a valuable addition for common use cases, even in its current state. What I suggest is an incremental approach. We can proceed with merging this PR to offer this feature to users. This way, we can provide immediate value to users while working towards a more comprehensive feature.

I am keen to hear your thoughts on this @helloanoop

Olian04:
@snippetkid I'm sorry but I fail to see how quoting another inadequate implementation (Insomnia) is a good argument for our implementation to also be inadequate.

I would accept this PR as an incremental implementation if it parsed all flags but warned the user when an unsupported flag was detected.

helloanoop:
@Olian04 A lot of users are waiting for curl support #338
We will consider this as an incremental implementation. 

> I bring up this flag in particular since my current use case for bruno requires some requests to be made insecurely.  However it only supports it as a global option, not on a request by request basis.

Can you raise an issue for this. I am not sure how other apps do this at a request level (via UI). We should be able to clearly support an api that can turn of ssl verification at the request level `req.disableSslVerification()`

helloanoop:
Nice work @snippetkid !

Merged !! This will go out in the upcoming 1.1.0 release

snippetkid:
Thanks for the merge @helloanoop 🚀 I'll keenly follow this feature and related issues and come up with ways to improve the curl parsing. Insomnia, Hoppscotch and RecipeUI  have their own curl parsing, so I guess we should write our own parser and replace the `parse-curl` package in the future if it has shortcomings.

PR #863, opened 2023-11-02
feat: add class when collection name is hovered during drag and drop

### LINKED ISSUE #849: No UI feedback when moving a request between folders in a collection
There's no feedback to the user when dragging  a request between folders as shown above. Either a border around the destination folder or highlighting it (as VSCode does in the video below)would be appreciated

**Current Behavior**

https://github.com/usebruno/bruno/assets/71697384/a3228839-71b2-49d7-bee4-3d94db478f6f

**Expected Behavior (taken from VSCode)**

https://github.com/usebruno/bruno/assets/71697384/faa57044-8079-44cd-a949-b4092cb92327

mikaoelitiana:
Resolves #849

# Description

During drag and drop, the target collection has a `isOver` property provided by the `collect` function 

https://github.com/usebruno/bruno/blob/c46fbfbde39b025c5e1692477508c2162b65f4aa/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js#L60-L62

We use that to add a new class to the collection item and that class has the same property as the existing `:hover` 

### Contribution Checklist:

https://github.com/usebruno/bruno/assets/674667/4ff5bd47-fdcb-4ebd-abf7-bef91477b653

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/StyledWrapper.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/StyledWrapper.js
@@ -33,7 +33,8 @@ const Wrapper = styled.div`
       overflow: hidden;
     }
 
-    &:hover {
+    &:hover,
+    &.item-hovered {
       background: ${(props) => props.theme.sidebar.collection.item.hoverBg};
       .menu-icon {
         .dropdown {
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
@@ -84,7 +84,8 @@ const CollectionItem = ({ item, collection, searchText }) => {
   });
 
   const itemRowClassName = classnames('flex collection-item-name items-center', {
-    'item-focused-in-tab': item.uid == activeTabUid
+    'item-focused-in-tab': item.uid == activeTabUid,
+    'item-hovered': isOver
   });
 
   const scrollToTheActiveTab = () => {


glokta1:
The highlighting works now but I'd want to see the request inside the new destination folder instead of being hidden

mikaoelitiana:
> The highlighting works now but I'd want to see the request inside the new destination folder instead of being hidden

On my previous example, the destination folder was empty but here is what it looks like when there is something in there. Is that what you are looking for?

https://github.com/usebruno/bruno/assets/674667/ab23a0bc-c6ef-43fa-897c-b1723204c7c2

glokta1:
Ah I see. Two things:

1. It would be better to show the request in its new folder, even if it was previously empty, no?

2. Highlighting the destination folder AND its existing contents is better feedback than highlighting every list item (see VSCode example in linked issue)

helloanoop:
@glokta1 Agree with you. We'll keep https://github.com/usebruno/bruno/issues/849 open and improve it.

mikaoelitiana:
@glokta1 to clarify what you are expecting: 

> It would be better to show the request in its new folder, even if it was previously empty, no?

Here you would like the dragged request to be displayed inside a target folder __before__ it is actually dropped there if I am getting it right?

glokta1:
No, *after*. See the original vid you posted where after dragging the request into the target folder, I don't actually see my request in the folder unless it already has existing requests (your 2nd vid)

mikaoelitiana:
@glokta1 actually in my first video, the target folder was empty and __collapsed__, but if the folder was expended, it would display as expected : 

https://github.com/usebruno/bruno/assets/674667/2b06b316-c5b0-430f-9f37-48ee9a33ffbd

Issue #853, opened 2023-11-01
Can't create environment variable with a "-" in the name


ljohnston:
I have environments with variable names like `some-service-url`. This has worked in the past. I just upgraded to 1.0.1 and it seems that dashes are not allowed in variable names. Is this intentional or is this a bug?

<img width="350" alt="image" src="https://github.com/usebruno/bruno/assets/361524/f67d1ff6-23c7-407e-80d8-19153bc9dcc5">

Note that this creates issues when importing a Postman collection (where variable names with dashes _are_ allowed). After importing a Postman collection that has requests defined as `GET {{some-service-url}}/api/v1/...`, for example, you can't create an environment variable in Bruno that will allow that request to be sent. To get things to work, you'd first have to modify every request imported from the collection to remove the dashes.

n00o:
I think this intended and was added on merge request #791. This is something that I also noticed because I have to pass a header that uses "-" for authorization and couldn't because both headers and environment variables can't have that character.

I would think the regex statement used for validation should be `^(?!\d)[\w-]*$` instead of `^(?!\d)\w*$`. Although there could be other characters that Postman and others allows that we should add for compatibility reasons.

helloanoop:
@ljohnston We will add support back or `-`. See https://github.com/usebruno/bruno/issues/682#issuecomment-1790236386

> To get things to work, you'd first have to modify every request imported from the collection to remove the dashes.

Will release a patch fix soon, meanwhile as a temporary workaround - you can open the collection folder in vscode/any editor and do a find and replace 

![image](https://github.com/usebruno/bruno/assets/9350733/f880feb0-e257-4187-9de2-0e4ad7e5bc88)

Issue #682, opened 2023-10-19
Variable Names - Enforce Strictness


helloanoop:
Enforce strict variable naming rules.

A variable name must start with a letter or `_`
A variable name cannot start with a digit
A variable name can only contain alpha-numeric characters, `_` (a-z, A-Z, 0-9, _ )

If a variable is found to not adhere to these rules, display an error in the UI. The user shouldn't get frustrated as to why the request its not working as expected.

Variable Naming rules apply to
1. Env Vars
2. Request Vars
3. Collection Vars
4. Dynamic Vars `bru.setVar()` and `bru.getVar()`

Also see: https://github.com/usebruno/bruno/issues/345

mirkogolze:
Hi @helloanoop, @Its-treason,

In the implementation am missing the "_" defined in the requirement. And "-" should be allowed too.

helloanoop:
> missing the "_" defined

We should add this immediately

> And "-" should be allowed too.

The `-` is atypical for a variable name (in the javascript world), but I see that other apps (#853) use it. I am fine with loosening the restriction around `-`

lujop:
This is greatly impacting us because we have collections from previous versions that heavily use variable names with `-`.

helloanoop:
@lujop This will be fixed in our upcoming release 1.1.0 which will be live in the next few hours.

helloanoop:
@lujop You can use `-` in the latest release 1.1.0

Thanks @Its-treason for getting this done !

PR #866, opened 2023-11-02
fix(#853): Allow - in variable names & small refactor

Its-treason:
# Description

Closes: #853 & #682

Allow `-` in variable names, update error messages & refactor the regex into a constant.


helloanoop:

PR #903, opened 2023-11-06
fix: Fix Content-Security-Policy config

Its-treason:
# Description

Fixes this electron startup error:

```
(node:9856) UnhandledPromiseRejectionWarning: Error: Each line must end in a semicolon
    at exports.setContentSecurityPolicy (C:\Users\Timon\projects\bruno\node_modules\electron-util\index.js:150:9)
(Use `electron --trace-warnings ...` to show where the warning was created)
(node:9856) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```
This happened because the last line was missing a semicolon.

I also updated the CSP because, the app would not start with the old CSP.

### DIFF
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -14,16 +14,18 @@ const { loadWindowState, saveBounds, saveMaximized } = require('./utils/window')
 
 const lastOpenedCollections = new LastOpenedCollections();
 
+// Reference: https://content-security-policy.com/
 const contentSecurityPolicy = [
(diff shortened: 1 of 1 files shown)


helloanoop:
Thanks for fixing this @Its-treason !

Issue #901, opened 2023-11-06
Unwanted field appears in the headers


jeanchristopheMICHEL:
When I create a new simple GET request, without body and without any field in header, the request sending add "Content-Type: application/json".
You can do that, by creating a request and use "generate code" (with any format you want).

Thanks,

helloanoop:
Agreed, We need to fix the `Code Generator` not to add additional headers.

martinsefcik:
@helloanoop I guess this issue was fixed by #938 and can be closed.

anusreesubash:
Closing this issue as this has been fixed. Please reopen if the issue persists.

PR #280, opened 2023-10-02
Feature/code generator

Beedhan:
I've added a code generator based on the discussion in #58.

PR #938, opened 2023-11-09
fix(#901): fix headers in code generator

adarshlilha:
# Description

Closes - #901 
While debugging the issue stated in #901, I found another issue(custom headers not being shown in code-generator at all, possibly missed in #280)

This PR addresses both of these:
* unwanted header - `content-type` added in code generator by default
* custom header not added in code generator at all

Snapshot: - pst
Screenshot 2023-11-10 at 03 33 29


helloanoop:
Nice. Thanks for taking care of this @adarshlilha !

PR #971, opened 2023-11-15
fix(#964): Allow "." in variable names + make error message more consistent

### LINKED ISSUE #964: Can't create environment variable with a "." in the name (Postman import issue)
Similar to #853 ... We have Postman environments with variable names like `some.service.url`. This creates issues in Bruno when importing a Postman collection (where variable names with dots are allowed). After importing a Postman collection that has requests defined as `GET {{some.service.url}}/api/v1/...`, for example, you can't create an environment variable in Bruno that will allow that request to be sent. To get things to work, you'd first have to modify every request imported from the collection to remove the dots.

nelup20:
Closes #964 

Thanks for creating this project btw! I was getting really disappointed with the alternatives (Insomnia, Postman, etc.) before stumbling upon Bruno on HN, and it's exactly what I was looking for.

|Before|After|
|-------|------|
|image|image|
|image|image|
||image|

# Description
Added `.` to the Regex, removed the now unnecessary check with the old regex & changed the error messages to be a bit more consistent.

Also added literal-segment notation with `[]` around the var name when interpolating since Handlebars doesn't allow `.`

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/utils/common/regex.js
+++ b/packages/bruno-app/src/utils/common/regex.js
@@ -1,1 +1,1 @@
-export const envVariableNameRegex = /^(?!\d)[\w-]*$/;
+export const envVariableNameRegex = /^(?!\d)[\w-.]*$/;
--- a/packages/bruno-js/src/bru.js
+++ b/packages/bruno-js/src/bru.js
@@ -1,7 +1,7 @@
 const Handlebars = require('handlebars');
 const { cloneDeep } = require('lodash');
 
-const envVariableNameRegex = /^(?!\d)[\w-]*$/;
+const envVariableNameRegex = /^(?!\d)[\w-.]*$/;
 
 class Bru {
   constructor(envVariables, collectionVariables, processEnvVars, collectionPath) {
@@ -64,7 +64,7 @@ class Bru {
     if (envVariableNameRegex.test(key) === false) {
       throw new Error(
         `Variable name: "${key}" contains invalid characters!` +
-          ' Names must only contain alpha-numeric characters, "-", "_" and cannot start with a digit.'
+          ' Names must only contain alpha-numeric characters, "-", "_", "." and cannot start with a digit.'
       );
     }
 
@@ -75,7 +75,7 @@ class Bru {
     if (envVariableNameRegex.test(key) === false) {
       throw new Error(
         `Variable name: "${key}" contains invalid characters!` +
-          ' Names must only contain alpha-numeric characters and cannot start with a digit.'
+          ' Names must only contain alpha-numeric characters, "-", "_", "." and cannot start with a digit.'
       );
     }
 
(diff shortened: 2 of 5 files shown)


nelup20:
Hi @helloanoop 👋, when I opened this PR I tested everything manually and everything seemed alright, but I had some questions about the e2e & unit tests:

- Do the playwright tests fail at the moment for everyone on the main branch? Or is it just for me? I suspected it's a local issue just on my side, but if not I was thinking of looking into them.
- I was thinking of adding more unit tests in a separate PR to the `bruno-electron` package, maybe also to `bruno-js`. What do you think? Is there a specific test coverage % we'd like to reach?

packages/bruno-app/src/utils/common/regex.js
```diff
@@ -1 +1 @@
-export const envVariableNameRegex = /^(?!\d)[\w-]*$/;
+export const envVariableNameRegex = /^(?!\d)[\w-.]*$/;
```
bbbco, line 1: I believe `.` needs to be escaped in order to be considered a literal `.`, otherwise, `.` refers to _any_ character.

packages/bruno-js/src/bru.js
```diff
@@ -1,7 +1,7 @@
 const Handlebars = require('handlebars');
 const { cloneDeep } = require('lodash');
 
-const envVariableNameRegex = /^(?!\d)[\w-]*$/;
+const envVariableNameRegex = /^(?!\d)[\w-.]*$/;
```
bbbco, line 4: Why don't we import this from `packages/bruno-app/src/utils/common/regex.js`?

packages/bruno-app/src/utils/common/regex.js
nelup20: Within a character set/class (`[]`) the `.` is treated as a literal instead of the usual "match any character" 👍

(side note: I used both https://regexr.com/ & https://regex101.com/ while testing, and the former showed no issues with this regex, but the latter displayed an error for the `-` which can be solved by escaping it. Not sure why there's a difference between them since they should both use the browser's JS regex engine)

packages/bruno-js/src/bru.js
nelup20: Great point, I agree and I'd also prefer to import it from one place instead of duplicating it. I wasn't sure whether to refactor it in this PR or a separate PR (not sure how it would be preferred in this project).

In this specific case, the `bruno-app` package is the standalone Next.js app, whereas the `bruno-js` package is a library used by the other packages in the codebase, so I think we can export the regex from `bruno-js` (or maybe another common/util package).

nelup20: I'm thinking of refactoring after all, otherwise I'll create a separate PR to keep this one small/focused
@helloanoop what do you think?

PR #1032, opened 2023-11-23
feat(#303): Default Environment

### LINKED ISSUE #303: [Feature] Default environment
[JIRA](https://usebruno.atlassian.net/browse/BRU-2528)

Currently, when you open a request for collection, you always have to first set the environment.

I suggest we have a default environment that will be used upon opening a request collection if user didn't set anything. It can be settable either in collection properties (selecting existing environment) or in environment config (ticking a default checkbox).

This way, executing requests can be made to work immediately upon opening them and CLI will work without mandatory `--env` parameter.

n00o:
Add environment to the presets tab which will set the default environment on opening a collection. Show default environment inside of environment selector. Also add cli support for default environments.

# Description

This is to fix #303 by adding the property called `environment` inside of `Collection > Presets`. The collection descriptor file, `bruno.json` will now have the preset structure of:
```json
{
  "presets": {
     "defaultUrl": "",
     "defaultRequestType": "",
     "environment": null
  }
}
```
The preset environment will only change the environment under these conditions:
* The collection is first being opened
* The collection is being opened again and the current session had the environment set to `No Environment` (if its something else then it won't set the environment)

In the future we can rely on something like `lastEnvironment` to override the preset environment if people set it in their preferences.

https://github.com/usebruno/bruno/assets/115114979/09145231-3f86-48be-9af7-0bf5e65df067

The environment selector now has a helpful indicator showing what the default environment is by the yellow `(Default)` value at the end of it. Any old collections that don't have a preset environment or the environment can't be found in the current list of environments will default to `No Environment` insuring that nothing breaks. 

Some behavior that I need some guidance on @helloanoop, would of the formatting of `(Default)` indicator and whether to have display next to `No Environment` if an environment preset is defined but not found. Currently I am setting it to `No Environment` in the environment selector and showing an error message on collection open when environment can't be found.

Now Bruno cli will try to look for the new `environment` preset value and set it as the environment if none is given.
Default_Environment_cli

I think this is a very helpful QOL change that will make opening and testing APIs a quick and easy process.
### Contribution Checklist:


Napam:
I personally would like something like this, is there any plans to incorporate such functionality? @helloanoop

naman-bruno:
Hi @n00o, thank you so much for the PR.

We’ve actually already addressed this by snapshotting the last selected environment (see PR #3215), which is now persisted on every launch.

thanks again for contributing!

PR #1043, opened 2023-11-23
fix(#521): Allow "context" as the name of a key/var in a JS expression

### LINKED ISSUE #521: Error accessing response property with 'context' in the name of the key
I'm trying to set a post response variable using `res.body`. In this case, the JSON response has a property named `login-context`, so I'm trying to access it like so:

```js
res.body["login-context"]
```

When I attempt to run the request, I receive the following error:

> Error invoking remote method 'send-http-request': SyntaxError: Identifier 'context' has already been declared

I also receive the same error if I use `res.body["context"]` (which isn't an actual property in the response), but `res.body.context` completes successfully (though it evaluates to `undefined` since it doesn't exist).

CleanShot 2023-10-10 at 17 15 48@2x

nelup20:
Closes #521 

# Description
Changed the name of the param used in the compiled JS function since "context" would clash with any key in the response which also contains "context".

|Before|After|
|-------|------|
|image|image|
||image|

### Contribution Checklist:

### DIFF
--- a/packages/bruno-js/src/utils.js
+++ b/packages/bruno-js/src/utils.js
@@ -18,8 +18,8 @@ const JS_KEYWORDS = `
  * ```js
  * res.data.pets.map(pet => pet.name.toUpperCase())
  *
- * function(context) {
- *   const { res, pet } = context;
+ * function(_BrunoNewFunctionInnerContext) {
+ *   const { res, pet } = _BrunoNewFunctionInnerContext;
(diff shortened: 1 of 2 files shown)


nelup20:
**Update**: Unfortunately I won't have time anymore to contribute to Bruno (and subsequently follow up on this PR in case of issues). I've allowed edits for maintainers so feel free to take over this PR or close it as you see fit 👍

Issue #1002, opened 2023-11-20
Error invoking remote method 'send-http-request': TypeError: Cannot read properties of undefined (reading 'toUpperCase')


cyrilgeorge153:
Getting error added in title while using Digest Auth in Bruno.

![image](https://github.com/usebruno/bruno/assets/52770689/371104f6-d79f-4e27-a6e6-6e807fa34a48)

rahul-kumar-sdet:
Similar issue is faced while using any JS function like Match etc in the script section

PR #1047, opened 2023-11-24
Fix(#1002): Digest Error

n00o:
This is to fix toUpperCase() error and uri path not being correct.

# Description
This is to fix(#1002). It appears that some sites that don't send the algorithm in the `www-authenticate` header so we can just default to MD5 so that it works for more sites. Also Digest Auth requires that the parameter uri needs to be a partial path instead of the full path so that has been fixed. Maybe some sites are more flexible but it's better to follow this process.
  
Digest_touppercase_error

### Contribution Checklist:


helloanoop:
Thank you for the fix @n00o !

Issue #1035, opened 2023-11-23
boolean response not displayed


corentingosselin:
Using postman

![image](https://github.com/usebruno/bruno/assets/8132994/4fa73c50-b645-4937-9e74-01718cdffb0f)

Using bruno

![image](https://github.com/usebruno/bruno/assets/8132994/c2917609-8dd2-46a8-a08d-74580904fb52)

PR #1055, opened 2023-11-25
fix(#1035): fix falsy check of response

adarshlilha:
Closes: #1035 

while rendering a API response, all falsy values were ignored
this is fixed with this PR
Screenshot 2023-11-25 at 14 23 58


helloanoop:
Thanks for taking care of this @adarshlilha !

PR #1061, opened 2023-11-26
fix(RequestPane): Fix Mardown Table styling

Its-treason:
# Description

Updated table styling in docs preview.

Before:

grafik

After:

grafik


helloanoop:

PR #1070, opened 2023-11-27
feat(#946): Send request even when protocol isn't specified in URL

### LINKED ISSUE #946: Connexion to localhost without http specified
Hello,
I'm trying Bruno v1.1.1 and I found out that specifying `http://` is mandatory. I wanted to know if this was intended because it is no mandatory in postman or insomnia.
I created simple API to try Bruno that I run locally on the port 8080 and here is what I get:

- `GET http://localhost:8080/` works fine
- `GET http://127.0.0.1:8080/` works fine
- `GET localhost:8080/` throws an error: Error invoking remote method 'send-http-request': AxiosError: Unsupported protocol localhost:
- `GET 127.0.0.1:8080/` throws an error: Error invoking remote method 'send-http-request': Error: connect ECONNREFUSED 127.0.0.1:80

Thanks for your work anyway, this is a really cool software

nelup20:
Closes #946 

# Description
This is something which I also missed from Postman, now users can send a request without specifying the protocol.
We now prepend `http://` to the URL when preparing the request if there's no protocol.

|Before|After|
|-------|------|
|<video src="https://github.com/usebruno/bruno/assets/30027205/6bfe4d66-68bc-4c6c-92e6-70bac0efadc1"></video>|<video src="https://github.com/usebruno/bruno/assets/30027205/c8e09786-c908-4191-b9e9-e3449d5062a2"></video>|

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/prepare-request.js
+++ b/packages/bruno-electron/src/ipc/network/prepare-request.js
@@ -70,14 +70,14 @@ const setAuthHeaders = (axiosRequest, request, collectionRoot) => {
   return axiosRequest;
 };
 
-const PROTOCOLS = ['http://', 'https://', 'wss://'];
+const protocolRegex = /([a-zA-Z]{2,20}:\/\/)(.*)/;
 
 const prepareRequest = (request, collectionRoot) => {
   const headers = {};
   let contentTypeDefined = false;
   let url = request.url;
 
-  if (PROTOCOLS.find((protocol) => url.startsWith(protocol)) === undefined) {
+  if (!protocolRegex.test(url)) {
     url = `http://${url}`;
   }
 
@@ -102,8 +102,8 @@ const prepareRequest = (request, collectionRoot) => {
 
   let axiosRequest = {
     method: request.method,
-    url: url,
-    headers: headers,
+    url,
+    headers,
     responseType: 'arraybuffer'
   };
 
--- a/packages/bruno-electron/src/ipc/network/prepare-request.js
+++ b/packages/bruno-electron/src/ipc/network/prepare-request.js
@@ -70,9 +70,16 @@ const setAuthHeaders = (axiosRequest, request, collectionRoot) => {
   return axiosRequest;
 };
 
+const PROTOCOLS = ['http://', 'https://', 'wss://'];
+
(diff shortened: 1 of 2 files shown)


packages/bruno-electron/src/ipc/network/prepare-request.js
```diff
@@ -70,9 +70,16 @@ const setAuthHeaders = (axiosRequest, request, collectionRoot) => {
...
 };
 
+const PROTOCOLS = ['http://', 'https://', 'wss://'];
+
 const prepareRequest = (request, collectionRoot) => {
   const headers = {};
   let contentTypeDefined = false;
+  let url = request.url;
+
+  if (PROTOCOLS.find((protocol) => url.startsWith(protocol)) === undefined) {
```
Its-treason: Using a regex to check if the URL contains a protocol would be better than a static list. 

If somebody used an unsupported protocol like `ftp://` the protocol would be updated here and it causes a weird error.

packages/bruno-electron/src/ipc/network/prepare-request.js
```diff
@@ -95,7 +102,7 @@ const prepareRequest = (request, collectionRoot) => {
 
   let axiosRequest = {
     method: request.method,
-    url: request.url,
+    url: url,
```
Its-treason: You can simplify this: `url: url,` -> `url,`

nelup20:
Thanks for reviewing @Its-treason! I've also added an extra test case that I forgot about.

Its-treason (approved):
Looks good

helloanoop:
Merged!

Nice work @nelup20 !

martinsefcik:
@helloanoop @Its-treason Shouldn't it be changed also in cli package because otherwise there will be inconsistency. Such a request will be working in app but not in cli.

helloanoop:
> Shouldn't it be changed also in cli package because otherwise there will be inconsistency.
Agreed.

@nelup20 I also had to revert the change as this fails when the url has vars in it. Ex: `{{baseUrl}}/echo/json` where `baseUrl` is an env var.

You need to move this logic such that this happens after the interpolation of url has been completed.

nelup20:
Sorry I indeed missed both of those, I'll fix them in a new PR

PR #1106, opened 2023-11-30
Fix: Update CSP to allow posthog

Its-treason:
Updated Electron's Content-Security-Policy to include `app.posthog.com`. This fixes telemetry data not being sent.


helloanoop:
Thank you @Its-treason ! 💛

Issue #1076, opened 2023-11-28
Request disappears from UI list when percent sign (%) in password field


PrimeF:
When the password field of the basic auth configuration contains a percent sign %, the request is not recognised and shown in the app.

Example request:
```
meta {
  name: ABC
  type: http
  seq: 1
}

get {
  url: https://abc.com/
  body: none
  auth: basic
}

auth:basic {
  username: USER
  password: A%
}
```

By removing the percent sign and restarting the app or moving the request's JSON file to another location and back the request pops up.

**Bruno version used**: v1.2.0

bpoulaindev:
Seems to be related to the bru-lang package, currently having a look and trying to identify the source of the problem

likwidoxigen:
Also unable to save when a Multipart Form param contains newlines. Example is [Oracle WebCenterRest](https://docs.oracle.com/en/cloud/saas/human-resources/24b/tutorial-hdl-rest/#Oracle-WebCenter-REST) where the sample is un-minified json.

sanjai0py:
Closing this as it cannot be reproduced in the latest version of Bruno [v1.20.1](https://github.com/usebruno/bruno/releases/tag/v1.20.1) 

https://github.com/usebruno/bruno/assets/84461672/6e4bda8b-0c26-4e4b-b7d3-6f5647295404

likwidoxigen:
> Closing this as it cannot be reproduced in the latest version of Bruno [v1.20.1](https://github.com/usebruno/bruno/releases/tag/v1.20.1)


Should I open a new issue for the newlines in parameters preventing saves?

sanjai0py:
Hey @likwidoxigen, could you please check again? In the latest version of Bruno, multipart form data supports multiline values.

https://github.com/usebruno/bruno/assets/84461672/433c63d7-42f3-4764-acab-75664482b407

helloanoop:
@likwidoxigen If you are still facing issues, please raise a new issue.

Issue #571, opened 2023-10-13
Bug: unable to set request bodies with colon characters in their names


rombert:
Hi,

I work on applications that make heavy use of namespacing property names with colons, e.g. _sling:propertyType_. I imported one collection for such an app from Insomnia and noticed that the parameters are displayed incorrectly in the UI. 

![image](https://github.com/usebruno/bruno/assets/21237/e4974eb5-b36e-45ba-9557-f5f12a005581)

The data is stored in the file as

```
body:form-urlencoded {
  message: The Apache Sling REST API is awesome!
  createdBy:
  sling:resourceType: pospai/component/comment
  sling:redirect: /pospai/home/welcome/jcr:content/comments.html
}
```

Naively trying to escape the colon values did not work, both

```
  sling\:resourceType: pospai/component/comment
```

and 

```
  "sling:resourceType": pospai/component/comment
```

look for the first `:` character to identify the field name.

Is there a way to have field names include the colon character?

SoulKa:
The issue was that the imported (and also any manually written) parameter values have not been encoded thus resulting in invalid bruno files. Fix is created.

rombert:
Thanks for the fix @SoulKa ! With version 0.27.2 I can manually adjust the fields after the import so that the property is named `sling:resourceType`, but the initial import is still incorrect. While my immediate problem is solved, I've attached my Insomnia export - [Insomnia_2023-10-13.json](https://github.com/usebruno/bruno/files/13113535/Insomnia_2023-10-13.json) in case there is interest in fixing the import as well.


You can look at home _Create comment (John Doe)_ is imported, for instance.

SoulKa:
Thanks for your collection! I could recreate your issue and created a fix for it. When importing the collection, the parameters not look as expected:

![grafik](https://github.com/usebruno/bruno/assets/15342503/bd5cabe9-e6bf-4cf0-acfe-38203112449d)

rombert:
Awesome, thanks @SoulKa !

helloanoop:
@SoulKa I had to revert the fix as this was causing #1339 and #1076 Commit: https://github.com/usebruno/bruno/commit/a90467255535ba3c740363d5aa226bbede8b14ee

This fix is going in the `1.8.0` release.

### So how do we solve this issue

Before we approach a solution, I think we should never auto encode / decode values. It should always be configurable (See #732)

I see 2 Solutions

#### Toggle Inside Request Settings
[...]

PR #1181, opened 2023-12-07
Bugfix/request disappearing

bpoulaindev:
# Description
Solving issue #1076 
Added override decodeURIComponentSafe function to prevent % character being escaped

image

### Contribution Checklist:


helloanoop:
Thank you @bpoulaindev !

I have some comments
* Can you remove the `bun.lockb` file. You can add it to `.gitignore`
* Can we move the `decodeURIComponentSafe` to a new file `utils.js` and also add a unit test to it ?

bpoulaindev:
Eventually why would we need to decode the URI for a password ? I can't see how we could have a function that can preserve an uri and a password at the same time without breaking it

helloanoop:
@bpoulaindev The `decodeURIComponent` was added in this commit https://github.com/usebruno/bruno/commit/ddddab811290c82e15a7008c4027792fe7772bf7

It's related to https://github.com/usebruno/bruno/issues/571

bpoulaindev:
I eventually opted for non-decoding passwords and keeping encodeURI for every other field

packages/bruno-lang/v2/src/bruToJson.js
```diff
@@ -105,7 +105,7 @@ const mapPairListToKeyValPairs = (pairList = [], parseEnabled = true) => {
   }
   return _.map(pairList[0], (pair) => {
     let name = _.keys(pair)[0];
-    let value = decodeURIComponent(pair[name]);
+    let value = name === 'password' ? pair[name] : decodeURIComponent(pair[name]);
```
helloanoop, line 108: @bpoulaindev We need to solve this in a generic way and should not hardcode this to apply to any key-val pair that has the name `password`

The `mapPairListToKeyValPairs` is used to parse all key val pairs (headers, assertions, vars, query params)

baptpln: @helloanoop how could you have a generic function that work both on a password (which can be of any form), a random string and an encoded URL with params ?
Edit : we could encode everything in base 64 but i'm afraid it would hurt performance

helloanoop:
Closing this as the issue has already been fixed.

Issue #1112, opened 2023-12-01
[BUG] Generate code doesn't support variables


survivant:
When I try to generate code from the Query it will generate an error if the URL contains variables in it

![image](https://github.com/usebruno/bruno/assets/191879/fa2e1e7a-b78e-4e54-bc8b-f4102c9be735)
![image](https://github.com/usebruno/bruno/assets/191879/79b336de-e69d-41fc-a465-68dcc4a7d82d)

and the generated code should include the variables that are in the query.  They could be passed as a Header too

survivant:
here an example with a variable in the headers
![image](https://github.com/usebruno/bruno/assets/191879/d3736530-e81b-409f-abc6-026747461984)

survivant:
and we could add a copy button like
![image](https://github.com/usebruno/bruno/assets/191879/01bd82b5-19f2-467c-ba13-a8fba2fe8293)

afonsocarlos:
I'm having the same issue but for Environment variables with spaces in their values.

survivant:
@afonsocarlos can you test my with package here is the issue is there too ?  https://github.com/survivant/bruno/tree/BrunoSD I regroup few fixes there

survivant:
@afonsocarlos  I have this with my branch 
![image](https://github.com/usebruno/bruno/assets/191879/a33635d2-5ca9-4c03-b6ae-d849508c9724)
![image](https://github.com/usebruno/bruno/assets/191879/c998acab-8647-4161-a4d6-f19801d6ec74)
![image](https://github.com/usebruno/bruno/assets/191879/97c7175c-1405-4e5a-898e-92ac11021b24)
![image](https://github.com/usebruno/bruno/assets/191879/7df30974-958e-41c8-846d-20d210b9f95b)

I'll add your fix, but I'm just curious, are you get the error.  I was suppose the encoreURI the url, so the params are included

![image](https://github.com/usebruno/bruno/assets/191879/5e0df1b7-11c5-4929-a04c-c5e2b36b5f90)

afonsocarlos:
I couldn't find where you're calling encodeURL in the codebase, but try to set an environment variable containing spaces and that may replicate the issue I faced

survivant:
Can you post your. Bru file? We learn by our mistakes, so I want to find if
I forgot this usecase, because I posted an example with space in the param
name, but for the env variable it did not accept a variable with space, but
for the value yes.

[...]

PR #1207, opened 2023-12-11
fix(#1112): generate code for parameter values with spaces

afonsocarlos:
# Description

This fixes the error "Error generating code snippet" when Environment parameters have values with spaces

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/utils/codegenerator/har.js
+++ b/packages/bruno-app/src/utils/codegenerator/har.js
@@ -51,7 +51,7 @@ const createPostData = (body) => {
 export const buildHarRequest = ({ request, headers }) => {
   return {
     method: request.method,
-    url: request.url,
+    url: encodeURI(request.url),
     httpVersion: 'HTTP/1.1',
     cookies: [],
     headers: createHeaders(headers),


packages/bruno-app/src/utils/codegenerator/har.js
```diff
@@ -38,7 +38,7 @@ const createPostData = (body) => {
       mimeType: contentType,
       params: body[body.mode]
         .filter((param) => param.enabled)
-        .map((param) => ({ name: param.name, value: param.value }))
+        .map((param) => ({ name: param.name, value: encodeURIComponent(param.value) }))
```
martinsefcik: Theoretically param name can also contains spaces or some other characters which should be encoded (at least I'm not aware of such RFC which does not allow it). So don't we want to apply the same also for param name ?

afonsocarlos: We certainly could. I've never come across such case, but escaping the name should do no harm anyways.

afonsocarlos: @martinsefcik on second thought, it makes more sense to just encode the whole URL. That would catch any corner cases. Does that make sense for you?

sanjai0py:
Closing this PR as another [PR](https://github.com/usebruno/bruno/pull/1827) with the same changes has already been merged. Thanks for working on this.

PR #1240, opened 2023-12-18
feat: OpenAPI import: create a default environment

### LINKED ISSUE #1239: [Feature] OpenAPI Import: make server protocol and host easily changeable
Currently import from OpenAPI constructs a default server from the spec and hardcodes it in each request. This makes it difficult to change later to a different environment or to fill out at all if the spec does not provide any servers.

In my opinion, the server protocol and host should be extracted to some sort of variables to make them easily changeable.

AndrejsK:
# Description

Fixes #1239

The previous solution constructed a default server baseUrl and directly concatenated it into each request of the collection. This makes it difficult to change it later through the UI if it is wrong or if the same OpenAPI spec can be used in multiple environments.

This solution instead creates a default environment with two variables "protocol" and "host", which are filled according to the previously constructed default baseUrl. This makes it very easy to change for all requests while preserving the server information extracted from the spec.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/utils/importers/openapi-collection.js
+++ b/packages/bruno-app/src/utils/importers/openapi-collection.js
@@ -31,9 +31,8 @@ const readFile = (files) => {
 };
 
 const ensureUrl = (url) => {
-  let protUrl = url.startsWith('http') ? url : `http://${url}`;
   // replace any double or triple slashes
-  return protUrl.replace(/([^:]\/)\/+/g, '$1');
+  return url.replace(/([^:]\/)\/+/g, '$1');
 };
 
 const buildEmptyJsonBody = (bodySchema) => {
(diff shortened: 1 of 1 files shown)


nikolaihen:
Any updates on getting this merged...? 

This is a pain when working with large APIs with different environments...

bSolt (changes requested):
Your solution is not any better than the way it currently works. I need the whole host to come from a variable, and your solution would break that. You can just replace your hosts in your collection this way with your own script if you need it. That's what I ended up doing anyway

AndrejsK:
Right, but this way you can change it through the UI, that's better than writing a script, if you ask me. Would you be happy with this PR if the whole base url was a single variable?

TheBlackReverand:
I make a PR with my proposition #2088 with that you can create and describe multiple environment.

helloanoop:
Closing this PR as this has been addressed in #3267

Issue #1039, opened 2023-11-23
[Feature Request] Flag a request to be exclude from test runner


stgonzales:
It would be good to have a way to mark each request to be exclude from the test runner. This could be by adding a meta flag or in a blacklist format.

helloanoop:
Yes, we will support this.

I am thinking of supporting a more generic feature called `tags`
Here is how I am envisioning it

- A tag is essential a string. Ex: `regression`, `sanity`
- Each request can have multiple tags
- When running the collection, we should be able to run tests where only requests with a certain set of tags get included

Here is how it would look in the bru file.

```groovy
meta {
  name: Get Users
  type: http
  seq: 1,
  tags: [
    sanity,
    regression
  ]
}
```

The GUI Collection runner will support options to only run requests which match a certain set of tags.

In the CLI, we can have something like
```bash
bru run --tags=sanity,regression
```

Let me know if you have feedback or suggestions

cc @dcoomber @jeff-edmondson-pci @rabestro @mj-h @Its-treason @martinsefcik

rabestro:
@helloanoop 
The concept of tags is handy, and it is supported by many testing frameworks. I suggest considering the option not only to run tests that contain a specific tag but also to exclude tests with a particular tag. For example, in the Karate framework, there is no special annotation, but you can add a tag named "ignore" and then run tests while excluding the ones with the "ignore" tag.

KirkMunroSagent:
> @helloanoop The concept of tags is handy, and it is supported by many testing frameworks. I suggest considering the option not only to run tests that contain a specific tag but also to exclude tests with a particular tag. For example, in the Karate framework, there is no special annotation, but you can add a tag named "ignore" and then run tests while excluding the ones with the "ignore" tag.

Maybe something like this then?

`bru run --tags=sanity,regression --exclude-tags=ignore`

That way it's not specific to a hard-coded tag name like `ignore`, and the collection that is run is the aggregate of all tests that match at least one of the tags in the `--tags` argument, minus any tests that match at least one of the tags in the `--exclude-tags` list.

brvaland:
@helloanoop do we have any estimate when this feature will be available as we integrate our collection with CICD and would like to exclude few POST request for PRODUCTION environment to avoid adding data.

Is there a workaround which I can explore via script to move to next/skip request in the runner.?

irharrier2:
@helloanoop your suggested solution is perfect. Any idea when are you going to deliver it?

[...]

PR #1314, opened 2024-01-03
Feature: Add an optional param to cli to only run requests that have tests

jeff-edmondson-pci:
# Description
- In a CI/CD environment there exists a use case of only wanting to run requests that have tests. 
- Added an optional param `--tests-only` to the CLI arg list. If this is set it will only add request that have tests. 

#### Collection 
image

#### Result 
image

### Contribution Checklist:


ahmad-elafify-pci:
+1

mj-h:
Possibly related: https://github.com/usebruno/bruno/issues/1039

This PR creates a difference between the way the collections are run in then UI vs the CLI. If this PR is merged, I propose to file a follow-up issue to add a similar option to the UI.

However, there are a few questions that this PR raises:
- should a file where the tests are all commented out be included? (To me, the answer would be "no",  but this PR would include it)
- should a file that has assertions but no tests be included? (To me, the answer is "yes", but this PR would skip such a file)
- should a file that has tests in the post-response script be included?
  - this is messy. The post-response-script and the test-script are run separately, but they do pretty much the same thing. The UI gets confused when I move tests from one to the other. I think this should be cleaned up first before merging this PR -- there should not be two places where post-response scripts are being defined. @helloanoop : What do you think?

Lastly: This PR promotes a collection-writing-style that I would call "bag-of-requests", where some are used for exploratory-testing, and others are used for automatic testing, and the difference between the two is given implicitly.

I would propose a different solution to this problem: In your collection, sort your requests into appropriate subfolders. Then, we should change the runner so that you can do: 

```
bru run my-collection --folder "subfolder with all the tests"
```
I think that would lead to cleaner collections, and give you more flexibility overall.

jeff-edmondson-pci:
> I would propose a different solution to this problem: In your collection, sort your requests into appropriate subfolders. Then, we should change the runner so that you can do:

@mj-h Yes I do agree that it would be easier to just throw all of the tests into a folder but I don't think that is very realistic for complex collections, such as having a very large API collection which many developers are using & contributing to. It is organized by api routes. The structure would be:

```
Super Cool Sports API
└───Teams
 │   └─── League 1
 |          └─── test1.bru
 │   └─── League 2
└───Games
 │   └─── Scores
```

Which matches the url path `/teams/league-1/test1` for the resource.

Now if we wanted to have the CI/CD run the requests only with tests but only for a given folder as you are suggesting (correct me if I am wrong) we run into the issue of the API collection not remaining organized. 
The two work arounds for this could be 
1. Have duplicated request between original location and the tests folder. This is not ideal for a variety of reasons
2. Just move the request into the tests folder from it's original location. Again not ideal. 

```
Super Cool Sports API
└───Teams
 │   └─── League 1
 │   └─── League 2
└───Games
 │   └─── Scores
└───Tests
 |   └─── test1.bru

```

rabestro:
> Open to suggestions

(1) I have many tests with [assertion only](https://github.com/rabestro/pig-latin-rest/blob/main/bruno-test/json-response/Translate%20text%20to%20Pig%20Latin.bru). I have a test with [post-script and assertion](https://github.com/rabestro/pig-latin-rest/blob/main/bruno-test/xml-response/Translate%20text%20to%20Pig%20Latin.bru). 

The proposed implementation skipped them.

(2) The suggested implementation raises the cyclomatic complexity of the code which is not ideal. The implementation involves adding another if/then case, but it is not a recommended approach. Instead, using strategies would be a better option. For instance, we can use a simple strategy (all scripts) and apply different strategies for other cases as needed.

rabestro:
@jeff-edmondson-pci you always have the opportunity to build your fork and use your version of Bruno with the features you need. 

From my point of view, there is at least a lack of documentation correction that would reflect the appearance of new features. 

I also think it’s right to first discuss the need for any additions before creating a pull request.

helloanoop:
Thanks @jeff-edmondson-pci for working on the PR and @mj-h and @rabestro  everyone for chiming in with comments

> should a file where the tests are all commented out be included? (To me, the answer would be "no", but this PR would include it)

@mj-h I agree. This is edge case though. We can iteratively improve this logic later by using the [decomment](https://www.npmjs.com/package/decomment) library to address this

> should a file that has assertions but no tests be included? (To me, the answer is "yes", but this PR would skip such a file)

@mj-h  I agree. An assertion is just a "declarative" way of writing a test. I see that this has been addressed in the PR.

> The post-response-script and the test-script are run separately, but they do pretty much the same thing. The UI gets confused when I move tests from one to the other. I think this should be cleaned up first before merging this PR

The separation is intentional. I agree that both these blocks (post response scripts and test scripts) do the same thing. The intent is to visually separate code related to "scripting" and "tests".
* scripting - things like parsing a response and updating some collection variables
* tests - assertions and tests

> I have many tests with [assertion only](https://github.com/rabestro/pig-latin-rest/blob/main/bruno-test/json-response/Translate%20text%20to%20Pig%20Latin.bru). I have a test with [post-script and assertion](https://github.com/rabestro/pig-latin-rest/blob/main/bruno-test/xml-response/Translate%20text%20to%20Pig%20Latin.bru). The proposed implementation skipped them.

@jeff-edmondson-pci This has been addressed in the PR

> The suggested implementation raises the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of the code which is not ideal. The implementation involves adding another if/then case, but it is not a recommended approach. Instead, using strategies would be a better option. For instance, we can use a simple strategy (all scripts) and apply different strategies for other cases as needed.

@jeff-edmondson-pci There is a plan to include [tags](https://github.com/usebruno/bruno/issues/1039) in the roadmap. I think the business use case of `run tests that have tests/assertions` is a valid one. Implementing the tags feature would take some time.

Also the `--test-only` arg being introduced in this PR is optional. My personal take would to be use the concept of `tags` (which is a pretty common patten across testing frameworks) when that feature lands in Bruno.

@dcoomber @Its-treason @martinsefcik @jeff-edmondson-pci @mj-h 

The PR is GTG from my side. Let me know if you have further comments. I am planning to have this merged over the weekend.

Its-treason (approved):

packages/bruno-cli/src/commands/run.js
```diff
@@ -215,12 +230,14 @@ const builder = async (yargs) => {
     .example(
       '$0 run request.bru --output results.xml --format junit',
       'Run a request and write the results to results.xml in junit format in the current directory'
-    );
+    )
+    .example('$0 run request.bru --test-only', 'Run all requests that have a test');
```
martinsefcik, line 237: On line 214 option name is `'tests-only` here in example `'test-only`

packages/bruno-cli/src/commands/run.js
```diff
@@ -200,6 +211,10 @@ const builder = async (yargs) => {
       type: 'boolean',
       description: 'Allow insecure server connections'
     })
+    .option('tests-only', {
+      type: 'boolean',
+      description: 'Only run requests that have a test'
```
martinsefcik, line 216: If it runs also in case there are assertions only I would extend this description to be clear when request is executed.

martinsefcik:
@helloanoop Personally I would rather see in Bruno more "standardized" support for tags instead of this feature. If there would be such support then I guess this `--tests-only` option is not needed.
Also If I put only something like `console.log('')` inside tests script then it automatically executes such request, but does it mean that such request has tests ? So for me it is a bit confusing, option description is short and not enough and I mostly agree with @rabestro

helloanoop:
Thanks everyone for the feedback and comments.
I have decided to merge this.

I agree that `--tests-only` (although optional) has some edge cases
- Like a request can have console.log('') inside tests script then it automatically executes such request, but does it mean that such request has tests ?
- What if a test block is commented out. Does it mean that such request still has tests ?

Its upto the user to ensure that
* they have tests in the request and
* they are not commented out and 
* they don't have code in the tests script that are not tests

PR #1328, opened 2024-01-05
feat: Added the ability to save requests on running

SeifOwleN:
# Description

The change adds another preference option in which when enabled, it saves any request when ran

bruno-saveReqOnRun

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/store/preferences.js
+++ b/packages/bruno-electron/src/store/preferences.js
@@ -17,6 +17,7 @@ const defaultPreferences = {
     },
     storeCookies: true,
     sendCookies: true,
+    saveRequestOnRun: true,
     timeout: 0
   },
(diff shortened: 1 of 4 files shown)


packages/bruno-electron/src/store/preferences.js
```diff
@@ -17,6 +17,7 @@ const defaultPreferences = {
     },
     storeCookies: true,
     sendCookies: true,
+    saveRequestOnRun: true,
```
rsilvr, line 20: I'm not sure, but might be worth considering: maybe the default value should be false to keep the same behavior for current users?

helloanoop:
Closing this PR. Reason being that auto-saving requests when run isn’t a broadly expected behavior.

Issue #1310, opened 2024-01-03
[Feature] : Settings on folder level


bpoulaindev:
It would be useful to have settings and other properties not only on the collection level but also on the folder level
Also referencing issue #1174

helloanoop:
Closing this as this functionality is now available in Bruno.

Issue #1174, opened 2023-12-07
Feature Request: Script and tests at folder level


tho-gru-38:
It would be very useful if it is possible to execute scripts or tests at folder level.

My use case is that the service i am testing runs on a system where changes are applied very frequently. Therefore I have the demand to repeat all requests in a folder in case an error occurred. But because I can not execute at folder level I can only rerun the complete collection. this is frustrating because the collection needs one hour or more but the changes are applied ever five minutes.

If scripts and tests are available at folder level I can count errors an retries at this level. So I can decide where to skip using the new bru.setNextRequest feature.

kyleboe:
I commented on #1334 as well but I think this may be a solid case to remove "Folders" altogether and allow for Collections to be nested.

sanjai0py:
Closing this issue as the [PR](https://github.com/usebruno/bruno/pull/2529) adding the ability to add folder level Headers, Scripts, and Tests has already been merged into Bruno and released in the latest [v1.20.0](https://github.com/usebruno/bruno/releases/tag/v1.20.0)

PR #1334, opened 2024-01-05
[Feature] : Settings on folder level

bpoulaindev:
# Description

**Referencing issue #1310** 

Add settings at the folder level similarly to collection settings
- multiple settings tab can be opened
- only headers have been added in those settings
- headers var with the same name as one from the parent will override the parent's one in the request

image

### Contribution Checklist:


fulvio-diller:
It would be nice to have body and tests also..

bpoulaindev:
@fulvio-diller Yeah i'll add those once headers are validated and officially stable

kyleboe:
A thought from a design perspective: Should the concept of "Folders" be removed and should "Collections" be able to be nested? I'm not sure if that has already been discussed but it seems like that is the desired functionality as expressed in #1174 & #1310

Personally, my team would greatly benefit from being able to nest Collections to run a hierarchy of tests. So much so, that if we had that functionality, we would swap from Postman to Bruno immediately.

tho-gru-38:
Hi @kyleboe,

I do not think that nesting collections is a good idea for several reasons:

1. How should bru.setNextRequest() implement a jump between nested collection?
2. What is the meaning of environments? Currently environments are stored at collection level. If a collection is nested there might exist environments too. To be able to run separately (standalone) environments must exist. On the other hand the values in the nested collection might be different from the values in the outer collection.
3. Currently it is not possible to move/copy a request from one collection to another collection.

To my understanding the concept of folder exist to create hierarchical tests. Therefore I want to be able to store scripts, tests ... at folder level.

Kind Regards
Thomas

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -138,6 +139,61 @@ export const saveCollectionRoot = (collectionUid) => (dispatch, getState) => {
...
+};
+export const retrieveDirectoriesBetween = (pathname, parameter, filename) => {
+  const parameterIndex = pathname.indexOf(parameter);
+  const filenameIndex = pathname.indexOf(filename);
+  if (parameterIndex === -1 || filenameIndex === -1 || filenameIndex < parameterIndex) {
+    return [];
+  }
+  const directories = pathname
+    .substring(parameterIndex + parameter.length, filenameIndex)
+    .split('/')
```
Its-treason, line 182: `PATH_SEPERATOR` should be used here and 5 lines below for Windows compatible

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -150,9 +206,26 @@ export const sendRequest = (item, collectionUid) => (dispatch, getState) => {
...
 
-    sendNetworkRequest(itemCopy, collection, environment, collectionCopy.collectionVariables)
+    const folderDatas = itemTree.reduce((acc, currentPath) => {
+      const folder = findItemInCollectionByPathname(collectionCopy, currentPath);
+      if (folder) {
+        acc = mergeRequests(acc, folder.root.request);
+      }
+      return acc;
+    }, {});
+    const mergeParams = mergeRequests(collectionCopy.root.request, folderDatas);
```
Its-treason, line 259: For new collection `collectionCopy.root` can be undefined resulting in an error here.

helloanoop:
Merged to `feat/folder-level-settings` branch where I will be continue further work on this.

Issue #1320, opened 2024-01-04
Only sends the last element of POST array


deimios:
Passing arrays as POST body variables requires repeating the variable name and appending square brackets.
Example: 
- item[] = 1
- item[] = 2
- item[] = n

So the sent/urlencoded POST data should look like:
`item%5B%5D=1&item%5B%5D=2&item%5B%5D=n`

However Bruno only sends:
`item%5B%5D=n`
Effectively it is ignoring all the prior instances of the same variable name.

The request settings I used:
![Bruno-Array-Screenshot](https://github.com/usebruno/bruno/assets/2050407/8bc4c4b0-4477-40e0-aa20-d25ee054397d)

helloanoop:
This will be available in upcoming version `v1.29.0`

PR #1355, opened 2024-01-09
fix(#1320): send all values of same key in post req

adarshlilha:
fixes - #1320 

Currently, if multiple data keys of the same name are sent in a request, bruno only takes the last one
This is because params key/value pairs are stored in an obj, so it gets overwritten with new values (old value is discarded)

Fix:
* Updated the code for `prepare-request` to check if the key is already present In resultant obj, convert the val to arr of values

for the below request:
Screenshot 2024-01-10 at 02 33 42

`item` is an arr with 2 values
Screenshot 2024-01-10 at 02 35 14

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,17 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Debug Main Process",
+      "type": "node",
+      "request": "launch",
+      "cwd": "${workspaceFolder}/packages/bruno-electron",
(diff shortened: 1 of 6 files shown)


adarshlilha:
IMO this change should work fine

Additionally the response formatting might need changes if we want to show the response arr instead of each items
```js
{
  "item[0]": "1",
  "item[1]": "2",
  "item1": "3",
  "id": 201
}
```

codegsaini:
It works

Screenshot from 2024-12-07 20-24-56

Issue #1363, opened 2024-01-10
Bruno parses incorrectly URL used to make ServiceNow JSON API request


JC001010:
Discovered issue while using the ServiceNow JSONv2 API ([link](https://docs.servicenow.com/bundle/tokyo-application-development/page/integrate/inbound-other-web-services/concept/c_DataModificationAPI.html))

In short, bruno does not parse correctly request URLs that have empty Query params or params whose value look like Query params. For instance, the following URL is used in ServiceNow to update a ticket whose number us INC00123. 

https://your_instance.service-now.com/incident.do?JSONv2&sysparm_action=update&sysparm_query=number=INC00123

 If I add that URL to bruno and then save the request or add a new 'Query' param, bruno will try to parse the URL and it will not parse the  'sysparm_query=number=INC00123' section of the URL correctly. It will also add 'JSONv2' as an empty param, which is expected, but it will then remove that param from the request URL any time it tries to re-parse the URL (see screenshots)

When I add the URL to a bruno request:
![before request](https://github.com/usebruno/bruno/assets/65236958/97e81372-7722-4009-8aef-82a855b9ae16)

After I change anything in the request and bruno re-parses the URL:
![after any request modification](https://github.com/usebruno/bruno/assets/65236958/fb9632e9-05d6-4b0e-96fc-882b37bff9fe)

Thank you

PR #1370, opened 2024-01-11
fix(#1363): add query param even when value is missing

adarshlilha:
fixes: #1363 

Current behaviour: 
Currently, bruno incorrectly strips off a query param if the `value` or `name` for the same is empty/missing

this has been fixed with this PR

New behaviour:
if query param `Name` field is not empty, append the same in the URL (remove check for value to be present)
if query param `Name` is missing, do not append it to the URL

https://github.com/usebruno/bruno/assets/13575704/4405f2d1-6c66-4ffe-a124-25cd9e5d6d40

### DIFF
--- a/packages/bruno-app/src/utils/url/index.js
+++ b/packages/bruno-app/src/utils/url/index.js
@@ -33,8 +33,13 @@ export const stringifyQueryParams = (params) => {
 
   let queryString = [];
   each(params, (p) => {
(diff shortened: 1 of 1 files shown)


Its-treason (approved):
Looks good and Bruno follows the behavior of postman with this PR

helloanoop:
Thank you for working on this @adarshlilha !

Issue #732, opened 2023-10-23
[Bug] - URLEncode edited Query parameters from the dedicated table/tab on change


l0new4nder3r:
I imported requests a while ago now, and some were imported with query parameters like dates:

`?from=2023-10-23T00%3A00%3A00%2B00%3A00&to=2023-10-23T23%3A59%3A59%2B00%3A00`

They work fine when I run them.
The Query tab allows me to see what the current decoded values are (nice for human eyes! ❤️).
![image](https://github.com/usebruno/bruno/assets/27144372/1f32f66e-b487-40cd-a94f-52363d002c58)

But when I modify my "from" date, by example starting a day earlier, in the Value field of the Query tab, then the URL is immediately altered, and the query params are no longer encoded - making my request fail, and me losing the previously encoded value that was running fine before 😞 

`?from=2023-10-22T00:00:00+00:00&to=2023-10-23T23:59:59+00:00`

I can ctrl-z to get back to it, and manually fix it, but altering the encoded value itself in the main URL can sometimes be tedious.
And we are not far from having something a nice user experience here - my suggestion : when editing any value within the "Query" tab, automatically apply URLEncode while updating the resulting URL.

l0new4nder3r:
@Its-treason 
Fair enough! 
I was unaware of such a possible use case.
As long as it is possible and easy to use, totally fine by me :)

Could it be some box to tick in the Query tab? 
One per line, would be weird though.
And having this option set globally would defeat the possibility to have URLEncode and not URLEncode in the same instance.

What do you think? @helloanoop any idea of what the design could be?

Its-treason:
One checkbox per line is definitely too much, I was thinking about putting a checkbox under the right side of the table.

CarlosLecval:
Hi, this is what I worked on

https://github.com/usebruno/bruno/assets/61945879/0a38a6e9-1a46-4487-90fb-60912fb28a35

Darwin1337:
This issue is very frustrating and it would be nice to see it fixed!

riekusr:
Also in need of this

l0new4nder3r:
Temporary workaround using this "script > pre-request" line

`req.setUrl(req.getUrl().replace(/(.+?\?).+/,'$1') + encodeURIComponent(req.getUrl().replace(/.+?\?/,'')));`

helloanoop:
[...]

PR #1374, opened 2024-01-12
fix(#732) Option to encode query params

CarlosLecval:
# Description

solves: #732 

Basically just added a checkbox to toggle the option to encode the query parameters as discussed in the issue

https://github.com/usebruno/bruno/assets/61945879/04c23ced-6939-4ddf-83cc-86adfaaabeca

### DIFF
--- a/packages/bruno-app/src/components/RequestPane/HttpRequestPane/index.js
+++ b/packages/bruno-app/src/components/RequestPane/HttpRequestPane/index.js
@@ -16,6 +16,7 @@ import Tests from 'components/RequestPane/Tests';
 import StyledWrapper from './StyledWrapper';
 import { get } from 'lodash';
 import Documentation from 'components/Documentation/index';
+import EncodeQueryParams from '../QueryParams/EncodeQueryParams/index';
 
 const HttpRequestPane = ({ item, collection, leftPaneWidth }) => {
   const dispatch = useDispatch();
@@ -132,6 +133,11 @@ const HttpRequestPane = ({ item, collection, leftPaneWidth }) => {
         {focusedTab.requestPaneTab === 'body' ? (
           <div className="flex flex-grow justify-end items-center">
             <RequestBodyMode item={item} collection={collection} />
+          </div>
+        ) : null}
+        {focusedTab.requestPaneTab === 'params' ? (
+          <div className="flex flex-grow justify-end items-center">
+            <EncodeQueryParams item={item} collection={collection} />
           </div>
         ) : null}
       </div>
--- a/packages/bruno-app/src/components/RequestPane/QueryParams/EncodeQueryParams/StyledWrapper.js
+++ b/packages/bruno-app/src/components/RequestPane/QueryParams/EncodeQueryParams/StyledWrapper.js
@@ -1,0 +1,11 @@
+import styled from 'styled-components';
+
+const Wrapper = styled.div`
+  font-size: 0.8125rem;
+
+  .label-item {
+    color: ${(props) => props.theme.colors.text.yellow};
+  }
+`;
+
+export default Wrapper;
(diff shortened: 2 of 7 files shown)


Fabrice-Deshayes-aka-Xtream:
Hello. This PR would be very useful. @helloanoop, do you plan to merge this feature?

CarlosLecval:
I'd be glad to fix the conflicts or change anything for this to be merged 👍

Benoit12345:
Hello @helloanoop,
Is it planned to (fix conflicts +) merge this feature in next release please ? It would be really great !

divinvishnu:
Hey [CarlosLecval](https://github.com/CarlosLecval), are you planning on merging this in soon? It is a much welcome PR, so please ask if you need any help getting things resolved for this PR.

Benoit12345:
Hello,
just a little up to know if this feature will be available in one of the next releases or if there're some blocking points about it ?

Benoit12345:
hi @helloanoop,
1 year later PR creation, could you tell us if this feature is going to be merged or cancelled please ?

coluzziandrea:
it's a bit sad to see this stalling here, this feature will be very useful

merzeri:
Adding on to this. It's tremendously frustrating for it to make assumptions about encoding or not-encoding strings that don't seem to be an issue with Insomnia. In my case, I'm making requests to an API in which most endpoints expect a `criteria` parameter to be given a JSON object containing the query details -- sort of a flattening of GraphQL to JSON REST APIs. Something the request is encoding prevents this from working correctly, netting a 400 error from the remote Tomcat server.

helloanoop:
After reviewing the related issues, I believe the root cause of many of them is an inconsistency in how Bruno handles URL encoding between the below two components
- URL Input Box
- Query Parameters Table

Currently, Bruno automatically decodes the URL from the input field using URLSearchParams. However, when changes are made in the query parameters table, those changes are not encoded before updating the URL in the input field. This inconsistency leads to confusion and creates a number of edge cases.

Another challenge is that we cannot determine whether the URL entered in the input field is already encoded. Given this uncertainty, it is not ideal for Bruno to assume the URL is encoded and then proceed with decoding while populating the query parameters table.

**Solution**
To resolve this, we are considering the following approach:
1. Bruno will not perform any encoding or decoding in the UI. The URL input field and the query parameters table will remain identical.
2. At the time of sending the request, Bruno will encode the query parameters using JavaScript's encodeURIComponent function. This will be the default behavior, but users will have the option to disable encoding via a setting if needed.

**Tradeoff:**
With the current implementation, when users input an encoded URL, Bruno decodes it before displaying it, making it easier to read. With the new approach, encoded URLs will remain as entered, which may reduce readability in some cases.
Ex: `?from=2023-10-23T00%3A00%3A00%2B00%3A00&to=2023-10-23T23%3A59%3A59%2B00%3A00`

Benoit12345:
thanks for your answer @helloanoop .
When you talk about a setting, I hope you thought to a request setting because, as you said:
> At the time of sending the request, Bruno will encode the query parameters using JavaScript's encodeURIComponent function.

So it's request sending time which is important. The encoding may have to be used for a request and not for another.
If it is the case (and a easy way to set if from Bruno' interface with a right click or something like that), it seems to be fine from my point of view.

merzeri:
This sounds like an excellent move on the project! I'm not sure what Insomnia does by comparison (as it is not open-source after all) but I imagine it is very much something like this.

Benoit12345:
I think this MR could be closed since the release 2.8.0 proposed a settings by request allowing to automatically encode/decode query parameters (cf. #5089  )

maintainer-bruno:
Hey @CarlosLecval,

Thanks so much for taking the time to contribute and open this PR! 🙌

The issue this PR aimed to address has since been handled in a different way through [#5089](https://github.com/usebruno/bruno/pull/5089), so we’ll be closing this one to avoid duplication. That said, we really appreciate your effort and initiative—please don’t hesitate to contribute again in the future!

Thanks again!

Issue #922, opened 2023-11-08
CLI 2 critical severity vulnerabilities


bourpie:
When installing CLI on Node V. 18.14.1, I get this message

```
npm WARN deprecated vm2@3.9.19: The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider 
migrating your code to isolated-vm.

changed 146 packages, and audited 147 packages in 11s

34 packages are looking for funding
  run `npm fund` for details

2 critical severity vulnerabilities
```

Its-treason:
`vm2` has been deprecated for a while now, and we already planned to update to a different library here #263.

The vulnerabilities in `vm2` allow for malicious code to escape the sandbox. I think if you only run trusted request scripts it should be fine for now, but we will definite address this in the future.

chriswarkentin:
I would say that this should be fairly high priority if you're looking to drive adoption. I'm evaluating Bruno to use at my workplace and things like that will definitely raise eyebrows.

camba1:
Is there any progress on this? I love Bruno, but unfortunately cannot use the CLI until this is addressed and thus cannot use it to automate API testing

Its-treason:
> Is there any progress on this? I love Bruno, but unfortunately cannot use the CLI until this is addressed and thus cannot use it to automate API testing

Not really, I opened a second PR that replaced vm2 with the native vm module. But this did not get any attention yet.

qury:
Are there any plans from the main contributors to fix any of the critical bugs?

As of today this is what i get from npm audit:
`44 vulnerabilities (31 moderate, 9 high, 4 critical)`

Critical packages being:
- vm2
  - vm2 Sandbox Escape vulnerability - https://github.com/advisories/GHSA-cchq-frgv-rjh5
  - vm2 Sandbox Escape vulnerability - https://github.com/advisories/GHSA-g644-9gfx-q4q4

- underscore
   - Arbitrary Code Execution in underscore - https://github.com/advisories/GHSA-cf4h-3jhx-xvhq
- minimist
   - Prototype Pollution in minimist - https://github.com/advisories/GHSA-vh95-rmgr-6w4m
   - Prototype Pollution in minimist - https://github.com/advisories/GHSA-xvch-5gv4-984h

treynolds011:
I know that there has been some discussion of getting this fixed but I would love to start using the cli in my environment but can't until this is fixed.

cemerick1:
[...]

PR #1400, opened 2024-01-16
fix(#263): Replace vm2 with fork of vm2 to fix security issues

### LINKED ISSUE #263: Deprecate VM2
We've had great success with vm2 in running scripts inside a sandbox.
The project [vm2](https://www.npmjs.com/package/vm2) however has been currently discontinued and not recommended for production usage.

The recommended alternative is [isolated-vm](https://www.npmjs.com/package/isolated-vm)

We want to eventually move to an alternative that is being maintained. I don't want to rush at the moment. There are plenty more things that stack higher in priority. 

Here is how an our tasklist to get this done should look like
- [ ] a solid test suite that covers a lot of scripting capabilities
- [ ] deprecate `vm2`, move to a `isolated-vm`, but allow users to fallback to `vm2` via toggle if their tests start start failing
- [ ] once the new vm is stable, drop `vm2`

Its-treason:
# Description

Replaced `vm2` with `@n8n/vm2` a fork of vm2 that fixes 2 security vulnerabilities.

Closes #263 #922

---

PSA: While this closes sandbox escape vulnerabilities, you should always make sure not to run random/untrusted scripts. Bruno allows script's to access `node:fs` and other modules (If enabled in bruno.json) so be careful with what you run.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4041,6 +4041,41 @@
         "node": ">=12"
       }
     },
+    "node_modules/@n8n/vm2": {
+      "version": "3.9.23",
+      "resolved": "https://registry.npmjs.org/@n8n/vm2/-/vm2-3.9.23.tgz",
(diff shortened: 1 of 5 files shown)


martinsefcik (approved):

helloanoop:
Thank you @Its-treason !

PR #1405, opened 2024-01-17
fix(#1000): Convert too long numbers into String for collection vars

### LINKED ISSUE #1000: Long integer gets truncated and altered
I have a REST endpoint written in Java Spring Boot. When I call it via Bruno, my path variable get truncated, see:

grafik

1. In the background is my debugger session in IntelliJ, exactly where the call is intercepted.
2. In the foreground is Bruno and you can see the passed in parameter for my `agency`.
3. It seems that the original value of `2017236094450943841` entered in Bruno gets somehow altered to be `2017236094450943700`.
4. When I mimic the exactly same call via `curl`, I don't see this truncation.

Is this a bug?

I'm using
```
Bruno: v1.1.0
electron: 21.1.1
chrome: 106.0.5249.103
node: 16.16.0
v8: 10.6.194.17-electron.0
```

Its-treason:
# Description

Closes #1000

Convert too long numbers into a string for collection vars so they don't get changed.

Before:

Screenshot 2024-01-17 225517

After:

Screenshot 2024-01-17 225221

### DIFF
--- a/packages/bruno-js/src/utils.js
+++ b/packages/bruno-js/src/utils.js
@@ -92,7 +92,12 @@ const evaluateJsTemplateLiteral = (templateLiteral, context) => {
   }
 
   if (!isNaN(templateLiteral)) {
-    return Number(templateLiteral);
+    const number = Number(templateLiteral);
+    // Check if the number is too high. Too high number might get altered, see #1000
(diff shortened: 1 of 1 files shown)


helloanoop:
Thank you @Its-treason !

Issue #960, opened 2023-11-13
[Bug] Cannot override Collection Basic Auth with request specific Bearer Token


kpau:
I have a collection with Auth set to `Basic Auth`:
![image](https://github.com/usebruno/bruno/assets/1583751/7a5d80e2-5422-4a81-aea6-c22250188885)

In the collection there is a GET Request, which has a `Bearer Token` auth:
![image](https://github.com/usebruno/bruno/assets/1583751/0ac41898-299c-41b5-b69e-84f3e7911519)

When I execute the request and observe the "Timeline" tab on the right, I see that it still uses the `Basic Auth` credentials:
![image](https://github.com/usebruno/bruno/assets/1583751/2e446ab7-11b0-4a73-a755-3aaa48f10d17)

I would expect that it will use the `Bearer Token` Auth specified in the request.

If I disable the Collection Auth (set it to `No Auth`), then the request correctly uses the `Bearer Token` auth:
![image](https://github.com/usebruno/bruno/assets/1583751/403d6889-2ca6-46cd-8bae-e915b9cc5ecb)

I am using Bruno v1.1.1

n00o:
You can fix this by using the Collection's **Headers** tab and add a `Authorization` header and make the value:
`Bearer {{accessToken}}`

Then every request inside of this collection will use that same header for its authentication. This is not the most ideal solution but fortunately others have voiced this issue too in #816 and it should be fixed in #907.

btgs-0:
@kpau if you're able, please try out my feature branch https://github.com/btgs-0/bruno/tree/feature/auth-from-collection and let me know if it works for you

kpau:
@n00o , Thanks for the info. But my issue is a bit different. I understand that `No Auth` uses Collection auth, but in my case I am explicitly setting another type of Auth and it is still using the Collection auth. The issues are related so the root cause _might_ be the same.

@btgs-0 , Thanks for the PR. If I have time today, I'll test the branch. 

Even if the issue is fixed with this PR, it is good to have both issues documented for future reference.

btgs-0:
thanks for testing my PR @kpau. And thanks @n00o for connecting the dots, my PR incidentally fixes this issue by making the collection auth inheritance explicit

karloluis:
It appears that https://github.com/usebruno/bruno/pull/1667 addresses this by explicit requirement of inheritance.

sanjai0py:
[...]

PR #1409, opened 2024-01-18
FIX: Request>Auth>Bearer configuration does not override if Collection>Auth>Basic is set.

karloluis:
# Description

Fixes https://github.com/usebruno/bruno/issues/1536 and https://github.com/usebruno/bruno/issues/960

The way axios handles the `auth` property on the request configurations makes it such that it will always take precedence over the `headers.authorization` for the headers object. This was causing issues where the Bruno Collection has Basic Auth and the Request being run has Bearer Auth.

## Reproducing

1. In Collection>Settings>Auth select Basic
2. In a Request>Auth select Bearer
3. Spin up an http server; preferably with a debugger and a breakpoint on the endpoint
4. Inspect the `req.headers` property the server received

This will result in `Basic ========`; in this fix it should be `Bearer =======`

## Alternative implementation

Instead of inverting logic and doing conditional validations consider potentially mutating the request object.
By doing `delete request.auth` on the condition for the request on bearer mode the bug can be avoided. Although I tend to shy away from mutating objects this way as it may have unintended side-effects.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-cli/src/runner/prepare-request.js
+++ b/packages/bruno-cli/src/runner/prepare-request.js
@@ -1,4 +1,4 @@
-const { get, each, filter } = require('lodash');
+const { get, each, filter, isEmpty } = require('lodash');
 const decomment = require('decomment');
 
 const prepareRequest = (request, collectionRoot) => {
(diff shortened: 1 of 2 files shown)


packages/bruno-cli/src/runner/prepare-request.js
karloluis, line 1: Best read in split diff view.

karloluis:
Appears to be fixed by https://github.com/usebruno/bruno/pull/1667

PR #1461, opened 2024-01-27
feat(#1460): string interpolation

### LINKED ISSUE #1460: String Interpolation
String Interpolation

### Goal
Today our interpolation logic is duplicated across multiple packages (App, Electron, CLI)
The goal is to centralize a single source of truth for all interpolation logic.

### Considerations
* We want to be flexible in terms of key naming conventions
  * #931 
  * #964 
* We plan to support Nested environments in the future
  * #292 
  * #1098

### Moving away from handlebars
I think its time to move away from handlebars.
We don't need the full power of handlebars and we want write a custom interpolation function that meets our needs instead of shoehorning handlebars to match our use case.

### Related Tickets
* #1143
* #1148
* #1150
* #1367
* #1321
* #1410

helloanoop:
# Description

Closes #1460

### Contribution Checklist:

- [x] The pull request only addresses one issue or adds one feature
- [x] The pull request does not introduce any breaking changes
- [ ] I have added screenshots or gifs to help explain the change if applicable.
- [x] I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).
- [x] Create an issue and link to the pull request.


packages/bruno-common/src/utils/index.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    const output = flattenObject(input);
+    expect(output).toEqual({});
+  });
+
+  it('should handle an object with nested empty objects', () => {
+    const input = { a: { b: {}, c: { d: {} } } };
+    const output = flattenObject(input);
+    expect(output).toEqual({});
+  });
+});
```
martinsefcik, line 39: Probably these two additional test cases could be added:

```javascript
it('should handle an object with duplicate keys - dot notation used to define the last duplicate key', () => {
  const input = { a: { b: 2 }, 'a.b': 1 };
  const output = flattenObject(input);
  expect(output).toEqual({'a.b': 1});
});

it('should handle an object with duplicate keys - inner object used to define the last duplicate key', () => {
  const input = { 'a.b': 1, a: { b: 2 } };
  const output = flattenObject(input);
  expect(output).toEqual({'a.b': 2});
});
```

to ensure that always last defined key wins and it does no matter if it is defined by dot notation or by inner object.

martinsefcik (approved):

packages/bruno-common/src/utils/index.spec.ts
helloanoop: Done: https://github.com/usebruno/bruno/commit/2a6bfabc302ec83298f3ae03eb33b0b15eb9d843

helloanoop:
Merged. Thanks for the review @martinsefcik !

Issue #968, opened 2023-11-15
Auth: Cookies


Tanja-4732:
This project _really_ needs to support storing and automatically re-sending cookies, just like a browser would.

I'll work on an PR for this:

I'll work on cookies [in my fork](https://github.com/Tanja-4732/bruno) since you seem to be working on auto-save.

I hope to have some progress by the end of the week; will (hopefully) open a draft-PR until then

_Originally posted by @Tanja-4732 in https://github.com/usebruno/bruno/discussions/385#discussioncomment-7571305_

### Tasks
- [x]  Basic Cookie Support
- [ ] CLI Support
- [x] Delete a cookie in UI
- [ ] Edit a cookie in UI
- [ ] Add a cookie in UI
- [ ] Add a new domain/url in UI

nu111:
I am trying to move away from Insomnia or Postman, I would like to try Bruno, but I can't without cookies. I'd love to see this feature!

solsticesurfer:
+1 on cookie support. It's the only thing preventing my team from dumping Postman and moving over to Bruno. Far too many systems rely on a session cookie as part of their authentication flow.

chriswinter90:
+1 for many at our company. We use sessions on some of our node apps and the session id changes every time because the cookie is not saved causing a new session each time so no user info is persisted.

Tanja-4732:
Yes yes, I'll try to get started on this till about Monday,  
though I'm not very familiar with the codebase yet

joe-gre:
I'm trying to get cookie based auth working, but I don't think its sending the cookie header. Here is what I have in the Headers tab, which works in Postman but not Bruno: 

![image](https://github.com/usebruno/bruno/assets/22748761/21615b52-aecf-4805-9d3d-ca1d925c3d9f)

I tried setting the session cookie to valid string literal session ids too, both encoded and not encoded. Same issue.

Also FYI this is what I have in the post script to decode the cookie response and set it as a context var. This part is working well.
```
const header = res.getHeader("set-cookie")?.[0];
if (!header) return;

const session = header.substring(
  header.indexOf("=") + 1,
  header.indexOf(";")
);
bru.setVar("session", decodeURIComponent(session));
console.log(`session: ${bru.getVar("session")}`)
```

joe-gre:
[...]

PR #1463, opened 2024-01-28
"Add" cookies feature. Support for cookie propagation in case if target page returns 302.

r-polunin:
Related to:
https://github.com/usebruno/bruno/issues/968

If we have a chain of requests, i.e: request -> 302 re-direct -> 302 re-direct -> page.
Cookies from the first re-direct response are not propagated further.
This happens because of the way how axios works:
https://github.com/axios/axios/issues/3862

During the request preparation, if "Save cookies" checkbox is set, maxRedirections for axios instance will be set to 0 and further re-direction requests will be triggered re-cursively in error handler, new interceptors will be storing and enriching cookies.

Cookies UI:
New element allowing manual addition of the cookies is implemented. Table key is changed from domain to domain + path.

### DIFF
--- a/packages/bruno-app/src/components/Cookies/StyledWrapper.js
+++ b/packages/bruno-app/src/components/Cookies/StyledWrapper.js
@@ -11,6 +11,26 @@ const Wrapper = styled.div`
       user-select: none;
     }
   }
+
+  .settings-label {
+    width: 80px;
+  }
+
+  .textbox {
+    border: 1px solid #ccc;
+    padding: 0.15rem 0.45rem;
+    box-shadow: none;
+    outline: none;
+    transition: border-color ease-in-out 0.1s;
+    border-radius: 3px;
+    background-color: ${(props) => props.theme.modal.input.bg};
+    border: 1px solid ${(props) => props.theme.modal.input.border};
+
+    &:focus {
+      border: solid 1px ${(props) => props.theme.modal.input.focusBorder} !important;
+      outline: none !important;
+    }
+  }
 `;
 
 export default Wrapper;
--- a/packages/bruno-app/src/components/Cookies/index.js
+++ b/packages/bruno-app/src/components/Cookies/index.js
@@ -2,8 +2,9 @@ import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import Modal from 'components/Modal';
 import { IconTrash } from '@tabler/icons';
-import { deleteCookiesForDomain } from 'providers/ReduxStore/slices/app';
+import { deleteCookiesForDomain, addCookiesForURL } from 'providers/ReduxStore/slices/app';
 import toast from 'react-hot-toast';
+import { useFormik } from 'formik';
 
 import StyledWrapper from './StyledWrapper';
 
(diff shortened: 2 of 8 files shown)


martinsefcik:
1. I didn't check this PR in detail but I would expect to be implemented also for bruno-cli and not only for bruno-app.
2. It is better for two things (cookie addition and redirection modification) to do two PRs and not to merge it into one PR.
    It is first note mentioned in checklist when creating PR: `The pull request only addresses one issue or adds one feature.`
4. Didn't you check this https://www.npmjs.com/package/http-cookie-agent mentioned in axios issue thread you posted https://github.com/axios/axios/issues/3862#issuecomment-1681387359 ? If I understood it corectly this agent is handling cookies and we would not need to do redirections manually in such case (?) If it is like that then I like it more.

r-polunin:
Hi @martinsefcik, thanks for review,

>1) I didn't check this PR in detail but I would expect to be implemented also for bruno-cli and not only for bruno-app.
>
Hmm, let me check this on the weekend. Seems as there is no cookie support for cli version at all, possibly can go as a separate PR.

>2) It is better for two things (cookie addition and redirection modification) to do two PRs and not to merge it into one PR.
It is first note mentioned in checklist when creating PR: The pull request only addresses one issue or adds one feature.
> 
Ok, will split

>3) Didn't you check this https://www.npmjs.com/package/http-cookie-agent mentioned in axios issue thread you posted https://github.com/axios/axios/issues/3862#issuecomment-1681387359 ? If I understood it correctly this agent is handling cookies and we would not need to do redirections manually in such case (?) If it is like that then I like it more.
>
Yep, I've checked and played with this one a little.
I've used "export const createCookieAgent(http.Agent);
It was working for my case, but as I see, default httpAgent is overriden in Bruno:
https://github.com/r-polunin/bruno/blob/00c3f42589693c86af618d4b445ee394696e74fc/packages/bruno-electron/src/ipc/network/index.js#L172
I'm not sure how will it behave, if we wrap current custom Agent in Cookie Agent will check on the weekends .

In my implementation, one of issues I didn't consider -> recursive redirects. They won't stop, as there is no max redirects count property anymore, just recursive calls. This one will require a fix If decision is made to keep custom implementation and not the lib.

Will check both approaches on the weekends.

r-polunin:
@martinsefcik, good day!
Had some time today to checked the cookie agent library, and faced couple of challenges, with such wrapper:
```js
const wrapInCookieAgent = (AgentClass, ...args) => {
  const CookieAgentClass = createCookieAgent(AgentClass);
  const jar = getCookieJar();
  return new CookieAgentClass({ cookies: { jar }} , ...args);
}
```

1) In library's overriding class, Params are sent in super constructor as ...params.
```js
constructor(...params: ConstructorParams) {
      const { cookies: cookieOptions } = (params.find((opt) => {
        return opt != null && typeof opt === 'object' && 'cookies' in opt;
      }) ?? {}) as CookieAgentOptions;

      super(...(params as Params));
```
https://github.com/3846masa/http-cookie-agent/blob/a2f3b4e0bc9a33e20bc5cbcc395488abbee611f6/src/http/create_cookie_agent.ts#L53

It messes with SocksProxyAgent constructor which is used: 
image
As it treats "CookieAgentOptions" required for CookieAgent as a first param.
Do you know any clean way to overcome such limitation?

2) Bruno has 2 checkboxes: "Store" & "Send" cookies now. 
Both On/Off can be supported.
"Store" - Off, "Send" - On can be supported too (create new jar instance, copy from the global state)
But "Store" - On, "Send" - Off can't, as logic is hidden behind private method:
https://github.com/3846masa/http-cookie-agent/blob/a2f3b4e0bc9a33e20bc5cbcc395488abbee611f6/src/http/create_cookie_agent.ts#L98

If there is no way to handle these challenges in a good way, I can:
1) Write a similar custom agent to cover our case.
2) Fork lib rep & try to discuss with author required extensions.
3) Keep interceptors approach and handle retries.

What would you suggest to be the best way?

naman-bruno:
Hey @r-polunin, thanks for taking the time to contribute! 
Closing this PR.

PR #1475, opened 2024-01-29
feat: visualize better when a row is disabled

banjo:
# Description

The checkbox currently has no visual indication whenever it is pressed, this PR adds a slightly lower opacity and light gray background color to the row to make it clear that it is disabled.

CleanShot 2024-01-29 at 22 37 53

This applies to all places where it can be disabled (I think).

Also, my IDE automatically sorted the imports 🫣

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/CollectionSettings/Headers/index.js
+++ b/packages/bruno-app/src/components/CollectionSettings/Headers/index.js
@@ -1,18 +1,18 @@
-import React from 'react';
-import get from 'lodash/get';
-import cloneDeep from 'lodash/cloneDeep';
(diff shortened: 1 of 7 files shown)


martinsefcik:
@banjo  Question is if delete button opacity should be also changed as this button is doing the same action and it does not matter if row is enabled or disable.  Also checkbox opacity is also questionable. It looks a bit to me as these two action components are also disabled but they are not. What about changing "opacity" for whole row but not for these two components ? But I guess It's need to be done differently using background, text styling and not using opacity because these two components are children of row.

PR #1480, opened 2024-01-30
feat: prettify json body automatically

banjo:
# Description

Currently, the json body is by default in "string" mode, and you have to press the prettify button to fix it. 

This update automatically prettifies the JSON on mount IF it is a valid JSON. This means that, for example, if you add a new curl import, it will default to a prettified view. And if it is not a valid JSON-format, it will do nothing. 

Here is an example:

auto-format

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/RequestPane/RequestBody/RequestBodyMode/index.js
+++ b/packages/bruno-app/src/components/RequestPane/RequestBody/RequestBodyMode/index.js
@@ -1,13 +1,15 @@
-import React, { useRef, forwardRef } from 'react';
-import get from 'lodash/get';
 import { IconCaretDown } from '@tabler/icons';
 import Dropdown from 'components/Dropdown';
-import { useDispatch } from 'react-redux';
+import get from 'lodash/get';
 import { updateRequestBodyMode } from 'providers/ReduxStore/slices/collections';
-import { humanizeRequestBodyMode } from 'utils/collections';
-import StyledWrapper from './StyledWrapper';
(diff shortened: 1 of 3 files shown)


banjo:
I'm not really sure why the tests here fail, anyone knows why?

The `Publish test report` step fails in the pipe.

```
Run dorny/test-reporter@v1
Action was triggered by pull_request: using SHA from head of source branch
Check runs will be created with SHA=6cbde0cd915fd7102077e34a806c7[14](https://github.com/usebruno/bruno/actions/runs/7871199178/job/21473980886?pr=1480#step:7:15)a1747f96a
Listing all files tracked by git
Found 646 files tracked by GitHub
Using test report parser 'java-junit'
Creating test report Test Report
  Processing test results for check run Test Report
  Creating check run Test Report
Error: HttpError: Resource not accessible by integration
```

helloanoop:
Thanks for taking time to work on this @banjo .

One issue I see with this approach is that - this essentially force prettifies the body every time the request body is opened, leaving no option for the user to format the body as per their needs.

> This means that, for example, if you add a new curl import, it will default to a prettified view. And if it is not a valid JSON-format, it will do nothing.

I think trying to auto prettify the curl json body before creating a request still makes sense

helloanoop:
As mentioned [here](https://github.com/usebruno/bruno/pull/1480#issuecomment-2307765344), I don't think its a good idea to force prettify json on mount.

That said, auto prettifying the curl json body during import seems reasonable, and we'd be happy to consider a PR for that.

Closing this PR

Issue #1516, opened 2024-02-04
Case-sensitive renaming


starwalkn:
When I try to rename a query with changes only in the case of the letter, then after saving the name remains the same.

PR #1528, opened 2024-02-06
allowed case-sensetive name change for folders and requrests.

Exxetler:
# Description

Changed two lines to allow renaming folders and requests to the same name with different-case letters. Fixes [1516](https://github.com/usebruno/bruno/issues/1516)

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/collection.js
+++ b/packages/bruno-electron/src/ipc/collection.js
@@ -309,7 +309,7 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
       if (!fs.existsSync(oldPath)) {
         throw new Error(`path: ${oldPath} does not exist`);
       }
-      if (fs.existsSync(newPath)) {
+      if (fs.existsSync(newPath) && path.basename(oldPath) === path.basename(newPath)) {
         throw new Error(`path: ${oldPath} already exists`);
       }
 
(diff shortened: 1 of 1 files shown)


packages/bruno-electron/src/ipc/collection.js
```diff
@@ -309,7 +309,7 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
       if (!fs.existsSync(oldPath)) {
         throw new Error(`path: ${oldPath} does not exist`);
       }
-      if (fs.existsSync(newPath)) {
+      if (fs.existsSync(newPath) && path.basename(oldPath) === path.basename(newPath)) {
```
Its-treason, line 312: While this works to allow to change the request's name when just changing a character case, it also allows for request to overwrite each other when renaming.

This problem has been discussed in #531 and I added a possible solution for this here: https://github.com/usebruno/bruno/pull/531#issuecomment-1888068214

This could be implemented like this:

```js
      if (fs.existsSync(newPath)) {
        // Try to resolve both paths, if they reference the same file we are on a case-insentive filesystem like NTFS and are trying to rename a file
        if (fs.statSync(newFilePath).ino !== fs.statSync(oldFilePath).ino) {
          throw new Error(`path: ${oldPath} already exists`);
        }
      }
```

This kinda needs some testing with network drive, to check if [ino](https://nodejs.org/api/fs.html#class-fsstats) is present everywhere

Its-treason (changes requested):

Exxetler:
Alright, reading through #531 it seems a lot more involved than I anticipated. I'm only capeable of testing on windows right now, I'll close the PR and might come back to this issue at a later time. Thanks for the resoponse.

Issue #1101, opened 2023-11-30
[BUG] Vars in request override the environment variables definitely


luc-freyermuth:
Hello, I am here to report a bug I found when starting to use bruno at work !

## Description of the issue

I create an environment variable, lets say `project_id` and set a value to the variable, for example `10`. I enable this environment and create a request that uses this env variable in its path. eg: `http://localhost:3000/projects/{{project_id}}/users`. When I run the request, everything works, the value `10` is reflected in the URL.

Then, I create a var of the same name in the Vars > Pre Request section. Lets says `project_id` with value `20`. When I send the request, now the value `20` is reflected in the URL as expected.

Then, I delete my Pre Request Variable and send the request again. **The value `20` is still reflected in the URL**. And the value will be reflected everywhere, even in other requests. This seems to indicate that the global state of the application has been affected by my request Vars. At this point, the only solution I found to reset my variable to its original value is to restart bruno completely.

## Expected behavior

As a user, I would expect that data specified in the Vars tab of one request should not affect the application global state, so that when I change requests or delete the Vars line, the behavior goes back to where it was before i created the Vars line.

## My 2 cents as a dev

I didn't take the time to take a look at bruno's source code, so I could be really wrong. This behavior gives the feeling that global state is used and modified while processing the request. This could become an issue in future developments, as it couples the action of sending the request with the use case (sending a request by clicking "send" in an item of a collection). This could make some future requests harder to implement (eg. creating a runner that sends multiple requests from a file) because you'll have to control and remember all the side effects to avoid unexpected behavior ! 

And thanks for all your work on bruno ! :)

SoulKa:
Duplicate of #1526

epestela:
I must agree that is a quite annoying issue. 
Encountered this quite fast tbh while testing out Bruno
because of the unexpected behaviors that this can create I think it's quite critical to address this

SoulKa:
Deactivating the vars now works as expected with the fix #1539. However, deleting still doesn't work!

samuel-deal-tisseo:
Maybe a proper solution is to use a different syntax for vars and env values.

But this would be a massive breaking change

PR #1539, opened 2024-02-07
Bugfix #1101 Vars in request override the environment variables definitely

peerb:
Bugfix #1101

# Description

Introduced new method to bru.js for deleting request variables. This will allow the user to enable and disable request variables for requests. 

Bugfix #1101

### Contribution Checklist:

### DIFF
--- a/packages/bruno-js/src/bru.js
+++ b/packages/bruno-js/src/bru.js
@@ -66,6 +66,21 @@ class Bru {
     this.collectionVariables[key] = value;
   }
 
+  deleteVar(key) {
+    if (!key) {
+      throw new Error('Deleting a variable without specifying a name is not allowed.');
+    }
+
+    if (variableNameRegex.test(key) === false) {
+      throw new Error(
+        `Variable name: "${key}" contains invalid characters!` +
+          ' Names must only contain alpha-numeric characters, "-", "_", "."'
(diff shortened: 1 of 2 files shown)


peerb:
@helloanoop any suggestions from your side regarding this bugfix?

epestela:
any news on this case?

vladdeSV:
If there is no environment `{{version}}`, will the request variable be empty if disabled?

epestela:
@helloanoop this case seam to be solves in 1.20.2

@vladdeSV in 1.20.2
- if the variable is defined in the URL, it's replaced by an escaped variable tag 
  - {{undefined}} becomes %7B%7Bundefined%7D%7D
- if it's in the request header it's maintained with no repmacement
  - {{undefined}} remains {{undefined}}
currently you must define an empty variable to have it removed in the query

helloanoop:
Thanks for the bump @epestela !

Thanks for your work on this @peerb 

Th issue has been fixed in `v1.20.2` release
Before v1.20.2 the request variables were being persisted and were getting leaked into the global scope.

The release also includes support for `bru.deleteVar()` api to delete the runtime vars.

We have also added support for Folder Level Variables which gets applied to all requests under the folder.

PR #1667, opened 2024-02-26
feat(#BRU-18) : "inherit" auth mode for inheriting auth to requests from collection

lohxt1:
Inherit auth to requests from collection

[inherit_auth_mode.webm](https://github.com/usebruno/bruno/assets/25679466/cc95329d-41d5-47ae-9b9a-7c1667dde14f)

[inherit_auth_mode_cli.webm](https://github.com/usebruno/bruno/assets/25679466/568128af-ce45-4cda-b237-e6c1691e36e0)


helloanoop:
Merged!

Thanks for taking care of this @lohxt1 !

helloanoop:
@lohxt1 Just a FYI, I have made changes to gracefully handle older collections which are using `none` and still inherit collection auth.

Plan is to give some time to for users to migrate (and broadcast these via x/discord/communit calls) before causing CI pipelines to break.

See https://github.com/usebruno/bruno/commit/1cf8a2f3f1d7f2f321c8e96b2ec0c2fed71ba371

Tydax:
> @helloanoop is there a way to make `none` actually None? I want to be able to use inherit, but have a route that needs to not inherit.

I can confirm that I have the same issue, it seems that `No auth` on a request does not “override” the collection authentication setting and defaults to an `inherit` behaviour. I think this is a bug.

For now, I got it working by manually defining an `Authorization` header in the _Headers_ tab in the collection's settings and then adding another `Authorization` header with a blank value on the request that should not use it.

helloanoop:
Breaking change will be shipped in `v1.16.0`
After which **None means None**

melroy89:
Two tests are failing tho..?

slowjoe007:
Also figured this in #2057. The tests appear to fail due to this change when executed via `bru` CLI. The request `auth/inherit auth` is working inside the GUI though.

melroy89:
I don't understand why a failing CICD run was merged, causing regression. But yea.

PR #1796, opened 2024-03-14
Fix: OAuth2 Access Token request is sent as GET

### LINKED ISSUE #1795: BUG: OAuth2: Access Token request is sent as GET
Bruno 1.11.0
When setting the OAuth2 auth method on a GET request, Bruno attempts to obtain Access Token from token endpoint also using GET. The [specs](https://www.rfc-editor.org/rfc/rfc6749#page-21) however mandates that 

> The client MUST use the HTTP "POST" method when making access token requests.

#1003

pietrygamat:
# Description

The client MUST use the HTTP "POST" method when making access token requests.

### Contribution Checklist:

Fixes #1795

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -203,6 +203,7 @@ const configureRequest = async (
         interpolateVars(requestCopy, envVars, collectionVariables, processEnvVars);
         const { data: authorizationCodeData, url: authorizationCodeAccessTokenUrl } =
           await resolveOAuth2AuthorizationCodeAccessToken(requestCopy, collectionUid);
+        request.method = 'POST';
         request.headers['content-type'] = 'application/x-www-form-urlencoded';
         request.data = authorizationCodeData;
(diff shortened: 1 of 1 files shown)


packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -203,6 +203,7 @@ const configureRequest = async (
         interpolateVars(requestCopy, envVars, collectionVariables, processEnvVars);
         const { data: authorizationCodeData, url: authorizationCodeAccessTokenUrl } =
           await resolveOAuth2AuthorizationCodeAccessToken(requestCopy, collectionUid);
+        request.method = 'POST';
```
pietrygamat, line 206: The request here is just the token endpoint request, right? I am not overwriting the user's collection request, I hope?

francoischaix-circeo:
The POST request is also valid for every flow. however this change only does it for authorization flow.

francoischaix-circeo (commented):
This applies the request method only to the authorization flow. But other flows are also requred to be sent as POST in every case. this will not fix the client_credentials and password flow

Issue #1800, opened 2024-03-14
GraphQL query editor prettify icon lays on top of environment config modal 


devon-wolf:
When editing a GraphQL request, the "prettify" icon button does not render below the environment config modal as expected - it remains visible when the modal is open.

In this screenshot, it renders above the value for the `jwt` variable:
<img width="976" alt="Screenshot 2024-03-13 at 8 15 25 PM" src="https://github.com/usebruno/bruno/assets/33987744/788bc6ba-3b6f-4056-8b04-e477a255dbab">

This may be something I can make a PR for myself, but in case I don't have the time I hope someone can pick it up. It seems like a z-index change would fix it.

Thanks!

helloanoop:
Thanks for the PR @devon-wolf !

This will go out in the `1.13.0` release. Please subscribe to [Discord](https://discord.com/invite/KgcZUncpjq) announcement channel to stay notified on the releases.

PR #1801, opened 2024-03-14
GraphQL query editor prettify icon lays on top of environment config modal #1800

ialloyd:
This PR adjusts the `z-index` of the "Prettify" button to prevent it from overlapping with the environment config modal.


devon-wolf:
This seems to be intended to address #1800, but is not acting on the correct element. This issue is also addressed in PR #1821.

Issue #1799, opened 2024-03-14
GraphQL query dark mode - cursor is black when it should be white


devon-wolf:
When editing a GraphQL query in dark mode, the cursor color is black and very difficult to see. In regular HTTP requests in dark mode the cursor is white.

In this screenshot, the cursor can be (barely) seen on line 5 after the closing curly brace:
<img width="596" alt="Screenshot 2024-03-13 at 8 14 08 PM" src="https://github.com/usebruno/bruno/assets/33987744/3eecc59f-3634-4202-b295-8d1bb1dea3e3">

I may be able to make a PR for this myself, time permitting, but figured I'd open this issue in the meantime in case I can't get to it. Thank you for such a lovely project!

sanjai0py:
PR #1823 merged.

PR #1802, opened 2024-03-14
GraphQL query dark mode - cursor is black when it should be white #1799

ialloyd:
In dark mode, the cursor in the GraphQL query editor is black, making it hard to see. This issue aims to change the cursor color to white for better visibility in dark mode.


devon-wolf:
Related to issue #1799 - this is also resolved in a more complete way in PR #1823

Issue #1761, opened 2024-03-11
Using a `/` in an Environment Name causes error


christopher-nash:
I ran into this while attempting to import a Postman Environment. I had an environment I called "Beta / Staging" in Postman which would result in an `An error occurred while importing the environment`. After some experimenting with the json I removed the `/` character and it imported fine.

To reproduce:
1. Click `Configure` for the environments
2. Click `+ Create`
3. Set the Environment Name to `Some / Thing` (any name which includes a `/` will cause the error)
4. Click the `Create` button
5. Note the error `An error occurred while created the environment` is displayed

Setting the environment to include other special characters such as `~` or `=` or even `\` works fine. It only appears to happen with `/`.

I tried looking in the developer console and I don't see any errors showing up there. I'm not sure where to go looking for generated log files that might give additional details on the actual error root cause.

leeclarke:
I encountered the same issue with a colon as well in the title. Snippet of the environment file I was trying to import:
I edited the file and removed the the ":" and it imported to verify it was the colon
```json
{
	"id": "af02b0ad-2048-1f15-b544-e8011d2e0ca7",
	"name": "3: CAT",
	"values": [
		{
			"key": "base_url",
			"value": "https://example.com/",
			"enabled": true
		},
	 ]
}
```

busy-panda:
Fixed for any character not allowed in filenames

anusree-bruno:
This issues has been fixed. Invalid characters will be automatically converted to - on import.

PR #1804, opened 2024-03-14
[Bug] Using a / in an Environment Name causes error in Postman Import #1761

ialloyd:
This PR fixes a bug that caused environment import to fail if the name contained a /. The importPostmanEnvironment function now replaces / with _ in environment names, allowing successful import.

### DIFF
--- a/packages/bruno-app/src/utils/importers/postman-environment.js
+++ b/packages/bruno-app/src/utils/importers/postman-environment.js
@@ -31,8 +31,9 @@ const importPostmanEnvironmentVariables = (brunoEnvironment, values) => {
 };
 
 const importPostmanEnvironment = (environment) => {
+  const sanitizedEnvironmentName = environment.name.replace(/\//g, '_');
   const brunoEnvironment = {
-    name: environment.name,
+    name: sanitizedEnvironmentName,
     variables: []
   };
 


npu78954:
This PR fix the bug https://github.com/usebruno/bruno/issues/1761

Could it be merged please?

packages/bruno-app/src/utils/importers/postman-environment.js
```diff
@@ -31,8 +31,9 @@ const importPostmanEnvironmentVariables = (brunoEnvironment, values) => {
 };
 
 const importPostmanEnvironment = (environment) => {
+  const sanitizedEnvironmentName = environment.name.replace(/\//g, '_');
```
igor-topolski, line 34: This will not work for `Some / Environment / Name` as it'll result in `Some _ Environment / Name`, also there are more forbidden characters. 

Any character that makes an invalid file name will throw an exception (`/` for Windows, but Unix-based systems have more symbols).

Pragadesh44-Bruno:
Hey @ialloyd , thank you for raising this PR. In our latest versions, we have fixed the same bug you mentioned. If there are any invalid characters in the environment filename, they will be replaced with a “-”.
Thanks for the contribution!



https://github.com/user-attachments/assets/442fec23-9eda-46a9-b148-e1d1bd4d25a9

Issue #1651, opened 2024-02-22
[Bug] Error importing Postman Collection with [ ] symbols in it's name


acarrascosa:
I've been using bruno with no problem, but seems to have some problems importing some collections, they create well in the Bruno Collections folder, but then they does not appear in Bruno itself

acarrascosa:
The error seems to be when you try to import a postman collection with "[ ]" in it's name. I've just removed the symbols and works perfectly

anusree-bruno:
This has been fixed in latest release. Closing the issue, feel free to reopen if this still happens to you.

PR #1811, opened 2024-03-14
[Bug] Error importing Postman Collection with [ ] symbols in it's name #1651

ialloyd:
Fixes an issue with importing Postman collections that have square brackets in their names. This change sanitizes the collection name by removing square brackets, preventing import errors and crashes. Resolves issue #1651.

### DIFF
--- a/packages/bruno-app/src/utils/importers/postman-collection.js
+++ b/packages/bruno-app/src/utils/importers/postman-collection.js
@@ -53,6 +53,10 @@ const convertV21Auth = (array) => {
   }, {});
 };
 
+function sanitizeName(name) {
+  return name.replace(/\[|\]/g, ''); // This will remove square brackets
+}
+
 const importPostmanV2CollectionItem = (brunoParent, item, parentAuth) => {
   brunoParent.items = brunoParent.items || [];
   const folderMap = {};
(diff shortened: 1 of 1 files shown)


helloanoop:
@bpoulaindev Can you review this?

bpoulaindev (approved):
Works fine on my side, this would also replace a single [ or ] in the name but it's fine.
My IDE suggests a small change in the syntax but the function's behavior remains identical : 
image
Work fine even on a hardcore name : 
image

packages/bruno-app/src/utils/importers/postman-collection.js
```diff
@@ -54,7 +54,11 @@ const convertV21Auth = (array) => {
   }, {});
 };
 
-const importPostmanV2CollectionItem = (brunoParent, item, parentAuth, options) => {
+function sanitizeName(name) {
+  return name.replace(/\[|\]/g, ''); // This will remove square brackets
```
igor-topolski, line 58: This will not work for `Some / Environment / Name` as it'll result in `Some _ Environment / Name`; also, there are more forbidden characters. 

Furthermore, this is prone to introduce a silent crash if the user has two environments following the same pattern, e.g.  `Some / Env` and `Some [ Env` with results in the same name - `Some  Env`.

I proposed a detailed error instead to inform user of the invalid characters (https://github.com/usebruno/bruno/pull/2200)

naman-bruno:
Hey @ialloyd , thanks for your contribution!
Closing this PR.

Issue #1553, opened 2024-02-08
Prettify not working when “JSON body” contains variable


bogmart:
Hello,
I noticed a bug related to “Prettify” functionality when I put variables in 
![Screenshot (251)_bruno_err](https://github.com/usebruno/bruno/assets/36724163/7a05ca37-0270-4285-a52a-7f6976996f39)
a “JSON body”.

Example:
I have a variable named " input_idx" (it is an integer).
The “Prettify” for this “JSON body” is not possible:
{"method": "test", "params": {"input" : {{input_idx}} }}

![Screenshot (251)_bruno_err](https://github.com/usebruno/bruno/assets/36724163/15f7a364-156b-4298-b13b-675ee145f8db)

nlassaux:
Are you actually able to access the variable? I can't manage to get bruno to work at all if I pass a variable the way you did

bogmart:
> Are you actually able to access the variable? I can't manage to get bruno to work at all if I pass a variable the way you did

Yes, the variable is correctly expanded/replaced by Bruno with its value (in my case, Bruno sends a  HTTP POST). Only the prettify functionality is affected.
I just retested with Bruno v1.11.0, on WSL Ubuntu 22.04.

nlassaux:
Got it, thanks. Super weird, in my case Bruno sends the variable with the curly brackets to the backend. I'm closely monitoring #1812 in case that fixes it at the same occasion

bogmart:
Update: on version v1.17.0, the prettify generates an invalid JSON from my example.

```
{
  "method": "test",
  "params": {
    "input": {
      {input_idx
      }
    }
  }
}
```

nathanmcgarvey-modopayments:
Also see https://github.com/usebruno/bruno/issues/3019

helloanoop:
PR - #5113 has been merged that fixes this issue. This will be available in our upcoming - `v2.14.0` release scheduled to go out in last week of October

PR #1812, opened 2024-03-14
Prettify not working when “JSON body” contains variable #1553

ialloyd:
This PR enhances the JSON “Prettify” functionality to handle variables within the JSON body. The function now replaces variables with unique dummy values for prettification and then restores them, ensuring correct parsing and formatting of the JSON body.


asonkeri:
This works fine with unquoted variables e.g. `"hello": {{var}}` but fails when variables are inside quotes like `"hello": "{{var}}"`. The quoted variable case gets turned into `"hello": ""dummyValue0""` which json-bigint fails to parse as it is not valid JSON.
Seems that `jsonc-parser`, as used in #1831, would be able to properly format even this double quoted value.

helloanoop:
Closing this as this has been addressed by #5113

Issue #1910, opened 2024-03-23
Var used in POST's request body (JSON body) not being identified and evaluated correctly


nlassaux:
I have a Var that's declared, but I can't manage to use it as part of a json body.

<img width="1323" alt="The PropertyDetails Var is defined and instantiated" src="https://github.com/usebruno/bruno/assets/1917120/58ebf5b8-c909-48db-935f-ed6226388603">

Now I'm trying to use it in that JSON body:

<img width="1323" alt="The Var is not being sent correctly" src="https://github.com/usebruno/bruno/assets/1917120/3aa7b64a-f363-4b75-9493-d83625e2c8e3">

The documentation doesn't show any example for POST requests + JSON body. But do we agree that this variable should be sent in theory?

dw-0:
Sounds a bit like the issue #1635 

If possible, you could try and downgrade to v1.6.1 and try again. So you could at least verify that bug.
I tried with every version after v1.6.1 and all versions after it seem to be broken in that regard.

nlassaux:
I'm using v1.12.2, adding it to the issue's description! The new features were good though :(

dw-0:
>  The new features were good though :(

Yes :/ i really liked them as well. But i had to downgrade as my requests really depend on that feature to work :/

nlassaux:
Thank so much for your help, downgrading as well over here then!

Its-treason:
@dw-0 @nlassaux I implemented a fix in my fork: https://github.com/Its-treason/bruno/releases/tag/nightly that should hopefully fix your problem. Could test it locally and give feedback on whether it fixes your Issue? The fork can safely be installed next to mainline Bruno.

dw-0:
@Its-treason Sounds good! I will give it a try tomorrow and report back. Thank you.

dw-0:
@Its-treason Hey! First short feedback after the first short try this morning. I wasn't able to test anything really, as the program just showed a white page when i clicked on the little eye icon on the left side of the environment dropdown. Same happend when clicked on "configure" in the env dropdown. Env variables seem to be used in the request url, but not in the body.
![image](https://github.com/usebruno/bruno/assets/31533186/08becfc3-3694-4ee8-a1e3-ba7a2e485d01)
[...]

Issue #1635, opened 2024-02-21
[BUG] [>1.6.1] Array variables created with pre-request scripts are not treated as variables in Body.


Nerkho:
I recently noticed an issue that happens with version 1.7.0 and onwards.

The issue is similar to https://github.com/usebruno/bruno/pull/1217, but also affects the JSON request body and as far as I noticed only with variables created by a pre-request script.
The issue doesn't appear with version 1.6.1 and lower.

In the pre-request script, we create an array to be used in the request body.

```javascript
const project = bru.getEnvVar("PROJECT");
const roles = (bru.getEnvVar("ROLES").split(",")).map(element => {
  return element.trim();
});

var project_roles = [];

for (let i in roles) {
  project_roles.push("\""+project+"_"+roles[i]+"\"");
}

bru.setVar("roles", project_roles);
```

Request body :
```json
{
  "foo" : {{roles}}
}
```

In the response, we see the that the variable was treated as a plain value.
```json
{
  "args": {},
  "data": "\"{\\n  \\\"foo\\\" : {{roles}}\\n}\"",
  "files": {},
  "form": {},
  "headers": {
    "x-forwarded-proto": "https",
    "x-forwarded-port": "443",
    "host": "postman-echo.com",
    "x-amzn-trace-id": "Root=1-65d5c392-33b6d41c1866a3384c94b2c7",
    "content-length": "29",
    "accept": "application/json, text/plain, */*",
    "content-type": "application/json",
    "request-start-time": "1708508050633",
    "user-agent": "axios/1.6.7",
    "accept-encoding": "gzip, compress, deflate, br"
  },
  "json": null,
  "url": "https://postman-echo.com/post"
}
```

This only seems to impact Arrays.
If I replace my pre-request script with a simple string...

```javascript
bru.setVar("roles", "bar");
```

It works fine :
```json
{
  "args": {},
  "data": "\"{\\n  \\\"foo\\\" : bar\\n}\"",
  "files": {},
  "form": {},
  "headers": {
    "x-forwarded-proto": "https",
    "x-forwarded-port": "443",
    "host": "postman-echo.com",
    "x-amzn-trace-id": "Root=1-65d5c439-78924a0e68d4e1394df59f22",
    "content-length": "23",
    "accept": "application/json, text/plain, */*",
    "content-type": "application/json",
    "request-start-time": "1708508217412",
    "user-agent": "axios/1.6.7",
    "accept-encoding": "gzip, compress, deflate, br"
  },
[...]

PR #1923, opened 2024-03-25
Fix interpolation with Objects and Arrays

Its-treason:
# Description

Fixes: #1910 & #1635

Update how variable interpolation works, so the behavior with objects and arrays is more as expected.

- Interpolation of objects and arrays now works
- Try to use the default `toString`-Method if available
  - Except for the default Object and arrays

Test collection: https://cdn.discordapp.com/attachments/693228572286124085/1221951301404065873/json-testing_2.zip?ex=661471d5&is=6601fcd5&hm=277420a68a493567de437403f133622d186642f689e761d0438306b7579497ad&

Before:

image

After:

image

Pre-Build binary with the fix: https://github.com/Its-treason/bruno/releases/tag/nightly

### DIFF
--- a/packages/bruno-common/src/interpolate/index.ts
+++ b/packages/bruno-common/src/interpolate/index.ts
@@ -12,7 +12,23 @@
  */
 
 import { flattenObject } from '../utils';
-import cancelTokens from '@usebruno/app/src/utils/network/cancelTokens';
+
+function serializeObject(obj: Object) {
+  // Check if the object has a `toString` method like `Moment`
+  // Don't do it with arrays they serialize weirdly
+  if (typeof obj.toString === 'function' && !Array.isArray(obj)) {
+    try {
+      const result = obj.toString();
+      // The default object becomes '[object Object]' string
+      if (result !== '[object Object]') {
+        return result;
+      }
+    } catch {}
+  }
+
+  // Everything else will be json encoded
+  return JSON.stringify(obj);
+}
 
 const interpolate = (str: string, obj: Record<string, any>): string => {
   if (!str || typeof str !== 'string' || !obj || typeof obj !== 'object') {
@@ -30,13 +46,7 @@ const interpolate = (str: string, obj: Record<string, any>): string => {
 
     // Objects must be either JSON encoded or convert to a String via `toString`
     if (typeof replacement === 'object') {
-      // Check if the object has a `toString` method like `Moment`
-      if (typeof replacement.toString === 'function') {
-        try {
-          return replacement.toString();
-        } catch {}
-      }
-      return JSON.stringify(replacement);
+      return serializeObject(replacement);
     }
 
     return replacement;
--- a/packages/bruno-common/src/interpolate/index.ts
+++ b/packages/bruno-common/src/interpolate/index.ts
@@ -13,7 +13,7 @@
 
 import { flattenObject } from '../utils';
 
-const interpolate = (str: string, obj: Record<string, any>, isJsonBody = false): string => {
+const interpolate = (str: string, obj: Record<string, any>): string => {
   if (!str || typeof str !== 'string' || !obj || typeof obj !== 'object') {
     return str;
   }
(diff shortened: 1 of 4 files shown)


mjhcorporate:
Hmm... I'm a bit confused that for string interpolation, you have to put double-quotes into the string yourself (`"{{my_string}}"`), but for arrays and objects, you don't have to put `[ ]` and `{ }` around them.

Of course, fixing this by removing the need to put `"` around the string would break backwards-compatibility in a really bad way, and it would break otherwise needed functionality where you just want to plop a raw string into some body. (I sometimes need to send invalid JSON as part of my tests).

The more I think about this: Trying to solve serialization for the user is really tricky. Whatever we do, it should be as simple as possible, so that the user can have a good mental-model of what is going on.

Its-treason:
> Hmm... I'm a bit confused that for string interpolation, you have to put double-quotes into the string yourself (`"{{my_string}}"`), but for arrays and objects, you don't have to put `[ ]` and `{ }` around them.
> 
> Of course, fixing this by removing the need to put `"` around the string would break backwards-compatibility in a really bad way, and it would break otherwise needed functionality where you just want to plop a raw string into some body. (I sometimes need to send invalid JSON as part of my tests).
> 
> The more I think about this: Trying to solve serialization for the user is really tricky. Whatever we do, it should be as simple as possible, so that the user can have a good mental-model of what is going on.

Yeah, I also thought about this and the first implementation was like this (Nothing must be quoted). But then I saw dw-0 and other wrapping strings with double quotes.

packages/bruno-common/src/interpolate/index.ts
```diff
@@ -13,19 +13,44 @@
 
 import { flattenObject } from '../utils';
 
+function serializeObject(obj: Object) {
+  // Check if the object has a `toString` method like `Moment`
+  // Don't do it with arrays they serialize weirdly
+  if (typeof obj.toString === 'function' && !Array.isArray(obj)) {
```
mjhcorporate, line 19: I think this can lead to really misleading behavior. Consider the case where an object has a custom `toString` method: It would be called, but only, if the object is at the top level. If it is nested, then the default `JSON.stringify()` method is used, which leads to a different result.

Example: 
```js
// object on top-level
const today = moment();
console.log(today.toString());
// -> Wed Mar 27 2024 09:43:06 GMT+0100

// object nested:
const nestedExample = {"nestedExample": today}
console.log(JSON.stringify(nestedExample))
// -> {"nestedExample":"2024-03-27T08:43:06.249Z"}
```

See [https://jsfiddle.net/17Lzbjwm/1/](https://jsfiddle.net/17Lzbjwm/1/)

Proposal for simpler logic for `{{thing}}` replacement:
- if `thing` is a string, do a literal replacement
- if `thing` is anything else, use `JSON.stringify(thing)`
  - in case the user does not like the result, they have to do it themselves in prerequest or postrequest scripts.

Its-treason: I see, the whole point of calling `toString` was, because `JSON.stringify` would add double quotes around strings. I'll look into it later the day.

dw-0:
Any update here? I think the issue is still present?

PhilosoCoder:
It is still present, could You please fix the interpolation? Still doesn't work with arrays and objects...

packages/bruno-common/src/interpolate/index.ts
helloanoop: Agree @mjhcorporate
Its how we have done in #4519

@Its-treason Agree that some objects need to be stringified via `.toString()` instead of `JSON.stringify()`
As I mentioned in https://github.com/usebruno/bruno/pull/4498#issuecomment-2813840183, I think the path forward to do this would be to introduce piping support in interpolation `{{ now | toString }}`

helloanoop:
Thanks for the work on this @Its-treason 

Closing this PR as this has been implemented in #4519

Issue #1080, opened 2023-11-28
Use system CA certificates and/or allow specifying additional CA certificates


brki:
On a Mac, I try to access a resource that uses a custom CA certificate; it fails with:
```
Error invoking remote method 'send-http-request': Error: self signed certificate in certificate chain
```

I  already have that CA certificate in my keychain, so I can access corporate websites that use that CA certificate in a browser.

I am aware of the possibility to disable SSL verification, but don't like that idea. And neither do the security guys where I work.

I saw the issues https://github.com/usebruno/bruno/issues/739 and https://github.com/usebruno/bruno/issues/211 , which have a similar title and intention, but those people were satisfied when they discovered they could uncheck the SSL/TLS Certificate Validation checkbox.

Ideally, bruno would use the system CA certificates.

Insomnia has the same issue of not using the system certificates, but it allows specifying a CA certificate file to use. A solution like that would work for me too.

Is there by chance already some environment variable that I could use when launching bruno, so that it would look in a given file for certificates?  In https://github.com/usebruno/bruno/issues/739, `NODE_EXTRA_CA_CERTS` was mentioned (didn't work, apparently).

And thanks for bruno!

brki:
I see it's possible to specify a custom CA certificate now 😎 .

Thank you!

For ease of use, having the system CA certificates used would still be an improvement.

j-lebek:
First, let me express my satisfaction with this tool. Bruno is a valuable tool! Thanks for creating and maintaining!

I use custom CA certificates for some calls, too. I figured that I can also provide one file containing _multiple_ CA certificates, which is very helpful.

What I am missing though, is having the option to go with the default set of commonly accepted CA certificates and **add** a few more CA certificates for some specific systems.

At the moment, I get myself going by either ticking off `Use custom CA certificate` when connecting to systems that use official certificates – and re-enable this option (which fortunately keeps the file reference ❤️ ) when connecting to systems using certificates issued by private CAs.

I would appreciate, if there was a choice of either extending or replacing the list of trusted CAs.
Thank you!

slowjoe007:
Addressed in #1863

PR #1937, opened 2024-03-26
feat: Allow default truststore extension on bru CLI (#1080)

slowjoe007:
# Description

The change in feature #1863 did not cover the CLI feature of Bruno, `bru`. This feature adds that part and introduces a new command line option to `bru` CLI to control the way custom CA certificates are handled: replace _or_ extend the default truststore.

The new CLI option is called `--extend-truststore` and documented in both synopsis and readme.
This new option only is in effect, if a custom CA certificate is specified via option `--cacert`.

Screenshot 2024-03-26 212154

Screenshot 2024-03-26 212922

### Contribution Checklist:

   #1080

### DIFF
--- a/packages/bruno-app/src/components/CollectionSettings/Auth/OAuth2/PasswordCredentials/index.js
+++ b/packages/bruno-app/src/components/CollectionSettings/Auth/OAuth2/PasswordCredentials/index.js
@@ -20,7 +20,7 @@ const OAuth2AuthorizationCode = ({ item, collection }) => {
 
   const handleSave = () => dispatch(saveCollectionRoot(collection.uid));
 
-  const { accessTokenUrl, username, password, scope } = oAuth;
+  const { accessTokenUrl, username, password, clientId, clientSecret, scope } = oAuth;
 
   const handleChange = (key, value) => {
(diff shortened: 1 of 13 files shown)


mjhcorporate:
Thank you for adding this to the CLI!

A question to the general audience: Would it maybe be better if the default behavior would be to extend the truststore, and have a command-line-switch to limit the certificates to the one that was provided? In essence, replace `--extend-truststore` with `--ignore-truststore`?

That way, the "standard" use case ("make https requests and not have the user think about it") works more robustly, but we still retain full functionality.

It's just a thought -- personally, I'm fine either way.

slowjoe007:
Totally agree, @mjhcorporate!  

I'd love to see this custom CA feature work the way you describe, Martin. When it was introduced, it replaced the default truststore entirely. This does not feel natural, but it might be required in certain cases.  

When I added the option to keep the default CA certificates in place, I decided to not invert the logic that existed before.  

What's your take on this, @helloanoop?

slowjoe007:
Perfect. If time permits, I'll take care of this (both CLI and GUI) in the next few days.

PR #1947, opened 2024-03-27
fix: Collection variable viewer

Its-treason:
Due to changes in https://github.com/usebruno/bruno/pull/650 collection variables would be passed as an object but were expected to be an array. Collection variables are now converted to an array.

Fixes: #1940

---

Also, we should later remove the `Show secret variable values` button for Collection.

image

PR #2022, opened 2024-04-05
fix: Collection variable viewer

npu78954:
This is a bugfix for usebruno#1940 

Due to changes in usebruno#650 collection variables would be passed as a object but was expected to be an array. Collection variables are now converted to an array.


npu78954:
duplicated with PR https://github.com/usebruno/bruno/pull/1947

Issue #1003, opened 2023-11-20
Auth: OAuth 2.0


helloanoop:
Parent Issue: #119

Support OAuth 2.0

fuxx:
Really important feature <3

and-rose:
So far this is the only thing keeping me attached to Postman. Really keen to see this implemented.

Rens660:
We would love to move to Bruno as well, but since we are requiring `grant_type=authorization_code` we can't yet. Is there maybe a progress update on this feature? 🙌

tsteckenborn:
Dependend on what you need you might want to check [this](https://github.com/usebruno/bruno/discussions/385#discussioncomment-7813010) as a workaround for e.g. the Client Credentials flow.

@Rens660 - I've got quite a similar one for Authorization Code. You'll need to do one manual step in the browser, but atleast it returns e.g. the corresponding link as part of an error message. Would you be interested in that?

Rens660:
Hi @tsteckenborn, 
Unfortunately I am stuck with using Authorization Code, but if you would share your workaround for if, I am definitely interested, thanks!

helloanoop:
I will be building OAuth 2.0 functionality in this upcoming live coding stream with Hussain on Dec 16th
https://www.youtube.com/watch?v=ZtNGbglTjps

tsteckenborn:
> Hi @tsteckenborn, Unfortunately I am stuck with using Authorization Code, but if you would share your workaround for if, I am definitely interested, thanks!

Take a look if [that](https://github.com/usebruno/bruno/discussions/385#discussioncomment-7838309) helps to ease the usage

mjschlosser8:
Can't wait for this!

andifalk:
> Pretty much the only thing tying me to insomnia. We use the implicit flow on our end.

FYI: Implicit flow is deprecated, should not be used any more (replaced by Authorization code + PKCE)

42shadow42:
> > Pretty much the only thing tying me to insomnia. We use the implicit flow on our end.
> 
> FYI: Implicit flow is deprecated, should not be used any more (replaced by Authorization code + PKCE)

Thanks for the heads up, though I don't see Authorization code + PKCE supported here yet either. I'll have to check to see if we have migration plans from Implicit flow.

billbrinck:
Is there any progress on the OAuth2 support?

fuxx:
> Is there any progress on the OAuth2 support?

[...]

PR #2051, opened 2024-04-10
 fix(#1003): content type issue for client credentials & password credentials grant types -- missing client id & secret for password grant type 

lohxt1:


helloanoop:
Merged! Thanks @lohxt1 !

Issue #1999, opened 2024-04-03
OAUTH2 auth is successful but token endpoint is returned instead of api endpoint


wmlele:
Hello,

trying to access an OAUTH2/OIDC protected endpoint, but the Response contains the output on the OAUTH Token request, instead of the protected API endpoint (which is never called despite a successful oauth):

![Screenshot from 2024-04-03 17-51-33-1](https://github.com/usebruno/bruno/assets/1617799/aca04079-4955-43cc-92ff-8cd2797194f2)

The `{{oidc_issuer}}/me` (userinfo) endpoint is never called, the output is from the successfull call to `/token`

This is what insomnia returns for the exact same configuration

![Screenshot from 2024-04-03 17-51-44-1](https://github.com/usebruno/bruno/assets/1617799/05263744-9355-47b0-a56c-dd02df552b04)

Not sure if it's a mistake on my side or something different from my expectations.

Thanks

guilhermeferrari:
+1

diogenesc:
Facing the same issue here

wmlele:
Just a follow up on this. Of course you *can* use bruno for an oauth/oidc protected api endpoint. 
Just define an _empty_ request with oauth2 authentication,  and store the `access_token` as a variable, then just do a bearer token auth on the actual APIs using the stored `access_token`.
Not sure though this is the way it was intended?

diogenesc:
> Just a follow up on this. Of course you _can_ use bruno for an oauth/oidc protected api endpoint. Just define an _empty_ request with oauth2 authentication, and store the `access_token` as a variable, then just do a bearer token auth on the actual APIs using the stored `access_token`. Not sure though this is the way it was intended?

Probably not, seems wrong if we think how others API clients works.

ryanholden8:
Any update on this? Seems this is a confirmed bug based on all the referencing from #2148?

diogenesc:
+1

nmartinon:
+1 My team also just can't use bruno because of this

fekitibi:
+1

deinspanjer:
I'm really excited with the potential of Bruno here but I'm running into the same issue.
If I try to set up Oauth2 as the auth directly in a request, the actual request doesn't go through, only the token issuing request.

If I try to set up Oauth2 at the collection level, and set a request to inherit, I see the note:
Note: You need to use scripting to set the access token in the request headers.

But I'm not sure how exactly to do that.
[...]

PR #2058, opened 2024-04-10
feat: OAuth2: Implement OAuth 2.0 Implicit Grant

### LINKED ISSUE #1901: Authorization Code not consumed
I have a problem with setting up an authentication via authorization code (grant_type).
I have provided parameters:
1. a callback URL, 
2. Authorization URL
3. Client ID
4. Scope read write
When I click Get Access Token a small window will pop up and I can log in into app.
But then nothing happens, this window in Postman is being closed and the token is stored.
Here in Bruno something is different. 
How should I use it?

image

### LINKED ISSUE #2056: OAuth2: Implement OAuth 2.0 Implicit Grant

### Describe the feature you want to add

Bruno being primarily a testing tool should support OAuth2 implicit flow despite it being obsolete - because the fact that no **new** apps **should** use it does not mean no **legacy** apps we are working with do.

The feature has been requested previously here https://github.com/usebruno/bruno/issues/1003#issuecomment-1863309018, https://github.com/usebruno/bruno/issues/119#issuecomment-1805679198 and https://github.com/usebruno/bruno/issues/1901 .

[implicit.webm](https://github.com/usebruno/bruno/assets/2108093/68ac75e2-c811-46e9-b432-63a829192bb4)
Screenshot from 2024-04-10 22-38-35

pietrygamat:
Implements[ OAuth2 Implicit Grant authorization](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2)

Screenshot from 2024-04-10 22-38-35

# Description

### Contribution Checklist:

resolves #2056
resolves #1901

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -12,7 +12,6 @@ const { ipcMain } = require('electron');
 const { isUndefined, isNull, each, get, compact, cloneDeep, forOwn, extend } = require('lodash');
 const { VarsRuntime, AssertRuntime, ScriptRuntime, TestRuntime } = require('@usebruno/js');
 const prepareRequest = require('./prepare-request');
-const prepareCollectionRequest = require('./prepare-collection-request');
 const prepareGqlIntrospectionRequest = require('./prepare-gql-introspection-request');
 const { cancelTokens, saveCancelToken, deleteCancelToken } = require('../../utils/cancel-token');
 const { uuid } = require('../../utils/common');
@@ -31,9 +30,9 @@ const { shouldUseProxy, PatchedHttpsProxyAgent } = require('../../utils/proxy-ut
 const { chooseFileToSave, writeBinaryFile, writeFile } = require('../../utils/filesystem');
 const { getCookieStringForUrl, addCookieToJar, getDomainsWithCookies } = require('../../utils/cookies');
 const {
-  resolveOAuth2AuthorizationCodeAccessToken,
-  transformClientCredentialsRequest,
-  transformPasswordCredentialsRequest
+  oauth2AuthorizeWithAuthorizationCode,
+  oauth2AuthorizeWithClientCredentials,
+  oauth2AuthorizeWithPasswordCredentials
 } = require('./oauth2-helper');
 const Oauth2Store = require('../../store/oauth2');
 const iconv = require('iconv-lite');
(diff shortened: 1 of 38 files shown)


pietrygamat:
FYI, the mockup is from
```
meta {
  name: implicit
  type: http
  seq: 6
}

get {
  url: http://localhost:8080/realms/bruno/protocol/openid-connect/userinfo
  body: none
  auth: oauth2
}

auth:oauth2 {
  grant_type: implicit
  callback_url: bruno://auth
  authorization_url: http://localhost:8080/realms/bruno/protocol/openid-connect/auth?login_hint=bruno-user
  client_id: brunoclient
  scope: openid
}
``` 
against locally started Keycloak.
For reproducable environment, place the following realm-export.json file in an empty directory and start Keycloak in Docker: 
```
docker run -p 8080:8080 \
-e KEYCLOAK_ADMIN=admin \
-e KEYCLOAK_ADMIN_PASSWORD=admin \
-v $(pwd):/opt/keycloak/data/import:z quay.io/keycloak/keycloak:24.0.1 \
start-dev --import-realm
```
[realm-export.json](https://github.com/usebruno/bruno/files/14937930/realm-export.json)

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -232,6 +233,11 @@ const configureRequest = async (
         request.data = passwordData;
         request.url = passwordAccessTokenUrl;
         break;
+      case 'implicit':
+        interpolateVars(requestCopy, envVars, collectionVariables, processEnvVars);
+        const { accessToken } = await getOAuth2ImplicitToken(requestCopy, collectionUid);
+        request.headers['Authorization'] = `Bearer ${accessToken}`;
+        break;
```
pietrygamat: @lohxt1 , please help me understand this part - the `request` here is what user configured in the client, and we only want to add the authorization to it. Why do all other grant types modify the request's method, or data, and not simply add ```request.headers['Authorization'] = `Bearer ${accessToken}`;```?

Am I completly missing the point of what's going on, or is this a bug in the implementation of all other grant types?

lohxt1: @pietrygamat

By default, the request.method value is whatever that's set in the query URL bar. It could be GET, POST, DELETE, or any other method. This needs to be overridden to POST.

The client_id and client_secret can be sent either via headers or the request body for all grant types.
In the current implementation, they are explicitly sent through the request body along with the rest of the input values from the Auth tab. We do have plans to add a dropdown select so that users can choose between headers and the request body.

Hope that answers your question.

pietrygamat: @lohxt1 
I would expect that requests to auth url and token url are separate from main collection request. Current behaviour is to overwrite information about original user's request and leads to bugs such as this: https://github.com/usebruno/bruno/issues/2060 or https://github.com/usebruno/bruno/issues/1999.

I thought maybe somewhere in this file the request is copied, repurposed as authorization request,  executed, logged in Timeline tab and used to extract access_token which is then added to original request exactly as specified by the user, but it is not happening - the code here leaves the request modified and original one is lost. 

Now, it cannot be the desired outcome, can it?

lohxt1: @pietrygamat 

I understand your point. But in order to reuse the generated access_token, you would need to automatically store it somewhere in the app as a variable and reuse it in the original request, correct? However, the intention behind the current implementation was to let the user decide how they want to store and use the access_token through scripting and not blackbox it.

If you want the user's original request (which includes the access token) specified in the top query URL bar to execute automatically right after the OAuth2 authorization process is completed in the background, you can achieve this with some modifications to the existing implementation.

However, you would lose the ability to run post-response scripts specific to the OAuth2 flow, such as saving the access_token in your desired environment variable or collection variable. This is because the post-response script runs on the latest response body and not on the response bodies of any intermediate requests; like the OAuth2 authorization requests.

pietrygamat: It would make sense to create a predefined and documented variable for token data and allow to reference it from pre/post/collection level scripts. It would for example store:
`access_token`, `refresh_token`, `token_type`, `expires_in` and `scope` and live in collection level. It would offer the same scripting flexibility, but also allow for more intuitive UI (resolving referenced bugs) and serve as a starting point for implementing any logic for automatic token management (e.g. automaic attempts to use refresh token when access_token is expired or skip token request if one is already stored). 
With this approach Implicit Flow would not be the odd one out just because it cannot even return access_token in a response body ;).  
Not sure however how to best make it work.

pietrygamat:
While functional on its own, this may be superseded by #2077 where more complete overhaul of oauth2 is proposed, resulting in a more consistent user experience between different oauth2 grants.

s-kocher:
Thanks @pietrygamat to help on the implicit flow ! @helloanoop is there some blocking point preventing the PR to be merged and be soon available in future release ?

helloanoop:
@pietrygamat 

We have merged the latest OAuth 2.0 UX revamp to the main branch.
I see that this PR has too many conflicts (We did a lot of refactor during the revamp including deleting files)

I'd suggest cutting a new branch and copying the changes.

helloanoop:
@pietrygamat

Let me know if you'd have bandwidth to update as per my comment https://github.com/usebruno/bruno/pull/2058#issuecomment-2747592317

If not, we can have someone from Bruno Team pick this up.

helloanoop:
Closing this. New PR for Implicit Grant is being tracked at #4307

@pietrygamat Please review if you get some time this week.

Issue #1907, opened 2024-03-23
Accessibility issues on the Welcome page


shrilakshmishastry:
List of accessibility issues on the welcome page:

1. bruno icon shouldn’t be accessible as it’s decorative
2. The wrong HTML is used for the button, line numbers 62 ,67, 71 in the Welcome component. All of these need to be buttons or link
3. in all links SVG should be aria-hidden

List of accessibility issues on sidebar footer icons

1. wrong HTML used
2. wrong roles used

List of accessibility issues in Modal
1. When the modal is open windows under a modal dialog are accessible by the keyboard

bpoulaindev:
I initially had a PR aiming to replace the icons, use buttons, improve accessibility with keyboard navigation... : #1359 this is not really planned for the moment, too many core functionalities to work on. But feel free to send a PR to fix what you mentioned

shrilakshmishastry:
> I initially had a PR aiming to replace the icons, use buttons, improve accessibility with keyboard navigation... : #1359 this is not really planned for the moment, too many core functionalities to work on. But feel free to send a PR to fix what you mentioned

fixed here https://github.com/usebruno/bruno/pull/1906

PR #2076, opened 2024-04-14
fix: accessibility issue modal

shrilakshmishastry:
# Description

### Accessibility issue in Modal 

**Ally pattern of modal:**
Modal is a window overlaid on either the primary window or another dialog window. Windows under a modal dialog is inert. `That is, users cannot interact with content outside an active dialog window.` Inert content outside an active dialog is typically visually obscured or dimmed so it is difficult to discern, and in some implementations, attempts to interact with the inert content cause the dialog to close. 

[read more](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)

**Issue:**
When the modal is open windows under a modal dialog are accessible by the keyboard.

[Link to issue description](https://github.com/usebruno/bruno/issues/1907)

**Fix:**
 Use `dialog` tag and call `showModal()` method on the mount of Modal component 
 
 **Screen recording**
 
 | then | now |
 |--|--|
|https://github.com/usebruno/bruno/assets/29778698/20ac9709-7967-4ccb-a580-578fd9ece1fc |https://github.com/usebruno/bruno/assets/29778698/00c0b4e0-c5f9-4889-8c4c-9f3001ed946d |

### Contribution Checklist:

- [] **The pull request does not introduce any breaking changes**

### DIFF
--- a/packages/bruno-app/src/components/Modal/StyledWrapper.js
+++ b/packages/bruno-app/src/components/Modal/StyledWrapper.js
@@ -3,6 +3,9 @@ import styled from 'styled-components';
 const Wrapper = styled.dialog`
   color: ${(props) => props.theme.text};
 
+  max-width: unset;
+  max-height: unset;
+
   &.modal--animate-out {
     animation: fade-out 0.5s forwards cubic-bezier(0.19, 1, 0.22, 1);
 
--- a/packages/bruno-app/src/components/Modal/StyledWrapper.js
+++ b/packages/bruno-app/src/components/Modal/StyledWrapper.js
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 
-const Wrapper = styled.div`
+const Wrapper = styled.dialog`
   color: ${(props) => props.theme.text};
 
   &.modal--animate-out {
--- a/packages/bruno-app/src/components/Modal/index.js
+++ b/packages/bruno-app/src/components/Modal/index.js
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import StyledWrapper from './StyledWrapper';
+import { useRef } from 'react';
 
 const ModalHeader = ({ title, handleCancel, customHeader }) => (
   <div className="bruno-modal-header">
@@ -69,6 +70,7 @@ const Modal = ({
   closeModalFadeTimeout = 500
 }) => {
   const [isClosing, setIsClosing] = useState(false);
+  const modalRef = useRef(null);
   const escFunction = (event) => {
     const escKeyCode = 27;
     if (event.keyCode === escKeyCode) {
@@ -99,11 +101,11 @@ const Modal = ({
   }
 
   useEffect(() => {
-    document.getElementsByTagName('dialog')[0].showModal();
+    modalRef?.current?.showModal();
   }, []);
 
   return (
-    <StyledWrapper className={classes} onClick={onClick ? (e) => onClick(e) : null}>
+    <StyledWrapper ref={modalRef} className={classes} onClick={onClick ? (e) => onClick(e) : null}>
       <div className={`bruno-modal-card modal-${size}`}>
         <ModalHeader title={title} handleCancel={() => closeModal({ type: 'icon' })} customHeader={customHeader} />
         <ModalContent>{children}</ModalContent>
--- a/packages/bruno-app/src/components/Modal/index.js
+++ b/packages/bruno-app/src/components/Modal/index.js
@@ -97,6 +97,11 @@ const Modal = ({
   if (hideFooter) {
     classes += ' modal-footer-none';
   }
+
+  useEffect(() => {
+    document.getElementsByTagName('dialog')[0].showModal();
+  }, []);
+
   return (
     <StyledWrapper className={classes} onClick={onClick ? (e) => onClick(e) : null}>
       <div className={`bruno-modal-card modal-${size}`}>
--- a/packages/bruno-app/src/components/Modal/index.js
+++ b/packages/bruno-app/src/components/Modal/index.js
@@ -1,6 +1,5 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import StyledWrapper from './StyledWrapper';
-import { useRef } from 'react';
 
 const ModalHeader = ({ title, handleCancel, customHeader }) => (
   <div className="bruno-modal-header">
@@ -81,6 +80,7 @@ const Modal = ({
   const closeModal = (args) => {
     setIsClosing(true);
     setTimeout(() => handleCancel(args), closeModalFadeTimeout);
+    modalRef.current.close();
   };
 
   useEffect(() => {
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,8 @@
         "packages/bruno-graphql-docs"
       ],
       "dependencies": {
-        "json-bigint": "^1.0.0"
+        "json-bigint": "^1.0.0",
+        "lossless-json": "^4.0.1"
       },
       "devDependencies": {
         "@faker-js/faker": "^7.6.0",
@@ -36,7 +37,6 @@
     },
     "node_modules/@alloc/quick-lru": {
       "version": "5.2.0",
-      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=10"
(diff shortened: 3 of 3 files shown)


PushpenderSaini0:
Can you remove the `package-lock.json` changes from this PR !

packages/bruno-app/src/components/Modal/index.js
```diff
@@ -97,6 +97,11 @@ const Modal = ({
   if (hideFooter) {
     classes += ' modal-footer-none';
   }
+
+  useEffect(() => {
+    document.getElementsByTagName('dialog')[0].showModal();
```
Its-treason: I think `useRef` should be used here instead of accessing the DOM directly.

Its-treason (changes requested):

packages/bruno-app/src/components/Modal/index.js
PushpenderSaini0: Avoid `document.getElementsByTagName` it's generally not a good practice to access dom directly in react.

shrilakshmishastry: ok

shrilakshmishastry:
> Can you remove the `package-lock.json` changes from this PR !

I tried [here](https://github.com/usebruno/bruno/pull/2076/commits/defbd9bf0c0e58825a0da2fd49ab52a19810ce47) but it didn't get updated 😬

packages/bruno-app/src/components/Modal/index.js
shrilakshmishastry: done here https://github.com/usebruno/bruno/pull/2076/commits/d467d837efb0a4390ee35a12d85da8690eb9a70d

shrilakshmishastry: https://github.com/usebruno/bruno/pull/2076/commits/d467d837efb0a4390ee35a12d85da8690eb9a70d

PushpenderSaini0:
> > Can you remove the `package-lock.json` changes from this PR !
> 
> I tried [here](https://github.com/usebruno/bruno/pull/2076/commits/defbd9bf0c0e58825a0da2fd49ab52a19810ce47) but it didn't get updated 😬

The file is changed in 2 commits this one will be a bit tricky !

1 ) Checkout the file from the main branch
2) Commit the file (this version is from main branch)
3) (Optional) Squash the changes to make the history clean !

```bash 
git checkout main -- package-lock.json
git commit -a -m "revert package-lock.json"
git rebase -i HEAD~4
```

PushpenderSaini0 (approved):

ad1992 (commented):
Hi @shrilakshmishastry as far as understand the issue (from description and screen recordings) is that focus is going out of the active modal/dialog when navigating through the keyboard and hence you are trying to make sure that the focus stays in active dialog. Please let me know if my understanding is correct.

However the solution of triggering `showModal` when mounted isn't clear to me and how is it fixing this issue ?
We should instead have a keydown handler for modal component which makes sure that on keyboard navigation the focus always keeps looping (in order) inside the active dialog unless esc is pressed ?
A small suggestion regarding the screen recording - you can directly upload the `.mov` files so reviewers don't have to download the files to view the screen-recordings :)

For reference - I came across this PR in https://twitter.com/shrilakshmihg/status/1779420061721035153

shrilakshmishastry:
> Hi @shrilakshmishastry as far as understand the issue (from description and screen recordings) is that focus is going out of the active modal/dialog when navigating through the keyboard and hence you are trying to make sure that the focus stays in active dialog. Please let me know if my understanding is correct.
> 
> However the solution of triggering `showModal` when mounted isn't clear to me and how is it fixing this issue ? We should instead have a keydown handler for modal component which makes sure that on keyboard navigation the focus always keeps looping (in order) inside the active dialog unless esc is pressed ? A small suggestion regarding the screen recording - you can directly upload the `.mov` files so reviewers don't have to download the files to view the screen-recordings :)
> 
> For reference - I came across this PR in https://twitter.com/shrilakshmihg/status/1779420061721035153

Hey @ad1992 thanks for the suggestion!!

yes, the focus is going out of the modal, hence the fix.

calling `showModal` in useEffect is to set `dialog` open. It's a way around not using JS to handle focus the one you mentioned `throw key-down event`.


The intention of using the `dialog` tag is its ability to handle focus traps in an HTML declarative way not by JS. 

you can read more about it here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

The doc says you need to call `showModal` method in the caller component. Since we are handling showing Modal via js (mounting it to DOM) on button click there is no straight way we can access `dialog`. Hence calling `showModal` in useEffect so `dialog` provides the expected behavior when it's in `open` state and act as `Modal`.

packages/bruno-app/src/components/Modal/index.js
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react';
 import StyledWrapper from './StyledWrapper';
+import { useRef } from 'react';
```
ad1992:
```suggestion
import React, { useEffect, useState, useRef } from 'react';
import StyledWrapper from './StyledWrapper';
```
Since there are react imports already so better to club with it

ad1992 (approved):
suggested a minor change otherwise looks good @shrilakshmishastry

I am not sure if there is a way to test the changes directly in the browser so I haven't tested it however the code looks good!

Additionally, I think you probably might be targeting the rest of the clean-up in separate PR (eg closing the modal via dialog close and removing ESC handler etc)

shrilakshmishastry:
> suggested a minor change otherwise looks good @shrilakshmishastry
> 
> I am not sure if there is a way to test the changes directly in the browser so I haven't tested it however the code looks good!
> 
> Additionally, I think you probably might be targeting the rest of the clean-up in separate PR (eg closing the modal via dialog close and removing ESC handler etc)

closing modal is taken care in `closeModal()` method and esc handling is already present in the existing code

ad1992:
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

@shrilakshmishastry as per the docs, I think you can remove the `ESC` handling as it shouldn't be needed anymore and taken care of by the dialog itself.

shrilakshmishastry:
> > https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
> 
> @shrilakshmishastry as per the docs, I think you can remove the `ESC` handling as it shouldn't be needed anymore and taken care of by the dialog itself.

We need it, as there are cases where we needn't close the modal on the click of `esc` key. There is one prop the modal accepts and based on that action on the `esc` key-down is taken care 

https://github.com/usebruno/bruno/blob/d490b1b5fc2cea0d8c0aa60d4ab07210f4d647cc/packages/bruno-app/src/components/Modal/index.js#L67C2-L67C20

packages/bruno-app/src/components/Modal/StyledWrapper.js
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 
-const Wrapper = styled.div`
+const Wrapper = styled.dialog`
   color: ${(props) => props.theme.text};
```
Its-treason, line 4: Could you also add:

```css
  max-height: unset;
  max-width: unset;
```
To remove the weird gap around the backdrop?

shrilakshmishastry: could you please elaborate more

Its-treason: Dialogs now have a "weird" gap between the window border and the backdrop:
image

This is because the `dialog` Element has some weird default CSS. From the [chrome source](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/core/html/resources/html.css#1444):

```css
dialog:-internal-dialog-in-top-layer {
    /* https://html.spec.whatwg.org/multipage/rendering.html#flow-content-3 */
    position: fixed;
    overflow: auto;
    inset-block-start: 0;
    inset-block-end: 0;
    max-width: calc(100% - 6px - 2em);
    max-height: calc(100% - 6px - 2em);
    /* https://github.com/w3c/csswg-drafts/issues/6939#issuecomment-1016679588 */
    user-select: text;
    visibility: visible;
}
```

Unsetting the default `max-width` and `max-height` will remove the gap.

```suggestion
  color: ${(props) => props.theme.text};
  
  max-height: unset;
  max-width: unset;
```

shrilakshmishastry: thanks !!

fixed it now

shrilakshmishastry: @Its-treason can you check now

Its-treason (approved):

Its-treason (changes requested):
image

I found that the "toasts" are now in the background, behind the dialog/backdrop instead of the foreground. This is probably due to how the `dialog` element works: https://stackoverflow.com/a/41816764

shrilakshmishastry:
> image
> 
> I found that the "toasts" are now in the background, behind the dialog/backdrop instead of the foreground. This is probably due to how the `dialog` element works: https://stackoverflow.com/a/41816764

Seems intersting, I will look into this. Thanks

shrilakshmishastry:
> > image
> > I found that the "toasts" are now in the background, behind the dialog/backdrop instead of the foreground. This is probably due to how the `dialog` element works: https://stackoverflow.com/a/41816764
> 
> Seems intersting, I will look into this. Thanks

The toast needs to be attached to or decendent of `top layer` element when triggered from them. I have rised the issue in react-hot-toast. Not sure whether it should be fixed from them or here. I will look into this furthur 
https://github.com/timolins/react-hot-toast/issues/355

helloanoop:
@anusreesubash Please review this?

PR #2131, opened 2024-04-18
fix: error boundary adding return-to-app and force-quit options

bijinab:
# Description

This PR fixes the annoying blank screen in the event of an uncaught error which left the user with no option other than force quit. With this fix, options to return to the App and Force Quit is added.


helloanoop:
Merged. Thanks @bijinab !

Issue #2124, opened 2024-04-18
Issue with Parameters Passed in the URL


Chibane690:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

I added a parameter "q" directly into the URL bar to be executed without going through the query array as illustrated in the image below:

![Capture](https://github.com/usebruno/bruno/assets/80396129/4862498c-248b-4038-a89c-af5fafd02966)

When I added another parameter from the query array, in my case it was the "options" attribute with the value "concise", it modified the value of the "q" attribute that I associated at the beginning, which caused a bug as illustrated in the image below:

![Capture2](https://github.com/usebruno/bruno/assets/80396129/27349e66-866a-4ed4-8494-c2a6c8c5cfba)

### Screenshots/Live demo link

![Capture](https://github.com/usebruno/bruno/assets/80396129/2f904355-063e-47e4-a314-3e8a053f5e1c)
![Capture2](https://github.com/usebruno/bruno/assets/80396129/8513fbdc-ab83-461e-95a0-05490227de22)

Gruummy:
Are you sure that your uri .... parameters are formed correctly ?

[URIs include components and subcomponents that are delimited by characters in the "reserved" set. These characters are called "reserved" because they may (or may not) be defined as delimiters by the generic syntax, by each scheme-specific syntax, or by the implementation-specific syntax of a URI's dereferencing algorithm. If data for a URI component would conflict with a reserved character's purpose as a delimiter [emphasis added], then the conflicting data must be percent-encoded before the URI is formed.](https://www.rfc-editor.org/rfc/rfc3986#section-2.2)

I think you have a problem with the "==" in your string. The "sign" is a reserved sub-delim character
... and you need to do the percent-encoding / url-encoding

https://www.rfc-editor.org/rfc/rfc3986#section-2.1

Probably Bruno is not handling / showing the data for this reason not correctly in your first screenshot

pietrygamat:
I think this is a serious bug and important feature to fix, but it comes with a bunch of design decisions to make. The URL field should probably validate it's contents and notify user if the value is not correctly formatted URL - and what @Chibane690 inputs is in fact invalid. 

[...]

PR #2139, opened 2024-04-19
fix: Issue with Parameters Passed in the URL(#2124)

pietrygamat:
This PR fixes query parameter parsing from URL field. The '=' should be allowed within query parameter value. While first equals sign separates parameter name from its value, any subsequent occurrences are valid and should not be discarded.

Additionally in curent implementation the last query parameter value incorrectly contains what should be URI fragment (anything that follows #), which this PR also addresses.

fixes: #2124 
fixes: #2731

Before:
Screenshot from 2024-04-19 19-33-27

After:
Screenshot from 2024-04-19 19-32-30

### DIFF
--- a/packages/bruno-app/src/utils/url/index.js
+++ b/packages/bruno-app/src/utils/url/index.js
@@ -18,12 +18,17 @@ const hasLength = (str) => {
 };
 
 export const parseQueryParams = (query) => {
-  if (!query || !query.length) {
+  try {
+    if (!query || !query.length) {
+      return [];
+    }
+
+    return Array.from(new URLSearchParams(query.split('#')[0]).entries())
+      .map(([name, value]) => ({ name, value }));
+  } catch (error) {
+    console.error('Error parsing query params:', error);
     return [];
   }
-
-  return Array.from(new URLSearchParams(query.split('#')[0]).entries())
-    .map(([name, value]) => ({ name, value }));
 };
 
 export const parsePathParams = (url) => {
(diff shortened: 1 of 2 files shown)


packages/bruno-app/src/utils/url/index.js
```diff
@@ -18,10 +18,14 @@ export const parseQueryParams = (query) => {
     return [];
   }
 
-  let params = query.split('&').map((param) => {
-    let [name, value = ''] = param.split('=');
-    return { name, value };
-  });
+  let params = query
```
MasterLambaster: would it make sense to use `URLSearchParams` in this case that'd follow URL Standard?

pietrygamat: The suggestion makes sense, I'll take a look. To consider here is that the URL field and query-params fields are updated based on each-other's value. That means during typing character by character, we may find ourselves in a state where the url is NOT correctly formatted (e.g. we only typed in half of the url envoded value so far), so the resulting string will not validate as URLSearchParams, and crash the view, unless handled with a fallback behavior which would be what is already in this PR anyway,

pietrygamat: Also, based on discussion in #2191, there is a design decision to make, perhaps the proper fix is even more involved (like adding extra assumptions, validations and/or UI warnings).

sanjai0py (approved):

sanjai0py (approved):

PR #2148, opened 2024-04-21
fix: Check OAuth2 redirect URL for matching callback URL and authorization code in query parameters

### LINKED ISSUE #2147: BUG: OAuth2 Authorization Code Flow fails for intermediate redirects with an auth code

### Describe the bug

On similar lines as #1778, the authorization flow fails when there is a multi-step authorization, and one of the redirect URLs also contains `code` as a query parameter.

My example scenario where this happened:

Auth server: Keycloak using OIDC, backed by Google accounts as an ID provider
On successful auth, Keycloak sends an intermediate redirect to the following URL (Note the `code=` at the end):
```
https://<keycloak URL>/realms/<realm>/broker/google/endpoint?state=<state>&code=<code>&...
```

Since this URL has `code` as a query param but the URL is not the expected callback URL, Bruno throws the following error:

```
Error occurred in handler for 'send-collection-oauth2-request': Error: Invalid Callback Url: {url}
    at onWindowRedirect (/Users/dakshin/projects/personal/bruno/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js:37:18)
    at EventEmitter.willRedirectListener (/Users/dakshin/projects/personal/bruno/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js:78:7)
    at EventEmitter.emit (node:events:527:28)
```

_No response_

image

dakshin-k:
Fixes #2147

# Description

In an Authorization code flow, there may be multiple intermediate redirects before reaching the final one which matches the callback URL and has a code in the query params.

We should wait until we see a redirect URI that matches both the conditions. This fixes the issue where, when a redirect contains `code` as a query param but is not the final one (i.e., is not to the callback URL) an error is thrown saying the callback URL is invalid.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
+++ b/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
@@ -1,6 +1,10 @@
 const { BrowserWindow } = require('electron');
 const { preferencesUtil } = require('../../store/preferences');
 
+const matchesCallbackUrl = (url, callbackUrl) => {
+  return url ? new URL(url).host == new URL(callbackUrl).host : false;
+};
+
 const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
   return new Promise(async (resolve, reject) => {
     let finalUrl = null;
@@ -31,7 +35,7 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
 
     function onWindowRedirect(url) {
       // check if the redirect is to the callback URL and if it contains an authorization code
-      if (url && url.includes(callbackUrl)) {
+      if (matchesCallbackUrl(url, callbackUrl)) {
         if (!new URL(url).searchParams.has('code')) {
           reject(new Error('Invalid Callback URL: Does not contain an authorization code'));
         }
@@ -93,4 +97,4 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
   });
 };
 
-module.exports = { authorizeUserInWindow };
+module.exports = { authorizeUserInWindow, matchesCallbackUrl };
--- a/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
+++ b/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
@@ -30,12 +30,12 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
     });
 
     function onWindowRedirect(url) {
-      // check if the url contains an authorization code
-      if (new URL(url).searchParams.has('code')) {
-        finalUrl = url;
-        if (!url || !finalUrl.includes(callbackUrl)) {
-          reject(new Error('Invalid Callback Url'));
+      // check if the redirect is to the callback URL and if it contains an authorization code
+      if (url && url.includes(callbackUrl)) {
+        if (!new URL(url).searchParams.has('code')) {
+          reject(new Error('Invalid Callback URL: Does not contain an authorization code'));
         }
+        finalUrl = url;
         window.close();
       }
       if (url.match(/(error=).*/) || url.match(/(error_description=).*/) || url.match(/(error_uri=).*/)) {
--- a/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
+++ b/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
@@ -2,7 +2,7 @@ const { BrowserWindow } = require('electron');
 const { preferencesUtil } = require('../../store/preferences');
 
 const matchesCallbackUrl = (url, callbackUrl) => {
-  return url ? new URL(url).host == new URL(callbackUrl).host : false;
+  return url ? url.href.startsWith(callbackUrl.href) : false;
 };
 
 const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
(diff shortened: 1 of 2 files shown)


helloanoop:
@lohit-1 Can you review this?

packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
```diff
@@ -30,12 +30,12 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
     });
 
     function onWindowRedirect(url) {
-      // check if the url contains an authorization code
-      if (new URL(url).searchParams.has('code')) {
-        finalUrl = url;
-        if (!url || !finalUrl.includes(callbackUrl)) {
-          reject(new Error('Invalid Callback Url'));
+      // check if the redirect is to the callback URL and if it contains an authorization code
+      if (url && url.includes(callbackUrl)) {
```
pietrygamat: I noticed a case where this check may not be sufficient (which is the problem in this version as it was in the original):
My keycloak server has a redirect uri set to exactly: `https://example.com?param=value`. As keycloak expects me to use callback url exactly as configured, I do so in _bruno_, but after I provide correct credentials, I get redirected to `https://example.com/?param=value&code=...`, whis does not pass the check `url.includes(callbackUrl)` (extra `/` breaks this).  
Perhaps the string comparison using `includes` should be done on `URL.href`'s rather than plain string values, to have a level of normalization?
```
let strurl1 = 'https://example.com?par=val'
let strurl2 = 'https://example.com/?par=val'
strurl1 == strurl2 // ==>false
new URL(strurl1) == new URL(strurl2); // ==> false
new URL(strurl1).href == new URL(strurl2).href; // ==> true
new URL(strurl1).href // => https://example.com/?par=val 
new URL(strurl2).href // => https://example.com/?par=val 
```

dakshin-k: @helloanoop @lohit-1 Do we have any unit tests on this behaviour? I was not able to find any when creating the PR but it seems like there are several edge cases here so I would like to add some if they are not there

lohit-bruno: Hey @dakshin-k, there aren't any unit tests in place for this behaviour currently. 
Please feel free to add them. Thanks

dakshin-k: @lohit-1 Have added some test cases, please review.

I couldn't write a test for the entire component, including mocking all the `window` events, so I just extracted that check into a separate function and wrote tests for it.

packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
```diff
@@ -1,6 +1,10 @@
 const { BrowserWindow } = require('electron');
 const { preferencesUtil } = require('../../store/preferences');
 
+const matchesCallbackUrl = (url, callbackUrl) => {
+  return url ? new URL(url).host == new URL(callbackUrl).host : false;
+};
```
pietrygamat: Nope: my callback url is pointing to the same server that is auth url. It will pass this check and fail on missing 'code' even before the first redirect. 
Screenshot from 2024-04-22 23-48-42

dakshin-k: Oh man, there are _lots_ of edge cases here 😅 

Comparing the `href` as you suggested earlier will not work as the href includes any query params in the URL, i.e. while it solves the trailing `/` problem it breaks for the below case:

```
callbackUrl: https://callback.url/endpoint
testUrl       : https://callback.url/endpoint?code=abcd

// testUrl.href = https://callback.url/endpoint?code=abcd
```

This is the reason I compared against hostname instead of href. I currently don't have a better idea to implement this, will think about it and come back later.

Ideas are welcome if I'm missing something 😇

lohit-bruno: The function could be something similar to this @dakshin-k 

```
const matchesCallbackUrl = (url, callbackUrl) => {
  let { origin: urlOrigin, pathname: urlPathname } = new URL(url);
  let { origin: callbackUrlOrigin, pathname: callbackUrlPathname } = new URL(callbackUrl);

  urlPathname = urlPathname?.at(-1) == '/' ? urlPathname?.slice(0, -1) : urlPathname;
  callbackUrlPathname = callbackUrlPathname?.at(-1) == '/' ? callbackUrlPathname?.slice(0, -1) : callbackUrlPathname;

  const urlWithoutQueryParams = `${urlOrigin}${urlPathname}`;
  const callbackUrlWithoutQueryParams = `${callbackUrlOrigin}${callbackUrlPathname}`;

  return url ? urlWithoutQueryParams == callbackUrlWithoutQueryParams : false;
};

```

Edit: 
Not recommended, seems like too much of an overkill
The below solution should be enough

pietrygamat: Wouldn't this suffice?
`testUrl.href.startswith(callbackUrl.href)`

The spec mandates the code param is _appended_ the the query component, so it should be at the end. Also luckily the fragment component (the url part after #) is not allowed, so it does not have to be handled.

dakshin-k: Thanks, that worked! I have pushed the change, would someone mind testing it out once?

I faced a weird problem testing locally, not sure if I introduced a bug or found an existing one 😅 

I'm able to generate auth tokens successfully now, but I can't make any requests - When I click the send request button, it just shows me the response from my OAuth server with the access token, it doesn't actually make a request using that token or display that response:
 
Screenshot 2024-04-23 at 9 59 51 PM

pietrygamat: I wouldn't mind that. That's actually #1999

dakshin-k: Ah okay, thanks! 

So I guess that's it on this PR? Let me know if there's anything further to do!

P-de-Jong:
I have verified that the changes in this PR allow for using Auth0 with Bruno. Would appreciate for this to get merged and released so I can roll out Bruno in our company :)

image

gousse:
fix: #1003

helloanoop:
Merged!

Thanks @dakshin-k for working on this.
Thank you for the review @pietrygamat @lohxt1 !

This will be released in the build scheduled to go out on Tuesday, June 4 2024.

Issue #2152, opened 2024-04-23
Docs not support "http" markdown image


LCLLL:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

i found only support "https" image when use markdown grammar like " ![](https://xxx.jpg) ", but can't use " ![](http://xxx.jpg) "

### Screenshots/Live demo link

<img width="657" alt="iShot_2024-04-23_09 08 04" src="https://github.com/usebruno/bruno/assets/1538928/979a3560-d9c9-460b-ae96-1e1132cd90b6">

busy-panda:
Hi @LCLLL 

It should be fixed when the PR is merged.

PR #2171, opened 2024-04-25
bugfix(#2152): Docs not support "http" markdown image

busy-panda:
resolves: #2152

# Description

The error message was:

_Refused to load the image 'http://www.google.com/images/branding/googlelogo/2x/googlelogo_color_272x92dp.png' because it violates the following Content Security Policy directive: "img-src 'self' blob: data: https:"._

Electron was not allowing to load images over "http", only "https" was allowed.

I think it's reasonable to allow it (for images only).

https://github.com/usebruno/bruno/assets/139650490/ecaa04cd-ba32-4c75-8ae5-4ee961b3bb9f

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -24,6 +24,8 @@ const contentSecurityPolicy = [
   "font-src 'self' https:",
   // this has been commented out to make oauth2 work
   // "form-action 'none'",
+  // we make an exception and allow http for images so that
+  // they can be used as link in the embedded markdown editors
   "img-src 'self' blob: data: http: https:",
   "media-src 'self' blob: data: https:",
   "style-src 'self' 'unsafe-inline' https:"
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -24,7 +24,7 @@ const contentSecurityPolicy = [
   "font-src 'self' https:",
   // this has been commented out to make oauth2 work
   // "form-action 'none'",
-  "img-src 'self' blob: data: https:",
+  "img-src 'self' blob: data: http: https:",
   "media-src 'self' blob: data: https:",
   "style-src 'self' 'unsafe-inline' https:"
 ];


packages/bruno-electron/src/index.js
```diff
@@ -24,7 +24,7 @@ const contentSecurityPolicy = [
   "font-src 'self' https:",
   // this has been commented out to make oauth2 work
   // "form-action 'none'",
-  "img-src 'self' blob: data: https:",
+  "img-src 'self' blob: data: http: https:",
   "media-src 'self' blob: data: https:",
   "style-src 'self' 'unsafe-inline' https:"
```
andreassiegel: Looking at this, I wonder why HTTP would be fine for images but at the same time not for fonts, styles or other media.

I think, if the content security policy is changed, the change should be consistently applied to all kinds.

Its-treason: In this case the user can embed an Image from an HTTP-Server in the Docs. But users can't link any fonts etc. And I think we should use the most strict rules if possible.

end3rbyte: Yes I also thought about it but i didn't want to open security breaches without a need for it. It seems unlikely someone will include a font in the markdown editor, i don't think it's even possible. What do you think?

end3rbyte: Sorry i didn't see @Its-treason response before so i agree to only allow images because they can be linked in the markdown editor

andreassiegel: Makes perfect sense. 👍 
Maybe it's worth adding the reasoning to the code as an inline comment. 🙂

end3rbyte: Ok i can't now but i will later unless you can do it before.
Thanks

packages/bruno-electron/src/index.js
```diff
@@ -24,7 +24,7 @@ const contentSecurityPolicy = [
   "font-src 'self' https:",
   // this has been commented out to make oauth2 work
   // "form-action 'none'",
-  "img-src 'self' blob: data: https:",
+  "img-src 'self' blob: data: http: https:",
```
andreassiegel, line 29:
```suggestion
  // HTTP is intentionally allowed in the security policy to be able to embed images from HTTP servers.
  // In general, the most strict rules should be rules if possible,
  // i.e., the policy for fonts, styles, etc. allow only HTTPS as they cannot be embedded anyway
  "img-src 'self' blob: data: http: https:",
```

packages/bruno-electron/src/index.js
andreassiegel: I suggested a change because I can't directly edit the file.

end3rbyte: The comment is added now.
Thanks.

andreassiegel (approved):

Its-treason (approved):

helloanoop:
Merged. Thank you @busy-panda !

Thanks for the review @Its-treason @andreassiegel !

Issue #606, opened 2023-10-16
Autocomplete environment variables


roelofjan-elsinga:
I have quite a few environment variables and don't remember the name for most of them, so it'd be very helpful if the environment variables are autocompleted while using them in request headers and perhaps even the request body.

zeroskylian:
I think this is very useful, please add it. Just like postman, input `{` can prompt user

thetsajeet:
On top of this, I have a suggestion to have the autocomplete while adding new requests as well. Currently autocompletion of env variables are done after creating a request in the URL input, but maybe adding autocomplete in the URL input while creating this request section could also help:

![Image](https://github.com/user-attachments/assets/0bda017c-68e3-4a33-870d-b6e81c6097cf)

anusree-bruno:
@thetsajeet
Thanks for the feedback! Aside from your suggestion about adding autocomplete in the URL input while creating a new request, I believe the main issue has been completed.

thetsajeet:
Yes. Do you want me to create a new issue for it?

Issue #412, opened 2023-10-06
auto-completion for vars


cvmocanu:
Insomnia has a very useful feature, where you can use Ctrl+Space in any input box / text area, and you are presented with a list of variables that can be used.
Please implement this in Bruno also.

Both collection and environment variables should appear in the list, probably with a different icon.

jakenvac:
In the mean time, are there any docs that document where and how you can use the variables & with what syntax? (outside of scripting). I couldn't find any in the official docs.

FallenHero66:
I second this request, would be nice to have autocomplete on variables in the built-in bruno fields (URL field, headers, query, body ...)

> In the mean time, are there any docs that document where and how you can use the variables & with what syntax? (outside of scripting). I couldn't find any in the official docs.

I don't think there are, but you can use them like in Postman buy setting two curly brackets around the variable name: {{varname}} - if you do it right, they will be highlighted. You can specify their values either in the "vars" tab or in an environment, if you want to use them cross-request.

jakenvac:
@FallenHero66 Thanks for this! I've not had to use bruno (or any other http client) since I asked, but your timing is impeccable as I'll be using bruno today.

I'll experiment with the tip you've provided and see if I can contribute back some documentation to the project.

PR #2190, opened 2024-04-27
feat: Add variables auto completion

### LINKED ISSUE #1141: Autocompletion of headers value
Similar to postman, having an autocomplete for the value field of headers would be nice

MasterLambaster:
# Description
Add autocomplete for environment variables. 

https://github.com/usebruno/bruno/assets/134304/701515eb-9e95-4571-89f5-ef71ec33eb5f

The trigger is `{{` that shows all available environment variables. Supports all fields that have collection passed to it, i.e auth, headers etc. 
Might need sanity check for all the places where it's currently applied. 
Fixes #1141, #606, #412.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/SingleLineEditor/index.js
+++ b/packages/bruno-app/src/components/SingleLineEditor/index.js
@@ -80,14 +80,12 @@ class SingleLineEditor extends Component {
         'Shift-Tab': false
       }
     });
-    if (this.props.autocomplete) {
-      this.editor.on('keyup', (cm, event) => {
-        if (!cm.state.completionActive /*Enables keyboard navigation in autocomplete list*/ && event.keyCode != 13) {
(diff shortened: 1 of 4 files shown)


busy-panda:
I am having a little bug:

image

You can reproduce with the collection under /bruno/packages/bruno-tests/collection
with environment prod

The suggested value "process" shouldn't be here.

Also I don't think, you should display the value of the variable.
It will display secrets on the screen, which is not good in terms of security.

MasterLambaster:
@busy-panda `process` is an nested object that could be accessed, normally via dot notation. It's only visible if it's actually set.
Regarding hiding values: there are couple existing issues for the hiding secret values and it should be done on the global level rather this particular case.

acuenca-facephi:
Nice job, this should be included in the following releases.

G-Kevin:
I'd like to pump this topic as this would be a very useful feature. :)

PR #2164, opened 2024-04-25
Feature/oauth2 basic auth

pietrygamat:
# Description

With this change the OAuth2 authorization on request and collection level is now including additional configuration option for handling client credentials when accessing token endpoint. This applies to all currently supported grant_types.

The new default option is to sent the credentials as `Authorization: Basic` header:
```
auth:oauth2 {
  ...
  client_id: brunoclient
  client_secret: {{client_secret}}
  client_secret_method: client_credentials_basic
}
```

The old behavior to include them in the request body is still available.
```
auth:oauth2 {
  ...
  client_id: brunoclient
  client_secret: {{client_secret}}
  client_secret_method: client_credentials_post
}
```

### Contribution Checklist:

- [ ] **The pull request only addresses one issue or adds one feature.**
- [ ] **The pull request does not introduce any breaking changes**
- [ ] **I have added screenshots or gifs to help explain the change if applicable.**
- [ ] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ ] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

PR #2077, opened 2024-04-14
feat: Improvements to OAuth2-related user interface

pietrygamat:
# Description

Redesign of OAuth2 UI. Shows tokens and other information obtained from Authorization Server in separate (rudimentary) part of UI, decoupling them from information coming from actual user request. With this change it is now possible to inherit OAuth2 configuration from collection level.

This PR is a continuation of work started in #2061 where caching of access token was first implemented (currently pending review).

# Request level view:
- Clicking Get Access Token (or simplly running the request) results in bruno executing correct flow. 
- User may be prompted to provide password, depending on the grant type. 
- After the request completes, the oauth2 credentials are stored and displayed in collapsible section at the bottom of the screen.
- Any subsequent requests witin the same collection will reuse them. 
- Clearing cache resets everything back.

Screenshot from 2024-05-05 22-01-00
Screenshot from 2024-05-05 22-01-33
Screenshot from 2024-05-05 22-02-11
Screenshot from 2024-05-05 22-03-28
# Collection level
- The same controls are added to collection level Auth configuration.
- Clicking Get Access Token on collection level results in executing authorization request and updating credentials. 
- Inherit mode is now supported also when collection uses OAuth2. 
Screenshot from 2024-05-05 22-04-26
Screenshot from 2024-05-05 22-04-45

### Contribution Checklist:

- [ ] **The pull request only addresses one issue or adds one feature.**
- [ ] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

resolves #1999 
resolves #2074
resolves #1704 
resolves #2494

PR #2253, opened 2024-05-06
Feat: OAuth2: Show details of both token request and user request in the Timeline

### LINKED ISSUE #1933: Oauth2.0 - client Credentials -> Get Access Token =>not good URL?
When I use the client Credentials with Oauth2, i think that is not the good URL to use to get the token whith the button "Get Acces Token". 
Bruno-client_credentials

pietrygamat:
# Description

This PR builds on top of both  #2077 and #2164 . It enhances the timeline to show the details of both OAuth2 token request and user defined request.
This is done because: 
- Since #2077 the token request is executed in the background, and perhaps skipped, if bruno has a token cached, so there is no feedback to the user about what requests are executed
- Since #2164 the token request works in two modes (credentials in basic auth header  or request body) - this is another variable that we want the user to have better visibility of

When token request is executed, both requests are logged:
Screenshot from 2024-05-06 23-03-43

When using cached credentials (previously obtained access_token), only user request is shown:
Screenshot from 2024-05-06 23-19-15

Also, fixes #1933

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/Timeline/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/Timeline/index.js
@@ -6,6 +6,13 @@ import StyledWrapper from './StyledWrapper';
 const Timeline = ({ request, response }) => {
   const requestHeaders = [];
   const responseHeaders = typeof response.headers === 'object' ? Object.entries(response.headers) : [];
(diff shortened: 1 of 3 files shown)


leikoilja:
this is really needed

sreelakshmi-bruno:
Hi, thanks for your contribution. This feature is available as of v4.1.0, please try it out. Closing this PR for now.

PR #2272, opened 2024-05-09
feat: display raw query response unformatted for readability

DawoodMorris:
# Description

In the current Bruno release, when the query response is raw text, such as when the server returns an error and outputs programming language error messages, the response looks like the following image and, is not clearly readable (in my humble opinion):

image

The solution is to display the raw response unformatted if it is raw text or mixed with unformattable data as in the above screenshot. 

After modifying the `formatResponse` function in the `/bruno/packages/bruno-app/src/components/ResponsePane/QueryResult` component, to not format if the data is raw and not parsable, the display looks like the following and is readable (in my humble opinion):

image

**Good To Consider**
I have been using Postman for a long time. Then I recently updated it and it became very slow and sucked the burn out of me as it slowed my machine. I grew impatient as it was getting in my way. I began hunting for alternatives and came across Bruno. The only setback with Bruno was that I could not read the raw server responses during my work (you know the usual programming errors dance!). I tried to come up with a solution because in Postman when the response is raw text and not parsable, it is displayed clearly and is legible. 

My proposed feature/solution achieves the same effect in Bruno.

I hope this makes sense.

Thanks

Dawood


packages/bruno-app/src/components/ResponsePane/QueryResult/index.js
```diff
@@ -35,15 +42,14 @@ const formatResponse = (data, mode, filter) => {
...
-
     return safeStringifyJSON(parsed, true);
   }
 
   if (typeof data === 'string') {
     return data;
   }
-
-  return safeStringifyJSON(data);
+  //return safeStringifyJSON(data)
```
maxmarkus: I'd suggest to remove commented code.
```suggestion
  //return safeStringifyJSON(data)
```

packages/bruno-app/src/components/ResponsePane/QueryResult/index.js
```diff
@@ -18,15 +18,22 @@ const formatResponse = (data, mode, filter) => {
     return '';
   }
 
-  if (mode.includes('json')) {
+  let isTrueJSON;
+  try {
+    isTrueJSON = typeof JSON.parse(JSON.stringify(data)) === 'object';
+  } catch (error) {
+    console.log('Error parsing JSON: ', error.message);
```
maxmarkus: Console output is not required here, I think?
```suggestion
    console.log('Error parsing JSON: ', error.message);
```

DawoodMorris: Sure, that (the console.log) is not required. It was for inspection purposes. We can remove it.

helloanoop:
Merged! Thanks @DawoodMorris !

DawoodMorris:
Thanks @helloanoop!

DawoodMorris:
When should we expect it to make it to the release versions?

helloanoop:
This will be available in the upcoming `v1.29.0` release which is expected to go out tomorrow.

DawoodMorris:
Marvelous!

PR #2321, opened 2024-05-16
fix: multipart formdata interpolation

lohxt1:
~ multipart formdata interpolation
~ also considers scenarios when the body is set using a pre-request script

https://github.com/usebruno/bruno/assets/25679466/7b0bda6c-033a-4f4f-8603-460f1661b33b

https://github.com/usebruno/bruno/assets/25679466/fe75eb24-95ae-4b60-b83d-bce559c001c7


packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -186,6 +195,12 @@ const runSingleRequest = async function (
       request.data = qs.stringify(request.data);
     }
 
+    if (request.headers['content-type'] === 'multipart/form-data' && !request.__bruno__bodySetViaMethodCall) {
```
Its-treason, line 198: Instead of checking of the Body was set inside a script. Wouldn't it be better to check if the body is already a `FormData` object?

What if a user has a script like this:
```js
const originalBody = req.getBody();

const calculated = 'Imagine this is calculated';

req.setBody({
  ...originalBody,
  'script-val': calculated,
})
```
Now the plain object is sent to the server.

packages/bruno-electron/src/ipc/network/prepare-request.js
```diff
@@ -207,10 +207,11 @@ const prepareRequest = (request, collectionRoot, collectionPath) => {
...
 
   if (request.body.mode === 'multipartForm') {
+    axiosRequest.headers['content-type'] = 'multipart/form-data';
+    const params = {};
     const enabledParams = filter(request.body.multipartForm, (p) => p.enabled);
-    const form = parseFormData(enabledParams, collectionPath);
-    extend(axiosRequest.headers, form.getHeaders());
-    axiosRequest.data = form;
+    each(enabledParams, (p) => (params[p.name] = p.value));
+    axiosRequest.data = params;
```
Its-treason, line 214: This introduces the same breaking change from my PR, where `req.body` becomes a plain Key-Value object instead of a `FormData` object.

helloanoop:
We're closing this PR and will begin working on this issue from scratch.

PR #2324, opened 2024-05-16
Feat: Catch post request variable evaluation errors

### LINKED ISSUE #2005: [Request] When using post response variables, the actual response of the request should be returned even if the var fails to be set
## Problem

Currently when we have a post response variable configured, that expect a specific response format like `res.body.data.id` - it will fail whenever that particular format is not found.
There are cases where that format may be missing, for example due to expired token - that results with a different response from the API we call.
However, right now, with the variable configured, Bruno will show:
`Error invoking remote method 'send-http-request': TypeError: Cannot read properties of undefined (reading 'id')` 
instead of the actual response. To see the actual response the only way is to disable the variables(s) and call the endpoint again. 

This is how it looks in app:
image

## Suggestion 

When post response variables are in use, but the response does not return what the variable expects, it should still show the actual response. The error of the variable should be showed in a separate tab - similar to how test assertions are working - that a failed test does not replaces the response, but shows the information in a separate tab.

My apologies if this is somewhere mentioned already, but I couldnt find similar request with my search phrases.

rbideau:
fix #2005

# Description

- display post request variable evaluation errors in a toast, each individual variable error on a new line
- display the response body (was previously replaced by the an error "Error invoking remote method 'send-http-request': ..."

image

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -395,6 +395,10 @@ const registerNetworkIpc = (mainWindow) => {
           collectionUid
         });
       }
+
+      if (result?.error) {
+        mainWindow.webContents.send('main:display-error', result.error);
+      }
     }
 
     // run post-response script
--- a/packages/bruno-js/src/runtime/vars-runtime.js
+++ b/packages/bruno-js/src/runtime/vars-runtime.js
@@ -56,14 +56,27 @@ class VarsRuntime {
       ...bruContext
     };
 
+    const errors = new Map();
     _.each(enabledVars, (v) => {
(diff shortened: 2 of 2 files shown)


helloanoop:
Nice @rbideau !

@lohxt1 Lets target ti have this merged for next week's release.

prwnr:
@rbideau will this also work for post-response scripts? cause I noticed today that in case of their error the response is also not shown.

rbideau:
@prwnr I didn't test with post-response scripts but I don't think it will works with that.

I've prefered to kpe the PR small and focused on #2005 to help make it easier to review and merge.

helloanoop:
Merged.

Thanks for taking care of this @rbideau !

PR #2353, opened 2024-05-22
fix: collection import

lohxt1:
Fixed object spread operations for collection import


helloanoop:
Merged. Thanks @lohxt1 !

Issue #2367, opened 2024-05-27
Error invoking remote method 'send-http-request': TypeError [ERR_UNKNOWN_ENCODING]: Unknown encoding: ISO-8859-1


TamaraTorres:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Request with form urlencoded values, I was using version 1.10 and the request was working I upload to version 1.18 and now I'm getting this error ( no change on the request, neither ont the service , which in fact works on Postman ) 
![image](https://github.com/usebruno/bruno/assets/6197753/d33d2b22-ec89-4a54-b7ef-9e7d51013db2)

### Screenshots/Live demo link

![image](https://github.com/usebruno/bruno/assets/6197753/11208ebd-19e4-459e-a253-3c57465f8990)
![image](https://github.com/usebruno/bruno/assets/6197753/823c5850-8498-43eb-b9c6-8e18da2cbb06)

pbleuse-orange:
Hello,

I have same issue with same version on windows and linux.
CLI work.

Pascal

helloanoop:
This has been released in `v1.9.0`

Closing this issue. Please reopen if the issue persists.

pbleuse-orange:
> This has been released in `v1.9.0`
> 
> Closing this issue. Please reopen if the issue persists.

I think you would said 1.19.0 ;)

PR #2371, opened 2024-05-28
fix(#2367): handle response body decode

lohxt1:
Description: decoding response body based on the content-type charset value

content_type_charset
decoded_response_body


KWottrich:
Hoping this gets merged soon! I'm experiencing the same issue described in usebruno#2367

Issue #2376, opened 2024-05-28
trailing decimal points with a 0 are truncated. 


mattisking:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

We have a couple of API's that require a "number" field, with data validation rules for decimal.

Currently Bruno is trimming off 0's which is a real problem for us with monetary amounts. I'm not in a position to change the validation rules.

Very quick reproducible example:
POST application/json

{
"amount": 123.00
}

Bruno sends:
{
"amount": 123
}

This breaks data validation rules for the webapi that is requiring all amounts to be in decimal format with 2 decimal points: 0.00

I can attach more here if necessary, but I honestly think that's enough. This may exist as a bug somewhere and if it does, I apologize. I searched a few ways and didn't find it.

### Screenshots/Live demo link

![bruno-number-truncate](https://github.com/usebruno/bruno/assets/412168/3cc7ed02-ffc4-4d85-9d3c-adbd1f5096ee)

Most of the relevant information is redacted but this is pretty clear I think.

Issue #2362, opened 2024-05-24
Unquoted Variables in JSON Body Break Comments and Prettier


prestonguillot:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Using an _unquoted_ variable in the value of a field in a JSON body causes any comments in the body to be sent on the request instead of stripped. This also breaks the prettier.

e.g.

```
{
  "test": {{testVar}} //test
}
```

is sent as 

```
data "{\n \"test\": 1234 //test\n}\n"
```

and is prettied into

```
{
  "test": {
    {testVar
    }
  }     //test
}
```

whereas

```
{
  "test": "{{testVar}}" //test
}
```

is sent as

```
data {"test":"1234"}
```

and is unchanged by prettify.

### .bru file to reproduce the bug

```
meta {
  name: Demo
  type: http
  seq: 2
}

post {
  url: www.google.com
  body: json
  auth: none
}

body:json {
  {
    "test": {{testVar}} //test
  }
}

vars:pre-request {
  testVar: 1234
}
```

### Screenshots/Live demo link

<img width="772" alt="Screenshot 2024-05-24 at 10 25 54 AM" src="https://github.com/usebruno/bruno/assets/1053799/fc57e9dd-604c-4bf6-887d-8ba7629b2ec3">
<img width="733" alt="Screenshot 2024-05-24 at 10 26 20 AM" src="https://github.com/usebruno/bruno/assets/1053799/ca9974de-1383-4c23-8f4f-67cfb45c4af9">
<img width="455" alt="Screenshot 2024-05-24 at 10 26 03 AM" src="https://github.com/usebruno/bruno/assets/1053799/36cf4287-c99f-4598-a00e-e7480b2dd784">

helloanoop:
This has been fixed as a part of PR https://github.com/usebruno/bruno/pull/2773, and will go out in the v1.27.0 release.

Issue #1768, opened 2024-03-12
[BUG]: Decimal numbers `*.0` are converted to integers


AHOCHWE:
## Description
It seems that Bruno performs a conversion when a decimal number from the pattern `*.0`  is sent to the API. Bruno perform a conversion to an integer which leads to some issues in our API and their behavior because it expects a decimal number and not an integer.

![image](https://github.com/usebruno/bruno/assets/58166982/db17631c-d587-477b-8ffe-d85bc3924669)

mjhcorporate:
Unfortunately, this is a well-known problem with JSON (see, for example, [this note in the FHIR spec](https://build.fhir.org/json.html#decimal)). JSON only has on number type, and most serializers throw away trailing zeros and trailing decimal points.

In general, I highly recommend sending the numbers as strings if you care about the number format. Anything else is bound to break at some inconvenient time.

BUT... bruno should of course cover this use-case. Currently, for json-bodies, bruno builds a JS object and then serializes it. This changes several things:
- trailing zeros (and possibly the decimal point) are omitted
- the JSON formatting is stripped
- the order of the properties can potentially change (in reality, the order stays constant)
- duplicate properties are deduplicated (last property with same name "wins"

Naturally, this is what you want for proper json requests, but bruno should also be able to send slightly improper json requests! After all, it's a testing tool.

One possible workaround that _should_ work (but doesn't, because of a separate bug) is: Set the body-type to `text`, and manually set the `Content-Type` to `application/json`. That way, you have full control over the request body.

![image](https://github.com/usebruno/bruno/assets/49681105/fa3094d8-26de-4f61-a469-50e62824b410)

Unfortunately, there is a separate bug that sometimes (not always) encodes the body as json (see below). I think there is an existing issue for this, but I have to look it up.

---
To see the body that bruno sends, use netcat as "echo-server": `nc -l 8000`. You can then send a request to `localhost:8000`, and it will be printed to the command line.

Together with the text-body + manual content-type-header trick, this leads to:
```
> nc -l 8000
POST / HTTP/1.1
Accept: application/json, text/plain, */*
Content-Type: application/json
request-start-time: 1710330826355
User-Agent: axios/1.6.7
Content-Length: 17
Accept-Encoding: gzip, compress, deflate, br
Host: localhost:8000
Connection: close

{
  "num": 3.00
}
```

But sometimes, the result is something like this:
```
> nc -l 8000
POST / HTTP/1.1
Accept: application/json, text/plain, */*
[...]

Issue #396, opened 2023-10-05
Allow Comments in JSON Body


helloanoop:
This is very important to save useful info.

`request body`
```javascript
// this should still work despite being a comment line
{
  "access_token": "{{token}}",
  "clientVersion": "1.0.0.0"
}
``` 

Solution: Use a decommenting library suchas [decomment](https://www.npmjs.com/package/decomment) to strip comments.

Solutioning Credit: @pove 👊

helloanoop:
Yes. Thanks @DaPutzy 

I also added it to the release notes of [v0.20.0](https://github.com/usebruno/bruno/releases/tag/v0.20.0)

PR #2382, opened 2024-05-29
fix(#2376): send unmodified request json body

lohxt1:
fixes: [#2376](https://github.com/usebruno/bruno/issues/2376), [#2362](https://github.com/usebruno/bruno/issues/2362), #1768 

added a checkbox in the collection settings' presets tab to choose if the request body should be parsed and de-commented, or be sent as it is needs to be added

Screenshot 2024-06-19 at 3 30 11 PM

(part of the bruno config presets)

https://github.com/usebruno/bruno/assets/25679466/0e8b477e-2ee4-49b8-b057-7c39968b8837

currently bruno supports adding comments in the json body
and before the actual request is made the comments are removed and the body is parsed as json
ref: [PR](https://github.com/usebruno/bruno/issues/396)

### DIFF
--- a/packages/bruno-app/src/components/CollectionSettings/Presets/StyledWrapper.js
+++ b/packages/bruno-app/src/components/CollectionSettings/Presets/StyledWrapper.js
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 
 const StyledWrapper = styled.div`
   .settings-label {
-    width: 110px;
+    width: 170px;
   }
 
   .textbox {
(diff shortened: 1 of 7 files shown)


jwetzell:
Does this cover #2207, #2438 as well? I would think getting a less destructive JSON parsing library like suggested in on of those issues would be the better/more transparent solution for the end user.

helloanoop:
CLosing since this was addressed in PR https://github.com/usebruno/bruno/pull/2773

Issue #2446, opened 2024-06-12
OAuth2: The scope in case of `Password Credentials` grant type


nicolasnobelis:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

See attached screenshot:

The scope param is always passed (when according to the spec, it is optional).

The server responds with

```
{
  "error": "invalid_scope",
  "error_description": "Invalid scopes: "
}
```

This issue requires the same kind of fix as https://github.com/usebruno/bruno/pull/1807/files but for Password Credentials.

### Screenshots/Live demo link

![bruno_bug](https://github.com/usebruno/bruno/assets/28786956/ad3c8cf7-88fb-466f-94ad-cdadc7377ea2)

helloanoop:
The PR has been merged.
This is set to go out in the upcoming `1.20.0` release

PR #2447, opened 2024-06-12
fix: Pass the scope conditionally for OAuth2 Password Credentials

nicolasnobelis:
The scope param is always passed (when according to the spec, it is optional).

The server responds with

{
  "error": "invalid_scope",
  "error_description": "Invalid scopes: "
}

Fixes: https://github.com/usebruno/bruno/issues/2446.


pietrygamat:
The same is true for other authorization methods Perhaps it's worth updating different ones as well?

helloanoop:
Merged!

Thanks for working on this @nicolasnobelis !
Thanks for the review @pietrygamat @lohxt1 !

Issue #2439, opened 2024-06-11
Request with pathParam returning 400 code


Ukrainis:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

When using :pathParam in request, then getting response code 400. Same with hardcoded parameter giving correct response.
In attachment you will find two same requests, one with pathParam (returns 400), second without (returns 200 and correct data). And two screenshots of same requests.

### .bru file to reproduce the bug

Not working:
meta {
  name: Get user with pathParam
  type: http
  seq: 2
}

get {
  url: https://spring-boot-test-aplication.herokuapp.com/api/users/findById/:id
  body: none
  auth: none
}

params:path {
  id: 16
}
Same but without pathParam and working:
meta {
  name: Get user without pathParam
  type: http
  seq: 3
}

get {
  url: https://spring-boot-test-aplication.herokuapp.com/api/users/findById/16
  body: none
  auth: none
}

### Screenshots/Live demo link

![Using_pathParam](https://github.com/usebruno/bruno/assets/11504404/2141e62e-9d4b-4957-89ca-71686330166f)
![Without_pathParam](https://github.com/usebruno/bruno/assets/11504404/88169902-8b46-4bb8-8950-d5cb26f53d59)

fgrande:
On my Mac, Bruno 1.19.0 works fine, with a path param at the end of a call.
Not able to reproduce what @Ukrainis says, in both cases (with/without param) the application returns 404 (due to application not found)

layereight:
> We have this issue as well, it seems to only occur when the pathParam is the last element of the url.

Same here. We're always getting HTTP 400 as response.

Ukrainis:
> On my Mac, Bruno 1.19.0 works fine, with a path param at the end of a url. Not able to reproduce what @Ukrainis says, in both cases (with/without param) the application returns 404 (due to application not found)

This is working endpoint, just checked, gave same results, as described in a bug.

> We have this issue as well, it seems to only occur when the pathParam is the last element of the url.

Getting same, when this is not at end of the url.

sanjai0py:
Could you share more information by sending a .bru file or a JSON export? That would be really helpful for investigating the issue further.

Ukrainis:
Attached collection:
[Personal.json](https://github.com/user-attachments/files/15804208/Personal.json)

fgrande:
[...]

PR #2448, opened 2024-06-12
Fix for error with params

fgrande:
# Description

This would fix Issue #2439 
Looks like "params" property breaks the call, and this is filled with the params.


sanjai0py (commented):
Can you please add more details on how this fix works?

fgrande:
Consider those calls

https://spring-boot-test-aplication.herokuapp.com/api/users/findById/:id (set 16 as :id)
https://spring-boot-test-aplication.herokuapp.com/api/users/findById/16

The difference between them are the "params" tag in the request.

**_Request with params :id :_** 

mode: 'none',
  method: 'GET',
  url: 'https://spring-boot-test-aplication.herokuapp.com/api/users/findById/16',
  headers: {},
  params: [
    {
      name: 'id',
      value: '16',
      enabled: true,
      type: 'path',
      uid: '0Ndm3TH0VGhdOE0CgP4f6'
    }
  ],
  responseType: 'arraybuffer',
  script: {},


**_Request without param :_**

 mode: 'none',
  method: 'GET',
  url: 'https://spring-boot-test-aplication.herokuapp.com/api/users/findById/16',
  headers: {},
  params: [],
  responseType: 'arraybuffer',
  script: {},
  vars: {},
  assertions: [],


As you can see the difference lays in the params property. Since it is already decoded (the request.url is - at this point - already filled with the correct param) we don't need it anymore. And putting it, axios refuse the request because of an exception related to the actual url composed :

 _header: 'GET /api/users/findById/16?0[name]=id&0[value]=16&0[enabled]=true&0[type]=path&0[uid]=w4AGcAZKgr0VVwXrnBKt7 HTTP/1.1\r\n' +
      'Accept: application/json, text/plain, */*\r\n' +
      'request-start-time: 1718263487771\r\n' +
      'User-Agent: axios/1.6.7\r\n' +
      'Accept-Encoding: gzip, compress, deflate, br\r\n' +
      'Host: spring-boot-test-aplication.herokuapp.com\r\n' +
      'Connection: keep-alive\r\n' +
      '\r\n',

So, removing the params property, works fine. Maybe is not the cleaner solution, but I can give the idea of what's happening.

Regards
FabioG

fgrande:
Adding test files (related to the issue) 
[BrunoTest.zip](https://github.com/user-attachments/files/15815507/BrunoTest.zip)

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -499,6 +499,8 @@ const registerNetworkIpc = (mainWindow) => {
       let response, responseTime;
       try {
         /** @type {import('axios').AxiosResponse} */
+        // Looks like the "params" create problems if filled
+        request.params = [];
```
sanjai0py: After interpolation, we can remove the params values as they are already interpolated in the URL, making the approach simpler.

https://github.com/usebruno/bruno/blob/9b382fa0911536a0bfa27ae0d8a67007cc1364f7/packages/bruno-electron/src/ipc/network/interpolate-vars.js#L118

`  delete request.params;
`

sanjai0py (approved):

packages/bruno-electron/src/ipc/network/index.js
fgrande: @sanjai0py Thanks a lot for the hint (I'm still working on my knowledge of Bruno), I've just updated my PR. Let me know if it's ok, hope to be (somehow) helpful

helloanoop:
Thanks @fgrande for investigating the issue and coming up with the fix.

I feel the code related to deleting the path params is better to be kept inside the `configureRequest` instead of having it in interpolate vars function. I have asked @sanjai0py to make this change right away in a separate PR.

Closing this PR. Thanks again for your work on this.

Issue #2043, opened 2024-04-09
JSON body is quoted when invalid


GregoireBellon:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Hello, 

VERSION : 1.12.3. 
---

When sending a POST request with valid JSON in the body, the body is sent without any modification. 

When sending a POST request with invalid JSON in the body, the body is quoted before being sent.

This behavior only occurs when the selected content type is JSON. When sending text, invalid XML or SPARQL, the body is not quoted. There is no indication in the timeline view that the body is quoted.

--- 
I'm conscious that I'm complaining that Bruno prevents me from shooting myself in the foot, but I'm working on an API that, for legacy purposes, needs to accept requests containing invalid JSON in the body, and a text/json content type. And I can't test my implementation because of this behavior. 

A setting to configure this behavior, and some cues that it happens would be nice.

I would be glad to contribute to fix this issue !

### .bru file to reproduce the bug

[reproduce-bug.zip](https://github.com/usebruno/bruno/files/14918530/reproduce-bug.zip)

### Screenshots/Live demo link
Here are the bodies the requests recieved by a beeceptor endpoint : 
![Capture d'écran 2024-04-09 143041](https://github.com/usebruno/bruno/assets/45675498/cbf7a520-f1c1-4284-b389-2fe25bbac99a)
![Capture d'écran 2024-04-09 143059](https://github.com/usebruno/bruno/assets/45675498/6ec5b3e3-49ce-4bab-afb8-e51057476628)

As you can see, the data section doesn't show that quotes are being added to the request
![Capture d'écran 2024-04-09 143321](https://github.com/usebruno/bruno/assets/45675498/69ece7cd-5ef2-4a19-b3d9-03250e700154)
![Capture d'écran 2024-04-09 143237](https://github.com/usebruno/bruno/assets/45675498/49ffca96-d94e-4724-b4be-300ff5e13390)

Fabiensk:
I'm facing the exact same issue. With Bruno 1.16.1 (desktop, not CLI),  I could use in a Pre Request Script `req.setBody(Buffer.from("the text")) ` and the raw value without quotes was sent. With Bruro 1.17 this hack stopped working.

zlfn:
This issue has deteriorated after 1.27.0,
Bruno now **IMPLICITLY** sends quoted requests.
[...]

PR #2449, opened 2024-06-12
fix: invalid json body mistakenly quoted

GregoireBellon:
# Description
This pull request addresses the issue https://github.com/usebruno/bruno/issues/2043. 

It is also related to this one https://github.com/axios/axios/issues/4034

In my opinion, Bruno shouldn't patch a faulty JSON string, since it restrains its range of use, notably when working with poorly designed APIs. 

Moreover, other API clients (Postman, insomnia) don't have this behavior, so it could prevent one from adopting Bruno.

The fix consists in adding a transformRequest to the Axios instance that prevents the default behavior if the content type is "application/json" and the body is a string.


hotbinz:
yes,you are right,please merge this request as soon as possible

Issue #2496, opened 2024-06-21
Export Collection as Postman Collection


thannebauer:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

When exporting an Collection as Postman Collection and importing it in Postman, the URL is broken.
URL "https://github.com/usebruno/bruno/issues/new?assignees=&labels=bug&projects=&template=BugReport.yaml" is imported in Postman as "github.com/?assignees=&labels=bug&projects=&template=BugReport.yaml"
URL "{{apiurl}}\somewhere\somehow?param1=123" (apiurl is variable from environment) is imported as "/"

Comparing both Postman Json collections I find these main differences:
* request.url.protocol element is missing for "fix URL"
* request.url.path is exported as empty array by bruno, but filled by postman
* request.url.host is empty for "Test Variable URL"

After fixing this for testing purpose, I could import the collection with correct URLs in Postman. 

In addition an empty request named "bruno" is ex- and imported. This is caused by an additional empty element:
`    {
      "name": "bruno",
      "event": []
    },`

[
![compare_json_collections](https://github.com/usebruno/bruno/assets/157024608/90074b25-8072-443b-9e96-ff3b18559ccb)
](url)

### .bru file to reproduce the bug

Bruno collection: [Sandbox.zip](https://github.com/user-attachments/files/15927091/Sandbox.zip)

exported Postman collection: 
[Sandbox.json](https://github.com/user-attachments/files/15927094/Sandbox.json)

Same collection content exported from Postman: 
[Postman-Sandbox.postman_collection.json](https://github.com/user-attachments/files/15927118/Postman-Sandbox.postman_collection.json)

### Screenshots/Live demo link

Postman after import:
![image](https://github.com/usebruno/bruno/assets/157024608/a2fdff4e-4ce6-410c-8c5b-8989a6b7c67b)
![image](https://github.com/usebruno/bruno/assets/157024608/21ef43ce-d24b-493f-bc05-333dbbfdcd8b)
![image](https://github.com/usebruno/bruno/assets/157024608/6ae07903-460e-4a3e-b9e8-c3c89f633d9e)
What I have exported from Bruno
![image](https://github.com/usebruno/bruno/assets/157024608/4dfa9834-0da2-4c7c-a8c9-0e5f5361f907)
![image](https://github.com/usebruno/bruno/assets/157024608/7622f2ac-596d-4054-b792-cc72205123b3)

[...]

PR #2510, opened 2024-06-25
Fix URL generation bug in postman collection exporter #2496

gtlsgamr:
# Description

This PR fixes an issue mentioned in #2496 where exporting a postman collection which has variables in the url leads to getting only '/' in the place of full URL.  Now we have a check for template variables before generating URL hostname so that it does not try to create a URL variable anymore which led to the mentioned error.

**BEFORE**
Bruno -
image

Postman -
image

**AFTER**
Postman -
image

### Contribution Checklist:

- [x ] **The pull request only addresses one issue or adds one feature.**
- [x ] **The pull request does not introduce any breaking changes**
- [x ] **I have added screenshots or gifs to help explain the change if applicable.**
- [x ] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x ] **Create an issue and link to the pull request.**

### DIFF
--- a/packages/bruno-app/src/utils/exporters/postman-collection.js
+++ b/packages/bruno-app/src/utils/exporters/postman-collection.js
@@ -173,13 +173,18 @@ export const exportCollection = (collection) => {
 
   const generateHost = (url) => {
     try {
-      const { hostname } = new URL(url);
-      return hostname.split('.');
+      if (url.includes('{{') && url.includes('}}')) {
+        return url;
+      } else {
+        const { hostname } = new URL(url);
+        return hostname.split('.');
+      }
     } catch (error) {
       console.error(`Invalid URL: ${url}`, error);
       return [];
     }
   };
+
 
   const generatePathParams = (params) => {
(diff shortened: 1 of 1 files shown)


packages/bruno-app/src/utils/exporters/postman-collection.js
```diff
@@ -173,14 +173,19 @@ export const exportCollection = (collection) => {
 
   const generateHost = (url) => {
     try {
-      const { hostname } = new URL(url);
-      return hostname.split('.');
+      if (url.includes('{{') && url.includes('}}')) {
+        return url;
```
jbgomond, line 177: const { hostname } = new URL(url); has the effect of removing the path after the hostname. It will not work with that code. Also, we should add unit tests :)

jbgomond (changes requested):

helloanoop:
Hey @gtlsgamr !
thanks for taking time to work on this PR.

I am closing this PR as this has already been addressed in #3025

Issue #2539, opened 2024-06-30
New Assertion : lengthBetween


fgrande:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

I'd like to have a new assertion named lengthBetween, to validate a field's length (string or array) against an interval (now we have just an exact comparison, through "length").

I'm adding a PR for this.

### Mockups or Images of the feature

<img width="532" alt="image" src="https://github.com/usebruno/bruno/assets/811941/591339da-851a-467c-b412-8cb50e2b8321">

PR #2540

PR #2540, opened 2024-06-30
Added lengthBetween Assertion

fgrande:
# Description

Added a new assertion named lengthBetween, to validate a field's length (string or array) against an interval (now we have just an exact comparison, through "length").

Issue #2539 

### Contribution Checklist:

image

### DIFF
--- a/packages/bruno-app/src/components/RequestPane/Assertions/AssertionOperator/index.js
+++ b/packages/bruno-app/src/components/RequestPane/Assertions/AssertionOperator/index.js
@@ -6,33 +6,34 @@ import lightTheme from 'themes/light';
 /**
  * Assertion operators
  *
- * eq          : equal to
- * neq         : not equal to
- * gt          : greater than
- * gte         : greater than or equal to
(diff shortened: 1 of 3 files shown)


jwetzell:
Not sure I see why this is necessary when you can do something like this.
Screenshot_20240701_132138

fgrande:
@jwetzell I guess You're right, it's quite useless.... :-/

jwetzell:
@fgrande I wasn't sure if the existing stuff didn't work for all cases of length. I don't use the assertion stuff a whole lot, but even the length option puzzles me just because what I tried with `.length` already did what I expected and would do string length or array length.

fgrande:
@jwetzell you're right, at this point also length is quite useless since everything can be done with `.length` and the existing operators (equal, notEqual, gt, lt, between, etc...). At the same time, IMHO, assertions are here to simplify life and configurations, being used with ease... 
I closed the PR since is not adding value to Bruno. Useless ideas happens :-)

Issue #2573, opened 2024-07-06
Broken requests sequence after any save or move existing requests


SPIRIT-00:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

After 1.20.1 update was broken request sequences. On any save or move existing or new request value of "seq:"  always sets on 1.

First image: before editing/create any request. Second image: i edited request that was on 7th place. It moved on 1st place.
Same happens if move any request, all them seq sets on 1.

### Screenshots/Live demo link

![изображение](https://github.com/usebruno/bruno/assets/128896119/a9906b64-5346-4bfe-8bc8-b5420ecf02ef)
![изображение](https://github.com/usebruno/bruno/assets/128896119/c48da204-7b7b-474a-a18f-06327ca17740)

helloanoop:
Thank you @Its-treason for the fix. I have merged the PR

Will have a patch version 1.20.2 go out in the next few hours.

helloanoop:
@SPIRIT-00 @Its-treason 

This has been fixed in `v1.20.2`

SPIRIT-00:
Confirm, just tested it.

PR #2574, opened 2024-07-06
fix(#2573): Fix broken resequenzing of requests

Its-treason:
# Description

Fixes broken resequenzing of Requests. The `seq` property is in the Top-Level of the JSON instead of inside `meta`.


helloanoop:
Merged!

Thanks for the quick fix @Its-treason !

Issue #409, opened 2023-10-06
[FEATURE REQUEST] .env file per environment


skreutzberger:
As far as I understand the docs at https://docs.usebruno.com/secrets-management/dotenv-file.html and my own testing, it seems that I can just have one `.env` file per collection. 

"Normally" I would expect an additional functionality where the environment name matches the `.env` files name. So for example if I have in my collection `foo` the 2 environments `local` and `prod` then the following files would be good:

- `.env` -> to store general env vars which are used by all environments
- `local.env` (or `.local.env`) -> values in here override the ones in `.env` and are just used in the `local` environment
- `prod.env` (or `.prod.env`) -> values in here override the ones in `.env` and are just used in the `prod` environment

It would allow a more realistic scenario where you separate production and development secrets.

What do you think?

helloanoop:
So usually many things remain static in an env like hostname, port, etc.
The secrets itself are the only things that change and what we don't want to commit to git.

How about this?
`.env`
```javascript
LOCAL_AWS_ACCESS_KEY_ID=123
LOCAL_AWS_SECRET_ACCESS_KEY=123

PROD_AWS_ACCESS_KEY_ID=123
PROD_AWS_SECRET_ACCESS_KEY=123
```

`environments/local.bru`
```groovy
vars {
    AWS_ACCESS_KEY_ID = {{process.env.LOCAL_AWS_ACCESS_KEY_ID}}
    AWS_SECRET_ACCESS_KEY = {{process.env.LOCAL_AWS_SECRET_ACCESS_KEY}}
}
```

`environments/prod.bru`
```groovy
vars {
    AWS_ACCESS_KEY_ID = {{process.env.PROD_AWS_ACCESS_KEY_ID}}
    AWS_SECRET_ACCESS_KEY = {{process.env.PROD_AWS_SECRET_ACCESS_KEY}}
}
```

Related: https://github.com/usebruno/bruno/issues/122#issuecomment-1622854781

helloanoop:
@DaPutzy so true!

I have a day job as a tech lead. After my job, I have been pulling straight all-nighters for the last 7 days.
I try to get sleep for 4-6hrs
There is a release that goes out in the every single day for the past week 😂

But the community has been super helpful. They keep sending a lot of PR's and help improving Bruno.

And Thank you @DaPutzy , I see you have been commenting on tickets helping connect dots and trying to control the chaos!

helloanoop:
[...]

PR #2578, opened 2024-07-08
Feature/read multiple env files

minamz:
# Description
Enables loading multiple `<named>.env` files from the collection folder, instead of only the `.env` file. 
Use case: I have a project  which reads vars from multiple named .env files, and wanted Bruno to be able to access vars from the multiple files for experimentation/integration with the project.

Below images illustrate how variables are loaded to Bruno from 2 separate env files: .`env` and `localBru.env `.
Also the values are updated when files are changed and saved.
image
image
image
image

### Contribution Checklist:


gheiler-on:
I think in order to solve https://github.com/usebruno/bruno/issues/409 it should be possible to add a hierarchy, given you will have the same keys in different files, and you wan to be in control of which files should have preponderance over the other ones.
For that I don't see any other way more than adding a UI piece where you will select per environment which .env files to load on which sequence.
Just adding, in case you also want to support that issue, given I would be interested on it :)

minamz:
@gheiler-on thanks for the input; ~ 💚  i totally agree with that;  didn't really have a solution for duplicate keys as in my use case they were unique... 
ideally it would be possible to set a global env file, then set a specific env file for a collection; which would take priority over the global one, and so on.
sadly i'm not so much skilled on frontend at the moment so as to achieve that (but slowly learning ... 🥹  )

ElRige:
Hi, this feature would be very useful for my project. 
How about naming the files `.env.something` instead?
That's what's done on vuejs https://cli.vuejs.org/guide/mode-and-env.html#environment-variables or reactjs https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used

PR #2640, opened 2024-07-15
fix: active enviroment not updated after rename when there is single enviroment

therealrinku:
# Description
minor fix for when there is single environment and user renames it, active environment is not updated to the renamed environment which causes error when user tries to update or delete the active environment.

### Contribution Checklist:
- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ ] **Create an issue and link to the pull request.**

before:
https://github.com/user-attachments/assets/3da6d45b-88fa-436d-9b3e-817ee38ff58e

after:
https://github.com/user-attachments/assets/b8d4d9cd-fe4b-48f4-b0b7-7884cee9b79c

PR #2660, opened 2024-07-18
Revert "fix: active enviroment not updated after rename when there is single enviroment"

sanjai0py:
Reverts usebruno/bruno#2640

Cause it's causing this bug!

https://github.com/user-attachments/assets/e76db461-1fa5-46e1-8c08-597744016ae2


therealrinku:
@sanjai0py thanks for looking into that and sorry about that 😅

how about adding something like this, right after that if statement, 
it seems to fix the both issue, I tested here and looks good

if (prevEnvUids?.length === 1 && envUids?.length === 1) {
        setSelectedEnvironment(environments ? environments[0] : null);
  }

https://github.com/user-attachments/assets/67eb4770-95a1-4efd-b8d2-08a5d5ee8113

PR #2673, opened 2024-07-19
Feat/ Update node version used by project to match Electron

pietrygamat:
# Description
In this PR the node version advertised as used by the project is bumped to match what is packaged with previously updated electron. Also, `httpsnippet`  is updated, as it claims node 20 compatibiliy since version [v3.0.4](https://github.com/Kong/httpsnippet/releases/tag/v3.0.4).

fixes: #2710
relates to: #2668


pietrygamat:
@lohxt1 @sanjai0py could you take a look? It seems this should have been done as part of #2668.

PR #2724, opened 2024-07-30
fix: option to toggle on/off system proxy env variables

lohxt1:
system_proxy_checkbox

### DIFF
--- a/packages/bruno-app/src/components/CollectionSettings/ProxySettings/index.js
+++ b/packages/bruno-app/src/components/CollectionSettings/ProxySettings/index.js
@@ -7,17 +7,19 @@ import toast from 'react-hot-toast';
 import { IconEye, IconEyeOff } from '@tabler/icons';
 import { useState } from 'react';
 import { useMemo } from 'react';
+import { cloneDeep } from 'lodash';
 
 const ProxySettings = ({ proxyConfig: _proxyConfig, onUpdate }) => {
   const proxyConfig = useMemo(() => {
+    const proxyConfigCopy = cloneDeep(_proxyConfig);
     // backward compatibility check
-    if (_proxyConfig?.enabled == 'true') _proxyConfig.enabled = true;
-    if (_proxyConfig?.enabled == 'false') _proxyConfig.enabled = false;
+    if (proxyConfigCopy?.enabled == 'true') proxyConfigCopy.enabled = true;
+    if (proxyConfigCopy?.enabled == 'false') proxyConfigCopy.enabled = false;
 
-    _proxyConfig.enabled = ['string', 'boolean'].includes(typeof _proxyConfig?.enabled)
-      ? _proxyConfig?.enabled
+    proxyConfigCopy.enabled = ['string', 'boolean'].includes(typeof proxyConfigCopy?.enabled)
+      ? proxyConfigCopy?.enabled
       : 'global';
-    return _proxyConfig;
+    return proxyConfigCopy;
   }, [_proxyConfig]);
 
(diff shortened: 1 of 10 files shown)


byjrack:
So initial results are positive

If I have env:https_proxy configured and call to the proxy i am not getting a 400

If I only  have env:https defined, but not env:http (or vice a versa) and System:enabled I am getting an error that probably should be handled. 

`Error invoking remote method 'send-http-request': Error: Invalid system http_proxy`
`Error invoking remote method 'send-http-request': Error: Invalid system https_proxy`

There is a non-zero chance someone may only have one of the two configured and the app should be able to handle that.

If you have App and System enabled System seems to win.  If i enabled App and put a junk proxy name in there things still seemed to work, but my test machine is doing some other lab testing that may be influencing the outcome.  Not sure if these should be "manual" OR System using a radio button vs the Global:enabled System:enabled at the same time.

Going to try and test on a machine that may be a bit closer to standard on proxy.

byjrack:
Another thought is how/if the Auth section influences the use of System Proxy.

If Auth only applies to Global section than def would be better to disable the other so that folks know what configuration is active.

byjrack:
The Or case seems to work with System enabled Global is locked.  I might say the UX is a bit unclear as the Global enabled check looks different, but the rest of the fields look the same and are just readonly.  The Global remains "enabled" and content visible so folks might be confused on what is active at this moment.

I still get the error where is System is enabled and i only have one of the envvars populated (only http and not https or vice versa) fetch tries to use both. I would see that as a somewhat common occurrence so probably some additional logic when System config is read.

byjrack:
And just retested and no more error when only one of the system vars is set.  

So on the surface I think the patch works.  Seems to avoid axios electing the wrong client and trying to do a TLS handshake instead of a CONNECT.

byjrack:
@lohxt1 where are we at with this getting merged?

helloanoop:
@and0p This is really helpful. Using mode with options like on/off/system proxy feels much cleaner and more intuitive from a UX perspective.

We're nearly finished with the PR implementation and have already tested it with @byjrack. @lohxt1 will update the UX as part of this PR.

helloanoop:
We did some further improvements over this PR
See https://github.com/usebruno/bruno/pull/2977#issue-2496433030

All changes are now in the main branch.
This is schedule to go out in the upcoming release on Tue 3 Aug, 2024

helloanoop:
@byjrack It'd be great if you could test the changes in the main branch.
Please let us know if you find any issues

byjrack:
Thus far looking good.  Same general experience that I had w @lohxt1's fork.

Technically this implementation is more "Environment" than "System" in my mind.  For System I would say you would use [ProxyConfig](https://www.electronjs.org/docs/latest/api/structures/proxy-config) and then build logic around that.  That class grabs the OS globally scoped proxy settings rather than just the process env.  Bit of semantics I get it and the help as well as the actual implementation show folks that "System = use what is in these environment variables" so wouldn't worry about it.

Goal should be to allow users to avoid the axios assumption around https_proxy. Maybe long term getting back to the in-box fetch to simplify the supply chain would be a good goal, but a much bigger effort than this one.

helloanoop:
@byjrack This is now available in `v1.28.0` 🥳 🎉

byjrack:
Yup already have folks testing.  Had one report where disabling TLS using System Proxy still errored on trust to our internal CA.  I can't repro and using the Custom CA I think should avoid it as well. Just trying to track down root cause before opening a new Issue.

Issue #1753, opened 2024-03-11
BigInteger on response 


MeskaKeyne:
When recieve response , BigInteger is not parse correctly
thanks for help :-)
EDIT: sorry for my mistake on title

Its-treason:
We recently merged a fix for BigInt that is not yet released: #1710 that might fix your issue. You can either wait for the next Bruno release, clone the repository and start Bruno locally or [try a prebuild binary from my fork](https://github.com/Its-treason/bruno/releases/tag/nightly) it includes the BigInt fix.

MeskaKeyne:
Many thanks

Raffine64:
json response:
{
  "id": 1.0000000000001147e+21
}

still same issue, tested with Bruno v1.17.0

azzamsa:
@Its-treason I think it is not included in the latest release. Or Am I missing something?
Bruno still does not get it right.

Bruno:

![image](https://github.com/user-attachments/assets/9c8fdc90-e830-4f44-a94d-3c07c54c1cec)


Insomnia

![image](https://github.com/user-attachments/assets/620f919a-60d7-4ff7-a1d0-2ffe64b0d285)

Raffine64:
@azzamsa, problem still remains, tested with Bruno v1.22.0:

{
    "id": 1.0000000000000325e+21,

@Its-treason can you see that it's getting fixed ?

It's very annoying in our day-to-day operations.

Its-treason:
> @azzamsa, problem still remains, tested with Bruno v1.22.0:
> 
> { "id": 1.0000000000000325e+21,
> 
> @Its-treason can you see that it's getting fixed ?
> 
> It's very annoying in our day-to-day operations.

Yes, is just looked into it and found a fix. I need to do some testing and will submit a PR later

helloanoop:
PR https://github.com/usebruno/bruno/pull/2773 has been merged, This will go out in the v1.27.0 release.

Raffine64:
Hello,

I still have the same issue in responses.

tested with Bruno 1.30.1
{
  "id": 1.0000000000000104e+21,
...
}

Raffine64:
thanks for your answer @Its-treason.

but the problem is that now my responses are not pretty printed (not formatted).

Issue #2438, opened 2024-06-11
Bruno converts decimal numbers in JSON body to strings


mrtnbrst:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Bruno converts decimal numbers in JSON to strings as visible in the screenshot.

### .bru file to reproduce the bug

[example.json](https://github.com/user-attachments/files/15788467/example.json)

### Screenshots/Live demo link

![image](https://github.com/usebruno/bruno/assets/54993850/c9636822-6ec0-40e3-9e58-e3d13c2efc35)

mrtnbrst:
The attached file is a .bru file, but those are not allowed to be uploaded, so I changed the file extension.

jwetzell:
Seems to be an issue with "long" decimals as not all of the decimals in your example got converted to strings.

jwetzell:
The problem is [this line here](https://github.com/usebruno/bruno/blob/3db7d54acad3cccbe8282a2467302824f7241c23/packages/bruno-electron/src/ipc/network/prepare-request.js#L176) for someone looking to fix this. It is using a library to parse the JSON but that library is converting these long decimals into "BigNumber" type. @sanjai0py is the one who added this BigNumber parsing in so might be able to speak to the reasoning or if this side-effect was not known.

helloanoop:
@sanjai0py Can you comment on this?

helloanoop:
This has been fixed as a part of PR https://github.com/usebruno/bruno/pull/2773, and will go out in the v1.27.0 release.

PR #2736, opened 2024-07-31
Fix/bigint in response

Its-treason:
# Description

Use `lossless-json` to parse the response data and then later stringily so information about BigInts are not lost.

Fixes: #1753

### DIFF
--- a/packages/bruno-app/src/utils/common/index.js
+++ b/packages/bruno-app/src/utils/common/index.js
@@ -1,5 +1,6 @@
 import { customAlphabet } from 'nanoid';
 import xmlFormat from 'xml-formatter';
+import { stringify } from 'lossless-json';
 
 // a customized version of nanoid without using _ and -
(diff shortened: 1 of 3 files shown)


jwetzell:
Would be nice to throw it on the request side to to solve #2438

Its-treason:
> Would be nice to throw it on the request side to to solve #2438 as well

It looks like this is already working

image

jwetzell:
I am not seeing that behavior, the latest version of bruno is still converting large numbers in JSON body to strings.

helloanoop:
Merged! Thanks for taking care of this @Its-treason !

helloanoop:
Hey @Its-treason 

I reverted this merge since we are planning a release today.
I created https://github.com/usebruno/bruno/issues/2767 to discuss the lossless-json stuff

It appears we may have to fork axios.

Issue #2747, opened 2024-08-02
Add a show/hide privacy toggle to passwords and secrets in Auth options


nataliecarey:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

**User story**

As a user I want my authentication password/secrets to be hidden by default and visible if I choose to reveal them.

**The use case/reason**

I work on a system where we have to be pretty careful with credentials, I was screen sharing with other developers while trying to make an authenticated API call from Bruno but I couldn't have the password on the screen share at any time.  I ended up setting up the request without the password, disconnecting the screen share, entering the password, making the request, removing the password, starting the screen share again.  I had to do that cycle multiple times which got quite frustrating.

**Related PRs**

 - [Adding a privacy toggle to proxy settings](https://github.com/usebruno/bruno/pull/2127) - the same privacy toggle should be used
 - [Masking support for SingleLIneEditor](https://github.com/usebruno/bruno/pull/2240) - this isn't merged yet but it implements the functionality needed on the element that's used for Auth settings

### Mockups or Images of the feature

<img width="361" alt="Screenshot 2024-08-02 at 10 02 09" src="https://github.com/user-attachments/assets/0bc8cafa-7d11-4f8a-ae5d-ae9990728292">
<img width="368" alt="Screenshot 2024-08-02 at 10 02 16" src="https://github.com/user-attachments/assets/96de28e7-c118-4bea-98aa-320e1bd7fe7a">

sanjai0py:
Closing this issue as the [PR](https://github.com/usebruno/bruno/pull/2750) by you that resolves this issue has been merged into Bruno. Thank you for taking the time to work on this!

katzda01:
In my use case we use different environment variables, and because of this change, it made my life unnecessarily tedious, as I keep having to click on reveal just to see which environment placeholder is used, in order to run the required SSO bearer token request.

<img width="456" height="79" alt="Image" src="https://github.com/user-attachments/assets/3f18c4e1-bdbf-4847-9f4b-15b27a59dba0" />

It would be far better if the default show/hide were set in the configuration so that people who would like this shown by default can have that.

PR #2240, opened 2024-05-04
feat: masking support for SingleLineEditor

krummbar:
# Description

Fixes #2217

This change adds masking capabilities to a `SingleLineEditor`

**Issue**
As of now, an environment variable becomes read-only once it is flagged as secret. This is due to the fact, that a div overlay is rendered instead of an input field. In order to change a secret value, one has to unsecret it, edit it, and flag as secret again. This could cause secrets leaks during screen sharing sessions whenever a environment secret needs to be changed.

**Solution**
Now it uses the SingleLineEditor even if the variable is marked as secret, but each **rendered** character is altered to a `*`. The actual value is preserved. The masking functionality is only activated and used if `props.maskInput=true` for a SingleLineEditor, otherwise the existing behaviour is preserved.

**New environment settings behaviour**

https://github.com/usebruno/bruno/assets/1665841/8926e8e9-ce5f-470b-85c4-83b21efe61bd

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

PR #2750, opened 2024-08-02
Feature: Add a show/hide privacy toggle to passwords and secrets in Auth options

nataliecarey:
Relates to https://github.com/usebruno/bruno/issues/2747

# Description

**User story**

As a user I want my authentication password/secrets to be hidden by default and visible if I choose to reveal them.

**The use case/reason**

I work on a system where we have to be pretty careful with credentials, I was screen sharing with other developers while trying to make an authenticated API call from Bruno but I couldn't have the password on the screen share at any time. I ended up setting up the request without the password, disconnecting the screen share, entering the password, making the request, removing the password, starting the screen share again. I had to do that cycle multiple times which got quite frustrating.

Related PRs

- https://github.com/usebruno/bruno/pull/2127 - the same privacy toggle is used
- https://github.com/usebruno/bruno/pull/2240 - I've built my solution on top of this branch because the work that @krummbar has done really enables this work.  I would highly recommend merging that PR before this one.

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.** (note: this will be reflected better in the changed files after #2240 is merged)

**Screenshots**

Screenshot 2024-08-02 at 18 24 35
Screenshot 2024-08-02 at 18 24 22
Screenshot 2024-08-02 at 18 24 03
Screenshot 2024-08-02 at 18 23 45
Screenshot 2024-08-02 at 18 23 35
Screenshot 2024-08-02 at 18 23 29
Screenshot 2024-08-02 at 18 23 21


katzda01:
In my use case, I do not need to be careful because the password always contains a `{{ placeholder }}` variable. This is being set automatically in a different request that fetches a temporary bearer token so the "secret" wouldn't have even lasted for long to be exploited, even if it were visible through the {{ placeholder }}.

Placeholders are not secrets but now are treated as such, making it tedious to find out which bearer token request I'm supposed to run. It would be nice if this "secret" was an optional toggle in the settings to "Hide password by default".

I feel that the original request, while a valid point, could've at least included an option to undo this in the settings.

Issue #2712, opened 2024-07-26
BUG: UI is getting distorted while opening the Application.


Pragadesh-45:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

While opening the application, the user sees a distorted UI during the loading time.

### .bru file to reproduce the bug

Not Needed.

### Screenshots/Live demo link

Attached is the screenshot for your reference:

![image](https://github.com/user-attachments/assets/cda0492a-662d-4630-9ec6-848c825bbdfd)

PR #2765, opened 2024-08-06
Bugfix/window UI distortion electron

Pragadesh-45:
# Description

There was a small issue with the Windows version (maybe in Linux too) where, when we initially started the application, it showed a broken window. This has now been fixed by setting the window's show property to false during initialization, so it gets directly loaded.

It fixes: [#2712](https://github.com/usebruno/bruno/issues/2712)

### Contribution Checklist:

Before:

https://github.com/user-attachments/assets/85225ab6-a132-469a-8ec9-6c9a8465162e

After:

https://github.com/user-attachments/assets/e9c63e1f-4a8a-48b3-91cc-6f3ceca28086

### DIFF
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -50,7 +50,7 @@ app.on('ready', async () => {
     height,
     minWidth: 1000,
     minHeight: 640,
-    show: true,
+    show: false,
     webPreferences: {
       nodeIntegration: true,
       contextIsolation: true,
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -68,6 +68,7 @@ app.on('ready', async () => {
     mainWindow.maximize();
   }
 
+  mainWindow.show();
   const url = isDev
     ? 'http://localhost:3000'
     : format({


packages/bruno-electron/src/index.js
```diff
@@ -68,6 +68,7 @@ app.on('ready', async () => {
     mainWindow.maximize();
   }
 
+  mainWindow.show();
```
Its-treason: You Should probably listen for the `ready=to-show` Event of the mainWindow here and then call mainWindow.show(). The Ready Event is fired, when the content / side is loaded. See: https://www.electronjs.org/docs/latest/api/browser-window#showing-the-window-gracefully

Pragadesh-45: > You Should probably listen for the `ready` Event of the mainWindow here and then call mainWindow.show(). The Ready Event is fired, when the content / side is loaded

Yeah, I tried listening to the `ready-to-show` event, and I noticed that it was a proper approach as well. However, after implementing that method, the app's initial load time seems like increased (just visually). So, I decided to stick with the current implementation to mimic the faster load time. Thanks for the review; I'll update the PR.

Using `ready-to-show` event:

https://github.com/user-attachments/assets/db8f1d26-1541-41dd-b1c0-907669ad780c

helloanoop: Agreed that `ready=to-show` is the technically correct solution.
The additional lag seems to be introduced by the time taken to render the first page

Approving this for now as it seems to solve for the distortion seen in windows

helloanoop:
Merged. Thanks for taking care of this @Pragadesh-45 !

Issue #2767, opened 2024-08-06
Lossless Serialization of Decimals and Large Numbers in JSON


helloanoop:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

There are a number of issues related to
* Not able to handle large numbers
* Decimal numbers being converted as strings

We tried using [json-bigint](https://www.npmjs.com/package/json-bigint), but that didn't help.
We then tried using [lossless-json](https://www.npmjs.com/package/lossless-json), but that also failed at the axios layer with the below message

> Error occurred in handler for 'send-http-request': TypeError: Do not know how to serialize a BigInt
    at stringify (<anonymous>)
    at stringifySafely (/Users/anoop/bruno/bruno/node_modules/axios/dist/node/axios.cjs:1436:37)
    at Object.transformRequest (/Users/anoop/bruno/bruno/node_modules/axios/dist/node/axios.cjs:1496:14)
    at transform (/Users/anoop/bruno/bruno/node_modules/axios/dist/node/axios.cjs:1922:15)
    at Object.forEach (/Users/anoop/bruno/bruno/node_modules/axios/dist/node/axios.cjs:283:10)
    at Object.transformData (/Users/anoop/bruno/bruno/node_modules/axios/dist/node/axios.cjs:1921:11

### Related Issues:
* https://github.com/usebruno/bruno/issues/1753
* https://github.com/usebruno/bruno/issues/1768
* https://github.com/usebruno/bruno/issues/2376
* https://github.com/usebruno/bruno/issues/2438
* https://github.com/usebruno/bruno/issues/2043

### Related PRs:
* https://github.com/usebruno/bruno/pull/2006
* https://github.com/usebruno/bruno/pull/2236
* https://github.com/usebruno/bruno/pull/2382
* https://github.com/usebruno/bruno/pull/2736
* https://github.com/usebruno/bruno/pull/2449

### Solution
Do not compromise the fidelity of the Json string (by doing JSON parse and stringify) before the network request is made

helloanoop:
PR #2773 has been merged, This will go out in the `v1.27.0` release.

thomascaillier:
Hello,
There is still a bug in v1.34.1 where a too long value from API is displayed rounded.
Example : 1429599376132020**638** is displayed 1429599376132020**700**
However, when saving response to file (top right button), the .json contains the right value

helloanoop:
[...]

PR #2773, opened 2024-08-07
fix(#2767): Fix serilization issues of bigint in json body

helloanoop:
Fixes: #2767


helloanoop:
The serialization is now lossless. Just need to fix the deserialization next by providing a toggle at collection level setting to disable parsing response as json (when the content type is application/json)

image

packages/bruno-app/src/components/ResponsePane/Timeline/index.js
```diff
@@ -33,9 +31,9 @@ const Timeline = ({ request, response }) => {
           );
         })}
 
-        {requestData ? (
+        {request.data ? (
           <pre className="line request">
-            <span className="arrow">{'>'}</span> data {requestData}
+            <span className="arrow">{'>'}</span> data <pre>{request.data}</pre>
```
Its-treason: This causes a horizontal overflow, if the text is too long

packages/bruno-js/src/bruno-request.js
```diff
@@ -56,11 +56,31 @@ class BrunoRequest {
...
+
+    if (hasJSONContentType && !options.raw) {
+      return JSON.parse(this.req.data);
+    }
+
     return this.req.data;
   }
 
   setBody(data) {
+    if (typeof data === 'object') {
```
Its-treason: This should probably also check if the Object is a "plain" object with `data.constructor === Object` (or something else). Setting a FormData object as the body does not work now, but worked previously.

packages/bruno-js/src/bruno-request.js
```diff
@@ -56,11 +56,31 @@ class BrunoRequest {
...
 
-  getBody() {
+  /**
+   * Get the body of the request
+   * 
+   * @param {*} options 
+   * @param {boolean} options.raw - If true, return the raw body without parsing it
+   * @returns 
+   */
+  getBody(options = {}) {
```
Its-treason, line 97: What about `req.body` that is set inside the constructor? I think this should also be JSON parsed like req.getBody

packages/bruno-js/src/bruno-request.js
helloanoop: This makes sense. Will have this updated.

helloanoop: So I have updated the logic to also check if content type is json (based on headers).
This should cover the use case of people setting body as FormData object.

If someone has set the Content-Type as json in the header (or it gets auto added if they have chosen JSON as body in UI), and then they try to pass a non-plain object while invoking `setBody()` - they are essentially shooting themselves in the foot and we don't need to gracefully handle that.

The reason why are supporting the `raw` option is to support use cases where people want to still have the content type as json, but want to pass the json strings themselves instead of passing the json object

@Its-treason lmk if you differ or have further suggestions

```javascript
setBody(data, options = {}) {
    if (options.raw) {
      this.req.data = data;
      this.body = data;
      return;
    }

    const isJson = hasJSONContentType(this.req.headers);
    if (isJson && this.__isObject(data)) {
      this.body = data;
      this.req.data = this.__safeStringifyJSON(data);
      return;
    }

    this.req.data = data;
    this.body = data;
  }
```

packages/bruno-js/src/bruno-request.js
helloanoop: 💯 good catch (this could have resulted in breakages for folks who were accessing `req.body` in scripts)

Have updated the code with a comment

```javascript
constructor(req) {
    // existing code

    /**
     * We automatically parse the JSON body if the content type is JSON
     * This is to make it easier for the user to access the body directly
     * 
     * It must be noted that the request data is always a string and is what gets sent over the network
     * If the user wants to access the raw data, they can use getBody({raw: true}) method 
     */
    const isJson = hasJSONContentType(this.req.headers);
    if (isJson) {
      this.body = this.__safeParseJSON(req.data);
    }
  }
```

packages/bruno-app/src/components/ResponsePane/Timeline/index.js
helloanoop: Nice catch. Fixed

helloanoop:
@Its-treason I have addressed all your comments and have also made some updates. Let me know if you have further comments

@lohxt1 will take over from here and add further changes on this PR to add a toggle in `bruno.json` to support disabling  the behaviour of bruno automatically parsing response as json (if the response a valid json). This is aimed at solving the problem around the response text losing its fidelity during the JSON.parse operation

`bruno.json`
```
{
    parseResponseJson: true
}
```

helloanoop:
Another option that I am think to solve problems for folks who don’t want us to to automatically parse response as json (where the response loses fidelity in some cases) is to provide an api that can disable this.

```
req.disableParsingResponseJson()
```

They can use this then at request level scripting, or apply it to folder level or collection level scripts as they see fit.

We can consider providing a UI toggle much later if there is demand for it.

helloanoop:
The PR is GTG from my side.

Folks who don’t want us to to automatically parse response as json (where the response text loses fidelity in some cases) can use the new api method `req.disableParsingResponseJson()`

They can use this at request level scripting, or apply it to folder level or collection level scripts as they see fit.

@Its-treason @jwetzell let me know if you have any further comments. I plan to have this merged tomorrow.

MeskaKeyne:
hello, fix seems not working in 1.27.0

eg
{
"id": 1.0000000000000493e+21,
    "dateFrom": "2019-08-08",
}

PR #2782, opened 2024-08-08
disable response json parse flag

lohxt1:


packages/bruno-js/src/bruno-request.js
```diff
@@ -167,6 +167,10 @@ class BrunoRequest {
   __isObject(obj) {
     return obj !== null && typeof obj === 'object';
   }
+  
+  disableResponseJsonParse() {
+    this.req.__brunoDisableResponseJsonParse = true;
```
helloanoop: I recommend renaming this as `.__brunoDisableParsingResponseJson`

PR #2791, opened 2024-08-09
updated headers, vars merge functions

lohxt1:
~ removed the `i.uid === request.uid` check which is always false
~ updated the request headers & vars merge logic


helloanoop:
Closing this as this will be addressed as a part collection variables PR

Issue #2806, opened 2024-08-11
search api


CPU-IDA:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

I hope URI can be searched here, thank you

### Mockups or Images of the feature

<img width="676" alt="无标题" src="https://github.com/user-attachments/assets/41d759b9-c11e-4ae2-b2b9-0772631cc9a7">

[JIRA](https://usebruno.atlassian.net/browse/BRU-1199)

N-Ziermann:
I've just taken a look at the implementation and it seems to be simple to implement.
I would be open to work on this 😄

sreelakshmi-bruno:
This feature is now available via `cmd+k`/`cntrl+k` search. Closing the issue since it's been resolved. Thanks!

PR #1230, opened 2023-12-16
feat(#1229) Include folder names in search

BrandonMurch:
# Description

When the folder name matches the search query, it should be displayed. 
If the folder is displayed, all children should also be displayed. 
This feature can be toggled on/off by a folder button in the search bar.

Screenshot 2023-12-16 at 12 23 14 pm
Screenshot 2023-12-16 at 12 23 26 pm
Screenshot 2023-12-16 at 12 24 25 pm

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

PR #2808, opened 2024-08-11
Feature/filter by url

N-Ziermann:
# Description

This PR aims to resolve this Issue: https://github.com/usebruno/bruno/issues/2806.
It enables users to not only filter their request by name, but also by URL.

Screenshot 2024-08-11 at 15 57 22

I also added unit tests for the function I edited and therefore needed to add a custom _jest-config_ to the _bruno-app package_ that uses the _compilerOptions.paths_ property from the _jsconfig_.

This was necessary because jest doesn't automatically read that property from the _jsconfig_ file and therefore couldn't understand the _utils/common_ imports from the _collections/index_ file used in _search.js_.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/jest.config.js
+++ b/packages/bruno-app/jest.config.js
@@ -1,0 +1,10 @@
+const { compilerOptions } = require('./jsconfig.json');
+const { pathsToModuleNameMapper } = require('ts-jest');
+
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  modulePaths: [compilerOptions.baseUrl],
+  moduleNameMapper: pathsToModuleNameMapper(compilerOptions.paths),
+  transform: { '^.+\\.(js|jsx)$': 'babel-jest' }
+};
--- a/packages/bruno-app/src/utils/collections/search.js
+++ b/packages/bruno-app/src/utils/collections/search.js
@@ -3,7 +3,10 @@ import filter from 'lodash/filter';
 import find from 'lodash/find';
 
 export const doesRequestMatchSearchText = (request, searchText = '') => {
-  return request.name.toLowerCase().includes(searchText.toLowerCase());
+  return (
+    request.name.toLowerCase().includes(searchText.toLowerCase()) ||
+    request.request.url.toLowerCase().includes(searchText.toLowerCase())
+  );
 };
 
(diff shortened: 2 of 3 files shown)


helloanoop:
Hey @N-Ziermann Thanks for working on this.

One issue with this approach is that the search results displayed do not indicate the context for why they were matched. This becomes problematic when there are many collections loaded with a variety of requests.

N-Ziermann:
Hey @helloanoop Thanks for your reply.

I understand your concern.
I'd suggest we highlight the matching part of the request name using \<mark\> tags if the name matches
and maybe turn the outline of the request path yellow (same yellow as when searching for something in the browser) if the path matches (to prevent issues with the syntax highlighting).

What do you think?

tlaloc911:
There is another unmerges PR related to filtering #1230 , to allow folder name search, and it added a button to alternate between regular and by folder search, you culd add something similar

N-Ziermann:
Thanks for pointing that out @tlaloc911 
I'm afraid a toggle between different filters might not scale well. Having more than 2 options (as 2 are already described in the PR you just linked) would need a lot of space and might make it even more confusing at some point.

What I could see ad an option is a filter (kebab menu) to control which filters you want to have active combined with the indicator I described previously.
This would allow users to completely turn off filters they dont like whilst still seeing which attribute caused the filtered value to be visible

N-Ziermann:
bruno mockup

Here is a little illustration of what I mean. It's of course not to scale and the colors aren't ideal, but it should get the point across.
If this is something that would improve this feature for you @helloanoop then I would love to implement it (with proper scaling and better looking colors of course 😉 )

helloanoop:
@tlaloc911  @N-Ziermann I am thinking of a collection level search that not just filters by url, but much more like search inside body, scripts, tests etc

Let me know your thoughts.

Here is a draft PR I have on this - #3823

Demo

https://github.com/user-attachments/assets/5778b1f0-9891-4dbf-90f0-bcc53ebb43ae

ghandic:
This would be super useful - any updates?

sreelakshmi-bruno:
This functionality is available in Bruno via the `cmd+k` option. We are also planning an enhancement around filtering which is being tracked in #6283.

helloanoop:
Closing this PR as we currently don't agree with the design proposed here.

Issue #2682, opened 2024-07-22
Cookie Management does not work while running whole folder/collection


mosotu88:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

**If I run all requests of a folder, the cookie management does not work same as when I run the requests one by one**

**Steps to reproduce with a simple example:**

1. Delete all cookies in bottom left cookie menu.
2. Add two requests to a folder. Example: GET https://github.com
3. Run folder

**Expected:**
Cookies are provided in request headers when running folders
Example:
First response will set cookie (in example: response header includes "< set-cookie: _gh_sess=..." )
Second response will not set a cookie as it should already be provided

**Actual.**
Cookies are not provided when running folders
Example:
First response of GET https://github.com sets session cookie (in example: response header includes "< set-cookie: _gh_sess=..." )
Second response of GET https://github.com also sets session cookie, eventhough it was already set before (in example: response header includes "< set-cookie: _gh_sess=..." )

If I run the requests one by one instead of running the folder, the cookie management works as expected. 

### .bru file to reproduce the bug

Attached example collection with example requests and a test in the second
[Cookie Folder Run Test.zip](https://github.com/user-attachments/files/16334986/Cookie.Folder.Run.Test.zip)

### Screenshots/Live demo link

see provided bruno collection in attachments to reproduce the bug.

sanjai0py:
Hey @mosotu88, thank you for taking the time to report this issue. It looks like there may be a bug. I will investigate and get back to you!

helloanoop:
Just merged the PR - #2951 
This will be available in `v1.28.0`

Issue #1332, opened 2024-01-05
[BUG] Cookie not saved on a HTTP 302 response


nico-m-64:
Hello !

> *Best wishes for the new year, wishes for a great future of this great project 🎉 !*

I observe a behaviour I suppose wrong, here is the scenario :

- in Bruno, I configure a request with some headers to authenticate on my server
- the request is sent, the server checks, and if OK, it sets a cookie, and answers with a 302 http code to redirect elsewhere on the same domain
- Bruno receives the response 302 and follows it, but the cookie that comes with this response is not stored and re-sent when redirect

**Here is what Bruno do** : (follows redirection in background and displays the final 200, with no cookie)

<img width="399" alt="Capture d’écran 2024-01-05 à 14 08 43" src="https://github.com/usebruno/bruno/assets/91060911/10b34ced-6b30-4ead-bca5-da3e427c7df9">

**Here is what it should be done**, I suppose : (store the cookie sent with the 302 answer, and then put it in the redirection call)

![Capture d’écran 2024-01-05 à 14 58 51](https://github.com/usebruno/bruno/assets/91060911/0decaf77-b0f8-42dd-960b-1fe5a3239cdb)

trevex:
Just tested bruno for a personal use-case as well and can confirm I ran into the same issue.

the-eric-kwok:
I'm facing the same issue

j-lebek:
Same issue for HTTP 301 (e.g. bit.ly links).

j-lebek:
When a redirect (HTTP 301 or 302) is received, Bruno follows it automatically (unless you specify `req.setMaxRedirects(0)` in the _Pre Request_ script, in which case Bruno stores the cookies of the initial response (unless run in a Runner, see #2950).

A workaround for you, @nico-m-64, might be to explicitly stop on the redirect response using `setMaxRedirects`in the _Pre Request_ script, use the _Post Request_ script or tests to store the redirect location (response header `location`) in a variable and manually run a subsequent request, which takes the target URL from the variable set before. That way the cookie will be stored and used on eligible requests.

There is more incorrect behavior in Bruno when it comes to cookies and redirects:
If a redirect is followed (default behavior), then cookies are stored neither for the initial response nor for the final response after following to the redirect location.

Example:
https://bit.ly/3AHXx8s is a redirect to https://www.cnn.com
Both locations set their own cookies (and in fact https://www.cnn.com wants to redirect again, which I suppress in this example for brevity).
```shell
[...]

PR #2820, opened 2024-08-13
Added cookie support to CLI requests

matthewdickinson:
# Description

Right now, Bruno CLI doesn't support cookies. Many of our collections include an "Auth" request that we can use to get a auth cookie that will be used on subsequent requests. However, without cookies, we can't run the same collections (e.g. treat them as a test suite) without adding complicated pre-request scripts to fake the cookie first.

This adds a `--use-cookies` flag that (if set to `true`) creates a cookie store for this invocation of `bru`. Hence, no existing use-cases should be impacted, but this feature would be available for those who wish to use it.

This also should work in both standard and Proxy (HTTP or SOCKS) configurations.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-cli/package.json
+++ b/packages/bruno-cli/package.json
@@ -35,6 +35,7 @@
     "decomment": "^0.9.5",
     "form-data": "^4.0.0",
     "fs-extra": "^10.1.0",
+    "http-cookie-agent": "^6.0.5",
     "http-proxy-agent": "^7.0.0",
     "https-proxy-agent": "^7.0.2",
     "inquirer": "^9.1.4",
@@ -43,6 +44,7 @@
     "mustache": "^4.2.0",
     "qs": "^6.11.0",
     "socks-proxy-agent": "^8.0.2",
+    "tough-cookie": "^4.1.4",
     "vm2": "^3.9.13",
     "xmlbuilder": "^15.1.1",
     "yargs": "^17.6.2"
--- a/packages/bruno-cli/src/commands/run.js
+++ b/packages/bruno-cli/src/commands/run.js
@@ -207,6 +207,11 @@ const builder = async (yargs) => {
       description:
         'The specified custom CA certificate (--cacert) will be used exclusively and the default truststore is ignored, if this option is specified. Evaluated in combination with "--cacert" only.'
     })
+    .option('use-cookies', {
+      type: 'boolean',
+      default: false,
+      description: 'Automatically save and sent cookies with requests'
+    })
     .option('env', {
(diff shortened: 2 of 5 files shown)


nikischin:
Fully agree with @helloanoop

However, Please keep in mind that there are still bugs in our current cookie implementation in the app, just like #2682, #2950 or #1332 so I suppose while migrating the current app implementation to CLI, you might first want to address those possibly or as least beware that those functions are currently not working as expected and therefore also will not when migrated to CLI.

matthewdickinson:
I will see what I can do to change the implementation to be closer to the GUI's implementation. However, I only have time to look at this every few weeks, but I might not get back to it until the end of the month.

Thanks!

packages/bruno-cli/src/utils/cookies.js
matthewdickinson, line 1: This is (almost) an exact copy of the util in the electron app, but I'm not sure where the mainainers would like to centralize this. `bruno-common` seems like it would make sense, but right now it doesn't seem to have any dependencies this this would require.

packages/bruno-cli/src/utils/cookies.js
```diff
@@ -0,0 +1,100 @@
...
+    setCookieHeaders = Array.isArray(headers['set-cookie'])
+      ? headers['set-cookie']
+      : [headers['set-cookie']];
+    for (let setCookieHeader of setCookieHeaders) {
+      if (typeof setCookieHeader === 'string' && setCookieHeader.length) {
+        addCookieToJar(setCookieHeader, url);
+      }
+    }
+  }
+}
```
matthewdickinson, line 91: I coped this from `packages/bruno-electron/src/ipc/network/index.js` since it seemed like it could be centralized to this util once they both used the same util.

matthewdickinson (commented):
@helloanoop @nikischin Would you be willing to look at this again? I switched it to use the (exact) same logic as the Electron app. 

It seems like the util could be centralized, but I left a question for you about that.

nikischin:
> @helloanoop @nikischin Would you be willing to look at this again? I switched it to use the (exact) same logic as the Electron app.
> 
> It seems like the util could be centralized, but I left a question for you about that.

Hi @matthewdickinson, thank you so much, I really appreciate that! I would be willing to test the cli, however, I haven't used the cli before and also I have no idea how to run the cli version of this branch and not the one installed with the latest version.

I would definitely benefit from this PR though, as if I want to use the cli in the future, I will definitely need this feature. Hope it gets merged soon! @lohxt1 is assigned for review :)

PR #2915, opened 2024-08-24
feat: Replace dictonary i18n with react-i18next

Its-treason:
# Description

Replaces the dictionary / custom i18n implementation with the well-known library `react-i18next` as shortly discussed in the community call on Wednesday.


helloanoop:
Merged! Thanks @Its-treason !

PR #3024, opened 2024-09-04
bugfix/Rename item with same name

### LINKED ISSUE #2995: Chaning only uppercase/lowercase in a file results in an error.

### Problem
Renaming a request fails if you only change uppercase/lowercase since a request with the same name already exists. This makes sence since we can not have the same file name twice inside the folder. As a user I would however expect the programm to be able to rename the file if I made a typo (f.ex if I want to rename myREquest to myRequest)

### Current Solution
When making a Caps typo the only solution is to rename it to another name and rename it again which is anoying. (myREQuest -> myRequest1 -> myRequest)

### Proposed Solution
When editing a file if the file exists but is the same file renaming it to the same file name (with different upper and lower cases) should be fine. 

_No response_

image

panda7789:
# Description
Fixes #2995 
In windows(case insensitive OS) when first is written into new path and than old deleted, windows rewrites same file and than deletes it. 
So when order of operations is swapped, old file is always deleted and than new is created with new name.

### Contribution Checklist:

https://github.com/user-attachments/assets/3873cc92-4550-4a65-bcf3-a94038f2e4db

### DIFF
--- a/packages/bruno-electron/src/ipc/collection.js
+++ b/packages/bruno-electron/src/ipc/collection.js
@@ -335,12 +335,15 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
       if (isWSLPath(newPath)) {
         newPath = normalizeWslPath(newPath);
       }
-
-      if (!fs.existsSync(oldPath)) {
-        throw new Error(`path: ${oldPath} does not exist`);
-      }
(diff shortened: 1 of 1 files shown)


helloanoop:
This issue has been fixed by in PR https://github.com/usebruno/bruno/pull/3171 and will be released in the upcoming v1.32.0 release next week.
Closing this PR.

Issue #3033, opened 2024-09-05
OAuth 2.0 auth Get Access Token button results in errors when state is passed in


oleg-codaio:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

When authenticating to the Dropbox API (https://www.dropbox.com/developers/documentation/http/documentation), using OAuth 2.0 authentication and hitting "Get Access Token" results in an `invalid_request` error due to `unknown field \"state\"`.

### Screenshots/Live demo link

<img width="322" alt="Screenshot 2024-09-05 at 12 53 29 AM" src="https://github.com/user-attachments/assets/e03fb150-4de6-4ade-abbb-b4320c8919bc">

PR #3034, opened 2024-09-05
Don't include state in OAuth 2 exchange flow

oleg-codaio:
# Description

Some APIs don't like extraneous properties such as state. Avoiding passing that in unless it's set to a non-empty value.

Fixes https://github.com/usebruno/bruno/issues/3033.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/oauth2-helper.js
+++ b/packages/bruno-electron/src/ipc/network/oauth2-helper.js
@@ -23,7 +23,7 @@ const resolveOAuth2AuthorizationCodeAccessToken = async (request, collectionUid)
   let requestCopy = cloneDeep(request);
   const { authorizationCode } = await getOAuth2AuthorizationCode(requestCopy, codeChallenge, collectionUid);
   const oAuth = get(requestCopy, 'oauth2', {});
-  const { clientId, clientSecret, callbackUrl, scope, state, pkce } = oAuth;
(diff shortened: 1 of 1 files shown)


packages/bruno-electron/src/ipc/network/oauth2-helper.js
```diff
@@ -29,9 +29,11 @@ const resolveOAuth2AuthorizationCodeAccessToken = async (request, collectionUid)
     code: authorizationCode,
     redirect_uri: callbackUrl,
     client_id: clientId,
-    client_secret: clientSecret,
-    state: state
+    client_secret: clientSecret
   };
+  if (state) {
+    data.state = state;
```
pietrygamat: Actually, I believe the state parameter should not be sent to access_token endpoint at all - it may be required by some services when directing user browser to authorization url, and at this stage of the flow bruno should at most verify if the state parameter was returned from `getOAuth2AuthorizationCode` - or maybe ignore it, because xsrf attacks the state parameter protects from are not a threat in Bruno integrated browser (may change if  #2650 is implemented though).

As a short term solution, I suggest removing line 33
``` 
state: state
```
and skip the `if` part entirely.

vaskevich: You might be right - updated!

pietrygamat (approved):
:+1:

PR #2898, opened 2024-08-22
fix: space in any param name is blocking the loading of the request file in GUI

pietrygamat:
# Description

Adjusted Bruno grammar so the `key` element does not disallow spaces (leading and trailing space is still trimmed).
fixes #2878
fixes #2810
fixes #3942

More flexible solution, i.e. adding option to quote and escape certain special characters proposed as an alternative to this in #3045 and #3178

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

PR #3045, opened 2024-09-05
fix: query parameters containing a colon (:) are not parsed correctly when changed

### LINKED ISSUE #2810: Params Query Bug

### Describe the bug

Adding an array key with space cannot be recognized when saving

meta {
  name: Get Users
  type: http
  seq: 2
}

get {
  url: http://laravel10.test/api/users?a=1
  body: none
  auth: none
}

params:query {
  test[iamRecognized]: Yey
  test2[iam not Recognized]: Nay
}

image

### LINKED ISSUE #2878: space in any param name is blocking the loading of the request file in GUI

[Jira](https://usebruno.atlassian.net/browse/BRU-1398)

### Describe the bug

When I use a space character ' ' in the param name in Params, Body, Headers tabs, it is blocking the loading of the file in the GUI.

meta {
  name: test1
  type: http
  seq: 3
}

get {
  url: http://url1?space in name
  body: none
  auth: none
}

params:query {
  space in name: 
}

When create a parameter with a space character in the name like "space in name", it is showing in GUI:
image

Action:
- I save the request
- I close Bruno GUI
- I open Bruno GUI => request is not visible anymore
image

### LINKED ISSUE #3037: query parameters containing a colon (:) are not parsed correctly when changed

### Describe the bug

I have the following request that is managed incorrectly in Bruno :
`{{BASE_URL}}/v2/en/sets?sort:field=name&sort:order=DESC&name=Dark&pagination:page=1&pagination:itemsPerPage=1`

When I input it in the URL bar everything is nice, but as soon as I save it get broken up by splitting it using the `:` while it should use the `=` in the params section.

and editing after that the params section change how the URL works.

```
meta {
  name: Get a list of sets
  type: http
  seq: 3
}

get {
  url: {{BASE_URL}}/v2/en/sets?sort:field=name&sort:order=DESC&name=Dark&pagination:page=1&pagination:itemsPerPage=1
  body: none
  auth: none
}

params:query {
  sort: field: name
  sort: order: DESC
  name: Dark
  pagination: page: 1
  pagination: itemsPerPage: 1
}

assert {
  res.status: eq 200
  res.body[0].id: eq swsh3
  res.body.length: eq 1
}

```

Repository : https://github.com/tcgdex/cards-database (_pull it using `--depth 1`_)

before saving : _Everything is fine_
image

After saving : _the url is find but the params are broken_
image

After editing the `Dark` param : _everything is broken_
image

### LINKED ISSUE #3942: Request (.bru file) not loaded if there's a space in a query param name

- [x] I use the newest version of bruno.

[Jira](https://usebruno.atlassian.net/browse/BRU-1398)

### Describe the bug

If there's a space in the **name** of any query string parameter, the `.bru` file doesn't load, and the request doesn't appear in the UI.

1. Save the attached `.bru` contents to a file inside a bruno directory and relaunch the app. Note that the request is not loaded.

1. Using an external editor, change the space between `sort descending` to an underscore and relaunch the bruno app. Note that the request now appears.

### Side Note
Why are QS params stored in 2 different places in the file? Why not store them in one place and then either encode or decode them at load time?

```
meta {
  name: Test Request
  type: http
  seq: 1
}

get {
  url: https://mydomain.com?sort%20descending=yes%20affirmative
  body: none
  auth: none
}

params:query {
  sort descending: yes affirmative
}
```

There's nothing to show because the GET request doesn't appear in the interface.

pietrygamat:
# Description

After saving the request, the colon (`:`) in a parameter name is being consumed by bru parser and what follows is treated as parameter value - so there must be a way to allow either quoting bru keys, or escaping problematic characters. In this PR I introduce a way to store dictionary keys for query params quoted. The quotes are discarded when loading collections, so they will be invisible in the UI.
fixes #3037 
fixes #2810 (alternative to #2898)
fixes #2878 (alternative to #2898)
fixes #3942 (alternative to #2898)

Automatic quoting is currently only applied when saving query parameters. It does not make a lot of sense for request headers, where colon is disallowed by HTTP spec. It may make sense to add in multipart-form body key-value table and vars table (#3178).

# Example
https://github.com/user-attachments/assets/4f3c530b-db79-4529-ac05-b10eaba8ddf1

```
params:query {
  normal-param: value
  "sort:field": name
  ~"sort:order": DESC
  ~"\"name\"\"\"": Dark
}
```

### Contribution Checklist:


packages/bruno-lang/v2/src/jsonToBru.js
```diff
@@ -4,6 +4,10 @@ const { indentString } = require('../../v1/src/utils');
 
 const enabled = (items = []) => items.filter((item) => item.enabled);
 const disabled = (items = []) => items.filter((item) => !item.enabled);
+const quoteKey = (key) => {
+  const quotableChars = [':', '"', '{', '}', ' '];
```
pietrygamat, line 9: The characters
```
:  {  } "
```
are important for bru syntax, and so they should be escaped by default. The question of space is open for discussion, as it may be solved in a different way (#2898) without forcing changes onto users' bru files.

Pragadesh-45 (approved):

Pragadesh-45:
@pietrygamat Will it be good if we quote all the keys instead of checking whether it consists any of this `[':', '"', '{', '}', ' ']`. But, it won't work as expected if we have `"key"` as part of the key. It will be nice to have your perspective on this.

The reason for this is to make everyone easily editing the key-value pair from bru files instead of having a thought of adding quotes when using `[':', '"', '{', '}', ' ']` as part of the keys itself.

pietrygamat:
@Pragadesh-45 when creating this PR I wanted to minimize the impact of the change on existing collections of users who do not need quoting (not using spaces or : and not having things broken), but I am not at all opposed to the idea of simply always applying the `"`s.  The but here is that the grammar should probably support unquoted keys anyway, so it can parse them on first load of old collections, and only update on save.

Yes, the users who used `"key"` with quotes already will be silently downgraded, and will have to reapply the quotes in UI (or add escaped quotes in bru files) - if we allow for escape for sequence just for quote and escape char, but not for any other characters. Assuming the collection will not fail to load, they should be able to spot the cause of breakage on their own, even if they miss the mention in README/release notes.

Pragadesh-45:
@pietrygamat, we recently had a discussion with the team about how to fix this issue, and we considered using the quoting logic. I had some doubts about why we need to include `{}` as part of the quoting logic. I also added `\t` to `quotableChars` because `\t` is considered a whitespace in Bru grammar. Additionally, I've updated the test as well. Could you please have a look at this PR?  #5332

helloanoop:
Closing this since the extended PR #3178 was merged.

Issue #3067, opened 2024-09-11
Path param not showing in Path params editors when using multiple env variable


jbuiquan:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

When I use two environment variable in my url, the path param is not added to the path param editor and i can't set a value.
When I use only one environment variable it works again.
I don't know why the path param gets stuck in red too.

I'm using the .deb package.

### .bru file to reproduce the bug

```
meta {
  name: Test
  type: http
  seq: 2
}

get {
  url: {{api_base_url}}:{{api_port}}/privileges/:id
  body: none
  auth: none
}
```

### Screenshots/Live demo link

With 2 variables
![image](https://github.com/user-attachments/assets/d57efd7d-d4db-48f0-906f-ac047581c2ca)

With 1 variable
![image](https://github.com/user-attachments/assets/ac722c32-6bf9-466b-9bc4-9430a32c9621)

apoorvyadav1111:
Hey, If you are not picking this up, I would like to work on this issue, Please assign this to me.

apoorvyadav1111:
Hi @sanjai0py, I have raised a PR #3107 to address this issue. I would appreciate your feedback.

helloanoop:
This will be available in upcoming `v1.29.0` release

abhishekkmr13:
I am still facing this issue on v1.40.0

My sample endpoint is:
{{host}}/user//{{user_id}}

I am getting below error:
405 Method Not Allowed


When I try inspecting the curl for this call then I don't find user_id correctly mapped to it.
curl --request GET \
  --url https://abc.com/user/ \
  --header 'accept: application/json'

PR #3107, opened 2024-09-15
#3067: handle multiple env vars

apoorvyadav1111:
# Description

This PR address #3067. The problem was when we create a new URL() with more than one env variable, it throws an error. But since bruno supports multiple environment variable, instead of throwing error, we try to split it directly and create path params.
image

image

### Contribution Checklist:

- [ x] **The pull request only addresses one issue or adds one feature.**
- [x ] **The pull request does not introduce any breaking changes**
- [ x] **I have added screenshots or gifs to help explain the change if applicable.**
- [ x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ x] **Create an issue and link to the pull request.**

### DIFF
--- a/packages/bruno-app/src/utils/url/index.js
+++ b/packages/bruno-app/src/utils/url/index.js
@@ -42,14 +42,14 @@ export const parsePathParams = (url) => {
     uri = `http://${uri}`;
   }
 
+  let paths;
+
   try {
     uri = new URL(uri);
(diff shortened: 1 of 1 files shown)


abhishekkmr13:
@helloanoop I am trying to pick my path-param value from environment variable but it is not picking it up:
image

PR #3110, opened 2024-09-16
chore: disableParsingResponseJson shim fn in safe mode

lohxt1:


helloanoop:
Thanks for taking care of this @lohxt1 !

Issue #3116, opened 2024-09-17
OpenAPI import: Malformed url of imported request


mjhcorporate:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

I have the following, simplified, OpenAPI 3.0 file:

```
openapi: 3.0.0
info:
  title: Car Rental API
  version: 1.0.0
  description: API for searching and booking rental cars

servers:
  - url: https://localhost:3000

paths:
  /health:
    get:
      summary: Health check endpoint
      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
                properties:
                  status:
                    type: string
                    example: "OK"
```
When I import this in Bruno, the single GET request is created with the following url: `https:/localhost:3000//health`.
- note the single slash between `https:` and `localhost`
- note the double slashes between `3000` and `health`

Curiously, sending this request works, albeit with the server complaining about a non-existing endpoint...

### Screenshots/Live demo link

![image](https://github.com/user-attachments/assets/4835f08d-d93a-4554-8c4b-223242fc36e3)

apoorvyadav1111:
Hi team, I have raised a PR for fixing this. #3121 .

Thanks,
Apoorv

anusree-bruno:
This issue has been fixed. Feel free to reopen the ticket if the issue persists

PR #3121, opened 2024-09-18
3116: fixed url slash formating from openapi

apoorvyadav1111:
# Description
This PR fixes the problem of parsing / when importing openApi spec yaml #3116 .

We introduce changes as per specs from open API as base URL not ending with `/` and rather the path begins with `/` . Also if the base URL are not provided, then baseUrl is set to `/`. 
image
image
Can find more details here https://swagger.io/docs/specification/api-host-and-base-path/

Result after changes:
image

### Contribution Checklist:

- [ x] **The pull request only addresses one issue or adds one feature.**
- [ x] **The pull request does not introduce any breaking changes**
- [x ] **I have added screenshots or gifs to help explain the change if applicable.**
- [ x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ x] **Create an issue and link to the pull request.**

### DIFF
--- a/packages/bruno-app/src/utils/importers/openapi-collection.js
+++ b/packages/bruno-app/src/utils/importers/openapi-collection.js
@@ -32,7 +32,7 @@ const readFile = (files) => {
 
 const ensureUrl = (url) => {
   // emoving multiple slashes after the protocol if it exists, or after the beginning of the string otherwise
-  return url.replace(/(^\w+:|^)\/{2,}/, '$1/');
+  return url.replace(/(^\w+:|^)\/{2,}/, '$1//');
 };
 
 const buildEmptyJsonBody = (bodySchema) => {
@@ -316,7 +316,7 @@ const getDefaultUrl = (serverObject) => {
       url = url.replace(`{${variableName}}`, sub);
     });
   }
-  return url.endsWith('/') ? url : `${url}/`;
+  return url.replace(/\/+$/,'');
 };
 
 const getSecurity = (apiSpec) => {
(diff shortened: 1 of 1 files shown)


packages/bruno-app/src/utils/importers/openapi-collection.js
```diff
@@ -32,7 +32,7 @@ const readFile = (files) => {
 
 const ensureUrl = (url) => {
   // emoving multiple slashes after the protocol if it exists, or after the beginning of the string otherwise
-  return url.replace(/(^\w+:|^)\/{2,}/, '$1/');
+  return url.replace(/(^\w+:|^)\/{2,}/, '$1//');
```
mjhcorporate, line 35: Do we really need to "fix" this for the user? Any helpful fix that we do has the chance of breaking a valid (?) use-case. For example, some file url look like this: `"file:///path/to/file` ([example](https://superuser.com/questions/352133/why-do-file-urls-start-with-3-slashes)). It's a bit contrived and likely won't come up in the real world, but there might be other use cases...

mjhcorporate (approved):
Looks good to me! This fixes my problem -- thank you for the quick PR! Just one note about possibly doing _less_ during the import, to not break some edge-cases.

packages/bruno-app/src/utils/importers/openapi-collection.js
apoorvyadav1111: Yes, this is an outlier and a rare case. If this is accepted to be merged, we can raise an enhancement issue for handling this case. That is if it is even needed for bruno. Thanks for you review.

naman-bruno:
Hey @apoorvyadav1111, thanks for your contribution!
Closing this PR.

PR #3123, opened 2024-09-18
fix: axios transform request bind

lohxt1:


sanjai0py:
Fixes https://github.com/usebruno/bruno/issues/3124

PR #3142, opened 2024-09-19
fix: multipart/form-data body interpolation

lohxt1:
changes:

~ the `formData body` will be available in json format during the pre-request script (`req.getBody()`).
   `json body` is interpolated after the pre-request step, and the `formData body` is then constructed from the json data.

Screenshot 2024-09-19 at 12 37 05 PM

---

~ if a constructed `formData body` is directly set in the pre-request script, the interpolation and the construction of `formData body` are skipped.

Screenshot 2024-09-19 at 12 31 24 PM


helloanoop:
@Its-treason Can you review this if you have some time over the weekend?
The implementation looks good to me.
We are fine to introduce the breaking change where `req.getBody()` returns an object instead of `FormData`

helloanoop (approved):

packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -195,6 +181,14 @@ const runSingleRequest = async function (
       request.data = qs.stringify(request.data);
     }
 
+    if (request?.headers?.['content-type'] === 'multipart/form-data') {
```
Its-treason, line 184: These two if-statements can be merged.

packages/bruno-cli/src/utils/common.js
```diff
@@ -14,7 +19,32 @@ const rpad = (str, width) => {
   return paddedStr;
 };
 
+const createFormData = (datas, collectionPath) => {
+  // make axios work in node using form data
+  // reference: https://github.com/axios/axios/issues/1006#issuecomment-320165427
+  const form = new FormData();
+  forOwn(datas, (value, key) => {
+    if (typeof value == 'object') {
```
Its-treason: I would personally invert the `if` to: `typeof value === 'string'` and return after `form.append` like this:

```js
if (typeof data === 'string') {
  form.append(key, value);
  return;
}

// Code for file here.
```
To make the code more readable

packages/bruno-tests/collection/bruno.json
```diff
@@ -15,7 +15,7 @@
     "bypassProxy": ""
   },
   "scripts": {
-    "moduleWhitelist": ["crypto", "buffer"],
+    "moduleWhitelist": ["crypto", "buffer", "form-data"],
```
Its-treason, line 18: We could add `form-data` to the default libraries.

Its-treason (approved):
I added some notes. But feature itself seems good.

packages/bruno-cli/src/utils/common.js
lohxt1: makes sense, will update the logic

packages/bruno-tests/collection/bruno.json
lohxt1: since `form-data` dep needs to be added to both vm2 and quickjs, i'l create a seperate pr for just that

helloanoop:
Merged. Thanks @lohxt1 for working on this!
Thank you @Its-treason for your review!

Issue #1116, opened 2023-12-01
[small bug] can't rename capitalised request into lowercase


adnan-smlatic:
If you have a request "Bruno" and want to rename it to "bruno" it doesn't work. You have to rename it to something else first and then "bruno"

helloanoop:
This has been addressed in PR #3171 and will be released in the upcoming `v1.32.0` release next week.

PR #1171, opened 2023-12-07
fix(#1116): allow rename with only case changes

Zomzog:
# Description

Should fix #1116 

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

PR #3171, opened 2024-09-23
Bugfix/rename with same name

Pragadesh-45:
fixes: #1116 #2995

# Description

This PR fixes the issue of renaming a request with the same name or a case-only difference (e.g., `Bruno` to `bruno`). This bug occurs because `fs.existsSync()`relies on the behavior of the file system (OS-Specific), which causes errors in Windows and macOS. You can check the details [here](https://github.com/usebruno/bruno/pull/3171/files#diff-06806ec8f9109775d07a5b223e78f2308d0811b7e78700305357adc7ec339a5dL342). 

This PR is an updated version of the PRs #531 , #1171 and #3024 by introducing a new utility function, `safeToRename`, to determine if it is safe to rename a file from `oldPath` to `newPath`. The function ensures safety by checking if the target path exists and whether the source and target files are the same, depending on the operating system (Windows/MacOS/Unix/Linux):

- **Unix/Linux/MacOS**: 
  - Compares the [inodes](https://www.redhat.com/sysadmin/inodes-linux-filesystem#:~:text=By%20definition%2C%20an%20inode%20is,filesystem%2C%20independent%20of%20the%20others.) of the files, as inodes uniquely identify files on these systems.

- **Windows (win32)**: 
  - Compares file birth time and size, and verifies if the absolute file paths are the same. This accounts for FAT-32 file systems that do not use inodes.

If the paths refer to the same file, renaming is considered safe. In case of any error (e.g., file system issues), the function logs the error and returns `false`.

**Note**: If you have `.bru` files like `a.bru` and `A.bru` on your Linux machine, the case-sensitive files will be displayed according to the behavior of the operating system, similar to how they are handled in VS Code.
 
### Contribution Checklist:

Before:

https://github.com/user-attachments/assets/5687d4f9-b402-42a8-8cbd-faaba17831e7

After:

https://github.com/user-attachments/assets/8389d9d7-d917-4fb4-949f-c29399da1420


helloanoop:
Merged! Thanks @Pragadesh-45 !
As usual, great writeup on what the PR does.

PR #3178, opened 2024-09-23
fix: unable to set request bodies with colon characters in their names

pietrygamat:


SoulKa:
What happens if I want to have the value `@file(bruno.json)` in my multipart form?

pietrygamat:
@SoulKa that part is not touched, so the value will be converted to a file, as it is currently
https://github.com/user-attachments/assets/4862bb1b-7f43-4487-8480-22ae7da0b225

Gavlar1:
Happy to raise a seperate bug if this is not the place, but this seems to have regressed as I have tried in multiple versions since 35, but even if I surround the entry in quotes in the .bru file, then when I check in the GUI then it still passes everything after the first colon into the Value rather than key.

pietrygamat:
@Gavlar1 it's not a regression, this PR was never merged, unfortunately.

bijin-bruno (approved):

helloanoop:
Merged. Thanks @pietrygamat for working on this and being patient :)

Thanks @Pragadesh-45 for reviewing this and the long hours.

pietrygamat:
@helloanoop @Pragadesh-45 will you please take a look at  #5462? Since the decision on how to handle the problematic characters is made, it is now the time to iron out some places that I missed in this PR, and now noticed on retest.

PR #3188, opened 2024-09-25
feat: remove pre-request var tooltip

lohxt1:
~ pre-request vars can now only be strings, hence removing the `template-literal` toopltip from the table header

Screenshot 2024-09-25 at 11 16 08 AM


helloanoop:
Merged! Thanks @lohxt1 !

PR #3105, opened 2024-09-14
Implement detection for runner

nikischin:
# Description
Implement a new function `bru.getExecutionMode()` to use inside scripting. Returns **single**, **runner** or **cli**.

Fixes #3104 
Related to #2397 
Includes changed from #1875 partially

Thank you @JorgeTrovisco this is a modified implementation of your PR. Thank you also @tlaloc911 as this also includes your changes.

image

The real benefit of this PR will be implemented with the merging of #2397, as this functionality has been extracted from this PR in order to keep it focussed on one feature. With both this and #2397 merged, you could do something like this in order to skip requests on runner runs but still have them for single tests.

```
if(bru.getExecutionMode() === 'runner'){
   bru.skipRequest();
}
```

Or you could even do something like this
```
if(bru.getVar("myAuthKey") === undefined){
  bru.skipRequest();
  
  if(bru.getExecutionMode() === 'runner'){
     bru.setNextRequest("getAuthToken");
  }
}
```

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

PR #3200, opened 2024-09-26
feat: request execution mode

lohxt1:
extension of pr #3105

moving the getExecutionMode fn to `req` object instead of `bru` object

`req.getExecutionMode()`

values: standalone | runner | cli

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -55,6 +55,7 @@ if (!SERVER_RENDERED) {
     'req.setMaxRedirects(maxRedirects)',
     'req.getTimeout()',
     'req.setTimeout(timeout)',
+    'req.getExecutionMode()',
     'bru',
     'bru.cwd()',
     'bru.getEnvName(key)',
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -43,6 +43,8 @@ const runSingleRequest = async function (
 
     request = prepareRequest(bruJson.request, collectionRoot);
 
+    request.__bruno__executionMode = 'cli';
+
     const scriptingConfig = get(brunoConfig, 'scripts', {});
     scriptingConfig.runtime = runtime;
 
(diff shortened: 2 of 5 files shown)


helloanoop:
@nikischin Can you review this PR?

nikischin (approved):
Excellent, thank you so much for this modified implementation! 

I reviewed code and tested in UI in both `developer` and `safe` mode and it does work as intended. I haven't tested in cli, so maybe someone can confirm for cli?

helloanoop:
Merged!

Thank you @lohxt1 and @nikischin !

helloanoop:
@nikischin I am thinking that for now, we should remove the `cli` from the options returned.
Even when the run is being done in CLI, it should still return `standalone` | `runner` based on whether a single request is being run or the a folder / collection is being run

An api to perhaps consider in the future is `bru.getPlatform()` that can return `app`|`cli` based on where the run is happening

Do you concur?

nikischin:
> @nikischin I am thinking that for now, we should remove the `cli` from the options returned.
> 
> Even when the run is being done in CLI, it should still return `standalone` | `runner` based on whether a single request is being run or the a folder / collection is being run
> 
> 
> 
> An api to perhaps consider in the future is `bru.getPlatform()` that can return `app`|`cli` based on where the run is happening
> 
> 
> 
> Do you concur?

Honestly, I have never used the cli so far. I might be interested in doing so in the future but did not have a need for it yet (if I would include it in some workflows, this would be the way to go).

So honestly I cannot really provide any valuable feedback on this one. If there's different ways of how to execute the cli I would prefer the version of having standalone and runner in both cli and ui with a separate getPlattform(). If however in cli it does not make a difference (let's say it's always a batch, but some times a batch of only one - this was what I assumed in my initial implementation) I would recommend to leave the implementation as is. Does this make sense? In the end you can decide better what's the best way to do, however if we change I would do so before releasing a new version as this would be a breaking change. I hope my feedback helped.

nikischin:
@helloanoop @lohxt1 let me know if I should modify the implementation in order to reflext the changes suggested or if you will do it:

> @nikischin I am thinking that for now, we should remove the `cli` from the options returned. Even when the run is being done in CLI, it should still return `standalone` | `runner` based on whether a single request is being run or the a folder / collection is being run
> 
> An api to perhaps consider in the future is `bru.getPlatform()` that can return `app`|`cli` based on where the run is happening

One detail which for me is unclear, if the `getPlatform()` would be attached to `bru` or the `req` object. From a user perspective it doesn't make a big difference, but for the implementation it could matter.

helloanoop:
Thanks for your feedback @nikischin !
We'll go with `req.getExecutionPlatform()` that returns `app`|`cli` based on where the request is being run

@lohxt1 will be make the updates for this.

PR #3222, opened 2024-09-28
feat: global env

lohxt1:
~ global environments

https://github.com/user-attachments/assets/2cd2b8af-cec0-4b6a-b9d4-8515035805bb

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/StyledWrapper.js
+++ b/packages/bruno-app/src/components/CodeEditor/StyledWrapper.js
@@ -10,6 +10,12 @@ const StyledWrapper = styled.div`
     flex: 1 1 0;
   }
 
+  /* Removes the glow outline around the folded json */
+  .CodeMirror-foldmarker {
+    text-shadow: none;
+    color: ${(props) => props.theme.textLink};
+  }
+
   .CodeMirror-overlayscroll-horizontal div,
   .CodeMirror-overlayscroll-vertical div {
     background: #d2d7db;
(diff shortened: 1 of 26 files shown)


codewell-learning:
Excellent feature and is very demanded. Nice work!

frdesnous:
Thanks a lot @lohxt1 , much appreciated!!! One question though: where are the global environment (.bru?) files stored? I created a new one but can't find the file

paulrudy:
> Thanks a lot @lohxt1 , much appreciated!!! One question though: where are the global environment (.bru?) files stored? I created a new one but can't find the file

I was looking too. Found it (on macOS at least) in `~/Library/Application Support/bruno/global-environments.json`

frdesnous:
Thanks @paulrudy. I think it would make more sense to be able to save it in a different place, like at the root of all collections for example.

paulrudy:
> Thanks @paulrudy. I think it would make more sense to be able to save it in a different place, like at the root of all collections for example.

I agree, and I'd prefer it to be in the same format as the local environments, to allow easy import (copy/paste) from local to global, and back.

frdesnous:
@lohxt1 I assume you are also planning to enable it in the bruno CLI and be able to run the collection with this global environment 
`bru run folder --env ../Global`

reinharn:
Will it soon be possible to configure the storage location of the global environments?

lundgreen-jyskebank-dk:
Hi. If I have a global environment called 'user' like:

CleanShot 2024-10-16 at 09 25 40@2x

Is it then possible to access the global variables at collection level? Like:

CleanShot 2024-10-16 at 09 30 19@2x

Usecase: ONLY set user/pwd in a global variable and then use it in all collections

yusufmalikul:
Thanks, I don't know if we have global variables.

FYI for linux the global variables stored in `.config/bruno/global-environments.json`

Issue #3229, opened 2024-09-30
[New Feature] List runtime variables / delete all runtime variables


tho-gru-38:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

I have a quite large collection which use the runtime variables a lot.

For the sake of run reproducible result I want to delete all runtime variables (set by `bru.setVar` function). I know that there is a `bru.deletVar` function which need the name of the variable to delete as parameter. I did not find a way to list all names of current runtime variables.

The ability of deleting all runtime variables will help in two situations:

1. When running folder or collection multiple time, the variables can be delete by a script.
2. If a collection consists of multiply folder, each top level folder can wipe all runtime variables. In this case the developer can be sure, that the result of a top level folder is not influenced by any preceding top level folder.

I also found the a function `bru.hasVar()` already exists. Unfortunately no documentation is available.

### Mockups or Images of the feature

I want to have one of  the following:

1. `bru.deleteAllVars()` - simply delete all runtime variables
2. `bru.listVars()` - return an array containing all names of runtime variables

The second solution enable the developer to achieve the first by building a loop over all variable names returned by `bru.listVars()` and delete each individually.

helloanoop:
Hey @tho-gru-38 !

Thanks for describing your requirement in detail.
Yes, I think the `bru.deleteAllVars()` function makes sense.

@lohxt1 Can you take a look at this?

helloanoop:
This has been addressed in #3231 and will be available in the `v1.33.0` release that is scheduled to go out on Thu, 4th Oct 2024

tho-gru-38:
Hi,

Thanks for implementing the new feature.

Kind Regards,
Thomas

PR #3231, opened 2024-09-30
feat: bru fns update

lohxt1:
#3229 

~ added `bru.deleteAllVars()` fn

~ added missing `bru.hasVar()` and `bru.hasEnvVar` fns to safe-mode


helloanoop:
Merged! Thanks @lohxt1 !

PR #2261, opened 2024-05-08
Feature: match postman-cli default when openapi definition does not include a server 

Tristano8:
# Description

Matches the [existing behaviour](https://github.com/postmanlabs/openapi-to-postman/blob/228c462c3f542e72e420c1e27fc9b5bfcb839a24/libV2/helpers/collection/generateCollectionFromOpenAPI.js#L92) of postman-cli's 'Import from openapi' with a couple of defaults:
 - uses `{{baseUrl}}` when no `servers` entry is found. This addresses https://github.com/usebruno/bruno/issues/1239
 - uses an empty string when `info` field is not present

Example result:
Screenshot 2024-05-08 at 11 06 27 AM

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [ ] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

PR #3267, opened 2024-10-08
Create default environments on openapi import using server url

anusreesubash:
Fixes: #1239 
Closes PRs: #1240, #2261
1. Instead of separate variables protocol and host, this change creates a single environment variable basePath
2. Loops through all server urls and create environments for each of them separately
3. Creates environments using the server description if that doesnot exists creates environment as Environment 1(2, 3, ...)

https://github.com/user-attachments/assets/b45b1cb8-99bd-4576-984f-d5603159c051


kevinpapst:
Looks fantastic, so I can finally use OpenAPI imports with Bruno. Up until now this missing piece is the one major pain point and the reason why I never really used Bruno since I purchased it months ago. Thanks for working on it 👍

kevinpapst:
@helloanoop this PR was supposed to close #2261 - see initial description - but it is still open

Issue #3265, opened 2024-10-08
Secret Variables Exposed on Hover Over Request Components


Pragadesh-45:
When hovering over a {{variable}} that is set as a secret value, the secret is currently displayed in the editor component. It would be better to show asterisks (***) in the hover text instead of revealing the secret value.

![image](https://github.com/usebruno/bruno/assets/1887439/d6557ed0-3114-42a7-8296-fb4f1676c20a)

_Originally posted by @JaceyPenny in https://github.com/usebruno/bruno/issues/1646#issuecomment-1989065065_

helloanoop:
@Pragadesh-45 Lets get this PR reviewed and closed this week.

PR #3268, opened 2024-10-08
Feat/displaying variable hints as secrets

Pragadesh-45:
fixes: #3265 

# Description

This PR addresses the issue of secret variables being displayed on hover in editor components by implementing the following logic:
 - Checks masked variables from environment variables, which support secret variables.
 - Verifies if lower-level variables (such as request or folder variables) override the secret variables, including cases with string interpolation.
 - Creates a (*****) node to hide secret values and prevent them from being displayed.

### Contribution Checklist:

Before:
image

After:

https://github.com/user-attachments/assets/7319a8e0-8dae-465a-a35c-f8fa0712ac92

### DIFF
--- a/packages/bruno-app/src/utils/codemirror/brunoVarInfo.js
+++ b/packages/bruno-app/src/utils/codemirror/brunoVarInfo.js
@@ -44,8 +44,11 @@ if (!SERVER_RENDERED) {
     const into = document.createElement('div');
     const descriptionDiv = document.createElement('div');
     descriptionDiv.className = 'info-description';
-
-    descriptionDiv.appendChild(document.createTextNode(variableValue));
+    if (options?.variables?.maskedEnvVariables?.includes(variableName)) {
+      descriptionDiv.appendChild(document.createTextNode('*****'));
+    } else {
+      descriptionDiv.appendChild(document.createTextNode(variableValue));
+    }
     into.appendChild(descriptionDiv);
 
     return into;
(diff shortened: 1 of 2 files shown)


fredx21:
Why do we want this? It's usefull to see the secrets.

Pragadesh-45:
Hey @fredx21!

The intention of this PR is to hide secrets when hovering over variables in the Editor, particularly during cases like screen sharing. We thought that this [Discussion](https://github.com/usebruno/bruno/discussions/1645) made sense, as it suggests hiding the variable only if it is marked as a secret. In some cases, we can't guarantee whether a variable is a secret since it is displayed when accessed. This will provide users with a strict way to view all secret variables in the same place—in this case, the Environment variable section. All secret variables can be viewed and edited in that area.

PS: _I acknowledge that in some cases, it may be useful to show the secret variables. It's truly subjective, and I believe the discussion mentioned above made sense, especially when we're on a client call and want to avoid exposing secret variables. For example, I used `{{host}}` as a secret. In most cases, it will be things like API keys, passwords, and JWT tokens. However, I think using `{{host}}` as an example may have caused a misunderstanding of the feature._

fredx21:
I understand that in some case it might be usefull to hide the secret variable, however, its more often usefull to have those secrets shown as this is mostly a developer-first tool. 

What do you think about adding a option in setting to allow user changing the default behavior?

Pragadesh-45:
> What do you think about adding a option in setting to allow user changing the default behavior?

This one will work great. Thanks for the nice suggestion, by the way!

helloanoop:
I have merged the PR.
@Pragadesh-45 please create a new github issue to track @fredx21 suggestion on adding a option in setting to allow user changing the default behavior.

PR #3300, opened 2024-10-14
feat: upgrade libraries for dependabot alerts

lohxt1:


packages/bruno-app/src/utils/common/polyfillResolver.js
```diff
@@ -0,0 +1,23 @@
+export const polyfillResolver = () => {
```
helloanoop: I recommend naming this `setupPolyfills`

```javascript
// polyfill required to make react-pdf
if (typeof Promise.withResolvers === "undefined") {
    if (typeof window !== 'undefined') {
      window.Promise.withResolvers = function () {
```

helloanoop:
Great work @lohxt1 !
Keep it up !!

This is a significant step forward in enhancing the security of the project

PR #3312, opened 2024-10-15
fix: add global env vars to the active collection

lohxt1:
~ merge `globalEnvironmentVariables` into the active collection and rebuild the `collections` immer proxy object, instead of adding it to the cloned `collection` object.


helloanoop:
Great work @lohxt1! Merged!

Issue #3233, opened 2024-10-01
Proxy Configuration Issue in Bruno v1.28+


giuseppemilicia:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Since version 1.28 of Bruno, when configuring "proxy enabled" and "proxy pass" for a specific domain within a collection, requests that bypass the proxy return the following error:
`Error invoking remote method 'send-http-request': Error: unable to verify the first certificate.`

If the proxy is disabled or when using version 1.27, everything works as expected.

### Screenshots/Live demo link

v1.31:
![image](https://github.com/user-attachments/assets/11e5acde-0c2d-49bc-b05e-9582d4da45f5)

christiangierschner:
![grafik](https://github.com/user-attachments/assets/998bc5a8-16e4-4984-addf-cdd341669a15)
same issue, but different error msg 1.28+
SSL is disabled -> not working

helloanoop:
I have merged the PR for this fix.

@Ved_Sharma please test the resolution once `v1.34.0` is live

helloanoop:
@giuseppemilicia, could you please check if the issue is resolved in version v1.34.0?

giuseppemilicia:
@helloanoop the issue has been fixed in version v1.34.0

PR #3317, opened 2024-10-16
Fixes Issue #3233: set httpsAgentRequestFields on request.httpsAgent when shouldProxy or shouldUseSystemProxy was false

giuseppemilicia:
Fixes Issue #3233

Previously, `httpsAgentRequestFields` were not applied to `request.httpsAgent` when `shouldProxy` or `shouldUseSystemProxy` was false, resulting in an improper agent configuration.


helloanoop:
Merged. Thanks @giuseppemilicia !
This will be available in the upcoming `v1.34.0` release

PR #3362, opened 2024-10-25
feat: revert graphiql, codemirro-graphql library versions

lohxt1:
~ will work on a seperate pr to fix the styling caused by the version upgrade


helloanoop:
@lohxt1 as discussed, lets try to fix the codemirror issues in the upgraded version over downgrading and introducing the dependabot alerts

PR #3472, opened 2024-11-12
fix: improve masking logic in MaskedEditor for large content handling (fixes #2842)

### LINKED ISSUE #2842: Bruno 1.24.0 is very slow when obfuscating tokens

### Describe the bug

Whenerer the Auth panel is oben with a obfuscated token, the application is VERY slow. When the token is cleartext, everything is as fast as normal.

_No response_

image

Pragadesh-45:
fixes: #2842 

# Description

This PR is implemented to improve the app's performance when masking text in the SingleLineEditor. The logic in PR #2240 created a new node for every masked text range. While this logic performs well when the content length is less than 1000 characters, its performance degrades as the length exceeds 1000 characters. In this PR, I've addressed these performance issues.

Changes made:
1. Added an early return if there are no lines in the editor.
2. Limited the previous logic to cases where the number of masked characters is less than 500. This is more efficient for editing small content, such as secret environment variables.
3. For content greater than 500 characters, implemented a line-by-line masking approach to improve performance.

Before:

https://github.com/user-attachments/assets/4b920159-1a16-4a29-b984-c460e5e6bb94

After:

https://github.com/user-attachments/assets/6c63dfea-5d0f-4c79-8e0a-80216c482573

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/utils/common/codemirror.js
+++ b/packages/bruno-app/src/utils/common/codemirror.js
@@ -52,17 +52,33 @@ export class MaskedEditor {
   /** Replaces all rendered characters, with the provided character. */
   maskContent = () => {
     const content = this.editor.getValue();
+    const lineCount = this.editor.lineCount();
+
+    if (lineCount === 0) return;
     this.editor.operation(() => {
       // Clear previous masked text
       this.editor.getAllMarks().forEach((mark) => mark.clear());
       // Apply new masked text
-      for (let i = 0; i < content.length; i++) {
-        if (content[i] !== '\n') {
-          const maskedNode = document.createTextNode(this.maskChar);
+
+      if (content.length <= 500) {
+        for (let i = 0; i < content.length; i++) {
+          if (content[i] !== '\n') {
+            const maskedNode = document.createTextNode(this.maskChar);
+            this.editor.markText(
+              { line: this.editor.posFromIndex(i).line, ch: this.editor.posFromIndex(i).ch },
+              { line: this.editor.posFromIndex(i + 1).line, ch: this.editor.posFromIndex(i + 1).ch },
+              { replacedWith: maskedNode, handleMouseEvents: true }
+            );
+          }
+        }
+      } else {
+        for (let line = 0; line < lineCount; line++) {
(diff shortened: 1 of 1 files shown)


packages/bruno-app/src/utils/common/codemirror.js
```diff
@@ -52,17 +52,33 @@ export class MaskedEditor {
...
+            this.editor.markText(
+              { line: this.editor.posFromIndex(i).line, ch: this.editor.posFromIndex(i).ch },
+              { line: this.editor.posFromIndex(i + 1).line, ch: this.editor.posFromIndex(i + 1).ch },
+              { replacedWith: maskedNode, handleMouseEvents: true }
+            );
+          }
+        }
+      } else {
+        for (let line = 0; line < lineCount; line++) {
+          const lineLength = this.editor.getLine(line).length;
```
sanjai0py, line 76: In the current logic, when the content length exceeds 500, the `lineCount()` function splits the editor value at each newline character. However, what happens if the input is a very long continuous string?

Pragadesh-45: > However, what happens if the input is a very long continuous string?

If the string is longer than 500 characters, the line count will be 1, so the for loop will be executed for only a single line (1x time)

Pragadesh-45:
Hey @philippthiele,  

Thank you for sharing your opinion. While working on this, I noticed that the app's performance was declining when the input exceeded 1000 characters. I realized the need to mask every item individually. I even tried improving performance by storing the positions and applying bulk masking, but it resulted in roughly the same performance.  

This led me to consider a different approach: creating a `maskedNode` for every line to improve performance. However, I noticed that this approach made it difficult to view and edit masked variables side by side, which is useful in many cases, such as editing environment variables.

That’s why I raised a PR with this logic—I just wanted to solve this problem. I’d love to hear your thoughts on it!

philippthiele:
@Pragadesh-45 thanks for the explanation, it is not a use case I use, but that explains it. Waiting eagerly for this fix to be merged!

helloanoop:
Thank you @Pragadesh-45 !

PR #3522, opened 2024-11-21
feat: add 'isNotEmpty' assertion operator

sanjai0py:
# Description

This PR adds support for the isNotEmpty assertion operator. The `isNotEmpty` assertion checks that a value is not null, not 0, not false, not an empty string, not an empty array, and not an empty object. The implementation includes:

- Checking that the value is not one of [null, 0, false, '', []].
- Ensuring arrays have a length greater than 0.
- Ensuring objects have at least one key.

---

### Contribution Checklist:

### DIFF
--- a/packages/bruno-js/src/runtime/assert-runtime.js
+++ b/packages/bruno-js/src/runtime/assert-runtime.js
@@ -350,12 +350,7 @@ class AssertRuntime {
             expect(lhs).to.be.empty;
             break;
           case 'isNotEmpty':
-            expect(lhs).to.not.be.oneOf([null, 0, false, '', []]);
-            if (Array.isArray(lhs)) {
-              expect(lhs).to.have.length.above(0);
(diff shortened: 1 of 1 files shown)


packages/bruno-js/src/runtime/assert-runtime.js
```diff
@@ -345,6 +349,14 @@ class AssertRuntime {
           case 'isEmpty':
             expect(lhs).to.be.empty;
             break;
+          case 'isNotEmpty':
```
helloanoop, line 352: Can we use `expect(lhs).to.not.be.empty;` ?

sanjaikumar-bruno: This validation works only if the LHS is not falsy. For example the `null isNotEmpty` assertion will pass in the current implementation but the suggested method results in the error: `empty was passed a non-string primitive null.`

Issue #3495, opened 2024-11-15
Allow response headers to be changed


drwelby:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

In https://github.com/usebruno/bruno/issues/2441 the request body could be updated. This is useful to convert less human-friendly response bodies into structures that are more legible.

However, since it appears the application relies on the `content-type` header to determine how to display the result, this mean that while it is possible to convert the response between formats, the contents will only render as the original format.

A common use case is with APIs that return a response in the GeoJSON format. This standard represents spatial objects like polygons and points on a map. It is helpful to be able to see the results as a map instead of as a string of characters. 

An example of this in Postman is https://github.com/OrdnanceSurvey/postmanGeoJSON. The GeoJSON response body is inserted into HTML code to be rendered in the response.

Bruno can render similar content in the response correctly already - a GET request of https://leafletjs.com/examples/quick-start/example.html works fine in the "web" preview in the response panel.

### Mockups or Images of the feature

The following functions would be exposed in responses, following the same syntax for requests:

```
res.setHeader()
res.setHeaders()
```

This would look like (skipping the HTML generation code):

<img width="1272" alt="Screenshot 2024-11-15 at 11 40 30 AM" src="https://github.com/user-attachments/assets/90afe2ec-1778-485d-bf8d-e7929bfb430e">

drwelby:
The additional functions seemed straightforward to add so I created a PR: https://github.com/usebruno/bruno/pull/3527

michaelblack:
Just chiming in that I would also like to see this functionality implemented. My team is running a bunch of tests that, for standards reasons, have to wrap HTML in a base64 string inside a json response and it would be nice to preview that HTML directly instead of introducing manual steps to pull it out and render it.

drwelby:
The "preview" button in the response format dropdown in Bruno 3 looks like it will let you see the response as HTML But with response scripts being broken in the mean time we can't test it yet.

PR #3527, opened 2024-11-21
Add functions to set response headers

drwelby:
# Description

This adds the functions to enable the features requested in https://github.com/usebruno/bruno/issues/3495

In short, this allows the ability to modify the headers in the response object `res`. This is primarily useful to take response contents and convert them to a different content type. Changing the `content-type` header allows Bruno to render the new response correctly. This reproduces the functionality found in Postman's "visualizer" concept.

Attached is a screenshot demonstrating converting a JSON response of geospatial data into a map view. 

### Contribution Checklist:

Screenshot 2024-11-21 at 10 23 30 AM


DKmen (approved):

Its-treason:
I don't believe solving the "incorrect response formatting" issue by modifying the response header is appropriate, just because the app currently determines formatting through headers.

The response should be treated as immutable and should not be altered solely for formatting purposes.

A better approach would be to create a `res.setFormatMode(formatting: string)` method (the exact API and naming can be refined) that sets an internal formatting flag. The frontend can then read this flag to determine the appropriate formatting, rather than relying on headers.

drwelby:
Point taken, but responses have been mutable since https://github.com/usebruno/bruno/pull/3076.

PR #3546, opened 2024-11-25
refactor: [DX-3251] - blacksmith runner and cache refactor

m-schrepel:
This PR updates workflow files to use blacksmith runner and the latest setup-node action.


Its-treason:
Are the changes to OAuth2 intentional?

Issue #3454, opened 2024-11-08
When renaming a request in bruno client, if we add 'DELETE/' it removes such request


Drachenfels:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

It's easy to reproduce.

1. Create a new request say "Hello World"
![selection-20241108143931](https://github.com/user-attachments/assets/b56cec47-a36f-4c10-9c81-0e1b611f8996)

2. Rename the request to say "Hello/DELETE World"
![selection-20241108144105](https://github.com/user-attachments/assets/487b978e-8755-44e8-82e7-74a07ceb441d)

3. Instead of renaming, the file will be removed from the hard drive
![selection-20241108144113](https://github.com/user-attachments/assets/693ace3e-9420-4233-9549-b563f6359cd8)

### Screenshots/Live demo link

![selection-20241108143931](https://github.com/user-attachments/assets/b56cec47-a36f-4c10-9c81-0e1b611f8996)
![selection-20241108144105](https://github.com/user-attachments/assets/487b978e-8755-44e8-82e7-74a07ceb441d)
![selection-20241108144113](https://github.com/user-attachments/assets/693ace3e-9420-4233-9549-b563f6359cd8)

ganesh-bruno:
Hey @Drachenfels 
Thanks for reporting this issue. 

I'm able to reproduce it. 
The error occurs when renaming the request to `req-name/DELETE` and we're working on it.

helloanoop:
I have merged the PR https://github.com/usebruno/bruno/pull/3274 that will now throw warnings and prevent file deletion.
This will go out in the upcoming `v1.36.0` release scheduled to be released on 4 Dec 2024

We are also working on supporting backward and forward slashes in request names. Please track https://github.com/usebruno/bruno/issues/2996 for updates regarding that.

PR #3274, opened 2024-10-09
Fix/invalid file name handling

Pragadesh-45:
fixes: #3266 #1401 #3454 

# Description

This PR implements the logic for validating filenames using a new utility function called `isValidFilename` to handle cases where filenames contain invalid characters `\ / : * ? " < > |` or end with a space (` `) or a period (`.`). This function has been added to the `rename-item`, thereby fixing issue #3266 , and to the `new-request` to improve error handling.

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

Before:

https://github.com/user-attachments/assets/0aaeb9d8-f731-4bbb-93fd-081b28ca94a3

After:

https://github.com/user-attachments/assets/276aac38-fb35-47e9-87f8-83faf5cbba58

PR #3581, opened 2024-12-02
Prevent data-loss on rename if writing new file fails

Osndok:
# Description

As it currently stands, bruno relies upon itself holding data in memory during rename. If a write operation fails for any reason, we have a very good chance of losing user data. If we do the delete operation after the write, then we instead only risk duplication of data (if the unlink fails), which seems axiomatically correct.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/collection.js
+++ b/packages/bruno-electron/src/ipc/collection.js
@@ -392,8 +392,9 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
       moveRequestUid(oldPath, newPath);
 
       const content = jsonToBru(jsonData);
-      await fs.promises.unlink(oldPath);
+      // we must write the new file first to avoid data loss if an exception occurs during write
       await writeFile(newPath, content);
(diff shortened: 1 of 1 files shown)


Osndok:
Related to #3454, #3274

Pragadesh44-Bruno:
Hey @Osndok,

Thanks for raising a PR addressing this. I agree with your point that it is generally best to call `unlink` after a successful rewrite to minimize the risk of data loss.

However, the specific block you're referring to is written to handle a unique issue on Windows. Due to the file system's case insensitivity, filenames like `Abc.bru` and `abc.bru` are considered the same. In such cases, attempting a direct rename can fail or produce unexpected behavior. To address this, the implementation unlinks the file first, followed by writing the new file.

Additionally, there are scenarios where atomic renames cannot be performed, such as when dealing with case-insensitive file systems or cross-device operations. This implementation ensures compatibility with these edge cases.

helloanoop:
@Osndok 

There are some edge cases we need to handle as Pragadesh mentioned. We are discussing internally this and will get back to you.

Issue #3627, opened 2024-12-10
 Changing the case in the folder name causes it to be deleted


f0leg:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

**playback steps**
1) create a folder with an uppercase letter. For example, "Name"
2) change the folder name from the first step by replacing the letter with lowercase. For example, "name"
**Expected result**
The folder name has been changed
**The actual result**
The folder has been deleted, along with all requests

version 1.36.1

### Screenshots/Live demo link

https://github.com/user-attachments/assets/51f3f584-0e39-4518-ade2-570a447b4d78

anusree-bruno:
Hi @f0leg, thanks for reporting this. We’re working on it and will include a hotfix to resolve the issue

naman-bruno:
Hey, closing this issue as it’s been fixed. Thanks!

PR #3629, opened 2024-12-10
fix: reorder file operations to ensure proper rename and cleanup (fixes: #3627)

Pragadesh-45:
fixes: #3627

# Description

This PR fixes a bug introduced in PR #3236, which addressed an issue where folders with subfolders couldn't rename the parent folder. The original fix was developed to address two issues: case-insensitive renaming, particularly on Windows and macOS. However, while the step in `tempDir` works fine, we end up with both `temp = folder (oldPath)` and `newPath = Folder (newPath)`, which causes an issue where the `tempDir` file gets stuck in the temporary directory. 

To resolve this, I have added logic to completely remove the contents of the old path before moving the `tempDir` files into the new path. This should fix both the case-insensitive rename issue and the problem of being unable to rename the parent folder.

### Contribution Checklist:


Pragadesh-45:
Addressed in PR: #3635

Issue #2229, opened 2024-05-03
Bruno CLI: The `punycode` module is deprecated


fkinnar:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

When running any request from CLI, I get the warning:
_(node:12425) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)_

Tested on Windows and Ubuntu (via WSL)

### Screenshots/Live demo link

![image](https://github.com/usebruno/bruno/assets/45601515/9cd09d78-0f47-421a-9d83-f38b4f8165ac)

pieterocp:
I want to say that this is now resolved since node/packages have been updated? Can't say which version has resolved it.

sreelakshmi-bruno:
Closing this ticket since this issue has been fixed. Please reopen if you find otherwise, thanks!

PR #3640, opened 2024-12-12
fix: disable deprecation warnings

chenrui333:
# Description

### Contribution Checklist:

----

fixes https://github.com/usebruno/bruno/issues/2229
relates to https://github.com/Homebrew/homebrew-core/pull/200885

### DIFF
--- a/packages/bruno-cli/src/commands/run.js
+++ b/packages/bruno-cli/src/commands/run.js
@@ -1,3 +1,5 @@
+#!/usr/bin/env -S node --no-deprecation
+
 const fs = require('fs');
 const chalk = require('chalk');
 const path = require('path');


nmss:
Wouldn't it be better to address the deprecation and remove/replace punycode instead of hiding the warning message ?

https://nodejs.org/api/punycode.html#punycode
> The version of the punycode module bundled in Node.js is being deprecated. In a future major version of Node.js this module will be removed. Users currently depending on the punycode module should switch to using the userland-provided [Punycode.js](https://github.com/bestiejs/punycode.js) module instead. For punycode-based URL encoding, see [url.domainToASCII](https://nodejs.org/api/url.html#urldomaintoasciidomain) or, more generally, the [WHATWG URL API](https://nodejs.org/api/url.html#the-whatwg-url-api).

pieterocp:
@nmss it'd be better, but suppression is one way to fix it in the short term, this does stop consumers of the cli from having to see it. We probably want to go with a better fix involving slight internal changes and probably upgrades of dependencies.

PR #3653, opened 2024-12-15
fix: bruno-app tests

lohxt1:


helloanoop:
Merged! Thanks @lohxt1 !

PR #3655, opened 2024-12-15
fix: rsbuild related error fixes

lohxt1:


helloanoop:
Merged! Thanks @lohxt1 !

PR #3676, opened 2024-12-18
fix: string json response error

pooja-bruno:
when an endpoint was returning a string in json response we were getting an error


Pragadesh-45 (approved):

MarcHagen:
It seems this gives us some problems now?

I've some endpoints that return the raw value `"ok"` with Header content-type application/json.
But the assertion `res.body eq ok` now fails.
`res.body eq "ok"` also fails
`res.body: eq ""ok""` succeeds... 

This could not be the expected behavior right?

Issue #481, opened 2023-10-09
script PostResponse: res is not a function 


gmajcen:
tried to save some ids to var in Post Response under scripts: `const ids = res('features.id')`

```
Error: Error invoking remote method 'send-http-request': TypeError: res is not a function
```

the same thing works with Vars.

helloanoop:
@gmajcen Yup, I am able to reproduce the bug. Will get this fixed.

Meanwhile, here is a workaround.

```javascript
const {get} = require('lodash');

const page = get(res.getBody(), 'page');
console.log('page', page)
```

![image](https://github.com/usebruno/bruno/assets/9350733/e888aa81-cb6e-4fc7-b235-297fe81b5510)

armand-janssen:
I created https://github.com/usebruno/bruno/pull/579 to fix it. It works on my machine 😄

doliG:
Hello, I have the same problem (v1.2.0). Any updates on this @helloanoop  ?


It seems somewhat misleading from a user's perspective because the documentation explicitly states that [the feature works](https://docs.usebruno.com/scripting/response-query.html) but it does not. Tell me if I can do anything to help 😄

Note: the workaround mentionned works fine

marshall-davis:
The `res` object doesn't seem to be created at all.

angiurgiu:
The fix doesn't work, I am still getting 'Error invoking remote method 'send-http-request': TypeError: res is not a function' on the latest Bruno version.

angiurgiu:
Another bug: the Post Request Console logs don't show up in the timeline. The Bruno tool is very difficult to use.

helloanoop:
@anusreesubash Please work with @sanjai0py and prioritise reviewing the PR #579

The `res` property should be a function which uses a custom query parser maintainer by us - see https://docs.usebruno.com/scripting/response/response-query

AcacioPN:
Any workarounds if res is not found at all? Or it not, anyone knows a version where this still worked?

anusree-bruno:
@AcacioPN 

Our team is working on this. 
Are you not able to access the response data using the work around anoop had suggested in this comment https://github.com/usebruno/bruno/issues/481#issuecomment-1753148441?

AcacioPN:
Hi

Yes, the workaround worked. I might have been doing something wrong, as res was recognized sometimes, and other times it wasn't.

[...]

PR #3710, opened 2024-12-30
feat: make `BrunoResponse` callable to access body data using expressions fixes (#481)

Pragadesh-45:
fixes: #481 

# Description

This PR introduces the ability to use the `res` object as a callable function to access body data directly. For example:  

```javascript
const page = res('page');
```

This functionality simplifies post-response scripting. Currently, developers has a  workaround:  

```javascript
const { get } = require('lodash');
const page = get(res.getBody(), 'page');
console.log('page', page);
```

With this PR, the `res` object can now be used both as an object and as a function, providing a more convenient way to scripts.
### Contribution Checklist:


packages/bruno-js/src/bruno-response.js
```diff
@@ -36,6 +44,10 @@ class BrunoResponse {
     this.body = data;
     this.res.data = data;
   }
+
+  call(expr, ...fns) {
+    return get(this.res.data, expr, ...fns);
```
Kenttleton: This doesn't satisfy the full feature. lodash.get does not support array filtering.

Kenttleton: This would be better off using JSONPath instead of lodash.

Kenttleton (changes requested):

Kenttleton:
Check out the examples given in the [docs](https://docs.usebruno.com/testing/script/response/response-query). The solution should satisfy the example inputs and outputs as written.

Pragadesh-45:
@Kenttleton Thanks for your valuable input. I will update this PR

Kenttleton:
I recommend using the JSONPath library instead of lodash as it gets you most of the way there. This lib is also used in Bruno to filter responses and has a similar syntax as the docs. Don't forget to add an optional callback to the signature for further refinements and please provide test cases for your implementation.

ramki-bruno (approved):

PR #3712, opened 2024-12-31
feat: open setting for folder and collection when

pooja-bruno:
# Description

when we are clicking on any collection or folder in the sidebar it should open that collection/folder setting

### Contribution Checklist:

https://github.com/user-attachments/assets/fa1c55e5-e10c-4c9f-a28b-6c7308b91687

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
@@ -128,13 +128,23 @@ const CollectionItem = ({ item, collection, searchText }) => {
       );
       return;
     }
+      dispatch(
+        addTab({
+          uid: item.uid,
+          collectionUid: collection.uid,
+          type: 'folder-settings'
+        })
+      );
+  };
+
+  const handleFolderCollapse = () => {
     dispatch(
       collectionFolderClicked({
         itemUid: item.uid,
         collectionUid: collection.uid
       })
     );
-  };
+  }
 
   const handleRightClick = (event) => {
     const _menuDropdown = dropdownTippyRef.current;
@@ -260,9 +270,6 @@ const CollectionItem = ({ item, collection, searchText }) => {
               })
             : null}
           <div
-            onClick={handleClick}
-            onContextMenu={handleRightClick}
-            onDoubleClick={handleDoubleClick}
             className="flex flex-grow items-center h-full overflow-hidden"
             style={{
               paddingLeft: 8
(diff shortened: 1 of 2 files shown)


rajib-raiyat:
@pooja-bruno, I noticed the recent addition of the feature to open settings for folders and collections when clicked in the sidebar on [v1.38.0](https://github.com/usebruno/bruno/releases/tag/v1.38.0). While I appreciate the effort, I'm curious about the reasoning behind this change. Could you provide some insight into the motivation for implementing this feature? It seems a bit disruptive to the current workflow, and I’d like to understand the benefits it brings.

Thanks!

CC: @lohit-bruno

pooja-bruno:
Hi Komol,

Thank you for reaching out and sharing your thoughts about the recent update.

We understand that this might feel like a shift in your workflow, and we appreciate your feedback. The main reason behind implementing this change was to address a common issue among users: many were unaware that folders and collections had customizable settings. By making these settings more accessible, we aim to help users discover and take full advantage of these features.

That said, we understand that this might feel disruptive to your workflow, and we’d love to hear more about how it’s impacting you. Could you share specific examples or scenarios where this change is causing challenges? Your input will help us refine the feature further and ensure it works better for everyone.

rajib-raiyat:
Hi @pooja-bruno,  

Thank you for explaining the motivation behind this feature. While I understand the intent to make users aware of customizable settings for folders and collections, the current implementation can lead to unintended friction in the workflow.  

Opening the settings tab each time a user clicks on a folder or collection feels excessive, especially considering that users typically interact with these elements primarily for navigation. This results in multiple unnecessary tabs being left open, disrupting the user's main task.  

Realistically, users only need to learn about the customizable settings once—either when the app is first installed, during a major update, or on rare occasions. To address this, I suggest implementing an **in-app showcase** or **guided onboarding experience**. These features can highlight specific functionalities with a distinctive and attractive overlay, ensuring users are aware of them without interrupting regular workflows.  

This approach strikes a better balance: it educates users about features while preserving the simplicity and fluidity of navigating folders and collections.  

To illustrate my concern, here’s a screenshot of my scenario: I was looking for the login API request, and after finding it, I noticed multiple unnecessary tabs that were opened while navigating folders and collections. This added overhead, as I had to manually close each tab afterward:  

image

pooja-bruno:
Hey @rajib-raiyat ,

Thanks for sharing your detailed feedback—it’s super helpful!

We totally get how opening multiple tabs while navigating folders or collections can be annoying. If that’s the main issue we’ve already made some changes so that you won’t have to deal with unnecessary tabs opening up anymore. You can check out the details in this PR: [#3831](https://github.com/usebruno/bruno/pull/3831)

Hopefully, this makes things smoother for you. Thanks for the suggestion about a guided onboarding or in-app showcase—we’ll keep it in mind for future updates.

If you still notice anything off or have more suggestions, feel free to share. We’re all ears! 

Thanks again for helping us make things better!

PR #3719, opened 2025-01-01
feat: bru util fns -- skipRequest, stopExecution, getTestResults, getAssertionResults, runRequest

lohxt1:
~ `bru.runner.skipRequest()` fn
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ skips the execution of the current request.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ works when included in the pre-request script.
~ `bru.runner.stopExecution()` fn
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ ends the collection run after completing the execution of the current request.
    
(the above two functions are only valid in a collection run context.)

~ `bru.getTestResults()` fn
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ returns a summary of the test results for the current request.
~ `bru.getAssertionResults()` fn
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ returns a summary of the assertion results for the current request.

~ `bru.runRequest(requestPathName)` fn
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ execute and obtain the results of a request in the collection.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;~ works when included in the pre-request, post-response or tests script.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -1184,20 +1184,6 @@ const registerNetworkIpc = (mainWindow) => {
               stopRunnerExecution = true;
             }
 
-            if (postRequestScriptResult?.skipRequest) {
-              // mainWindow.webContents.send('main:run-folder-event', {
-              //   type: 'runner-request-skipped',
-              //   error: 'Request has been skipped from post-response script',
-              //   responseReceived: {
-              //     status: 'skipped',
-              //     statusText: 'request skipped via post-response script'
-              //   },
-              //   ...eventData
-              // });
-              // currentRequestIndex++;
-              // continue;
-            }
-
             // run assertions
             const assertions = get(item, 'request.assertions');
             if (assertions) {


helloanoop (approved):

thiagomini:
Hey @lohxt1, is there any documentation for this `runRequest()` method? I installed the last version of Bruno, and it does not show up in my "Scripts" tab - where am I supposed to use them?

Its-treason:
> I'm not sure what I'm missing. Seems my request is not skipped.
> 
> image
> 
> I'm running Bruno 1.38.1

`bru.runner.skipRequest()` Only works inside the request runner. It does not really make sense when running a single request. You could throw an Error in the pre-request script, if you want to abort the request.

hartmark:
> > I'm not sure what I'm missing. Seems my request is not skipped.
> > image
> > I'm running Bruno 1.38.1
> 
> `bru.runner.skipRequest()` Only works inside the request runner. It does not really make sense when running a single request. You could throw an Error in the pre-request script, if you want to abort the request.

Ah, I was thinking it would always get skipped. Running from collection it indeed gets skipped.

image

Thanks!

zumbuschms:
I just tried these commands, but I seem to miss something. I have a collection with a bunch of sub folders. The first subfolder is to create a tenant in a multi tenant application, then there are a few folders for setup of that tenant and to run API tests and the last folder is to tear down the tenant again. I want to run the tear down commands only if the teardown folder is triggered in a separate runner task. 

Therefore I decided to register a variable during the setup phase and check for the existence of the variable in the first request of the tear down phase. If the variable to set to true, I want to skip the request, stop the runner and delete the variable. Based on the explaination of the new commands, I added this code to the pre-request script of the first tear down request:

```
if (bru.getVar("skipDeletion") == true) {
  bru.deleteVar("skipDeletion");
  bru.runner.stopExecution();
  bru.runner.skipRequest();
}
```

When I run the whole collection recursively, the first tear down request is skipped but the remaining requests are executed. If I reverse the order of the _stopExecution_ and _skipRequest_ commands, the request is not skipped and the runner is stopped after the first request. Are the stopExecution and skipRequest methods mutually exclusive and only the last called method is applied?

Issue #3754, opened 2025-01-08
[v1.38.0 regression] Can't upload certificate files anymore


emanuelle-jubinville:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

When trying to update key and certificate files of type "Cert" to my Bruno collection settings, I now have the error "certFilePath is a required field". 
It was possible to upload files in the v1.37.0. This looks like a regression. I also searched for more details in the documentation, but didn't find how to specify a path as requested. I tried from both a folder on my C drive and the download folder. I'm a Windows user (if it helps).

### Screenshots/Live demo link

![image](https://github.com/user-attachments/assets/6b1128e6-2f17-468f-845a-15eb8ecc7eab)

sreelakshmi-bruno:
Hi, thanks for reporting. We're working on this and will go out in a hotfix release soon. Thanks!

helloanoop:
Just merged the PR for the fix. We'll publish a hotfix with the next 2 hrs.

PR #3757, opened 2025-01-08
fix: file input element - replace the deprecated path prop

lohxt1:
fixes: #3754 

the nonstandard `path` property of the `Web File` object, added in early electron versions for convenience, has been removed starting with electron 32.0 due to security concerns and deviation from standards. It is now replaced by the [webUtils.getPathForFile](https://github.com/electron/electron/blob/main/docs/api/web-utils.md#webutilsgetpathforfilefile) method.

source: https://github.com/electron/electron/blob/main/docs/breaking-changes.md#removed-filepath


helloanoop:
Merged. Thanks @lohit-bruno !

Issue #3762, opened 2025-01-08
`Cannot find module 'iconv-lite'` for bruno-cli 1.38.2 release


chenrui333:
👋 for bruno-cli 1.38.2 release, seeing some runtime failure as 

```
==> /opt/homebrew/Cellar/bruno-cli/1.38.2/bin/bru --version
  node:internal/modules/cjs/loader:1409
    const err = new Error(message);
                ^
  
  Error: Cannot find module 'iconv-lite'
  Require stack:
  - /opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/src/utils/common.js
  - /opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/src/runner/run-single-request.js
  - /opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/src/commands/run.js
  - /opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/node_modules/yargs/index.cjs
  - /opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/src/index.js
  - /opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/bin/bru.js
      at Function._resolveFilename (node:internal/modules/cjs/loader:1409:15)
      at defaultResolveImpl (node:internal/modules/cjs/loader:1060:19)
      at resolveForCJSWithHooks (node:internal/modules/cjs/loader:10[65](https://github.com/Homebrew/homebrew-core/actions/runs/12675231749/job/35325971644?pr=203591#step:3:66):22)
      at Function._load (node:internal/modules/cjs/loader:1214:37)
      at TracingChannel.traceSync (node:diagnostics_channel:322:14)
      at wrapModuleLoad (node:internal/modules/cjs/loader:234:24)
      at Module.require (node:internal/modules/cjs/loader:1495:12)
      at require (node:internal/modules/helpers:135:16)
      at Object.<anonymous> (/opt/homebrew/Cellar/bruno-cli/1.38.2/libexec/lib/node_modules/@usebruno/cli/src/utils/common.js:1:15)
      at Module._compile (node:internal/modules/cjs/loader:1739:14)
      at Object..js (node:internal/modules/cjs/loader:1904:10)
[...]

PR #3767, opened 2025-01-09
fix: cli - added the missing iconv-lite import and removed other unused imports

lohxt1:
fixes: #3762

~ added the missing `iconv-lite` import for cli
~ removed unused library imports


helloanoop:
Merged. Thanks @lohxt1 !

PR #3768, opened 2025-01-09
fix: allow node core modules in developer mode

lohxt1:
~ enables usage of node's core modules in developer mode

node_core_modules_usage

### DIFF
--- a/packages/bruno-js/src/runtime/script-runtime.js
+++ b/packages/bruno-js/src/runtime/script-runtime.js
@@ -120,6 +120,7 @@ class ScriptRuntime {
       sandbox: context,
       require: {
         context: 'sandbox',
+        builtin: [ "*" ],
         external: true,
         root: [collectionPath, ...additionalContextRootsAbsolute],
         mock: {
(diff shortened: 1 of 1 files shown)


packages/bruno-js/src/runtime/script-runtime.js
```diff
@@ -120,6 +120,7 @@ class ScriptRuntime {
       sandbox: context,
       require: {
         context: 'sandbox',
+        builtin: [ "*" ],
```
ramki-bruno, line 123: maybe only if `allowScriptFilesystemAccess` is set? similar to how its done for the `fs` builtin-module?

packages/bruno-js/src/runtime/script-runtime.js
```diff
@@ -120,6 +120,7 @@ class ScriptRuntime {
...
         external: true,
         root: [collectionPath, ...additionalContextRootsAbsolute],
         mock: {
```
ramki-bruno, line 126: Note related to this PR. @helloanoop any reason why we use `mock` instead of `external`? for adding the default libs?

Pragadesh-45 (approved):

Issue #3723, opened 2025-01-03
Sending files using a multipart/form-data request fails with 1.37.0


bbak:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

I've a collection that uploads some Json files to some service, which was working flawlessly with 1.36.x but doesn't in 1.37.0.

The error message from the remote service hints at Bruno is not passing the contents of the specified file - in my case in "data" - to the service.

### .bru file to reproduce the bug

meta {
  name: Some Name
  type: http
  seq: 1
}

post {
  url: https://some.target.url/api/receive
  body: multipartForm
  auth: inherit
}

headers {
  Content-Type: multipart/form-data
}

body:multipart-form {
  id: 65e62dabcd3c585015b9f56f
  data: @file(C:\Temp\bruno_upload\some_file.json)
}

### Screenshots/Live demo link

<img width="1" alt="dot_black" src="https://github.com/user-attachments/assets/e8a83278-c50f-4247-ab47-1130d30e219f" />

mesopotamiaMD:
I'm a user and have just seen your case, as I have had a similar issue. In my case, it worked when i used relative path   

sourceFile: @file(../../Jobsfolder/test.txt) 

but you have to change it over the file mode. 

The relative path is added to the installation path, which makes no sense and needs to be fixed. Regards...

bbak:
Ah, interesting. Thanks.

Yet I believe, when using the UI to select a file, the absolute Path is written to the .bru file. Hence it needs to be fixed anyways.

bbak:
Hey, I already see a difference: I run Bruno in Developer Mode, since I need access to the filesystem. And I actually wonder how you can add a file from the filesystem while being in Safe (Sandbox) Mode. Anyway, here's a Video:

https://github.com/user-attachments/assets/6fe8123f-d8ef-4253-b41b-aba8079ab795

bbak:
Still present in 1.38.0

bill-wirz-iw:
I have also seen this behaviour. I was able to send a file using the `Content-Type: multipart/data` header (what @ganesh-bruno used). However, the data was not formatted with the correct boundaries and was rejected by the API. However, `Content-Type: multipart/form-data` does not work (same error as @bbak)

flleonx:
I'm experiencing the same problems as @bbak and @bill-wirz-iw in version 1.38.0

[...]

PR #3795, opened 2025-01-15
use lowercase header keys while making requests

lohxt1:
#3723 

https://datatracker.ietf.org/doc/html/rfc7230#section-3.2

workaround for the above issue:
use `content-type` instead of `Content-Type`

### DIFF
--- a/packages/bruno-electron/src/utils/collection.js
+++ b/packages/bruno-electron/src/utils/collection.js
@@ -7,7 +7,7 @@ const mergeHeaders = (collection, request, requestTreePath) => {
   let collectionHeaders = get(collection, 'root.request.headers', []);
   collectionHeaders.forEach((header) => {
     if (header.enabled) {
-      headers.set(header.name, header.value);
+      headers.set(header.name?.toLowerCase?.(), header.value);
       if (header?.name?.toLowerCase() === 'content-type') {
         contentTypeDefined = true;
       }
@@ -19,14 +19,14 @@ const mergeHeaders = (collection, request, requestTreePath) => {
       let _headers = get(i, 'root.request.headers', []);
       _headers.forEach((header) => {
         if (header.enabled) {
-          headers.set(header.name, header.value);
+          headers.set(header.name?.toLowerCase?.(), header.value);
         }
       });
     } else {
       const _headers = i?.draft ? get(i, 'draft.request.headers', []) : get(i, 'request.headers', []);
       _headers.forEach((header) => {
         if (header.enabled) {
-          headers.set(header.name, header.value);
+          headers.set(header.name?.toLowerCase?.(), header.value);
         }
       });
     }


helloanoop (approved):

Pragadesh-45 (approved):

Pragadesh44-Bruno (approved):

ramki-bruno:
Just checked the behavior of `curl` w.r.t. this and it doesn't seem to modify header-keys.
Which actually makes sense, even though the standard states that its supposed to be case-insensitive, we don't have to make sure of it.
IMO, its better to let it be up to the user and the server implementation.

> Check the end of the line above `<= Recv header..`
```
$ curl --trace - http://localhost:4000 -H "X-Foo: bar" -H "x-foo: Bar"
== Info:   Trying 127.0.0.1:4000...
== Info: Connected to localhost (127.0.0.1) port 4000 (#0)
=> Send header, 101 bytes (0x65)
0000: 47 45 54 20 2f 20 48 54 54 50 2f 31 2e 31 0d 0a GET / HTTP/1.1..
0010: 48 6f 73 74 3a 20 6c 6f 63 61 6c 68 6f 73 74 3a Host: localhost:
0020: 34 30 30 30 0d 0a 55 73 65 72 2d 41 67 65 6e 74 4000..User-Agent
0030: 3a 20 63 75 72 6c 2f 38 2e 31 2e 32 0d 0a 41 63 : curl/8.1.2..Ac
0040: 63 65 70 74 3a 20 2a 2f 2a 0d 0a 58 2d 46 6f 6f cept: */*..X-Foo
0050: 3a 20 62 61 72 0d 0a 78 2d 66 6f 6f 3a 20 42 61 : bar..x-foo: Ba
0060: 72 0d 0a 0d 0a                                  r....
<= Recv header, 17 bytes (0x11)
0000: 48 54 54 50 2f 31 2e 31 20 32 30 30 20 4f 4b 0d HTTP/1.1 200 OK.
0010: 0a                                              .
<= Recv header, 25 bytes (0x19)
0000: 43 6f 6e 74 65 6e 74 2d 54 79 70 65 3a 20 74 65 Content-Type: te
0010: 78 74 2f 68 74 6d 6c 0d 0a                      xt/html..
<= Recv header, 37 bytes (0x25)
0000: 44 61 74 65 3a 20 46 72 69 2c 20 31 37 20 4a 61 Date: Fri, 17 Ja
0010: 6e 20 32 30 32 35 20 30 38 3a 34 39 3a 35 33 20 n 2025 08:49:53
0020: 47 4d 54 0d 0a                                  GMT..
<= Recv header, 24 bytes (0x18)
0000: 43 6f 6e 6e 65 63 74 69 6f 6e 3a 20 6b 65 65 70 Connection: keep
0010: 2d 61 6c 69 76 65 0d 0a                         -alive..
<= Recv header, 23 bytes (0x17)
0000: 4b 65 65 70 2d 41 6c 69 76 65 3a 20 74 69 6d 65 Keep-Alive: time
0010: 6f 75 74 3d 35 0d 0a                            out=5..
<= Recv header, 28 bytes (0x1c)
0000: 54 72 61 6e 73 66 65 72 2d 45 6e 63 6f 64 69 6e Transfer-Encodin
0010: 67 3a 20 63 68 75 6e 6b 65 64 0d 0a             g: chunked..
<= Recv header, 2 bytes (0x2)
0000: 0d 0a                                           ..
<= Recv data, 19 bytes (0x13)
0000: 39 0d 0a 68 65 6c 6c 6f 2e 2e 2e 0a 0d 0a 30 0d 9..hello......0.
0010: 0a 0d 0a                                        ...
hello...
== Info: Connection #0 to host localhost left intact
```

ramki-bruno:
>  use content-type instead of Content-Type
@lohxt1 More details on how this is solving the issue would be helpful.

bbak:
Interestingly, according to [RFC9110 Section 5.1](https://datatracker.ietf.org/doc/html/rfc9110#section-5.1) (obsoletes RFC7230) these so-called field names are "case-insensitive". The [HTTP Field Name Registry](https://www.iana.org/assignments/http-fields/http-fields.xhtml), in turn, capitalizes the first character of every word and at times the whole field name.

By digging through the Axios sources, it seems like they [lowercase any field name in `normalizeHeader()`](https://github.com/axios/axios/blob/bad6d8b97b52c0c15311c92dd596fc0bff122651/lib/core/AxiosHeaders.js) internally while capitalizing the first character when constructing the HTTP message to send, so they look like in the HTTP Field Name Registry (see: `formatHeader()`).

From that perspective, it shouldn't make any difference. Yet most likely this is not the whole story.

ramki-bruno:
> Interestingly, according to [RFC9110 Section 5.1](https://datatracker.ietf.org/doc/html/rfc9110#section-5.1) (obsoletes RFC7230) these so-called field names are "case-insensitive". The [HTTP Field Name Registry](https://www.iana.org/assignments/http-fields/http-fields.xhtml), in turn, capitalizes the first character of every word and at times the whole field name.
> 
> By digging through the Axios sources, it seems like they [lowercase any field name in `normalizeHeader()`](https://github.com/axios/axios/blob/bad6d8b97b52c0c15311c92dd596fc0bff122651/lib/core/AxiosHeaders.js) internally while capitalizing the first character when constructing the HTTP message to send, so they look like in the HTTP Field Name Registry (see: `formatHeader()`).
> 
> From that perspective, it shouldn't make any difference. Yet most likely this is not the whole story.
> 
> The [npm:form-data package](https://github.com/form-data/form-data/blob/master/lib/form_data.js) seems to use lowercase headers, though (see: `FormData.prototype.getHeaders()`).

ok, that is interesting. Looks like all the implementation work under the assumption that its case-insensitive.
Not sure why axios internally is capitalizing it that way, maybe title-case kinda looks good?

ramki-bruno:
Anyway, I was checking why it regressed and looks like its due to the changes introduced in two PRs #3142 and #3586 which eventually created this break.
Basically, there are some inconsistencies w.r.t. case-sensitivity for checking the existence of `content-type` header and the form-data is not getting added to request if the header is added explicitly with uppercase characters in the key.

I think this PR should fix that inconsistency. Also, since making the header key/field-name lowercase shouldn't make any difference, I think we should be good to merge this.

ramki-bruno (approved):

PR #3830, opened 2025-01-17
Fix: Import collection modal closes when click on ENTER

sanish-bruno:
# Description

There are two modals in question, one is import modal and the modals that gets opened when we click on each of the type. our common modal component have event listeners for <b>ESC</b> , <b>ENTER</b> keys, outer modal has "onClose" which gets triggered when user press <b>ENTER</b> , this closes the outer modal, as the inner modals are within the outer modal, those also gets removed from the dom. 

currently i removed handleConfirm function for outer modal, so that Pressing on ENTER to not to trigger modal closing, ESC closes the modal as before. also made the handleConfirm prop optional

Additionaly, the  <b>handleConfirm</b> function we use within the callback for key-press event is not updated as it is not added as the dependency for useEffect, they continue to use the older handleConfirm, and wasn't triggering cloning or import if i press enter after putting in the url

https://github.com/user-attachments/assets/46e2c756-6e1e-4787-b884-daad5e90a0ea

### Contribution Checklist:

Screenshot 2025-01-17 at 12 32 43 PM
Screenshot 2025-01-17 at 12 32 48 PM
Screenshot 2025-01-17 at 12 32 54 PM

### DIFF
--- a/packages/bruno-app/src/components/Modal/index.js
+++ b/packages/bruno-app/src/components/Modal/index.js
@@ -62,7 +62,7 @@ const Modal = ({
   confirmText,
   cancelText,
   handleCancel,
-  handleConfirm,
+  handleConfirm = () => {},
   children,
   confirmDisabled,
   hideCancel,
@@ -92,7 +92,7 @@ const Modal = ({
   };
 
   useFocusTrap(modalRef);
-  
+
   const closeModal = (args) => {
     setIsClosing(true);
     setTimeout(() => handleCancel(args), closeModalFadeTimeout);
@@ -103,7 +103,7 @@ const Modal = ({
     return () => {
       document.removeEventListener('keydown', handleKeydown);
     };
-  }, [disableEscapeKey, document]);
+  }, [disableEscapeKey, document, handleConfirm]);
 
   let classes = 'bruno-modal';
   if (isClosing) {
--- a/packages/bruno-app/src/components/Sidebar/ImportCollection/index.js
+++ b/packages/bruno-app/src/components/Sidebar/ImportCollection/index.js
@@ -68,7 +68,7 @@ const ImportCollection = ({ onClose, handleSubmit }) => {
     );
   };
   return (
-    <Modal size="sm" title="Import Collection" hideFooter={true} handleConfirm={onClose} handleCancel={onClose}>
+    <Modal size="sm" title="Import Collection" hideFooter={true} handleCancel={onClose}>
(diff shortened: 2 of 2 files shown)


helloanoop (approved):

packages/bruno-app/src/components/Modal/index.js
```diff
@@ -103,7 +103,7 @@ const Modal = ({
     return () => {
       document.removeEventListener('keydown', handleKeydown);
     };
-  }, [disableEscapeKey, document]);
+  }, [disableEscapeKey, document, handleConfirm]);
```
ramki-bruno, line 106: This should have just `handleKeydown` itself in the dependenc-array, instead of manually finding all the dependencies of its dependencies.
Its still missing, `closeModal` at this moment.

packages/bruno-app/src/components/Modal/index.js
```diff
@@ -62,7 +62,7 @@ const Modal = ({
   confirmText,
   cancelText,
   handleCancel,
-  handleConfirm,
+  handleConfirm = () => {},
```
ramki-bruno, line 65: Just a quick check around the code will show that this was already optional and this change is not necessary.

sanish-bruno: i thought about just removing it initially, without adding a default value for the prop, it would have worked fine keypress "Enter" case, it is bounded by the if check, but no such condtional is there when is called from modal footer. @ramki-bruno

packages/bruno-app/src/components/Modal/index.js
sanish-bruno: i will have to move handleKeyDown to a useCallback function with handleConfirm, closeModal, disableEscKey as deps, before adding it as a dependency right? else, the useEffect will get called with each render?

packages/bruno-app/src/components/Modal/index.js
ramki-bruno: `undefined` for `onClick` would be absolutely fine

packages/bruno-app/src/components/Modal/index.js
ramki-bruno: > i will have to move handleKeyDown to a useCallback function with handleConfirm, closeModal, disableEscKey as deps, before adding it as a dependency right?

yes, that will be even better
<br/>
> else, the useEffect will get called with each render?

yes, but at least it won't have any unexpected behaviors

PR #3831, opened 2025-01-17
Feature: Replaceable tab management system

naman-bruno:
# Description

This PR introduces a restructuring of the tab management system. Previously, every file, folder, or collection would open in a new tab by default, leading to potential tab overload. With this update, tabs now operate in a replaceable mode, meaning new files or collections will replace the currently open tab unless explicitly made permanent. To make a tab permanent (or "sticky"), users can simply double-click on the file name or the tab itself

### Contribution Checklist:

- [] **Create an issue and link to the pull request.**

https://github.com/user-attachments/assets/c37594ef-a956-4f15-ab28-87d3e8bd4ac9


pooja-bruno:
LGTM

packages/bruno-app/src/providers/ReduxStore/slices/tabs.js
```diff
@@ -124,6 +152,13 @@ export const tabsSlice = createSlice({
       const collectionUid = action.payload.collectionUid;
       state.tabs = filter(state.tabs, (t) => t.collectionUid !== collectionUid);
       state.activeTabUid = null;
+    },
+    makeTabPermanent: (state, action) => {
+      const { uid } = action.payload;
+      const tab = find(state.tabs, (t) => t.uid === uid);
+      if (tab) {
+        tab.preview = false;
+      }
```
ramki-bruno, line 151: Add an error log in else-case as that should not happen ideally and will help indicate a bug.

packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
```diff
@@ -73,13 +73,13 @@ const RequestTab = ({ tab, collection, tabIndex, collectionRequestTabs, folderUi
   if (['collection-settings', 'collection-overview', 'folder-settings', 'variables', 'collection-runner', 'security-settings'].includes(tab.type)) {
     return (
       <StyledWrapper
-        className="flex items-center justify-between tab-container px-1"
+        className={`flex items-center justify-between tab-container px-1 ${tab.preview ? "italic" : ""}`}
```
ramki-bruno, line 76: [`classnames`](https://github.com/JedWatson/classnames) lib should make this easier

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -180,6 +218,12 @@ const CollectionItem = ({ item, collection, searchText }) => {
     }
   }
 
+  const handleDoubleClick = (event) => {
+    if(!isFolder){
+      dispatch(makeTabPermanent({ uid: item.uid }))
```
ramki-bruno: I think we can make the double-click on the Folder work.
- double-click opens the folder if not already open and makes it permanent
- single click is as usual, toggle open/close

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -97,12 +99,48 @@ const CollectionItem = ({ item, collection, searchText }) => {
       })
     );
   };
+  const getItemByUid = (uid, collection) => {
```
ramki-bruno: you can move this to `packages/bruno-app/src/utils/collections/index.js`

ramki-bruno: wait, looks like this function is not used anywhere right now, in that case you can remove it

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -97,12 +99,48 @@ const CollectionItem = ({ item, collection, searchText }) => {
...
+  
+    const isRequest = isItemARequest(item);
+  
+    // Determine whether to replace an existing tab
+    let replaceTabUid = null;
+  
+    // Find any replaceable tab that can be replaced
+    for (let tab of tabs) {
+      if (tab.preview) {
+        if (!collection) continue;
```
ramki-bruno: won't this check  be redundant since `collection` won't be a nullable anyways? Or is there any case where it can be a nullable?

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -97,12 +99,48 @@ const CollectionItem = ({ item, collection, searchText }) => {
...
+  
+        const tabItem = findItemInCollection(collection, tab.uid);
+        if (tab.type !== "collection-settings" && !tabItem) continue;
```
ramki-bruno: You can leave a comment explaining the above conditional(intuiting [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) takes energy).
Something like "Replace the tab only if its collection-settings or other items in the file tree"

ramki-bruno: also its not `tabItem` exactly, `collectionItem` might be better

packages/bruno-app/src/providers/ReduxStore/slices/tabs.js
```diff
@@ -19,31 +19,59 @@ export const tabsSlice = createSlice({
...
-        collectionUid: action.payload.collectionUid,
+        uid,
+        collectionUid,
         requestPaneWidth: null,
-        requestPaneTab: action.payload.requestPaneTab || 'params',
+        requestPaneTab: requestPaneTab || 'params',
         responsePaneTab: 'response',
-        type: action.payload.type || 'request',
-        ...(action.payload.uid ? { folderUid: action.payload.uid } : {})
+        type: type || 'request',
```
ramki-bruno, line 56: There are still places where `addTab` is called without `type`, not sure if it'll be good idea to default it to `request`

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
```diff
@@ -59,26 +60,51 @@ const Collection = ({ collection, searchText }) => {
     'rotate-90': !collectionIsCollapsed
   });
 
+  const scrollToTheActiveTab = () => {
```
ramki-bruno: Variable, Runner etc. still lack this, I guess that deserves a separate PR

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
```diff
@@ -59,26 +60,51 @@ const Collection = ({ collection, searchText }) => {
...
+  
+    // Find any replaceable tab that can be replaced
+    for (let tab of tabs) {
+      if (tab.preview) {
+        const tabItem = tab.type === "collection-settings" ? collection : findItemInCollection(collection, tab.uid);
+  
+        if (tab.type !== "collection-settings" && !tabItem) continue;
+  
+        if (tabItem?.draft) continue;
+
```
ramki-bruno: If possible, move this logic to a until function so that it can be documented(via comments) properly. Also to maintain consistent behaviour.

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
ramki-bruno: Also, move this to a util function and you can replace the other instance in `CollectionItem/index.js` with the same. Clearly, we will be using it in another place in future and we have to fix the vertical scroll as well. So better have it as a util function.

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
```diff
@@ -59,26 +60,51 @@ const Collection = ({ collection, searchText }) => {
...
     dispatch(collapseCollection(collection.uid));
-    
-    // Only open collection settings if not clicking the chevron
+  
     if(!isChevronClick) {
       dispatch(
         addTab({
-          uid: uuid(),
+          uid: collection.uid,
```
ramki-bruno, line 83: +1

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
```diff
@@ -166,7 +192,7 @@ const Collection = ({ collection, searchText }) => {
             className={`chevron-icon ${iconClassName}`}
             style={{ width: 16, minWidth: 16, color: 'rgb(160 160 160)' }}
           />
-          <div className="ml-1" id="sidebar-collection-name">
+          <div className="ml-1 w-full" id="sidebar-collection-name">
```
ramki-bruno: Please share on why this is done

packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
naman-bruno: Yep, I agree that a utility function would be useful here. However, I think we should handle it in a separate PR since this pattern is used in multiple places.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
naman-bruno: Removed

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
naman-bruno: Yep you are right, removed this check

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
naman-bruno: Added comment and renamed const

packages/bruno-app/src/providers/ReduxStore/slices/tabs.js
naman-bruno: There are only two places where we call `addTab` without a type, and both are used specifically to open request tabs. This condition has also been in place for a long time, so the approach seems reasonable.

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
naman-bruno: Previously, clicking on the chevron and the name had separate functions, and clicking on the blank space after the name didn’t register the click. To fix that, I made this change. However, the issue has already been addressed in a different branch by introducing isChevronClick.

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
naman-bruno: moving it to tab utils

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
ramki-bruno: > double-click opens the folder if not already open and makes it permanent

for now, double-click opens or closes along with making the tab permanent, since waiting to register the second click will add delay in response for the single-click and make the UI feel slower.

ramki-bruno (approved):

helloanoop (approved):

PR #3834, opened 2025-01-17
async bru file parsing based on request/collection size

lohxt1:
wip

~ based on worker threads


packages/bruno-app/src/components/RequestTabPanel/index.js
```diff
@@ -175,54 +182,105 @@ const RequestTabPanel = () => {
...
-                onSchemaLoad={onSchemaLoad}
-                toggleDocs={toggleDocs}
-                handleGqlClickReference={handleGqlClickReference}
-              />
-            ) : null}
-
-            {item.type === 'http-request' ? (
-              <HttpRequestPane item={item} collection={collection} leftPaneWidth={leftPaneWidth} />
-            ) : null}
+      {item?.partial ?
```
helloanoop: Let's move this logic to a new component under `RequestTabPanel` called as `RequestNotLoaded`
This should help contain the specific logic around loading large requests as well as keep the current code here short and easy to maintain.

I also recommend doing an early return instead of a conditional like below

```js
if(item?.partial) {
  return <RequestNotLoaded item={item} collection={collection}>
}
```

packages/bruno-app/src/components/RequestTabs/RequestTab/SpecialTab.js
```diff
@@ -13,6 +13,14 @@ const SpecialTab = ({ handleCloseClick, type, tabName }) => {
           </>
         );
       }
+      case 'collection-load-stats': {
```
helloanoop: This should be `collection-overview`
Please change it across all places where you are using `collection-load-stats`

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/CollectionItemLoading/index.js
```diff
@@ -0,0 +1,32 @@
+import { useSelector } from "react-redux";
```
helloanoop: I don't see this component - `CollectionItemLoading` used anywhere. Please delete it.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -294,12 +293,20 @@ const CollectionItem = ({ item, collection, searchText }) => {
...
+              {item?.partial && !item?.loading ? 
+                <IconRefresh size={18} className="w-fit mr-2" strokeWidth={1.5} />
+              : null}
+              {item?.loading ? 
+                <IconLoader2 className="animate-spin w-fit mr-2" size={18} strokeWidth={1.5} />
+              : null}
+              {!item?.partial && !item?.loading ? 
+                <RequestMethod item={item} />
+              : null}
               <span className="item-name" title={item.name}>
```
helloanoop, line 302: The logic above is slightly hard to read. Consider moving it to a component in the same file like below

```js
const ItemIcon = ({ item }) => {
  if (item?.loading) {
    return <IconLoader2 className="animate-spin w-fit mr-2" size={18} strokeWidth={1.5} />;
  }

  if (item?.partial) {
    return <IconRefresh size={18} className="w-fit mr-2" strokeWidth={1.5} />;
  }

  return <RequestMethod item={item} />;
};
```

helloanoop: Also, lets use `IconAlertTriangle` with the theme yellow color. We want to indicate an warning instead of showing an alert icon

image

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -211,6 +211,60 @@ const safeToRename = (oldPath, newPath) => {
   }
 };
```
helloanoop, line 213: Recommended
```js
const getCollectionStats = async (directoryPath) => {
  let size = 0;
  let filesCount = 0;
  let maxFileSize = 0;

  async function calculateStats(directory) {
    const entries = await fsPromises.readdir(directory, { withFileTypes: true });

    const tasks = entries.map(async (entry) => {
      const fullPath = path.join(directory, entry.name);

      if (entry.isDirectory()) {
        if (['node_modules', '.git'].includes(entry.name)) {
          return;
        }

        return await calculateStats(fullPath);
      }

      if (path.extname(fullPath) === '.bru') {
        const stats = await fsPromises.stat(fullPath);
        size += stats?.size;
        if (maxFileSize < stats?.size) {
          maxFileSize = stats?.size;
        }
        filesCount += 1;
      }
    });

    await Promise.all(tasks);
  }

  await calculateStats(directoryPath);

  size = sizeInMB(size);
  maxFileSize = sizeInMB(maxFileSize);

  return { size, filesCount, maxFileSize };
}
```

helloanoop (changes requested):
Please also make changes as we discussed to improve the worker thread implementation.

Issue #3820, opened 2025-01-16
Redirect does not check Proxy


koehdaniel:
### I have checked the following:

- [x] I use the newest version of bruno.
- [x] I've searched existing issues and found nothing related to my issue.

### Describe the bug

I am requesting a site that is excluded form the Proxy by setting the "Proxy Bypass" in the settings or the NO_PROXY Env-Var.
That site then sends a 302 Found-Redirect to a URL that needs a Proxy.
The automatic redirect from Bruno does not use a Proxy for the redirected Request and results in a timeout.

# Test Setup:  
* Request to https://github.com/usebruno/bruno/releases/download/v1.38.1/latest.yml
* Redirects with a 302 to https://objects.githubusercontent.com/github-production-release-asset-2e65be/542284380/be84cca3-08a2-4334-b3b3-f2ac73442ec2
* The Proxy is set to an invalid URL "invalid-proxy:1234" with bypassProxy to "github.com"

I am expecting this Test to fail, since the redirected URL should use the invalid Proxy.
The test will pass which shows, that the Proxy is not used for the redirected Request to objects.githubusercontent.com

### .bru file to reproduce the bug

Test-Case:
```
meta {
  name: TestRedirect
  type: http
  seq: 1
}

get {
  url: https://github.com/usebruno/bruno/releases/download/v1.38.1/latest.yml
  body: none
  auth: none
}

```

Proxy-Config:
 ```
 "proxy": {
    "enabled": true,
    "protocol": "http",
    "hostname": "invalid-proxy",
    "port": 1234,
    "auth": {
      "enabled": false,
      "username": "",
      "password": ""
    },
    "bypassProxy": "github.com"
  }
```

### Screenshots/Live demo link

![Image](https://github.com/user-attachments/assets/4fdf0ccc-2de8-44e6-a386-a2b03e6777b8)

![Image](https://github.com/user-attachments/assets/6a1540fe-dec6-4149-bf89-6e26c560088f)

koehdaniel:
fyi: I used a workaround to get this working for me in my reallife scenario.
The first requests uses `req.setMaxRedirects(0);` to fetch the URL from the header "location" and stores it in a variable.
Then there is a second request that uses the URL previously stored in the variable.
Now bruno successfully uses the Proxy for objects.githubusercontent.com.

PR #3846, opened 2025-01-20
Fix: API redirects does not use Proxy and redirect cookies are not stored

naman-bruno:
# Description

Issue: #3820, #1332
Fix: When requesting a site excluded from the Proxy by configuring the "Proxy Bypass" in the settings, the site sends a 302 Found redirect to a URL that requires a Proxy. However, the automatic redirect from Bruno does not utilize the Proxy for the redirected request, leading to a timeout. Also cookies returned by redirect api does not get stored

### Contribution Checklist:

https://github.com/user-attachments/assets/5930534a-9d1c-4eaa-b78f-c1695effb1cb


packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -317,6 +317,63 @@ const configureRequest = async (
     }
   }
 
+  axiosInstance.interceptors.response.use(
```
lohxt1: this interceptor logic could be moved into the existing interceptor logic in the `makeAxiosInstance` function ?

`ipc/network/axios-instance.js`

Issue #3864, opened 2025-01-22
Crash when moving folders


Its-treason:
### I have checked the following:

- [x] I use the newest version of bruno.
- [x] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Crash does not trigger on every move. But on most and with multiple collection.

### Screenshots/Live demo link

https://github.com/user-attachments/assets/331aed41-969d-405d-b020-c06d8f55937c

sanjaikumar-bruno:
Thank you for reporting this, Timon. I am currently debugging the issue.

sanjaikumar-bruno:
This issue is no longer reproducible in the latest version of Bruno.

PR #3866, opened 2025-01-23
Fix/ Handle Folder settings stale `uids` (fixes: #3864)

Pragadesh-45:
fixes: #3864 
# Description

This PR implements a fix for the app crash issue that occurs when headers are opened in folder settings, and then we try to move that folder. The issue is caused by using stale header UIDs in non-existent folders. The only action allowed in this case should be closing the stale tab.

Added auto-close to close stale tabs.

### Contribution Checklist:

Before:

https://github.com/user-attachments/assets/87c58eff-1005-4232-bf47-2cba8c184a76

After:

https://github.com/user-attachments/assets/a860a8b7-8353-4ce6-bfde-251425b2f6db


helloanoop (changes requested):
I am not happy with so many conditionals being introduced to fix this issue. I'd like to understand the solutioning here better.

@Pragadesh44-Bruno lets discuss this once you are back from vacation next week.

helloanoop:
Discussed with @Pragadesh-45 that we need to rework on the fix at the electron layer.
Closing this PR. Will re-open if needed.

PR #3867, opened 2025-01-23
Feat/oauth2 improvements

helloanoop:
OAuth2 Revamp


packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/StyledWrapper.js
```diff
@@ -11,6 +11,47 @@ const Wrapper = styled.div`
     border: solid 1px ${(props) => props.theme.input.border};
     background-color: ${(props) => props.theme.input.bg};
   }
+
+  .token-placement-selector {
```
pietrygamat, line 15: How about extracting  token placement with corresponding HeaderPrefix/QueryParam field into a separate component? Between CSS, change handlers, dropdownRefs and currently hardcoded translatable strings, it's a lot of duplication.

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/GrantTypeSelector/index.js
```diff
@@ -46,15 +45,33 @@ const GrantTypeSelector = ({ item, collection }) => {
...
+            accessTokenUrl: '',
+            username: '',
+            password: '',
+            clientId: '',
+            clientSecret: '',
+            scope: '',
+            credentialsId: 'credentials',
+            tokenPlacement: 'header',
+            tokenPrefix: 'Bearer',
+            tokenQueryParamKey: 'access_token',
```
pietrygamat: Both of these default values are saved in bru file, even if the `tokenPlacement` option only needs one. It would be better if unused values are not automatically pushed into request configurations.

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2TokenViewer/index.js
```diff
@@ -0,0 +1,169 @@
...
+    return `${mins}m ${secs}s`;
+  } else {
+    const hours = Math.floor(seconds / 3600);
+    const mins = Math.floor((seconds % 3600) / 60);
+    return `${hours}h ${mins}m`;
+  }
+};
+
+const ExpiryTimer = ({ initialExpiresIn }) => {
+  const [timeLeft, setTimeLeft] = useState(initialExpiresIn);
```
pietrygamat: This timer resets when changing tabs.

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/index.js
```diff
@@ -1,28 +1,55 @@
...
+    try {
+      await dispatch(fetchOauth2Credentials({ request: requestCopy, collection }));
+      toggleFetchingToken(false);
+    }
+    catch(error) {
+      console.error('could not fetch the token!');
+      console.error(error);
+      toggleFetchingToken(false);
+    }
+  }
```
pietrygamat: Indicate a token fetch success with a toast?

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/index.js
```diff
@@ -79,12 +117,21 @@ const OAuth2AuthorizationCode = ({ item, collection }) => {
...
+        <div className="flex items-center px-2.5 py-1.5 bg-indigo-50/50 dark:bg-indigo-500/10 rounded-md">
+          <IconSettings size={14} className="text-indigo-500 dark:text-indigo-400" />
+        </div>
+        <span className="text-sm font-medium">
+          Configuration
+        </span>
+      </div>
       {inputsConfig.map((input) => {
         const { key, label, isSecret } = input;
         return (
```
pietrygamat, line 206: Please, make this section collapsible. The Oauth2TokenViewer is so pretty to look at, the users may want to have in view all the time, but they may sometimes have to scroll down to the Get Token button.

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -1192,4 +1193,41 @@ export const hydrateCollectionWithUiStateSnapshot = (payload) => (dispatch, getS
...
+  })    
+}
+
+export const clearOauth2Cache = (payload) => async (dispatch, getState) => {
+  const { collectionUid, url, credentialsId } = payload;
+  return new Promise((resolve, reject) => {
+    const { ipcRenderer } = window;
+    ipcRenderer.invoke('clear-oauth2-cache', collectionUid, url, credentialsId).then(resolve => {
+      dispatch(collectionClearOauth2CredentialsByUrl({ collectionUid, url, credentialsId }));
+      resolve();
```
pietrygamat: I'm getting a toast: Resolve is not a function.

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/index.js
```diff
@@ -40,6 +67,11 @@ const OAuth2AuthorizationCode = ({ item, collection }) => {
           state,
           scope,
           pkce,
+          credentialsId,
+          tokenPlacement,
+          tokenPrefix,
+          tokenQueryParamKey,
+          reuseToken,
```
pietrygamat: There's a lot of places where the new fields need keeping track of. It seems it can be replaced with simply:
```
dispatch(updateAuth({
      mode: 'oauth2',
      collectionUid: collection.uid,
      itemUid: item.uid,
      content: {
        ...oAuth,
        [key]: value
      }
    }))

```

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/index.js
```diff
@@ -100,17 +147,94 @@ const OAuth2AuthorizationCode = ({ item, collection }) => {
...
+      </div>
+      <div className="flex items-center gap-4 w-full" key={`input-token-placement`}>
+        <label className="block min-w-[140px]">Add token to</label>
+        <div className="inline-flex items-center cursor-pointer token-placement-selector">
+          <Dropdown onCreate={onDropdownCreate} icon={<Icon />} placement="bottom-end">
+            <div
+              className="dropdown-item"
+              onClick={() => {
+                dropdownTippyRef.current.hide();
+                handleChange('tokenPlacement', 'header');
```
pietrygamat, line 288: Can these enum-like values: `header`, `url`  be externalized somehow? They are used in views, in bru files, schema definitions, soon - as keys in translation files. Maybe this refactor is a good opportunity to take care of that as well?

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  requestCopy.method = 'POST';
+  requestCopy.headers['content-type'] = 'application/x-www-form-urlencoded';
+  requestCopy.headers['Accept'] = 'application/json';
+  requestCopy.data = data;
+  requestCopy.url = url;
+
+  const axiosInstance = makeAxiosInstance();
+  const response = await axiosInstance(requestCopy);
```
pietrygamat: This request fails due to certificate validation issues (expired certificate), despite having disabled SSL/TLS Certificate verification. Note, the electron browser authentication does work (login prompt is shown). The problem is also that it fails silently - no toast, no timeline information about the cause.

pietrygamat: This request is sent with Header:
```
Authorization: 'Basic Og==' 
```
There are several places in the code like this, where _any_ auth is confused with Basic auth resulting in storing base64 representation of ":" in Authorization header.  

```
  // todo: we have things happening in two places w.r.t basic auth
  // need to refactor this in the future
  // the request.auth (basic auth) object gets set inside the prepare-request.js file
  if (request.auth) {
    const username = _interpolate(request.auth.username) || '';
    const password = _interpolate(request.auth.password) || '';
    // use auth header based approach and delete the request.auth object
    request.headers['Authorization'] = `Basic ${Buffer.from(`${username}:${password}`).toString('base64')}`;
    delete request.auth;
  }
  ```
Before implementing #2106 this is critical, because many servers will look at Basic Auth rather than post body for client credentials, which are part of the OAuth2 flow. I think this refactor is also the time to resolve these TODOs.

**Comment by pietrygamat** [comment]
The UI here looks great and token decoder is way ahead of what the competition offers, so make sure to post a promotional video after releasing this!
:1st_place_medal:

**Review by pietrygamat — CHANGES_REQUESTED** [review]

File: packages/bruno-electron/src/utils/oauth2.js [hunk 10]
```diff
@@ -125,34 +125,42 @@ const getOAuth2TokenUsingClientCredentials = async ({ request, collectionUid, fo
...
   requestCopy.headers['Accept'] = 'application/json';
-  requestCopy.data = data;
+  if (credentialsPlacement == "basic_auth_header") {
+    requestCopy.headers['Authorization'] = `Basic ${Buffer.from(`${clientId}:${clientSecret}`).toString('base64')}`;
+  }
+  else {
+    const data = {
+      grant_type: 'client_credentials',
+      client_id: clientId,
+      client_secret: clientSecret
```
pietrygamat: Only 
```
if (credentialsPlacement === "body"
```

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/PasswordCredentials/index.js
```diff
@@ -118,6 +129,31 @@ const OAuth2PasswordCredentials = ({ save, item = {}, request, handleRun, update
...
+            </div>
+            <div
+              className="dropdown-item"
+              onClick={() => {
+                dropdownTippyRef.current.hide();
+                handleChange('credentialsPlacement', 'basic_auth_header');
+              }}
+            >
+              Basic Auth Header
+            </div>
```
pietrygamat, line 197: Sending client creds in Authorization header is the preferred choice (servers MUST support it), and including them in request body is optional alternative (servers MAY support it). So it would make sense for Bruno to default to `basic_auth_header` and have it as a first option.

pietrygamat (changes requested):

bastian-worx:
Fwiw -- I am successfully using this code for our API and am looking forward to its availibility in a release 👍

NewtonChutney:
Hello, thanks for this feature!

I was recently checking out the documenation and came across oauth 2 experimental preview..
The doc redirects to the releases page in which new Windows preview builds aren't placed.
Please update the link in the banner [here](https://docs.usebruno.com/auth/oauth2-experimental/overview#:~:text=You%20can%20download) to this [page](https://github.com/usebruno/bruno-nightly-builds?tab=readme-ov-file#nightly-builds)

Issue #3147, opened 2024-09-19
Position on response body resets if moving to another tab/request


Nahuel92:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Hi.

When I issue a request and scroll down through the response body, if I switch to another request and come back to the first one (or if I click on response's headers, Timeline or Tests tab), Bruno lost track of the position where I left.

I added a bunch of screenshots to help you guys understand of the issue. Let me know if more clarification is needed.

I used Mockoon to simulate an API server, and https://json-generator.com/# to generate a test JSON response.

Thanks!

### Screenshots/Live demo link

![1 1](https://github.com/user-attachments/assets/55ad5f0d-00e3-4031-9b63-4131456b99ad)
![1 2](https://github.com/user-attachments/assets/3e7a5e63-a6e3-46f2-bbdb-26da62254b57)
![1 3](https://github.com/user-attachments/assets/338bca22-e50c-4da1-911f-bc3749c5bf52)
![2 1](https://github.com/user-attachments/assets/95ab02e8-7029-4ae6-a9b1-e64cdd4739e3)
![2 2](https://github.com/user-attachments/assets/410ad9f8-cb1f-4542-9e16-d1cc57d6c189)
![2 3](https://github.com/user-attachments/assets/bd3b6ffe-1d76-4287-a635-1baf9f3b10cb)

Ved504:
Hey @Nahuel92, We appreciate you highlighting this issue. Our team will look into it shortly.

terminal-overflow:
Hey! FYI I think this is the same issue as https://github.com/usebruno/bruno/issues/2742. I do agree though, it'd be great to have this looked into.

anup-dey:
Is there any update on the issue?

terminal-overflow:
@anup-dey I created a [PR](https://github.com/usebruno/bruno/pull/3902) in response to the related issue(s). I'm waiting a review/feedback from maintainers.

helloanoop:
Thanks for the PR in @terminal-overflow 
It has been merged 🎉 

Excited to have this go out in our upcoming release - `v2.11.0` schedule to go out on Thu Sep 9th 2025

kevinlarimar:
I'd like to report that this issue has not been fully resolved in v3.5.0. When using the GraphQL query builder, switching tabs and returning to the builder causes the scroll position to reset to the top and all expanded sections to collapse.

[...]

Issue #2742, opened 2024-08-01
Response Scroll Not Preserved When Changing Requests


denismihaylovsap:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

First I would like to thank you for the awesome software you are providing! We are using it daily and we are very happy with it.

The only issue that is very frustrating is that Bruno does not keep the scrolling in the Response box after you navigate to other requests. When we return a list in JSON and then you scroll to the middle and find an ID which is used for another request, when you go back to the list your scroll is gone and you need to find the record again.

### Mockups or Images of the feature

<img width="2359" alt="Screenshot 2024-08-01 at 20 19 47" src="https://github.com/user-attachments/assets/3c36c3f4-8289-4b6f-b381-65a0a3a87cf5">
<img width="2356" alt="Screenshot 2024-08-01 at 20 20 00" src="https://github.com/user-attachments/assets/1f098622-f7a8-464d-8db1-193f5f3d71b5">
<img width="2356" alt="Screenshot 2024-08-01 at 20 20 11" src="https://github.com/user-attachments/assets/82b42ba3-e32e-409c-831b-f9d7ac89dbc4">

SPIRIT-00:
Same thing in runner, hard to work when run about 100-200 request and look for errors.

Issue #3431, opened 2024-11-05
Track expansion state of tree views for tabs that are out of view


fschwiet:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

When I change tabs Bruno should recall the expansion state of the view within a tab when I return to that tab.

Let me describe an example to help make that clear. As part of debugging my app, I am trying to manually compare specific parts of different responses coming from my app. My collection has two requests defined. I run one, and then use the expand/collapse to find the part I want to compare. I then run the other request and then use the expand/collapse to find the part of the response I want to compare. However now when I switch back to view the results of the first request the expand/collapse state has been reset, so I don't have the part I wanted to compare ready anymore.

### Mockups or Images of the feature

![remember expansion state](https://github.com/user-attachments/assets/bf5b57c2-5f2f-48b3-8df1-a19979624913)

tapioko:
Duplicate of https://github.com/usebruno/bruno/issues/874 ?

fschwiet:
Yes, sorry my search must have missed it because I searched for expansion and not expanse.

Issue #874, opened 2023-11-03
Persist collapse/expand state in JSON response preview


ThenTech:
Currently collapsing a nested object in a JSON response works, but when switching to another tab and then back, this state is reset.
This is especially annoying when fetching a large list of objects that all have nested objects, these all become expanded again after switching tabs.

So I propose to persist this collapsed/expanded state between tabs and/or also add 2 buttons to collapse/expand all nodes in the JSON preview.

anusree-bruno:
This is fixed in latest version v3.4. Closing this as resolved.

PR #3902, opened 2025-01-28
feat: Persist response body scroll position across tabs

terminal-overflow:
# Description

Persists the response body's scroll position when changing tabs. When a user changes to another tab and then comes back to the original tab, the previous scroll position will be restored.

This aims to solve https://github.com/usebruno/bruno/issues/3147 and https://github.com/usebruno/bruno/issues/2742.

I appreciate there may be a few changes needed here needed but the general principle should be there.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -6,7 +6,7 @@
  */
 
 import React from 'react';
-import { isEqual, escapeRegExp } from 'lodash';
+import { isEqual, escapeRegExp, debounce } from 'lodash';
 import { defineCodeMirrorBrunoVariablesMode } from 'utils/common/codemirror';
 import StyledWrapper from './StyledWrapper';
 import * as jsonlint from '@prantlf/jsonlint';
@@ -124,6 +124,8 @@ export default class CodeEditor extends React.Component {
     this.cachedValue = props.value || '';
     this.variables = {};
     this.searchResultsCountElementId = 'search-results-count';
+
+    this.scrollY = props.initialScroll;
 
     this.lintOptions = {
       esversion: 11,
@@ -251,6 +253,8 @@ export default class CodeEditor extends React.Component {
     if (editor) {
       editor.setOption('lint', this.props.mode && editor.getValue().trim().length > 0 ? this.lintOptions : false);
       editor.on('change', this._onEdit);
+      editor.on('scroll', this._onScroll);
+      editor.scrollTo(null, this.scrollY);
       this.addOverlay();
     }
     if (this.props.mode == 'javascript') {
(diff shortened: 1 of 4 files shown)


terminal-overflow:
FoundI have noticed this issue where the response body's scrollbar duplicates itself and thought it was as a result of this PR. It isn't as this is reproducible on main. Marking the PR as ready for review.
Screenshot 2025-01-28 at 18 45 58

ruicardoso-bx:
I've been looking for this feature, thanks for implementing it

packages/bruno-app/src/components/CodeEditor/index.js
```diff
@@ -302,15 +306,21 @@ export default class CodeEditor extends React.Component {
...
   }
 
   componentWillUnmount() {
     if (this.editor) {
       this.editor.off('change', this._onEdit);
+      this.editor.off('scroll', this._onScroll);
       this.editor = null;
     }
 
+    this.props.updateTabScrollPos(this.scrollY);
```
maintainer-bruno: HI @terminal-overflow, 

Thanks for the PR! This is a nice feature, and we’d like to get it released as soon as possible.
 
To maintain a high level of abstraction in the CodeEditor component and avoid coupling it to specific features, can we move the scroll position persistence logic to the parent component(`QueryResultPreview`)? This helps ensure the CodeEditor remains generic and reusable.

As part of this, we can expose a simple onScroll handler in the editor:
```
onScroll = (event) => this.props.onScroll?.(event);
```

Thanks again for your contribution!

packages/bruno-app/src/components/ResponsePane/QueryResult/QueryResultPreview/index.js
```diff
@@ -66,9 +72,25 @@ const QueryResultPreview = ({
     if (disableRunEventListener) {
       return;
     }
+
+    dispatch(
+      updateResponsePaneScrollPosition({
```
maintainer-bruno: The send request action is triggered from multiple places, so updating the active tab state here alone wouldn't cover all cases.

I suggest handling the tab reset within the sendRequest action itself. We can dispatch an additional action from inside sendRequest to reset the active tab as needed.
```
    await dispatch(resetActiveTab({
      uid: state.activeTabUid
    }));
```
Please refer - https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js#L238

maintainer-bruno (changes requested):

packages/bruno-app/src/components/CodeEditor/index.js
terminal-overflow: Hi @maintainer-bruno, thanks for the response. This is certainly a more maintainable approach. Completed [here](https://github.com/usebruno/bruno/pull/3902/commits/dd2c6d82e6786b54279a4e9a0c71a6ee82ab0766).

packages/bruno-app/src/components/ResponsePane/QueryResult/QueryResultPreview/index.js
terminal-overflow: Thanks for the suggestion, this makes sense. Completed [here](https://github.com/usebruno/bruno/pull/3902/commits/f743aae755e270aa42c9aedeeba5030ac39af107). Let me know if there's anything else that requires changing.

naman-bruno (dismissed):

naman-bruno (approved):

ThenTech:
@terminal-overflow Now that the scroll position is persisted, is it possible to take a look at #874 (and duplicate #3431) to persist the response tree view collapse/expand state as well?

Issue #1693, opened 2024-02-29
When specifying "No Body" a Content-Type header is still sent


jsonfry:
When doing a POST, PUT, or PATCH with the "No Body" option selected, a `Content-Type` header is sent with the value of `application/x-www-form-urlencoded`.

There should be no content type header sent

[Jira](https://usebruno.atlassian.net/browse/BRU-1929)

sanjai0py:
Can you check if this issue is happening in the latest version of Bruno, @jsonfry?

orgrimarr:
Got same issue using bruno `v1.13.1` running on ubuntu (using apt repository release)

jsonfry:
Yes, still a probelm on 1.14.0. 

Here's a simple Python server that responses with the request headers as the body:

```python
from http.server import BaseHTTPRequestHandler, HTTPServer

class RequestHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        self._print_headers()

    def do_POST(self):
        self._print_headers()

    def _print_headers(self):
        self.send_response(200)
        self.send_header('Content-type', 'text/plain')
        self.end_headers()
        for header, value in self.headers.items():
            self.wfile.write(f"{header}: {value}\n".encode('utf-8'))

def run(server_class=HTTPServer, handler_class=RequestHandler, port=8000):
    server_address = ('', port)
    httpd = server_class(server_address, handler_class)
    print(f"Server running on port {port}")
    httpd.serve_forever()

if __name__ == "__main__":
    run()
```

And here it is running, notice the returned header value for Content-Type, even though "no body" is selected:

 <img width="1456" alt="Screenshot 2024-04-26 at 12 18 01" src="https://github.com/usebruno/bruno/assets/368400/b9b41b61-97b4-43c0-acf5-fea97089e518">

notaphplover:
Recreated with latest version so far (1.27.0)

sanjai0py:
Hey everyone, I was able to reproduce this bug. It seems like Axios is setting a default `'Content-Type': 'application/x-www-form-urlencoded'` header, which we should override. We will get this fixed soon!

https://github.com/axios/axios/blob/9588fcdec8aca45c3ba2f7968988a5d03f23168c/lib/defaults/index.js#L12-L14

howard-repos:
any update? still got the same problem in version 1.33.0

devjos:
Any update? Problem still persists in v1.38.1

betawait:
[...]

PR #3973, opened 2025-02-09
Fix: Allow empty Content-Type when no body #1693

betawait:
By default Axios will set the Content-Type for POST/PUT/PATCH requests to "application/x-www-form-urlencoded" if the Content-Type header is not specified.

This explicitly sets the content type to "false" when there the body mode is set to "none", and the user has not set an explicit content type themselves. Setting the content type to false directs Axios not to send a Content-Type header.

# Description

This PR fixes an issue where an unwanted Content-Type header would be added to a POST request with body mode set to "none".  Axios adds "application/x-www-form-urlencoded" as the default Content-Type for POST/PUT/PATCH requests, unless the content type is explicitly set to 'false'.

If content-type is explicitly set by a user then that will still be preserved.

The interpolate-vars.js file had to be modified to handle content-type being set to 'false', as it was always expecting a string, and creating an empty string otherwise.

This is my first time in the bruno codebase, so I'm not confident I've thought about other side-effects this could have. I'm also not sure about bruno-cli. I would welcome any advice on other places to check.
 
### Contribution Checklist:

- [] **The pull request does not introduce any breaking changes**


betawait:
Updated to fix conflict.

betawait:
Fixed conflicts.

naman-bruno (approved):

maintainer-bruno (approved):

frokev:
The Content-Type header is set to undefined when not included. It needs to be removed from the header collection.

If you look at the Timeline you see Content-Type is still included as undefined, when you have not included it as a header at all for a GET request. A third-party API we use does not allow this, so we can't use Bruno 😑

In pre-request script, if you try to set the Content-Type header to null, you'll get this error:
`Pre-Request Script Error
contentType.includes is not a function`

We should be able to control which headers are automatically applied by Bruno, such as User-Agent as well. If not, it quickly introduces issues when testing against APIs that doesn't handle edge-cases to a high standard.

Edit: I'm at version 2.6.1, and I see this deviates a lot from the issue, so I created a new one, https://github.com/usebruno/bruno/issues/5012 😊

Its-treason:
> The fix seems to just have set the Content-Type header to undefined. It needs to be removed from the header collection.
> 
> If you look at the Timeline you see Content-Type is still included as undefined, when you have not included it as a header at all for a GET request. A third-party API we use does not allow this, so we can't use Bruno 😑

I think this is kinda like https://github.com/usebruno/bruno/issues/4428 but the Header is actually sent as `undefined`.

frokev:
> > The fix seems to just have set the Content-Type header to undefined. It needs to be removed from the header collection.
> > If you look at the Timeline you see Content-Type is still included as undefined, when you have not included it as a header at all for a GET request. A third-party API we use does not allow this, so we can't use Bruno 😑
> 
> I think this is kinda like #4428 but the Header is actually sent as `undefined`.

It does seem like it mentions it. But I'm not sure the issue conveyed the issue as well. I'll leave mine open https://github.com/usebruno/bruno/issues/5012 I guess.

nickrussler:
I believe this broke the AWS Auth / Sigv4 signature calculation. Without setting any body type you now get a signature mismarch error from your AWS API Gateway:

```
 "{\"message\":\"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\\n\\nThe Canonical String for this request should have been\\n'GET\\n/redacted\\n\\naccept:application/json, text/plain, */*\\ncontent-type:\\nhost:redacted.com
```

I assume the `content-type: false` goes into the signature, while axios throws it out -> mismatch. Or some other mismatch.

anusree-bruno:
@nickrussler can you check this PR https://github.com/usebruno/bruno/pull/5099. This will be fixed in next release

betawait:
Hi @frokev . Sorry, I didn't realize that axios would still send out that content-type header.

I can't find a way to remove the header in Axios completely. It looks like it may be a bug: https://github.com/axios/axios/issues/5236#issuecomment-1993490225

but I'll look a bit more later.

betawait:
I double checked this and found that, although the content-type header still appears in the timeline, it is removed from the headers that are actually sent out in the request.

This is done in Axios here: https://github.com/axios/axios/blob/v1.x/lib/core/AxiosHeaders.js#L238

But this is after the request interceptor that constructs the timeline runs, so the incorrect header is shown.
timeline

I think that also explains the issue with AWS Auth / Sigv4 signature calculation. That is done with an interceptor in the aws4-axios library, so it is likely getting the request passed to it before the content-type header has been removed.

I can't find another way to remove the content-type header. The interceptors are too early in the process, and the default content-type is added after that.

nwin:
> I can't find another way to remove the content-type header. The interceptors are too early in the process, and the default content-type is added after that.
From a display perspective you can just purge the content-type if it is false and a boolean. I was just about to create another issue for this.

PR #3755, opened 2025-01-08
Feat: Move-Collection with Drag-and-Drop

sanjaikumar-bruno:
# Description

Fixes https://github.com/usebruno/bruno/issues/283

https://github.com/user-attachments/assets/2369c525-e0cb-4f91-b085-a105742e93ff

### Contribution Checklist:
- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

PR #3986, opened 2025-02-11
Highlight Drag-and-Drop drop placement in the Sidebar/File-tree

ramki-bruno:
Right now during drag-and-drop in the left sidebar/file-tree, there is no way to be sure that where the dragged item will be dropped. This changes should highlight the location.
There are few other improvements we can do, this is one of those.

Any better suggestions for colors would be welcome.

https://github.com/user-attachments/assets/ef0b1270-bb35-4e83-9f1c-f7e4b10ca97c

image
image

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/StyledWrapper.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/StyledWrapper.js
@@ -33,8 +33,7 @@ const Wrapper = styled.div`
       overflow: hidden;
     }
 
-    &:hover,
-    &.item-hovered {
+    &:hover {
       background: ${(props) => props.theme.sidebar.collection.item.hoverBg};
       .menu-icon {
         .dropdown {
@@ -43,6 +42,14 @@ const Wrapper = styled.div`
           }
         }
       }
+    }
+
+    &.dnd-drag-over {
+      background: ${(props) => props.theme.sidebar.collection.item.hoverBg};
+    }
+
+    &.dnd-hl-drop-after {
+      border-bottom: ${(props) => props.theme.sidebar.collection.item.dndPlacementHighlightBorder};
     }
 
     &.item-focused-in-tab {
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/StyledWrapper.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/StyledWrapper.js
@@ -12,15 +12,12 @@ const Wrapper = styled.div`
       transform: rotateZ(90deg);
     }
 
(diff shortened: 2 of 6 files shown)


packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/StyledWrapper.js
```diff
@@ -33,8 +33,7 @@ const Wrapper = styled.div`
       overflow: hidden;
     }
 
-    &:hover,
-    &.item-hovered {
+    &:hover {
```
ramki-bruno, line 36: looks like `:hover` got reused to be `.item-hovered` in #863 though it could've had a separate class,
I've split it to `.dnd-drag-over` to only have specific styles needed.

packages/bruno-app/src/components/Sidebar/Collections/Collection/StyledWrapper.js
```diff
@@ -12,15 +12,12 @@ const Wrapper = styled.div`
       transform: rotateZ(90deg);
     }
 
-    &.item-hovered {
+    &.dnd-drag-over {
       background: ${(props) => props.theme.sidebar.collection.item.hoverBg};
-      .collection-actions {
```
sanish-bruno, line 17: @ramki-bruno  why the .collection-actione were removed, while testing, i didn't any issues as such, were this lines unnecessary. just needs to be sure that the changes are not  impacting anything else. i see on later lines we have the same code written for hover

ramki-bruno: It is unnecessary, it got added [here](https://github.com/usebruno/bruno/pull/3755/files#diff-f70560e462dd50a0ff0bba7f2c61126753fe875e907b6619e5d125db399cfd79R15-R25) just days before this PR.
There, is copied form another place where its was already unnecessary. check the [above comment](https://github.com/usebruno/bruno/pull/3986/files#r1950281946) for detail on that.

ramki-bruno: For further confirmation and clarification, `.item-hovered` class here was used only when another item is dragged over a item
This style changes the bg-color and makes the 3-dot(kebab menu) appear. Changing the bg-color makes sense on drag-over, but showing the menu doesn't. It only exists because the style used for on-hover just got reused for drag-over.

sanish-bruno (approved):

packages/bruno-app/src/components/Sidebar/Collections/Collection/StyledWrapper.js
sanish-bruno: LGTM

PR #3999, opened 2025-02-11
fix: should be able to save the request after reverting the changes 

lohxt1:
This fixes the issue where the sequence was not getting updated while the request in draft state was moved


helloanoop (approved):

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -1767,6 +1767,9 @@ export const collectionsSlice = createSlice({
           // we don't want to lose the draft in this case
           if (areItemsTheSameExceptSeqUpdate(item, file.data)) {
             item.seq = file.data.seq;
+            if (item?.draft) {
+              item.draft.seq = file.data.seq;
+            }
             if (item?.draft && areItemsTheSameExceptSeqUpdate(item?.draft, file.data)) {
               item.draft = null;
             }
```
ramki-bruno, line 1775: the above two clauses can be combined to be more clearer
```js
            if (item?.draft) {
              if (areItemsTheSameExceptSeqUpdate(item?.draft, file.data)) {
                item.draft = null;
              } else item.draft.seq = file.data.seq;
            }
```

Issue #1676, opened 2024-02-27
[Bug] Cookies not sent


sebastien-f:
Version: 1.9.0

I have a collection with 2 requests inside, `login` and `foo`. When running `login` (POST, with data) I get a response with a set_cookie header : 

```
< set-cookie: access_token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJsb2dpbiI6InMuZmlub3RAY29tbS1pdC5pbyIsImlhdCI6MTcwOTA1MTcwM30.OeXYsgE5_U4ClUgvGi2vK0tJOiBBc96EMHBTcNnboqM; Path=/; HttpOnly; Secure
```

When running request `foo` (GET, not data), cookie is not sent despite it being listed in the cookies visible with the bottom-left cookie and get rejected because of missing access_token.

If I manually add the cookie header to the `foo` request, the request is properly authorized.

Also, cookie is properly updated each time I call `login`.

Both requests uses same protocol + url + port.

Both Store and Send Cookies automatically are enabled. I also tried to deactivate+save then activate+save as I was upgrading from 1.2.0.

I also tried using the runner, in case that could have an impact, but same result. 

On a side note, I tried looking at the [documentation](https://docs.usebruno.com/) on how to use the cookies, but could not find any reference.

codingskynet:
Could you try API with `http`? I found that Postman add `Secure` cookies on header even `http` request, but bruno strictly handles it. If trying `http` API with `Secure` cookie, it does not send cookie on `http` request.

sebastien-f:
@codingskynet thank you for the suggestion, but the api is hosted on https and there is no way to makes calls over http. Not sure what Postman has to do here. Did you meant Bruno ?

codingskynet:
Actually, I experienced like this on `https` server(maybe supports auto-upgrade to https from http):
Postman: If I send request on http, `SECURE` cookie can be sent.
Bruno: If I send request on http, `SECURE` cookie cannot be sent. On https, working well.

igronus:
I can confirm cookies seem to be missing even with non-secure http requests on a classical symfony application.

GimpMaster:
Just adding my thoughts here. It would be GREAT if Bruno sent secure cookies over a localhost so we don't have to either

a.) setup a secure certificate for localhost testing or
b.) conditionally set secure to off for local testing on a backend

Webrowser's already do this.

```
[...]

PR #4038, opened 2025-02-14
feat: Add RFC 6761–compliant localhost loopback checks so `secure` cookies work on localhost (fixes: #1676)

Chriss4123:
fixes: [#1676](https://github.com/usebruno/bruno/issues/1676)

# Description

### Contribution Checklist:

This commit extends how cookies are treated in secure contexts by fully recognizing `localhost` and loopback IPs as trustworthy origins, matching the de facto behavior of all modern browsers and RFC 6761. Previously, `tough-cookie` defaulted `secure` to `true` only for `https:` and `wss:` URLs, causing cookies with `secure` set to never be sent to `localhost`.

### What Changed

1. **New `trustworthy-util.js`**  
   - Implements `isPotentiallyTrustworthy(url)` by checking:
     - Schemes: `https`, `wss`, `file`
     - Loopback IPs: `127.0.0.1/8` and `::1`  
     - Hostnames: `localhost` and `*.localhost`  
   - Adapted from Chromium’s `IsLocalhost`, `IsLoopback` and `HostNoBracketsPiece`, located at:
     - [`IsLocalhost`](https://github.com/chromium/chromium/blob/main/net/base/url_util.cc#L457-L551)
     - [`IsLoopback`](https://github.com/chromium/chromium/blob/main/net/base/ip_address.cc#L228-L243)
     - [`HostNoBracketsPiece`](https://github.com/chromium/chromium/blob/main/url/gurl.cc#L440-L448)

2. **`cookies.js` Update**  
   - Passes the `{ secure: isPotentiallyTrustworthy(url) }` option to `cookieJar.getCookiesSync()`.  
   - This overrides `tough-cookie`’s built-in:
     ```js
     let secure = options.secure;
     if (secure == null && (context.protocol == "https:" || context.protocol == "wss:")) {
       secure = true;
     }
     ```
     We do everything `tough-cookie` does by default—plus treat localhost/loopback the same as modern browsers. No existing functionality is removed, only expanded.

### Testing this change

```python
from flask import Flask, request, make_response

app = Flask(__name__)

@app.route('/setcookie')
def set_cookie():
    resp = make_response('')
    resp.set_cookie('example', '', secure=True)
    return resp

@app.route('/getcookie')
def get_cookie():
    return 'Cookie is set' if 'example' in request.cookies else 'No cookie set'

app.run()
```
- **Before**: The `secure` cookie would **not** be sent to `/getcookie` after it was set by `/setcookie` due to `secure` being set to `True`.  
- **After**: The `secure` cookie would be sent to `/getcookie` which mirrors the behavior of modern browsers.

All modern browsers (Chrome et al.) and many other API testing clients (Postman et al.) handle `http://localhost` as a trustworthy origin. It only makes sense that this behavior also exists in Bruno.

### DIFF
--- a/packages/bruno-electron/src/utils/cookies.js
+++ b/packages/bruno-electron/src/utils/cookies.js
@@ -1,4 +1,5 @@
 const { Cookie, CookieJar } = require('tough-cookie');
+const { isPotentiallyTrustworthy } = require('./trustworthy-util');
 const each = require('lodash/each');
 
 const cookieJar = new CookieJar();
@@ -11,7 +12,9 @@ const addCookieToJar = (setCookieHeader, requestUrl) => {
 };
 
 const getCookiesForUrl = (url) => {
-  return cookieJar.getCookiesSync(url);
+  return cookieJar.getCookiesSync(url, {
+    secure: isPotentiallyTrustworthy(url)
+  });
 };
 
 const getCookieStringForUrl = (url) => {
--- a/packages/bruno-electron/src/utils/trustworthy-util.js
+++ b/packages/bruno-electron/src/utils/trustworthy-util.js
@@ -1,0 +1,102 @@
+const { URL } = require('url');
+const net = require('net');
+
+const isLoopbackV4 = (address) => {
+  // 127.0.0.0/8: first octet = 127
+  const octets = address.split('.');
+  return (
+    octets.length === 4
+  ) && parseInt(octets[0], 10) === 127;
+}
+
+const isLoopbackV6 = (address) => {
+  // dns relies on c-ares under the hood
+  // (https://nodejs.org/download/release/v0.10.29/docs/api/dns.html#dns_dns)
+  // which always compresses loopback addresses to [::1]
+  return (address === '::1');
+}
+
+const isIpLoopback = (address) => {
+  if (net.isIPv4(address)) {
+    return isLoopbackV4(address);
+  }
+
+  if (net.isIPv6(address)) {
+    return isLoopbackV6(address);
+  }
+
+  return false;
+}
+
+const isNormalizedLocalhostTLD = (host) => {
+  return host.toLowerCase().endsWith('.localhost');
+}
+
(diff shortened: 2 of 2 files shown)


naman-bruno (dismissed):

JoshuaHintze:
@sreelakshmi-bruno - Any chancel to get this merged in?

sreelakshmi-bruno:
This will be merged and will be included in the upcoming release which will likely go out on coming Tuesday.

sreelakshmi-bruno:
Apologies, we've postponed the release for this PR to April. Thank you for your patience!

ramki-bruno:
@Chriss4123 I see that [your PR on `salesforce/tough-cookie`](https://github.com/salesforce/tough-cookie/pull/498) is already merged. That alone is great news and thanks for your contributions.

Do you think we can just get straight to upgrading `tough-cookie` once their major release is out?

Chriss4123:
> @Chriss4123 I see that [your PR on `salesforce/tough-cookie`](https://github.com/salesforce/tough-cookie/pull/498) is already merged. That alone is great news and thanks for your contributions.
> 
> Do you think we can just get straight to upgrading `tough-cookie` once their major release is out?

Yes, that should work. Just keep in mind tough-cookie has been converted to TypeScript a while ago, however you are using an old version which still uses vanilla JS.

ramki-bruno:
Since `tough-cookie` v6 is still in RC, I think we can go ahead with merging this for now.
I'm also adding few tests for this in #4709. Will add a note as well regarding upgrading it.

ramki-bruno (approved):

ramki-bruno:
Moved the util to `@usebruno/common` and fixed it for CLI as well.

helloanoop:
@ramki-bruno I'd like to keep the `bruno-common` platform agnostic as it gets bundled. Can you move the logic to `bruno-requests` ?

ramki-bruno:
Please use Merge commit

naman-bruno (approved):

PR #4091, opened 2025-02-23
chore: version upgrade and fix posthog api key issue while building from source

lohxt1:


helloanoop:
@lohxt1 Can you fix the conflicts ?

Issue #4108, opened 2025-02-25
Keybindings are not working properly on Windows


Pragadesh-45:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### Bruno version

1.39.1

### Operating System

Windows

### Describe the bug

 - The minimize window shortcut is currently `Ctrl + Shift + Q,` but it should be assigned to close window.
 - The Hide Window function is not working on Windows (Ctrl+M)
 - The Zoom In shortcut (`Ctrl + Plus (+)`) is not working on Windows, although it works fine on Mac

### Screenshots/Live demo link

.

PR #4110, opened 2025-02-25
Fix/keybinding issues

Pragadesh-45:
fixes: #4108 

# Description

This PR implements support for adding global shortcuts to fix **Ctrl + Plus (+)** for zooming on Windows. It also properly updates the key mapping page to use **Close Bruno** instead of **Minimize Bruno**, and introduces support for minimizing the window with **Ctrl + M** and **Ctrl + H** instead of **Win + H**.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/providers/Hotkeys/keyMappings.js
+++ b/packages/bruno-app/src/providers/Hotkeys/keyMappings.js
@@ -5,10 +5,10 @@ const KeyMapping = {
   newRequest: { mac: 'command+b', windows: 'ctrl+b', name: 'New Request' },
   closeTab: { mac: 'command+w', windows: 'ctrl+w', name: 'Close Tab' },
   openPreferences: { mac: 'command+,', windows: 'ctrl+,', name: 'Open Preferences' },
-  minimizeWindow: {
-    mac: 'command+Shift+Q',
-    windows: 'control+Shift+Q',
-    name: 'Minimize Window'
+  closeBruno: {
+    mac: 'command+Q',
+    windows: 'ctrl+shift+q',
+    name: 'Close Bruno'
   },
   switchToPreviousTab: {
     mac: 'command+pageup',
--- a/packages/bruno-electron/src/index.js
+++ b/packages/bruno-electron/src/index.js
@@ -11,7 +11,7 @@ if (isDev) {
 }
 
 const { format } = require('url');
-const { BrowserWindow, app, session, Menu, ipcMain } = require('electron');
+const { BrowserWindow, app, session, Menu, globalShortcut, ipcMain } = require('electron');
 const { setContentSecurityPolicy } = require('electron-util');
 
 const menuTemplate = require('./app/menu-template');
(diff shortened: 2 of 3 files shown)


packages/bruno-electron/src/index.js
```diff
@@ -159,6 +159,23 @@ app.on('ready', async () => {
     }
     return { action: 'deny' };
   });
+  
+  // Quick fix for Electron issue #29996: https://github.com/electron/electron/issues/29996
+  globalShortcut.register('Ctrl+=', () => {
```
sanish-bruno, line 170: @Pragadesh44-Bruno do we support `Ctrl+-` currently,, for zooming out, how does it behave currently?

sanish-bruno: I guess this is probably fixed in later versions of electron, they have mentioned the same within this PR, don't know whether it is part of our version of electron https://github.com/aza547/wow-recorder/pull/691

packages/bruno-app/src/providers/Hotkeys/keyMappings.js
```diff
@@ -5,10 +5,10 @@ const KeyMapping = {
   newRequest: { mac: 'command+b', windows: 'ctrl+b', name: 'New Request' },
   closeTab: { mac: 'command+w', windows: 'ctrl+w', name: 'Close Tab' },
   openPreferences: { mac: 'command+,', windows: 'ctrl+,', name: 'Open Preferences' },
-  minimizeWindow: {
-    mac: 'command+Shift+Q',
-    windows: 'control+Shift+Q',
-    name: 'Minimize Window'
+  closeBruno: {
```
sanish-bruno, line 8: @Pragadesh44-Bruno should we be adding minimize window here too?

sanish-bruno (changes requested):

packages/bruno-electron/src/index.js
Pragadesh-45: > @Pragadesh44-Bruno do we support Ctrl+- currently,, for zooming out, how does it behave currently?

yes, we do support it but it only works on mac and linux. it doesn't work as expected in windows

packages/bruno-electron/src/index.js
```diff
@@ -159,6 +159,23 @@ app.on('ready', async () => {
...
+    mainWindow.webContents.setZoomLevel(mainWindow.webContents.getZoomLevel() + 1);
+  });
+
+  globalShortcut.register('CommandOrControl+M', () => {
+    if (mainWindow) {
```
maintainer-bruno: I believe this check can be removed, `mainWindow` is already initialized at line 75

packages/bruno-electron/src/index.js
maintainer-bruno: Does Ctrl+- have any adverse effects on other operating systems? If not, I’d recommend adding it.

Pragadesh-45: When I tested Ctrl+- on Mac, Windows, and Linux, I couldn't find any issues. Only Ctrl+= is causing a problem, and that too only on Windows.

packages/bruno-electron/src/index.js
Pragadesh-45: sure, i'll update it

bijin-bruno (approved):

naman-bruno (approved):

PR #4160, opened 2025-03-05
Added option to customize `userData` path on dev mode

ramki-bruno:
If `ELECTRON_APP_NAME` env-variable is present and its dev env, then the `appName` and `userData` path is modified accordingly.

e.g.
```sh
ELECTRON_APP_NAME=bruno-dev npm run dev:electron
```

Note: This doesn't change the name of the window or the names in lot of other places, only the name used by Electron internally.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -98,6 +98,16 @@ npm run dev:electron
 npm run dev
 ```
 
+#### Customize Electron `userData` path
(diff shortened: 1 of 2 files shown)


helloanoop:
@ramki-bruno Can we add a note in contributing.md to document this ?

ramki-bruno:
> @ramki-bruno Can we add a note in contributing.md to document this ?

Done

PR #4202, opened 2025-03-11
feat: openapi converter package

pooja-bruno:
# Description

This PR includes the `bruno-converters` package, which allows conversion between OpenAPI spec and Bruno collections. This package can be used in both `bruno-cli` and `bruno-app`.


packages/bruno-converters/src/utils.js
```diff
@@ -0,0 +1,190 @@
...
+const each = require('lodash/each');
+const get = require('lodash/get');
+
+/**
+ * Generate a UUID
+ * @returns {string} UUID
+ */
+const uuid = () => {
+  // https://github.com/ai/nanoid/blob/main/url-alphabet/index.js
+  const urlAlphabet = 'useandom26T198340PX75pxJACKVERYMINDBUSHWOLFGQZbfghjklqvwyzrict';
```
Ghostbird, line 11: Why is this here? Shouldn't the first line in this file be:

```javascript
const { customAlphabet, urlAlphabet } = require('nanoid');
```

It is exported by https://github.com/ai/nanoid/blob/c49cc113499f2b44ab52b80314e18043ad0f3f79/index.js#L5

Ghostbird:
I'd love to see this merged. I recently imported our API definitions into its repository so we can easily test things however, the current import functionality is too limited. It doesn't handle parameters well, it doesn't handle request bodies, it yields incomprehensible filenames based on `summary`. With a better converter and the option to control it from code, ideally we could automatically keep our Bruno files up-to-date with the evolution of our API.

PR #4290, opened 2025-03-20
Hotfix: Performance improvements in Response-preview

### LINKED ISSUE #4285: render long json response 100% CPU

### This bug is:

- [ ] annoying

### Bruno version

1.40

### Operating System

Linux Mint 22 / Ubuntu 24.04

### Describe the bug

After update Flatpak from v1.39.1 to v1.40 the parsing of a long json response (8500KB) kills the app. It's not responsive anymore. 
A subtask "/app/extra/bruno/bruno --type=renderer" is eating 100% CPU.
I reverted to v1.39.1 and the problem was gone.

_No response_

sorry, can't share more details.

ramki-bruno:
Partially Fixes #4285

- JSON-Format only first 1MiB of response to prevent blocking render
- Add a missing useMemo in QueryResult


packages/bruno-app/src/utils/common/index.js
```diff
@@ -28,11 +28,16 @@ export const waitForNextTick = () => {
...
-    doc,
-    jsoncFormat(doc, null, {insertSpaces: true, tabSize: 2})
-  );
-}
+  // Format only the first 1MiB of the doc,
+  // this is to prevent bigger responses from blocking the thread
+  // and making the UI unresponsive.
+  const maxFormatLegth = 1048576; // 2 ^ 20, 1MiB
+  const edits = jsoncFormat(doc, { offset: 0, length: maxFormatLegth }, { insertSpaces: true, tabSize: 2 });
+  const truncatedDoc = doc.substr(0, maxFormatLegth);
```
ramki-bruno, line 37: 🤔 refactoring..

helloanoop: I am not sure if we truncating part of the json is a good idea. We are degrading the search experience for search items which may be present in the parts of json which got truncated.

We are also degrading the folding functionality of the codemirror for the truncated parts.

packages/bruno-app/src/utils/common/index.js
```diff
@@ -28,11 +28,16 @@ export const waitForNextTick = () => {
 };
 
 export const prettifyJson = (doc) => {
-  return jsoncApplyEdits(
-    doc,
-    jsoncFormat(doc, null, {insertSpaces: true, tabSize: 2})
-  );
-}
+  // Format only the first 1MiB of the doc,
```
helloanoop: I am not sure if we truncating part of the json is a good idea. We are degrading the search experience for search items which may be present in the parts of json which got truncated.

We are also degrading the folding functionality of the codemirror for the truncated parts.

ramki-bruno: Its not possible to handle this without performance impact on the UI.
I've updated it to 5MiB with the new formatter.
To fix this properly, we have to give the users option to format the entire response on-demand.
For now this should at least unblock the users.

Search should still work, although it'll be slow.

helloanoop: > Its not possible to handle this without performance impact on the UI.

I feel we should not enforce the limit, if someone has a beefy system then we should not dictate the experience.

> To fix this properly, we have to give the users option to format the entire response on-demand.

💯

OpenOrangeData:
I have an issue that is at least similar to this one: In my case, not just the search is slow, but Bruno sometimes crashes (hangs indefinitely) in case of a large response (sometimes it "only" fails to respond a few minutes).
I'm wondering if the improvements made in this issue to the formatting of the response-preview may solve the problem or if this is a separate issue and I should open a new bug?
Is there any way I could test the improvements made for this issue to if my problem will be solved?

Note: If the use of Accept-Encoding would work for me (it doesn't, #4592), the response size would already be considerably smaller and Bruno would probably crash much less frequently.

bijin-bruno:
Closing this as the issue has been resolved using a different approach via PR https://github.com/usebruno/bruno/pull/6026

PR #4103, opened 2025-02-25
Fix: Inconsistent JSON parsing and formatting in `res.body` and Res-preview

Pragadesh-45:
Closes #4046
Closes #3761
Closes #2273
Related Issues #3758, #3594, #3558
Related PRs #3805 #3706 #3676 #2272 

Some of the Issues mentioned above have already been worked on and fixed but we had to re-look them and fix the root-cause and had to revert some of those changes.

This PR fixes
- Inconsistencies in parsing and formatting the JSON response where anything other than `typeof data === "object"` is considered a not a valid JSON which includes string, numbers, boolean and also `null`.
- Wrong interpretation of string wrapped in quotes in of RHS input of Assert table. They are supposed to be still plain strings, instead they were interpreted as string containing quotes.
- For the Response-preview, format the JSON string/raw-response without parsing(`JSON.parse`) to preserve the precession of numbers bigger than `Number.MAX_SAFE_INTEGER`.
This is just to make sure that the exact response given by the server is shown in the preview. The `res.body/res.getBody()` will still have the parsed data and precession will be lost anyway for those numbers
- Inconsistencies in response-preview formatting for invalid-JSON response while the `content-type` is still `application/json`

Also, added test requests with Asserts and Test scripts to validate the above across older versions and to check for any regressions.
|||
| ------------- | ------------- |
| image | image |

Issue #3889, opened 2025-01-27
res.setBody() not working with objects in developer


shakzz31:
### I have checked the following:

- [x] I use the newest version of bruno.
- [x] I've searched existing issues and found nothing related to my issue.

### Describe the bug

res.setBody() doesnt work when changing to Developer Mode.

getting the following error message:
`Error invoking remote method 'send-http-request': Error: An object could not be cloned.`

### Screenshots/Live demo link

none

anusree-bruno:
Hi @shakzz31,

Thank you for reporting this. I was able to replicate the issue on my end, and we will work on fixing it.

PR #4332, opened 2025-03-25
Revert "Fix: Prettify JSON for Res-preview without parsing to avoid J…

ramki-bruno:
…S specific roundings"

Duplicates #4331 for v1.x

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/QueryResult/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/QueryResult/index.js
@@ -10,7 +10,7 @@ import QueryResultPreview from './QueryResultPreview';
 import StyledWrapper from './StyledWrapper';
 import { useState, useMemo, useEffect } from 'react';
 import { useTheme } from 'providers/Theme/index';
-import { getEncoding, prettifyJson, uuid } from 'utils/common/index';
+import { getEncoding, uuid } from 'utils/common/index';
 
 const formatResponse = (data, dataBuffer, encoding, mode, filter) => {
   if (data === undefined || !dataBuffer) {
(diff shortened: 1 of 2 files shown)


helloanoop (approved):

lonnyzhang423:
After making these changes, a new issue arises: if I set custom JSON content through res.setBody(), the preview panel adds extra double quotes to the string.

image

helloanoop:
@ramki-bruno FYR- https://github.com/usebruno/bruno/pull/4332#issuecomment-2766040747

ramki-bruno:
@lonnyzhang423 I just checked this case after reverting the commit.
Not sure if you've noticed, but `res.setBody(...)` isn't even reflecting in the preview with the revert, same in v1.40.0 and v2.0.0.
In #4103 we relied on the raw response data for the preview which didn't account for `res.setBody()`

But the problem is more nuanced.
If you want to set the `res.body` to an object, you should be able to set the object directly like `res.setBody({foo: "bar"})` instead of stringifying it like you are using.
Looks like that is not working because it gives error in the preview panel when an object is passed as the argument(PFA).

And when you set the body to the string, in this case a stringified Object, the current behavior is expected because you are setting `res.body` to the JSON string itself where ideally it should be parsed value/Object. The preview panel just show the same string.

image
image

ramki-bruno:
We are still working on this and will address these issues,

IMO, a proper fix is to make `res.setBody({foo: "bar"})` work where you can pass the Object itself instead of stringifying it.
@lonnyzhang423 let us know if that'll work for your use-case.

lonnyzhang423:
> We are still working on this and will address these issues,
> 
> IMO, a proper fix is to make `res.setBody({foo: "bar"})` work where you can pass the Object itself instead of stringifying it.
> @lonnyzhang423 let us know if that'll work for your use-case.

Yes, the final solution works and meets the requirements, assuming it can be implemented properly.

ramki-bruno:
@lonnyzhang423 looks like there is already an open issue for this #3889

ramki-bruno:
@lonnyzhang423 just curious, what is your use-case where you have to modify the response to something other than the actual server response?

lonnyzhang423:
the server data is encrypted and must be decrypted before being properly displayed.

Issue #1884, opened 2024-03-21
[Feature Request] Streamed response


leobenkel:
Hello,

thank you for this great product.

I have one feature request, sometimes API respond with streamed messages, series of JSON line or other that comes over time. 

I think at the moment, Bruno wait for complete response before rendering to the UI.

Would be nice to see the messages as they comes to follow along.

Thank you very much. 
[Jira](https://usebruno.atlassian.net/browse/BRU-1215)

lujun59:
This feature is import for some LLM streaming APIs, thanks.
We found postman has supported this feature.  See blow:
![image](https://github.com/usebruno/bruno/assets/1212672/2a733760-a29c-49a2-9823-857a8e66300b)

bredmor:
Would also love to see this feature. I work heavily with IoT devices and have to document/test a lot of APIs that return `text/event-stream` content.

etler:
Looking at the code, this project uses axios for the requests under the hood. The problem with that is that axios does not support http streams due to being built on top of XHR which is a very outdated and essentially abandoned standard.

Axios would need to be replaced with modern `fetch` calls. Node supports fetch via a flag from v16, without a flag from v18, and it's fully stable as of v21. This project uses electron v21.1.1 which is on node v16 so to support fetch natively it would need to include the flag. The other options would be to upgrade electron to be able to use a newer version of node, or use a polyfill library like `node-fetch` which supports streaming.

So first the maintainer would need to approve switching from axios to fetch, a fetch implementation would need to be added so it can be used, and then axios would need to be replaced with fetch, and then a streaming implementation would need to be added.

helloanoop:
> This project uses electron v21.1.1 which is on node v16

We plan to move to electron v31 which uses node v18 asap.

> to support fetch natively it would need to include the flag. 

this makes sense

> the maintainer would need to approve switching from axios to fetch

this would be very hard to execute at this stage of the project. The approach I am thinking of is to allow the toggling the provider at the collection level. (default to axios, but allow to use fetch / curl ). This would need UI changes, also a placeholder in `bruno.json` to store this data.

mirkonasato:
> > the maintainer would need to approve switching from axios to fetch
> 
> this would be very hard to execute at this stage of the project. The approach I am thinking of is to allow the toggling the provider at the collection level. (default to axios, but allow to use fetch / curl ).

[...]

PR #4472, opened 2025-04-10
#1884 - Add support for text/event-stream content-type

davirxavier:
# Description

This pull request adds support for text/event-stream content-type requests. I decided to try adding support for these types of request because i found a few issues requesting it, with some people saying that the only reason they didn't migrate fully from postman is the lack of this content-type support. 

If i didn't follow any standards of the project or didn't add a needed change somewhere please let me know and i will be glad to push any fixes. 

## Changes

The code changes are really simple, in summary, i just set the responseType to stream in the axios request if the accept header is set to text/event-stream, send the incoming streamed data to the front-end by listening and treat the streamed data in the response body panel.

With these changes, if the user sends a request with an accept header of text/event-stream, the right arrow icon in the URL bar will change to an X icon and the streamed data will be concatenated to the response body panel. If the user clicks the X icon, the front-end will send an event to the back-end signaling for the stream to be closed and the data will stop being streamed.

- Added boolean: item.response.hasStreamRunning, will be true if there's a steam active at the moment. This is the variable that is used to check if there's a stream active.
- packages/bruno-app/src/components/RequestPane/QueryUrl/index.js: Add icon change from arrow to X if there's a stream active.
- packages/bruno-app/src/components/RequestTabPanel/index.js: Changed the handleRun function to cancel the current request if there is a stream running (instead of starting a new request).
- packages/bruno-app/src/providers/App/useIpcEvents.js - Added two new events:

1. main:http-stream-new-data - Will be emitted from the back-end to the front-end when there's new data to the stream.
2. main:http-stream-end - Will be emitted from the back-end to the front-end when the stream has been cancelled by the host.

- packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js - Added and changed a few collection actions:

1. requestCancelled - If there's a stream active, don't set the response to null, and set the item.response.hasStreamRunning variable to null, so the front-end will reset to there being no stream running.
2. Don't set the cancelTokenUid to null when the request is created, so it can be cancelled by the user.
3. Add the streamDataReceived action, this action will be ran when the main:http-stream-new-data event is received, it will concatenate the data received in the stream with the current displayed data.

- packages/bruno-app/src/utils/network/index.js - Copy the hasStreamRunning variable from the back-end response to item.
- packages/bruno-electron/src/ipc/network/index.js:

1. Add checks for the text/event-stream accept header in the request send handler, if this header is found set the axios responseType to "stream" and set the connection header to "keep-alive", so the connection isn't lost.
2. If the header is found, the first response data will be an empty string.
3. Before returning the response in the handler, add listeners for the on data received and stream ended callbacks from the axios stream object.
4. When the data callback is received, emit a main:http-stream-new-data event to the front-end with the new data.
5. If there's a stream ended callback, check if there's still a delete token saved for that request, if there is, the event was not initiated by a user, so emit main:http-stream-end to the front-end and remove the saved delete token for that request.
6. Skip the request run if we are running a whole collection (because the event stream will never end).

### Contribution Checklist:

- [ X ] **The pull request only addresses one issue or adds one feature.**
- [ - ] **The pull request does not introduce any breaking changes**

### DIFF
--- a/packages/bruno-app/src/providers/App/useIpcEvents.js
+++ b/packages/bruno-app/src/providers/App/useIpcEvents.js
@@ -14,10 +14,10 @@ import {
   collectionUnlinkDirectoryEvent,
   collectionUnlinkEnvFileEvent,
   collectionUnlinkFileEvent,
-  processEnvUpdateEvent,
+  processEnvUpdateEvent, requestCancelled,
   runFolderEvent,
   runRequestEvent,
-  scriptEnvironmentUpdateEvent
+  scriptEnvironmentUpdateEvent, streamDataEnded, streamDataReceived
 } from 'providers/ReduxStore/slices/collections';
 import { collectionAddEnvFileEvent, openCollectionEvent, hydrateCollectionWithUiStateSnapshot } from 'providers/ReduxStore/slices/collections/actions';
 import toast from 'react-hot-toast';
@@ -173,6 +173,14 @@ const useIpcEvents = () => {
       dispatch(collectionAddOauth2CredentialsByUrl(payload));
     });
 
+    const removeHttpStreamNewDataListener = ipcRenderer.on('main:http-stream-new-data', (val) => {
+      dispatch(streamDataReceived(val));
+    });
+
+    const removeHttpStreamEndListener = ipcRenderer.on('main:http-stream-end', (val) => {
+      dispatch(requestCancelled(val));
+    });
+
     return () => {
       removeCollectionTreeUpdateListener();
       removeOpenCollectionListener();
@@ -193,6 +201,8 @@ const useIpcEvents = () => {
       removeGlobalEnvironmentsUpdatesListener();
       removeSnapshotHydrationListener();
       removeCollectionOauth2CredentialsUpdatesListener();
+      removeHttpStreamNewDataListener();
+      removeHttpStreamEndListener();
     };
   }, [isElectron]);
 };
--- a/packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
+++ b/packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
@@ -274,28 +274,33 @@ export const collectionsSlice = createSlice({
       if (collection) {
         const item = findItemInCollection(collection, itemUid);
         if (item) {
-          item.response = null;
+          if (item.response?.hasStreamRunning) {
+            item.response.hasStreamRunning = null;
+          } else {
+            item.response = null;
+          }
+
           item.cancelTokenUid = null;
         }
       }
     },
     responseReceived: (state, action) => {
       const collection = findCollectionByUid(state.collections, action.payload.collectionUid);
-    
+
       if (collection) {
         const item = findItemInCollection(collection, action.payload.itemUid);
         if (item) {
           item.requestState = 'received';
           item.response = action.payload.response;
-          item.cancelTokenUid = null;
+          item.cancelTokenUid = item.response.hasStreamRunning ? item.cancelTokenUid : null;
 
           if (!collection.timeline) {
             collection.timeline = [];
           }
-    
+
           // Ensure timestamp is a number (milliseconds since epoch)
-          const timestamp = item?.requestSent?.timestamp instanceof Date 
-            ? item.requestSent.timestamp.getTime() 
+          const timestamp = item?.requestSent?.timestamp instanceof Date
+            ? item.requestSent.timestamp.getTime()
             : item?.requestSent?.timestamp || Date.now();
 
           // Append the new timeline entry with numeric timestamp
@@ -2161,12 +2166,21 @@ export const collectionsSlice = createSlice({
     updateFolderAuthMode: (state, action) => {
       const collection = findCollectionByUid(state.collections, action.payload.collectionUid);
       const folder = collection ? findItemInCollection(collection, action.payload.folderUid) : null;
-      
+
       if (folder) {
         set(folder, 'root.request.auth', {});
         set(folder, 'root.request.auth.mode', action.payload.mode);
       }
     },
+    streamDataReceived: (state, action) => {
+      const { itemUid, collectionUid, data } = action.payload;
+      const collection = findCollectionByUid(state.collections, collectionUid);
+
+      if (collection) {
+        const item = findItemInCollection(collection, itemUid);
+        item.response.data = data.data + (item.response.data || '');
+      }
+    }
   }
 });
 
@@ -2278,7 +2292,9 @@ export const {
   collectionGetOauth2CredentialsByUrl,
   updateFolderAuth,
   updateFolderAuthMode,
-  moveCollection
+  moveCollection,
+  streamDataReceived,
+  streamDataEnded,
 } = collectionsSlice.actions;
 
 export default collectionsSlice.reducer;
--- a/packages/bruno-app/src/components/RequestPane/QueryUrl/index.js
+++ b/packages/bruno-app/src/components/RequestPane/QueryUrl/index.js
@@ -5,7 +5,7 @@ import { requestUrlChanged, updateRequestMethod } from 'providers/ReduxStore/sli
 import { saveRequest } from 'providers/ReduxStore/slices/collections/actions';
 import HttpMethodSelector from './HttpMethodSelector';
 import { useTheme } from 'providers/Theme';
-import { IconDeviceFloppy, IconArrowRight, IconCode } from '@tabler/icons';
+import { IconDeviceFloppy, IconArrowRight, IconCode, IconX } from '@tabler/icons';
 import SingleLineEditor from 'components/SingleLineEditor';
 import { isMacOS } from 'utils/common/platform';
 import StyledWrapper from './StyledWrapper';
@@ -136,7 +136,14 @@ const QueryUrl = ({ item, collection, handleRun }) => {
               Save <span className="shortcut">({saveShortcut})</span>
             </span>
           </div>
-          <IconArrowRight color={theme.requestTabPanel.url.icon} strokeWidth={1.5} size={22} />
+          {
+            item.response?.hasStreamRunning ? (
+              <IconX color={theme.requestTabPanel.url.icon} strokeWidth={1.5} size={22} />
+            ) : (
+              <IconArrowRight color={theme.requestTabPanel.url.icon} strokeWidth={1.5} size={22} />
+            )
+          }
+
         </div>
       </div>
       {generateCodeItemModalOpen && (
--- a/packages/bruno-app/src/components/RequestTabPanel/index.js
+++ b/packages/bruno-app/src/components/RequestTabPanel/index.js
@@ -8,7 +8,7 @@ import ResponsePane from 'components/ResponsePane';
 import Welcome from 'components/Welcome';
 import { findItemInCollection } from 'utils/collections';
 import { updateRequestPaneTabWidth } from 'providers/ReduxStore/slices/tabs';
-import { sendRequest } from 'providers/ReduxStore/slices/collections/actions';
+import { cancelRequest, sendRequest } from 'providers/ReduxStore/slices/collections/actions';
 import RequestNotFound from './RequestNotFound';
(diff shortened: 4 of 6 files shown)


sanish-bruno:
@davirxavier what are your thoughts on,

- how we can handle `Accept: */*` ? currently we need to add `Accept: text/event-stream` as a header to the request to make it work?

- also while making a request, the elapsed time shown currently is of the first stream response, not the entire time elapsed to get all the responses, other api testing clients seems to show the entire time elapsed, what are your thoughts on that? 

- the send button get disabled only after we get the first streamed response, multiple clicks done followed by first click are currently resulting in multiple responses being received at the same time, for example,
if we are supposed to get 7 responses back per request, if we clicked 3 times in quick succession, we will end up populating the response pane with 21 responses, all the responses disappear as we get all responses, i have added a screen recording of the same

- post response script seems to be running just after the first response and not waiting for all the streamed responses, how do you think we can tackle this

@helloanoop @ramki-bruno would also love to hear your thoughts on these, how we should we approach these problems!

https://github.com/user-attachments/assets/d242ea2e-f2f9-46f3-be35-39a916669a2e

davirxavier:
@sanish-bruno Hello :)

> how we can handle Accept: */* ? currently we need to add Accept: text/event-stream as a header to the request to make it work?

- The accept: text/event-stream is important only so we can explicitly set the responseType property to "stream" in the axios request before the request is executed. I think i could probably make an automatic "upgrade" to the stream responseType by making one request normally, checking the response's headers with an interceptor for the Content-Type, cancelling the request and making another one with the appropriate responseType value (i don't think the responseType can be changed on the fly with axios), but that would increase complexity a bit though. Do you think it would be worth to include something like this?

> also while making a request, the elapsed time shown currently is of the first stream response, not the entire time elapsed to get all the responses, other api testing clients seems to show the entire time elapsed, what are your thoughts on that?

> the send button get disabled only after we get the first streamed response, multiple clicks done followed by first click are currently resulting in multiple responses being received at the same time, for example,
if we are supposed to get 7 responses back per request, if we clicked 3 times in quick succession, we will end up populating the response pane with 21 responses, all the responses disappear as we get all responses, i have added a screen recording of the same

-  These seem easy to fix, i can work on them this weekend.

> post response script seems to be running just after the first response and not waiting for all the streamed responses, how do you think we can tackle this

- Would it be better for the script to run after the connection is closed or after every event is received from the server?

sanish-bruno:
> * Would it be better for the script to run after the connection is closed or after every event is received from the server?

I think we should be running the post-response scripts, after we received all the required responses and the connection is closed, that has been the behaviour on other api clients, I believe we should be following the same.



> * The accept: text/event-stream is important only so we can explicitly set the responseType property to "stream" in the axios request before the request is executed. I think i could probably make an automatic "upgrade" to the stream responseType by making one request normally, checking the response's headers with an interceptor for the Content-Type, cancelling the request and making another one with the appropriate responseType value (i don't think the responseType can be changed on the fly with axios), but that would increase complexity a bit though. Do you think it would be worth to include something like this?

Current behaviour forces users to explicitly add `Accept: text/event-stream` to make the requests, considering the behaviours over the other api testing clients, users are able to make a requests without adding `Accept: text/event-stream`, I think we should be relying on response headers, instead of request headers to decide the handling of response.

@davirxavier  What are your thought on these?

davirxavier:
I agree with these points, i'm currently working on adjustments for them.

davirxavier:
@sanish-bruno I have made new commits fixing the discussed problems. Below is a list of the changes i had to made in the code. Apologies for the delay.

> how we can handle Accept: */* ? currently we need to add Accept: text/event-stream as a header to the request to make it work?

I've managed to make a full automatic "upgrade" to streamed responses, though i've had to change every request to the "stream" responseType. With this setting, requests will always return a stream object immediately instead of the received data for requests, even for responses that are not text/event-stream. As to not make too much changes to the current code, i've wrapped the stream in a promise, the promise will resolve with the full accumulated data when the stream closes and will reject if any error occurs. So, if the response has the text/event-stream content-type i will mark the response as a stream and return immediately (as was working before), otherwise, i will wrap the stream in the aforementioned promise and await it's completion for the full response data.

For the full folder/collection runner, i have made the same changes, but the requests that have text/event-stream responses will wait for one data event and then stop automatically, so the request will not hang infinitely.

I have tested these changes with text/event-stream responses, json, image, video and html and it seemed to work fine, though this will warrant more testing, because i changed how every request works.

> also while making a request, the elapsed time shown currently is of the first stream response, not the entire time elapsed to get all the responses, other api testing clients seems to show the entire time elapsed, what are your thoughts on that?

I have added a stopwatch to the response panel, it will show the elapsed time in real time. If the stream is closed it will show the full time the stream was active.

> the send button get disabled only after we get the first streamed response, multiple clicks done followed by first click are currently resulting in multiple responses being received at the same time

I have changed the button to not make more requests when the current request is loading. I think the same was happening to other types of requests too? So this will probably fix this for them as well.

> post response script seems to be running just after the first response and not waiting for all the streamed responses, how do you think we can tackle this

The script will only run after the stream closes completely now.

sanish-bruno:
@davirxavier Thank you so much for actively working on it and we really appreciate it. I have tested it myself, and most of it seems to work as expected. we will review it. most probably will have to test the behaviour over on cli. This needs be tested thoroughly as there are many changes in major parts.

we will be proactively following this. Thanks again! we will let you know if there is anything we probably need to care about.

packages/bruno-app/src/providers/App/useIpcEvents.js
```diff
@@ -14,10 +14,10 @@ import {
   collectionUnlinkDirectoryEvent,
   collectionUnlinkEnvFileEvent,
   collectionUnlinkFileEvent,
-  processEnvUpdateEvent,
+  processEnvUpdateEvent, requestCancelled,
```
sanish-bruno: can we format these import statements one followed by one on each line

packages/bruno-app/src/providers/App/useIpcEvents.js
```diff
@@ -14,10 +14,10 @@ import {
...
   runFolderEvent,
   runRequestEvent,
-  scriptEnvironmentUpdateEvent
+  scriptEnvironmentUpdateEvent, streamDataEnded, streamDataReceived
```
sanish-bruno: same here,  we will have to keep the function import one per each line!

sanish-bruno: from the code i get that, we are not using "streamDataEnded", i think we can remove it!

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -274,28 +274,36 @@ export const collectionsSlice = createSlice({
       if (collection) {
         const item = findItemInCollection(collection, itemUid);
         if (item) {
-          item.response = null;
+          if (item.response?.hasStreamRunning) {
+            item.response.hasStreamRunning = null;
+
+            const startTimestamp = item.response?.timeline?.[0].timestamp;
```
sanish-bruno: @davirxavier little confused on whether we should be relying on timeline.timestamp or not!

sanish-bruno: request cancelled can be modified to use requestSent.timestamp

```js
    requestCancelled: (state, action) => {
      const { itemUid, collectionUid } = action.payload;
      const collection = findCollectionByUid(state.collections, collectionUid);

      if (collection) {
        const item = findItemInCollection(collection, itemUid);
        if (item) {
          if (item.response?.hasStreamRunning) {
            item.response.hasStreamRunning = null;

            const startTimestamp = item.requestSent?.timestamp || Date.now();
            item.response.duration = startTimestamp ? Date.now() - startTimestamp : item.response.duration;
          } else {
            item.response = null;
          }

          item.cancelTokenUid = null;
        }
      }
    }
```

and probably in bruno/packages/bruno-electron/src/ipc/network/index.js 
within  requestSent object , we can add timestamp, there are 2 `requestSent` object within the file

```js
  const { data: requestData, dataBuffer: requestDataBuffer } = parseDataFromRequest(request);
      let requestSent = {
        url: request.url,
        method: request.method,
        headers: request.headers,
        data: requestData,
        dataBuffer: requestDataBuffer,
        timestamp: Date.now()
      }
```

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -320,6 +328,12 @@ export const collectionsSlice = createSlice({
       if (collection) {
         const item = findItemInCollection(collection, action.payload.itemUid);
         if (item) {
+          if (item.response && item.response.hasStreamRunning) {
```
sanish-bruno, line 599: @davirxavier may i know , why did we introduce this if statement? was it causing any errors without this check?  little confused as we were already setting item.response = null, just below it, why are we using a different clearing logic for streams?

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -2278,7 +2302,9 @@ export const {
   collectionGetOauth2CredentialsByUrl,
   updateFolderAuth,
   updateFolderAuthMode,
-  moveCollection
+  moveCollection,
+  streamDataReceived,
+  streamDataEnded,
```
sanish-bruno: @davirxavier i wasn't able to find this exported function [streamDataEnded] within the file, and we are also not using this function elsewhere, i guess we can remove this

sanish-bruno (changes requested):

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
sanish-bruno: let me know if it causing any errors, and steps to reproduce it

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
davirxavier: I have changed the stopwatch with this suggestion, all works well!

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
davirxavier: This logic is necessary so the whole request state is not deleted when the clear response button is pressed while the client is receiving events. When the clear response button is clicked with these changes it will just clear the received data, like the respective clear button in other clients (like postman).

davirxavier:
The mentioned issues have been resolved successfully.

motz0815:
Hi, just chiming into the discussion here. I'm currently locally testing this PR and found that if the stream is in JSON format, the result will be displayed immediately after the first response part, but just displays [object Object] throughout and after the stream ends.
image

sanish-bruno:
@davirxavier Thanks a lot for spending time on refining the PR. We have been looking through other API testing clients and how they handle SSE. And many of the clients support SSE as a separate entity, and have a specific UI to go with it. We are thinking of a holistic approach. Even though the PR addresses many things, we will still need to put in lot more effort to make it on par with other tools. 

We will be allocating time for the team to work on that. We will keep you on loop on the progress we make. We really appreciate the time you spend on this PR over the last few weeks, and the effort you put into solving each of the concerns. Thank you!

shwarcu:
Hi, any news about SSE support? is it still on the roadmap? 

I work with systems that use SEE and would love to see even a not-perfect solution that would just allow me to stick with Burno for all my tasks. There's been a significant amount of work put into this contribution, maybe it could be allowed for the time "the holistic solution" is being prepared?

helloanoop:
This is on our roadmap for November. @sid-bruno will be getting this merged this week.

sid-bruno:
Merging it into an internal branch for fixes and improvements over the original implementation

sid-bruno:
Progress can be tracked on https://github.com/usebruno/bruno/pull/6074

sid-bruno:
https://github.com/usebruno/bruno/pull/6074 is now merged, thanks for the contribution @davirxavier , this should be in one of the next releases

PR #4483, opened 2025-04-11
Fix: Mock/Random built-in variables related issues

ramki-bruno:
- Fix: Line-breaks in `$<mock>` built-ins are breaking JSON req body
- Fix: Falsy values from `$<mock>` built-ins are not getting interpolated


helloanoop:
@ramki-bruno Can we include a test in the collection test suite for mock vars ?

ramki-bruno:
> @ramki-bruno Can we include a test in the collection test suite for mock vars ?

Exhaustive test for all the built-in variables and interpolation across all request params(query-params, headers, etc.) is in the works and that's how we found this issue.

But this doesn't have to wait.

PR #4498, opened 2025-04-14
improve: interpolation with object handling

pooja-bruno:
# Description

This PR enhances the interpolation functionality in Bruno by improving how variables are processed

### Contribution Checklist:


Its-treason:
Could you also add a test die Date- and Moment-Objects? See: https://github.com/usebruno/bruno/issues/1910#issuecomment-2018994497 Those seem to behave differently with JSON.stringify

helloanoop:
Thanks @Its-treason !

I have added the tests to document the behavior here - https://github.com/usebruno/bruno/blob/main/packages/bruno-common/src/interpolate/index.spec.ts#L519-L573

We spent a lot of time mulling over whether to JSON.stringify for all objects (including primitive objects, as well as things like Date, moment)

In an ideal scenario we want `JSON.stringify` to work for primitive objects, but `.toString()` for Date and moment.
That approach, however, adds complexity and specificity to the interpolation logic.

Since most users expect objects to be JSON stringified, we've gone with that as the default for now. In the future, we're considering introducing a piping syntax to handle custom formatting—for example, `{{ now | toString() }}` for the `.toString()` use case.

helloanoop:
Closing this PR as this has been addressed in #4519

PR #4540, opened 2025-04-21
Create/Import collection UX improvements

ramki-bruno:
- Improved error-toast for creating and importing collections
  - Added a util for formatting the error form IPC
  - Updated Toast global styles to prevent text overflow.
    Whenever long file paths are shown, it overflows the Toast container.
- Fix: Import-collection select-location Modal closes on error
- Fix: Improve UX for selecting location when create/import collection
  - Allow editing the input path where previously the `<input>` is marked
    `readonly`.
    Also this will allow automating test using Playwright.
    
|||
|-|-|
|image|image|
|image|

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
+++ b/packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
@@ -113,7 +113,6 @@ const CreateCollection = ({ onClose }) => {
             id="collection-location"
             type="text"
             name="collectionLocation"
-            readOnly={true}
             className="block textbox mt-2 w-full cursor-pointer"
             autoComplete="off"
             autoCorrect="off"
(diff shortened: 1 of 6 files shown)


packages/bruno-app/src/components/Sidebar/CreateCollection/index.js
```diff
@@ -113,14 +115,16 @@ const CreateCollection = ({ onClose }) => {
             id="collection-location"
             type="text"
             name="collectionLocation"
```
helloanoop, line 117: I am not sure we need to remove `readonly`.

If its just for the playwright tests to work, we could conditionally allow editing location based on a env var that gets injected at runtime.

ramki-bruno: IMO, it does improve the UX
any reason it should not be editable?

Issue #4573, opened 2025-04-28
TypeError: Cannot read properties of undefined (reading 'toLocaleString')


Bradleykingz:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [x] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.2.0

### Operating System

Manjaro

### Describe the bug

Certain requests make the app crash

I'm not sure how to reproduce it, but it occurs from time to time, even on a fresh collection with ~3 requests.

Pretty sure it's something that began with a recent update though.

```
index.72ed0700.js:681 Object
static/diff2html.min.css:1 
        
        
       Failed to load resource: net::ERR_FILE_NOT_FOUND
lib-react.41929aa2.js:2 TypeError: Cannot read properties of undefined (reading 'toLocaleString')
    at yI (index.72ed0700.js:7:291893)
    at lb (lib-react.41929aa2.js:2:41685)
    at aY (lib-react.41929aa2.js:2:61437)
    at ot (lib-react.41929aa2.js:2:72858)
    at ua (lib-react.41929aa2.js:2:111999)
    at lib-react.41929aa2.js:2:111844
    at ul (lib-react.41929aa2.js:2:111852)
    at i3 (lib-react.41929aa2.js:2:109022)
    at uT (lib-react.41929aa2.js:2:128880)
    at MessagePort.z (lib-react.41929aa2.js:2:180013)
aR @ lib-react.41929aa2.js:2
index.72ed0700.js:684 Objecterror: TypeError: Cannot read properties of undefined (reading 'toLocaleString')
    at yI (file:///snap/bruno/69/resources/app.asar/web/static/js/index.72ed0700.js:7:291893)
    at lb (file:///snap/bruno/69/resources/app.asar/web/static/js/lib-react.41929aa2.js:2:41685)
    at aY (file:///snap/bruno/69/resources/app.asar/web/static/js/lib-react.41929aa2.js:2:61437)
    at ot (file:///snap/bruno/69/resources/app.asar/web/static/js/lib-react.41929aa2.js:2:72858)
    at ua (file:///snap/bruno/69/resources/app.asar/web/static/js/lib-react.41929aa2.js:2:111999)
    at file:///snap/bruno/69/resources/app.asar/web/static/js/lib-react.41929aa2.js:2:111844
[...]

PR #4613, opened 2025-05-07
Fix: ResponseSize component logic to handle size when it's undefined

sreelakshmi-bruno:
Fixes: #4573 

[Jira](https://usebruno.atlassian.net/browse/BRU-1091)

### Changes

- Added handling for undefined and invalid values for `response.size` in the **ResponsePane** and **ResponseSize** components.
- This check ensures that only valid numeric sizes, including `0`, are allowed.

#### Excluded values

The following values are now excluded to prevent crashes or unexpected behavior:

- `undefined`  
- `null`  
- `NaN`  
- `Infinity`  
- `-Infinity`  
- Non-numeric types (e.g., strings, objects)

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/ResponseSize/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/ResponseSize/index.js
@@ -3,7 +3,7 @@ import StyledWrapper from './StyledWrapper';
 
 const ResponseSize = ({ size }) => {
 
-  if (!size) {
+  if (!Number.isFinite(size)) {
     return null;
   }
 
--- a/packages/bruno-app/src/components/ResponsePane/ResponseSize/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/ResponseSize/index.js
@@ -2,14 +2,20 @@ import React from 'react';
 import StyledWrapper from './StyledWrapper';
 
 const ResponseSize = ({ size }) => {
+
+  if (!size) {
+    return null;
+  }
+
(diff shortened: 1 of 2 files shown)


packages/bruno-app/src/components/ResponsePane/ResponseSize/index.js
```diff
@@ -2,14 +2,20 @@ import React from 'react';
 import StyledWrapper from './StyledWrapper';
 
 const ResponseSize = ({ size }) => {
+
+  if (!size) {
```
naman-bruno: I don't think this null check is necessary, since we're already providing a default value with `const responseSize = response.size || 0;`, making the check redundant.

naman-bruno (changes requested):

packages/bruno-app/src/components/ResponsePane/ResponseSize/index.js
sreelakshmi-bruno: I agree it’s redundant in the current usage, but keeping the check adds a layer of safety in case someone uses the component in the future without properly defaulting the size prop, it helps prevent unexpected crashes.

sreelakshmi-bruno: Not sure if I should remove the handling of the size prop from the components using `ResponseSize` since it's been handled in `ResponseSize`.

helloanoop (approved):

PR #4662, opened 2025-05-13
feat: enhance axios instance with redirect handling and cookie management in CLI

sanjaikumar-bruno:


packages/bruno-tests/collection/redirects/Test Redirects With Cookies.bru
ramki-bruno, line 1: Test requests for redirects with `set-cookie` are already added in #4709
You can skip it here.

ramki-bruno (dismissed):

PR #4693, opened 2025-05-16
mv:  function into common

pooja-bruno:
# Description

Move `isValidValue` function into common.js file, also added test for this function

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/utils/common.js
+++ b/packages/bruno-electron/src/utils/common.js
@@ -137,6 +137,10 @@ const parseDataFromRequest = (request) => {
   return parseDataFromResponse(requestCopy);
 };
 
+const isValidValue = (val) => {
+  return typeof val === 'string' && val.length >= 0;
+}
+
 module.exports = {
   uuid,
   stringifyJson,
@@ -147,5 +151,6 @@ module.exports = {
   generateUidBasedOnHash,
   flattenDataForDotNotation,
   parseDataFromResponse,
-  parseDataFromRequest
+  parseDataFromRequest,
+  isValidValue
 };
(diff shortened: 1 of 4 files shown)


helloanoop:
@lohit-bruno I'd like to discuss this before this gets merged.

maintainer-bruno (dismissed):

packages/bruno-electron/src/utils/common.js
```diff
@@ -137,6 +137,10 @@ const parseDataFromRequest = (request) => {
   return parseDataFromResponse(requestCopy);
 };
 
+const isValidValue = (val) => {
```
helloanoop: I'm concerned that placing this function in `common/utils` might lead others to reuse it beyond encryption-related purposes, and potentially updating the logic without recognizing the encryption-specific nuances involved.

I see two approaches that I feel are better
1. Explicitly export `isValueEncryptable` and `isValueDecryptable` from `utils/encryption.js`. It's a bit redundant but ensures clarity and protects against future misuse.
2. Adjust `encryptString` and `decryptString` to simply return an empty string when one is passed.

Ex:
```javascript
function encryptString(str) {
  if (typeof str === 'string' && str.length === 0) {
    return ''
  }
```

helloanoop (changes requested):

pooja-bruno:
Hey @helloanoop, I have done the changes. Can you please merge this PR. Thanks!

helloanoop:
@pooja-bruno I recommend using the `fail-early` principle to make the condition checks more readable

```javascript
function decryptString(str) {
  if (typeof str !== 'string') {
    throw new Error('Decrypt failed: unrecognized string format');
  }
  if (str.length === 0) {
    return '';
  }
}

function encryptString(str) {
  if (typeof str !== 'string') {
    throw new Error('Encrypt failed: invalid string');
  }
  if (str.length === 0) {
    return '';
  }
}
```

naman-bruno (approved):

Issue #4664, opened 2025-05-13
Postman Importer: Improve Handling of "Inherit" for Folders and Requests


philll11:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.2.0

### Operating System

Windows 11 26100.3915

### Describe the bug

Currently, when importing Postman collections (v2.1 export format), Bruno's importer does not accurately translate Postman's "Inherit auth from parent", due to the missing "Inherit" Auth Type at the folder level. This leads to imported collections behaving differently than they did in Postman, requiring manual correction.

#### Postman's Auth Representation in Exports:
- Inherit Auth: If a Postman item (folder or request) is set to "Inherit auth from parent," its JSON representation in the export omits the auth property entirely.
- To contrast: If a Postman item is explicitly set to "No Auth," its JSON representation includes an auth property, e.g., auth: { "type": "noauth" }.

#### Bruno's Current Folder Auth Limitations:
- Bruno folders currently offer "No Auth" and "OAuth 2.0" but lack an explicit "Inherit" option.

#### Current Importer Behavior (The Issue):
- For Folders/Requests with Omitted auth (Postman "Inherit"): The importer seems to resolve the parent's auth at import time and applies it explicitly to the Bruno item. This means if the collection had Basic Auth, imported sub-folders and requests (that were "Inherit" in Postman) now have Basic Auth directly defined on them in Bruno.
- For Folders with auth: { "type": "noauth" } (Postman "No Auth"): The importer correctly maps this to Bruno's "No Auth" for the folder.
  - And as a result, requests that were "Inherit" in Postman are now explicitly set to "No Auth" in Bruno.

### Impact
- Items that should inherit auth (and thus adapt to changes at the collection/parent level in Bruno) instead have auth hardcoded. This makes managing auth at a higher level in Bruno ineffective for these imported items.
- Users migrating from Postman need to manually change the auth settings for many imported folders and requests back to an inheriting state, which is tedious and error-prone.

#### Part 1: Enhance Bruno Folder Authentication
- Add an "Inherit" option to the Auth tab for folders in the Bruno UI.
- Ensure Bruno's request execution logic correctly processes this "Inherit" state for folders, allowing auth to flow down from the collection.

##### For Imported Folders:
[...]

PR #4443, opened 2025-04-05
feat: add ability to select requests in runner

therealrinku:
# Description
add ability to select requests in collection runner allowing users to run the selected requests only instead of the all the requests in the collection.

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ ] **Create an issue and link to the pull request.**

https://github.com/user-attachments/assets/e27745f6-11f5-443e-913f-8906f1c3e7a3

Issue #4329, opened 2025-03-25
Select individual requests from collection for runner


RatkoZG:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] blocks me from using Bruno
- [x] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### Describe the feature you want to add, and how it would change your usage of Bruno

It would be great if, when running a collection, user can select/unselect some of request from runner. If have 20 requests in my collection, it would be great if I can unselect few of them from running, without editing collection.

sreelakshmi-bruno:
Hi @RatkoZG, that’s a great feature! I’ll get back to you on the timeline for implementation.

helloanoop:
This is now available in Bruno 🎉 

<img width="2278" height="1554" alt="Image" src="https://github.com/user-attachments/assets/16d1a612-4294-40b4-9278-8c450420c263" />

Issue #4303, opened 2025-03-21
Enhancement debug run result


nudasilva:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] blocks me from using Bruno
- [x] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### Describe the feature you want to add, and how it would change your usage of Bruno

After running a collection, i can get run result.
However, a list of all tests is displayed, without filter.

![Image](https://github.com/user-attachments/assets/6afb7eb5-f770-414b-a70d-c93adc1aa634)

Is it possible to be able to filter the results, like in Postman ?
For instance, select Failed, so that the whole request fails are displayed and we can focus on failed tests ?:

![Image](https://github.com/user-attachments/assets/d1d1704c-66cc-498c-bd3d-efb1bb13ed12)

nudasilva:
Thanks a lot.

glewe:
Great news

naman-bruno:
Hey, we’ve already added filters in Bruno, so I’m closing this issue. If you have any questions, feel free to ask.
thanks!

PR #4759, opened 2025-05-25
bug/improve-handling-of -Inherit-for-folders-and-requests

philll11:
Pull request resolves issue [Postman Importer: Improve Handling of "Inherit" for Folders and Requests](https://github.com/usebruno/bruno/issues/4664) bug.

**Key Changes:**

1. UI: Added "Inherit from Collection" Auth Option for Folders
    - Folders in Bruno can now explicitly be set to "Inherit from Collection" in their Auth tab.
    - This addresses a gap where folders previously lacked a distinct "inherit" state separate from "No Auth."
    
image

2. Importer: Correctly Map Postman's "Inherit Auth"
    - Modified the Postman importer (packages/bruno-converters).
    - When a Postman folder or request has its auth property omitted (indicating "Inherit auth from parent" in Postman), it is now imported into Bruno with its authentication mode set to "Inherit."
    - This prevents parent auth from being incorrectly hardcoded onto child items that should be inheriting.
3. Request Pane Logic: Allow "No Auth" to be Inherited Correctly
    - Ensured that if a parent item (collection or folder) is set to "No Auth," child items set to "Inherit Auth" will correctly inherit this "No Auth" state, matching Postman's behavior.
    - Previously, "No Auth" on a Bruno folder would break the inheritance chain differently.

**Unit Tests:**

- Added new unit tests to _collection-auth.spec.js_, _folder-auth.spec.js_, and _request-auth.spec.js_ within packages/bruno-converters to cover these scenarios, including:
    - Collection "No Auth."
    - Folder/Request "Inherit Auth" (absent auth property).
    - Folder/Request explicit "No Auth" (type: "noauth").
    - Multi-level inheritance scenarios.

### DIFF
--- a/packages/bruno-converters/package.json
+++ b/packages/bruno-converters/package.json
@@ -12,7 +12,7 @@
   ],
   "scripts": {
     "clean": "rimraf dist",
-    "test": "node --experimental-vm-modules $(npx which jest) --colors --collectCoverage",
+    "test": "jest --colors --collectCoverage",
     "prebuild": "npm run clean",
     "build": "rollup -c",
     "watch": "rollup -c -w",
--- a/packages/bruno-converters/src/postman/postman-to-bruno.js
+++ b/packages/bruno-converters/src/postman/postman-to-bruno.js
@@ -2,7 +2,7 @@ import get from 'lodash/get';
 import { validateSchema, transformItemsInCollection, hydrateSeqInCollection, uuid } from '../common';
 import each from 'lodash/each';
 import postmanTranslation from './postman-translations';
-import { invalidVariableCharacterRegex } from '../constants/index';  
+import { invalidVariableCharacterRegex } from '../constants/index';
 
 const parseGraphQLRequest = (graphqlSource) => {
   try {
@@ -45,7 +45,7 @@ const convertV21Auth = (array) => {
 
 const constructUrlFromParts = (url) => {
   if (!url) return '';
-  
+
   const { protocol = 'http', host, path, port, query, hash } = url || {};
   const hostStr = Array.isArray(host) ? host.filter(Boolean).join('.') : host || '';
   const pathStr = Array.isArray(path) ? path.filter(Boolean).join('/') : path || '';
@@ -53,9 +53,9 @@ const constructUrlFromParts = (url) => {
   const queryStr =
     query && Array.isArray(query) && query.length > 0
       ? `?${query
-          .filter((q) => q && q.key)
-          .map((q) => `${q.key}=${q.value || ''}`)
-          .join('&')}`
+        .filter((q) => q && q.key)
+        .map((q) => `${q.key}=${q.value || ''}`)
+        .join('&')}`
       : '';
   const urlStr = `${protocol}://${hostStr}${portStr}${pathStr ? `/${pathStr}` : ''}${queryStr}`;
   return urlStr;
@@ -86,31 +86,31 @@ const constructUrl = (url) => {
   return '';
 };
 
-const importScriptsFromEvents = (events, requestObject) => {
-  events.forEach((event) => {
+const importScriptsFromEvents = (postmanEvents, brunoRequestObject) => {
+  postmanEvents.forEach((event) => {
     if (event.script && event.script.exec) {
       if (event.listen === 'prerequest') {
-        if (!requestObject.script) {
-          requestObject.script = {};
+        if (!brunoRequestObject.script) {
+          brunoRequestObject.script = {};
         }
 
         if (event.script.exec && event.script.exec.length > 0) {
-          requestObject.script.req = postmanTranslation(event.script.exec)
+          brunoRequestObject.script.req = postmanTranslation(event.script.exec)
         } else {
-          requestObject.script.req = '';
+          brunoRequestObject.script.req = '';
           console.warn('Unexpected event.script.exec type', typeof event.script.exec);
         }
       }
 
       if (event.listen === 'test') {
-        if (!requestObject.tests) {
-          requestObject.tests = {};
+        if (!brunoRequestObject.tests) {
+          brunoRequestObject.tests = {};
         }
 
         if (event.script.exec && event.script.exec.length > 0) {
-          requestObject.tests = postmanTranslation(event.script.exec)
+          brunoRequestObject.tests = postmanTranslation(event.script.exec)
         } else {
-          requestObject.tests = '';
+          brunoRequestObject.tests = '';
           console.warn('Unexpected event.script.exec type', typeof event.script.exec);
         }
       }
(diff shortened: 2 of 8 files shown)


sanish-bruno:
@philll11 will you able to resolve the conflicts, there are been lot of changes post the lat commit.

packages/bruno-converters/package.json
```diff
@@ -12,7 +12,7 @@
   ],
   "scripts": {
     "clean": "rimraf dist",
-    "test": "node --experimental-vm-modules $(npx which jest) --colors --collectCoverage",
```
sanish-bruno: @philll11 can you help me understand why this change was made?

sanish-bruno (changes requested):

packages/bruno-converters/src/postman/postman-to-bruno.js
```diff
@@ -2,7 +2,7 @@ import get from 'lodash/get';
 import { validateSchema, transformItemsInCollection, hydrateSeqInCollection, uuid } from '../common';
```
sanish-bruno, line 2: @philll11 we have already pushed some changes and we moved away from if, else if chaining to switch statements for `processAuth` Function ,  also i see few formatting changes. I believe,  we most likely need to only change the behaviour where the children get assigned the parents auth automatically, and when there is no auth defined. 

```js
      const auth = i.request.auth ?? parentAuth;
      processAuth(auth, brunoRequestItem.request);
```
i can also see few naming changes, where 

```
collection --->. postmanCollection
requestObject ---> brunoRequestObject
```
etc

This will make the review a  bit difficult. as the diff is much larger. Is it possible for you to rework on the issue so that we can have the smallest diff possible, that just addresses the issue for inheritance.

I really appreciate the time you invested on this PR so far.

philll11: Hi @sanish-bruno Thanks for the feedback - you are right, the diff is larger than it needs to be.
I will simplify this PR to focus only on the inheritance fix as you suggested. I'll work from v2.6.1 and won't include the extra formatting, renaming, and refactoring changes.
I'll look to push the updated version over the weekend.

sanish-bruno: @philll11 Thanks, let us know once you raise the PR

packages/bruno-converters/package.json
philll11: node --experimental-vm-modules $(npx which jest) could not find the correct jest modules so I replaced it in favor of the standard jest config. I have since reverted this to reduce the area of impact.

packages/bruno-converters/src/postman/postman-to-bruno.js
philll11: @sanish-bruno I’ve narrowed the changes to focus just on fixing the inherit bug while also merging the recent 2.7.0 version.

One suggestion: the processAuth function is getting quite big and hard to follow. It might help to move the OAuth logic into a separate internal function.
Also, variable names like auth and requestObject are a bit unclear - it’s not easy to tell which one is from Postman and which is from Bruno.

With all that being said, I do understand the need to change as little as possible. Please let me know if this reduction is enough.

sanish-bruno: @philll11 Yes, you are absolutely right, It is important to have a refactor. Thanks for working on this, I am in the process of reviewing/testing it. Hopefully will get it merged soon.

packages/bruno-converters/src/postman/postman-to-bruno.js
```diff
@@ -140,126 +140,133 @@ const importCollectionLevelVariables = (variables, requestObject) => {
   requestObject.vars.req = vars;
 };
 
-export const processAuth = (auth, requestObject) => {
-  if (!auth || !auth.type || auth.type === AUTH_TYPES.NOAUTH) {
-    return;
-  }
+export const processAuth = (auth, requestObject, collection = false) => {
```
sanish-bruno, line 143: @philll11 could have renamed it to "isCollection" for better clarity

packages/bruno-converters/src/postman/postman-to-bruno.js
```diff
@@ -140,126 +140,133 @@ const importCollectionLevelVariables = (variables, requestObject) => {
...
+  // Handle folder/request specific "Inherit"
+  if (!auth) return; // Return as requestObject is a folder/request and has a default mode = inherit
+
+  // Handle folder/request specific "No Auth"
+  if (auth.type === AUTH_TYPES.NOAUTH) {
+    requestObject.auth.mode = AUTH_TYPES.NONE; // Set the mode to none
+    return; // No further processing needed
   }
 
+  let authValues = auth[auth.type] ?? [];
```
sanish-bruno, line 161: @philll11 what will happen if `auth object` extists, but `auth.type `is not defined, previously we used to return early, without setting an auth mode, and `as a default auth.mode will none`, and now it's look like we will be setting 
`requestObject.auth.mode` to be undefined. 

I feel like to handle this case, we should explicitly setting `requestObject.auth.mode = AUTH_TYPES.NONE` within the `switch statement default case`, would like your thoughts on this.

This will also help us when postman introduces a new auth type, we will be able to import them gracefully,
I am planning to raise a follow-up PR to introduce a small fix to address this issue. Just wanted to inform you!

Let me know if you see any potential issue with this approach

Followup PR #5068

packages/bruno-converters/src/postman/postman-to-bruno.js
sanish-bruno: i have handled it in the followup PR #5068

sanish-bruno:
Followup PR fixing auth.type being undefined and respective test cases #5068

packages/bruno-converters/src/postman/postman-to-bruno.js
philll11: @sanish-bruno great fallback/future proofing idea - I don't see any immediate issues with this approach.

packages/bruno-converters/src/postman/postman-to-bruno.js
```diff
@@ -140,126 +140,133 @@ const importCollectionLevelVariables = (variables, requestObject) => {
...
-          clientId: findValueUsingKey('clientId'),
-          clientSecret: findValueUsingKey('clientSecret'),
-          scope: findValueUsingKey('scope'),
-          state: findValueUsingKey('state'),
-          tokenPlacement: findValueUsingKey('addTokenTo') == 'header' ? 'header' : 'url',
-          credentialsPlacement: findValueUsingKey('client_authentication') == 'body' ? 'body' : 'basic_auth_header'
-        };
+
+      const postmanGrantType = findValueUsingKey('grant_type');
+      const targetGrantType = oauth2GrantTypeMaps[postmanGrantType] ?? 'client_credentials'; // Default
```
sanish-bruno, line 215: @philll11 any reason why we went ahead with `client_credentials` as the default? previously, according to tests we fallbacked to `authorization_code`, I think it is fine, what do you guys think? @maintainer-bruno @naman-bruno @lohit-bruno

philll11: @sanish-bruno `client_credentials` fallback felt like a more natural fit as it only requires the base OAuth fields defined in `baseOAuth2Config`, keeping the switch default logic cleaner. However, this does alter existing behaviour (default grant type = authorization code) so happy to revert if this is not something we want to change.

sanish-bruno: @philll11 I felt the same, it felt cleaner to have client_credentials as the default fallback. let's see what rest of the team thinks! while selecting from the UI we defaults to `authorization_code` that's the inconsitency i could see.

sanish-bruno:
Fixes #4825

philll11:
@sanish-bruno as an aside, do you have any other bugs or features that need attention? I'd really like to keep contributing!

sanishchirayath1:
> @sanish-bruno as an aside, do you have any other bugs or features that need attention? I'd really like to keep contributing!

@philll11 Feel free to go through the issues, and find few issues that you'll be interested to work on, we can have a conversation on what will be the expectations from the ticket!

philll11:
> > @sanish-bruno as an aside, do you have any other bugs or features that need attention? I'd really like to keep contributing!
> 
> @philll11 Feel free to go through the issues, and find few issues that you'll be interested to work on, we can have a conversation on what will be the expectations from the ticket!

@sanish-bruno I personally would love to improve the Runner as I use it for regression testing my APIs. I see there are many bugs and features (#3530 #4303 #4329), including some PRs (#4443 #3808) but nothing that looks like it is being actively worked on. What do you think about this topic?

sanish-bruno:
@philll11 the changes are now merged with #5107, I am closing this issue now

PR #4778, opened 2025-05-27
expose path parameters in js api

Skrivoo:
# Description

In a prescript it is possible to get every path parameters as an array of objects like
http://localhost:1234/customers/:customerId/something

`req.getPathParams()`

with this parameters like :customerId can be resolved in prescript. It is useful in some cases for example to have an escher auth (https://escherauth.io/) where the url needs to be compared in server side with the one witch the request contains and for this, every parameter and envVar must be resolved.


Skrivoo:
I'll reopen this with resolved conflict and a linked issue

PR #4782, opened 2025-05-28
Fix: Special URI characters in proxy username/password is giving error

ramki-bruno:
URI-encoding the _username_ and _password_ before creating the proxy URI which then gets passed to `HttpsProxyAgent` and `HttpProxyAgent` respectively.

### DIFF
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -211,8 +211,8 @@ const runSingleRequest = async function (
         let uriPort = isUndefined(proxyPort) || isNull(proxyPort) ? '' : `:${proxyPort}`;
         let proxyUri;
         if (proxyAuthEnabled) {
-          const proxyAuthUsername = interpolateString(get(proxyConfig, 'auth.username'), interpolationOptions);
-          const proxyAuthPassword = interpolateString(get(proxyConfig, 'auth.password'), interpolationOptions);
+          const proxyAuthUsername = encodeURIComponent(interpolateString(get(proxyConfig, 'auth.username'), interpolationOptions));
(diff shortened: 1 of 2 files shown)


packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -211,8 +211,8 @@ const runSingleRequest = async function (
         let uriPort = isUndefined(proxyPort) || isNull(proxyPort) ? '' : `:${proxyPort}`;
         let proxyUri;
         if (proxyAuthEnabled) {
-          const proxyAuthUsername = interpolateString(get(proxyConfig, 'auth.username'), interpolationOptions);
-          const proxyAuthPassword = interpolateString(get(proxyConfig, 'auth.password'), interpolationOptions);
+          const proxyAuthUsername = encodeURIComponent(interpolateString(get(proxyConfig, 'auth.username'), interpolationOptions));
+          const proxyAuthPassword = encodeURIComponent(interpolateString(get(proxyConfig, 'auth.password'), interpolationOptions));
```
maintainer-bruno, line 215: I think we should do the same for proxyHostname too.

maintainer-bruno (changes requested):

packages/bruno-cli/src/runner/run-single-request.js
ramki-bruno: Special characters like `%/#: ` are  invalid in _host_.
Things like Emoji doesn't require percentage-encoding.

ramki-bruno: I think we should be good w.r.t. `proxyHostname`.
Besides, right now one can add the port or even  path(for some reason) in the Hostname field itself and skip the Port field. Encoding it will break those use-cases.

PR #4800, opened 2025-05-29
Added validator to check if a given path is inside an open Collection

ramki-bruno:
- Added validator to check if a given path is inside an open Collection.
And added an sample validation for _new-request_ IPC event.
- Refactor: Renamed `Watcher` class and instance to `CollectionWatcher`.
The name `Watcher` sounds very generic, but in this case its tightly
coupled with watching Bruno Collection paths. So it makes sense to name
it accordingly.


ramki-bruno:
The refactoring is in a separate commit in case if you want a clearer diff.

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -58,6 +59,16 @@ const envHasSecrets = (environment = {}) => {
   return secrets && secrets.length > 0;
 };
 
+const validateCollectionSubPath = (path) => {
```
helloanoop: I feel a better name is `validatePathIsInsideCollection`

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -238,6 +249,7 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
         throw new Error(`${request.filename}.bru is not a valid filename`);
       }
       const content = await jsonToBruViaWorker(request);
+      validateCollectionSubPath(pathname);
```
helloanoop: I think we should be doing the validation before we call `jsonToBruViaWorker`

maintainer-bruno (approved):

naman-bruno (approved):

Issue #4763, opened 2025-05-26
Portable version and dependencies issue windows


ganesh-bruno:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [x] this feature was working in a previous version but is broken in the current release.

### Bruno version

1.39.0

### Operating System

Windows

### Describe the bug

Since the version after 1.34, the portable version on Windows creates files in user folders (e.g., AppData), which causes issues on machines with AppLocker enabled.

Version 1.34 works as expected, keeping all required files within the same folder as the binary.

Environment:

- Windows with AppLocker enabled

- Affected versions: After 1.34

### .bru file to reproduce the bug

NA

### Screenshots/Live demo link

![Image](https://github.com/user-attachments/assets/fc22128e-36ab-439f-b03d-a28c6d5a5d49)

anusree-bruno:
This issue will be fixed by https://github.com/usebruno/bruno/pull/4173, which provides an option to add custom installation path.

Pragadesh44-Bruno:
Hey @ganesh-bruno, in our latest release, we’ve addressed this issue and implemented a fix. Please check version [v2.6.0](https://github.com/usebruno/bruno/releases/tag/v2.6.0)

Pragadesh44-Bruno:
Closing this since it's completed

PR #4816, opened 2025-06-02
Feat/ ASAR unpack

Pragadesh-45:
Jira: [BRU-625](https://usebruno.atlassian.net/browse/BRU-625)
fixes: https://github.com/usebruno/bruno/issues/4763
### Description

This PR introduces the use of `asar.smartUnpack` and configures specific patterns for `asarUnpack` in the Electron build process to avoid packaging native modules and certain node packages into the ASAR archive. This change is necessary to ensure the app launches correctly even when the system's temp folder is locked or inaccessible.

### Changes Implemented

```js
asar: {
  smartUnpack: true
},
asarUnpack: [
  "**/*.node",
  "native_modules/**/*"
]
```

* Enabled `smartUnpack` to allow Electron to selectively unpack files that are known to cause runtime issues when kept inside the ASAR archive.
* Explicitly excluded `.node` files and the `native_modules` directory from ASAR packaging.

### Reason

Electron's ASAR packaging has known limitations, especially when it comes to native modules and certain Node.js APIs. These limitations can cause the app to fail to start in certain environments (e.g., when the temp folder is locked or restricted).

Relevant issues:

* Native modules (`*.node` files) require `process.dlopen`, which cannot work reliably inside an ASAR archive.
* Some APIs (e.g., `fs.open`, `child_process.execFile`) trigger unpacking to temporary directories at runtime, which fails when access to those directories is restricted.

By unpacking these files at build time, we avoid these issues entirely.

### References

* [Electron ASAR limitations documentation](https://www.electronjs.org/docs/latest/tutorial/application-packaging#limitations-of-the-node-api)
* [Related issues with native modules in ASAR](https://github.com/electron/asar/issues)

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/package.json
+++ b/packages/bruno-electron/package.json
@@ -28,6 +28,13 @@
       "node_modules"
     ]
   },
+  "asar": {
+    "smartUnpack": true
+  },
+  "asarUnpack": [
+    "**/*.node",
+    "native_modules/**/*"
+  ],
   "dependencies": {
     "@aws-sdk/credential-providers": "3.750.0",
     "@faker-js/faker": "^9.5.1",


packages/bruno-electron/package.json
```diff
@@ -28,6 +28,13 @@
       "node_modules"
     ]
   },
+  "asar": {
+    "smartUnpack": true
+  },
+  "asarUnpack": [
+    "**/*.node",
+    "native_modules/**/*"
```
naman-bruno, line 36: @Pragadesh-45 Could you clarify the reason for including `native_modules` in this context?

Pragadesh-45: @naman-bruno  I've added the 'native_modules' folder because it contains native C++ .node bindings, which is essential for the app to function correctly. These bindings should to be accessible at runtime to avoid potential issues

maintainer-bruno (approved):

packages/bruno-electron/package.json
maintainer-bruno: @Pragadesh-45 Why is this added to package.json instead of `packages/bruno-electron/electron-builder-config.js` ?

Issue #1473, opened 2024-01-29
Support custom request methods (such as vault's LIST request)


paul-at-cybr:
LIST is used as the primary request method on the Vault API's [List Roles](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#list-roles) endpoint.

A GET alternative is available since they know they're being weird, but it might be nice to support custom request methods on a general basis.

DutchForeigner:
In my work place, we are required to move away from Postman (to Bruno). The challenge for us is indeed that the LIST method is used with HashiCorp Vault. As a result we could not import our Vault collections from Postman (without renaming each LIST entry to GET) and existing 'Requests' with List are now not working.

So I would 2nd this Feature request to add custom methods similar to what is in Postman
![image](https://github.com/usebruno/bruno/assets/76488714/289393f6-b141-4684-9b1d-a7373094abfe)

Its-treason:
As a workaround: You can set custom request methods inside the pre-request-script:

```js
req.setMethod("SEARCH")
```

rubancar:
any update on this? I'm in the same situation needing the method LIST.

Its-treason:
> any update on this? I'm in the same situation needing the method LIST.

Nope, you still have to use the workaround, I mentioned in the previous comment

DutchForeigner:
FYI: For people who need the LIST method for use with HashiCorp Vault, you can replace LIST with **GET** + add **?list=true** to the URL (e.g. add a query to it)

Rockybilly:
I also need this for the PURGE method. Although it's not standard, I believe it's crazy to not have custom http method support for an app such as Bruno

paul-at-cybr:
🎉 🎉

helloanoop:
PR has been merged! This will be available in our upcoming `2.11.0` release scheduled to go out on 9th Sep 2025

https://github.com/user-attachments/assets/23b2d1cb-e9fc-4d26-9f1e-5db33a8aa771

PR #4841, opened 2025-06-05
feat: custom HTTP method

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-1139)

This PR introduces a feature that allows adding custom HTTP methods for requests. Important changes are made around the 'bru lang' schema. Now, a standard request using typical HTTP methods as referenced here [HTTP Methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods) is stored like this, maintaining previous behavior:

```
...
get {
  url: https://example.com/api/resource
  body: json
  auth: inherit
}
...
```

Requests with custom HTTP methods—i.e., methods other than those referenced in [HTTP Methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods)—will be stored differently:
```
...
http {
  method: custommethod
  url: https://example.com/api/resource
  body: json
  auth: inherit
}
...
```

https://github.com/user-attachments/assets/f86b2cf6-cf00-4ae3-9fbb-c2801df8712b

resolves https://github.com/usebruno/bruno/issues/1473


packages/bruno-converters/src/postman/postman-to-bruno.js
```diff
@@ -330,7 +337,7 @@ const importPostmanV2CollectionItem = (brunoParent, item, parentAuth, { useWorke
         seq: index + 1, 
         request: {
           url: url,
-          method: i?.request?.method?.toUpperCase(),
+          method: method,
```
sanish-bruno, line 362: @sanjaikumar-bruno we used to uppercase the method, any explicit reason, why are we not doing it now?

packages/bruno-lang/v2/src/bruToJson.js
```diff
@@ -768,7 +789,19 @@ const sem = grammar.createSemantics().addAttribute('ast', {
     return {
       docs: outdentString(textblock.sourceString)
     };
-  }
+  },
```
sanish-bruno: @sanjaikumar-bruno  why did we add httpcustom twice in this file? we already have the same in lines 409 - 420

packages/bruno-lang/v2/src/jsonToBru.js
```diff
@@ -43,23 +43,30 @@ const jsonToBru = (json) => {
...
+    if (isStandard) {
+      bru += `${http.method} {\n  url: ${http.url}`;
+      if (http.body && http.body.length) {
+        bru += `\n  body: ${http.body}`;
+      }
+      if (http.auth && http.auth.length) {
+        bru += `\n  auth: ${http.auth}`;
+      }
+      bru += `\n}\n\n`;
+    } else {
```
sanish-bruno: @sanjaikumar-bruno   I believe, for custom methods, we are choosing to go with a structure, where

```
http {
   method: custom-method
 }
```
currently with the latest curl parser, the following curl
```bash
curl --location --request POP 'http://example.com/file'
```
creates following bru file, i believe we should have an internal  discussion on this on how setup bru file  while importing it as a curl

```
meta {
  name: pop-custom-method
  type: http
  seq: 2
}

pop {
  url: http://example.com/file
  body: none
  auth: inherit
}
```

packages/bruno-converters/src/postman/postman-to-bruno.js
```diff
@@ -36,6 +36,12 @@ const isItemAFolder = (item) => {
   return !item.request;
 };
 
+const isValidHttpToken = (token) => {
```
sanish-bruno: @sanjaikumar-bruno  can we move this function to utils file? most probably we also will need this function within other converters too, right? like openapi etc. I can also see the same regex being used in

`packages/bruno-schema/src/collections/index.js` LINE 52

Probably can also consider to move it into bruno-commons if possible

packages/bruno-schema/src/collections/index.js
```diff
@@ -48,7 +48,8 @@ const varsSchema = Yup.object({
 
 const requestUrlSchema = Yup.string().min(0).defined();
 const requestMethodSchema = Yup.string()
-  .oneOf(['GET', 'POST', 'PUT', 'DELETE', 'PATCH', 'HEAD', 'OPTIONS', 'TRACE'])
+  .min(1, 'method is required')
+  .matches(/^[!#$%&'*+.^_`|~0-9a-zA-Z-]+$/, 'method must be a valid HTTP token')
```
sanish-bruno: @sanjaikumar-bruno  if  `isValidToken ` is placed in bruno common, we can changes here to validate using .test by passing the custom validation function

packages/bruno-schema/src/collections/requestSchema.spec.js
```diff
@@ -18,23 +18,61 @@ describe('Request Schema Validation', () => {
     expect(isValid).toBeTruthy();
   });
 
-  it('request schema must throw an error of method is invalid', async () => {
+  it('request schema must validate successfully - custom method', async () => {
```
sanish-bruno, line 21: We probably need to add test cases for `bru-lang` parsing once the internal discussion is done on how should we represent custom methods within bru files, we have very limited tests there. will help us make the system more robust @sanjaikumar-bruno 

`packages/bruno-lang/v2/tests`

packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/StyledWrapper.js
```diff
@@ -9,7 +9,6 @@ const Wrapper = styled.div`
 
   .method-selector {
     border-radius: 3px;
-    min-width: 90px;
```
sanish-bruno, line 12: @sanjaikumar-bruno  i see that we are using the `HttpMethodSelector` component  where we create new request, the ui seems to distort when we try to input custom method from there, also the input field seems to have a background color that is not needed here, probably can keep it `transparent`, so that it can be used everywhere?

 also we need to have a discussion on curl imports


https://github.com/user-attachments/assets/e13e7aed-56f5-47c0-a622-eef5d9528488

packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/index.js
```diff
@@ -1,52 +1,151 @@
...
+        </span>
+        <IconCaretDown className="caret" size={16} strokeWidth={2} />
+      </button>
+    </div>
+  );
+});
 
-  const handleMethodSelect = (verb) => onMethodSelect(verb);
+const HttpMethodSelector = ({ method = DEFAULT_METHOD, onMethodSelect }) => {
+  const [inputValue, setInputValue] = useState(method || DEFAULT_METHOD);
```
sanish-bruno: @sanjaikumar-bruno Do we need this, as we are already have a default value within params?
`const [inputValue, setInputValue] = useState(method || DEFAULT_METHOD)`
can you check if we can use this?
`const [inputValue, setInputValue] = useState(method)`

packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/index.js
```diff
@@ -1,52 +1,151 @@
...
     return (
-      <div ref={ref} className="flex w-full items-center pl-3 py-1 select-none uppercase">
-        <div className="flex-grow font-medium" id="create-new-request-method">
-          {method}
-        </div>
-        <div>
-          <IconCaretDown className="caret ml-2 mr-2" size={14} strokeWidth={2} />
-        </div>
+      <div className="flex flex-col w-full">
+        <input
```
sanish-bruno, line 27: `ctrl/cmd + s` is not working while inside the custom method input

packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/index.js
sanish-bruno: @sanjaikumar-bruno as we are storing the request as a draft, we don't need to have the above useState, we can actually derive the state from redux store, keeping a single source of truth, and more concise.

Example [Gist](https://gist.github.com/sanish-bruno/17baab128b7013e257855957f49a77c6)

packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/index.js
```diff
@@ -1,52 +1,151 @@
...
+    dropdownTippyRef.current?.hide();
+    setTimeout(() => {
+      inputRef.current?.focus();
+      inputRef.current?.select();
+    }, 0);
+  };
+
+  const handleKeyDown = (e) => {
+    if (e.key === 'Escape') {
+      setInputValue(method || DEFAULT_METHOD);
```
sanish-bruno: @sanjaikumar-bruno is it necessary to have this `|| ` while calling `onMethodSelect` I have removed it in the gist i shared, can you check whether it is causing any issues?

sanish-bruno:
> [JIRA](https://usebruno.atlassian.net/browse/BRU-1139)
> 
> resolves #1473
> 
>  Screen.Recording.2025-06-10.at.16.43.36.mov

@sanjaikumar-bruno update the PR description

packages/bruno-lang/v2/src/jsonToBru.js
sanjaikumar-bruno: I made some changes, now the bru looks like the following,

```
meta {
  name: Test Custom Method
  type: http
  seq: 1
}

http {
  method: custommethod
  url: https://example.com/
  body: none
  auth: inherit
}
```

sanish-bruno:
@helloanoop please review the custom method bru-lang implementation

packages/bruno-lang/v2/src/bruToJson.js
```diff
@@ -436,6 +437,26 @@ const sem = grammar.createSemantics().addAttribute('ast', {
       }
     };
   },
+  trace(_1, dictionary) {
```
sanish-bruno, line 487: @sanjaikumar-bruno trace was not there before? why did we add it now?

packages/bruno-common/src/utils/http-token.ts
```diff
@@ -0,0 +1,9 @@
+/**
```
sanish-bruno: @sanjaikumar-bruno if you have a index.ts file within within, it will be better to add them within that file, here the file name seems to be related to the function, that is not ideal, can you think about a better name for this file or move it into another file within utils were it make sense.

sanish-bruno (changes requested):

sanish-bruno:
@helloanoop will you be able to review the changes for custom method! as this has some changes for bru-lang

sanish-bruno:
@sanjaikumar-bruno please resolve the conflicts

sanish-bruno:
@sanjaikumar-bruno have you tested the behaviour on runner as well as cli? any endpoints that can aid in testing the custom methods?

sanish-bruno (approved):

packages/bruno-schema/src/collections/requestSchema.spec.js
```diff
@@ -18,23 +18,61 @@ describe('Request Schema Validation', () => {
...
 
-    return Promise.all([
-      expect(requestSchema.validate(request)).rejects.toEqual(
-        validationErrorWithMessages(
-          'method must be one of the following values: GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS, TRACE'
-        )
-      )
-    ]);
+    const isValid = await requestSchema.validate(request);
+    expect(isValid).toBeTruthy();
```
sanish-bruno, line 33: probably to.be.true is more accurate?

packages/bruno-lang/v2/src/bruToJson.js
```diff
@@ -435,6 +436,26 @@ const sem = grammar.createSemantics().addAttribute('ast', {
...
```
sanish-bruno, line 487: Please add comments before resolving review comments, this will give more clarity, when we revisit, Currently i don't have clarity on why did it got resolved

sanjai0py: Before we only listed the Trace method in the grammar but didn’t handle it in the parser. Now that we treat standard and custom HTTP methods differently in .bru files, we need the logic to parse Trace requests properly. This change makes sure Trace requests are supported and handled like the other standard HTTP methods.

sanish-bruno:
@sanjaikumar-bruno i guess the the unit tests are failing cause we have changed  from toBeTruthy to toBe(true), why validate function returns object? Probably something specific to Yup, what does it return when the validation fails? I guess we need to investigate and make adjustments accordingly

sanish-bruno (approved):

PR #4845, opened 2025-06-05
Fix: Close open Modals after each test to reusing `page` for next test

ramki-bruno:


playwright/index.ts
```diff
@@ -116,6 +116,11 @@ export const test = baseTest.extend<
     } else {
       await use(page);
     }
+
+    // Close any open Modals after each `test`
+    if(await page.locator('.bruno-modal').isVisible()) {
+      await page.locator('body').press('Escape');
```
maintainer-bruno, line 122: I think closing the modal after every test blocks may lead to unwanted behavior. This is executed after the `beforeAll` hook too.
It should be the responsibility of each test suite or test block to take care of closing any open modals when needed.

maintainer-bruno:
Closing the PR as invalid.

playwright/index.ts
ramki-bruno: > I think closing the modal after every test blocks may lead to unwanted behavior

Actually that's the point. Can you explain more on the "unwanted behavior"?

ramki-bruno: > It should be the responsibility of each test suite or test block to take care of closing any open modals when needed.

but, it doesn't hurt to close it when one of the tests doesn't resulting in further tests in the same worker to fail.

PR #4889, opened 2025-06-13
OAuth: Allow the authorizeUrl to be a subpath of the callbackUrl

kekru:
# Description

First of all: Thanks for this nice tool :)
This is a little bugfix, to allow an OAuth authorizeUrl to be a subpath of the callbackUrl.

We have the following scenario:

- App = CallbackUrl running on `https://apigateway-something.mycompany.dev`
- Keycloak = AuthorizeUrl running on `https://apigateway-something.mycompany.dev/auth/realms/my-realm/protocol/openid-connect/auth`

So the special case is here, that the authorizeUrl is a subpath of the callbackUrl.

In `packages/bruno-electron/src/ipc/network/authorize-user-in-window.js` -> `window.webContents.on('did-navigate', ...)` this causes the OAuth process to run `window.close()` to early. It directly closes, when opening authorizeUrl, because it thinks, authorizeUrl is already the callbackUrl.  
In fact, the user does not even see the window, because it directly closes.

As we see here, authorizeUrl is called, then directly the tokenUrl. Normally there should be another call in between, which is the 302 redirect.

image

This PR fixes this, by adding another check, that ensures, that the window does not directly close, in that situation.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
+++ b/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
@@ -1,16 +1,14 @@
 const { BrowserWindow } = require('electron');
 const { preferencesUtil } = require('../../store/preferences');
 
-const matchesCallbackUrl = (url, authorizeUrl, callbackUrl) => {
-  if (!url) {
-    return false;
-  }
-
-  if (url.href.startsWith(authorizeUrl.href)) {
-    return false;
-  }
-
-  return url.href.startsWith(callbackUrl.href);
+const matchesCallbackUrl = (url, callbackUrl) => {
+  const removeTrailingSlashRegex = /\/$/;
+  return url
+    && url.origin === callbackUrl.origin
+    && url.pathname.replace(removeTrailingSlashRegex, '') === callbackUrl.pathname.replace(removeTrailingSlashRegex, '')
+    && url.username === callbackUrl.username
+    && url.password === callbackUrl.password
+    && url.search.startsWith(callbackUrl.search);
 };
 
 const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session }) => {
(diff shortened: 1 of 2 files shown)


packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
```diff
@@ -1,8 +1,16 @@
...
+const matchesCallbackUrl = (url, authorizeUrl, callbackUrl) => {
+  if (!url) {
+    return false;
+  }
+
+  if (url.href.startsWith(authorizeUrl.href)) {
+    return false;
+  }
+
+  return url.href.startsWith(callbackUrl.href);
```
pietrygamat: If we are getting into specific scenarios, this breaks for:
Auth URL: https://apigateway.mycompany.dev/auth/realms/my-realm/protocol/openid-connect/auth
Redirect URL: https://apigateway.mycompany.dev/auth/realms/my-realm/protocol/openid-connect/auth-redirect

It does not seem necessary to include a third parameter in the method. How about just making the test a bit tighter - the normalized `URL.origin` and `URL.pathName` are expected to match exactly, only the `URL.search` part is allowed to match partially (OAuth params must be _appended_, so `startsWith` is ok):

```
const matchesCallbackUrl = (url, callbackUrl) => {
  return url 
           && url.origin === callbackUrl.origin
           && url.pathName === callbackUrl.pathName 
           && url.search.startsWith(callbackUrl.search);
}
```
To be extra pedantic, `url.username` and `url.password` might be checked as well, as they are not part of `.origin`.

pietrygamat (changes requested):

packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
kekru: Thanks for your feedback, I updated the code.

I also added a "remove trailing slash from url.pathname", so that these already given test case still works:

```
{ url: 'https://callback.url/endpoint/?code=abcd', expected: true },
```

pietrygamat: This looks good to me.

Issue #4388, opened 2025-04-01
OAuth 2.0 deviating from RFC when requesting token


swulf-ld:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying

### Bruno version

2.0.1

### Operating System

Manjaro 25.0.0 Zetar 6.12.20-2-MANJARO

### Describe the bug

According to the [RFC Section 4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3), when fetching the token via POST from the endpoint, only the following parameters are to be sent:

- grant_type
- code
- redirect_uri
- client_id

I'm trying to work with the DATEV API, and when sending an additional scope parameter with the token request, I receive a HTTP 400 with "invalid_request".

### Screenshots/Live demo link

Automatic from collection level, doesn't work:
![Image](https://github.com/user-attachments/assets/a6c3a0c4-e235-4165-ad5d-2c3806fdc6dd)

Manually, with same body params, doesn't work:
![Image](https://github.com/user-attachments/assets/32782bd9-e49e-4ae0-95ab-481cb778294e)

Manually, without scope, works:
![Image](https://github.com/user-attachments/assets/9b5d5c6b-9a5c-4206-b3d6-8160445d9903)

anusree-bruno:
Similar https://github.com/usebruno/bruno/issues/1790

anusree-bruno:
I was able to verify this. I will update here on the progress.

helloanoop:
We've dedicated time to look into this in our current sprint.

sreelakshmi-bruno:
Hi @swulf-ld , this issue is resolves in v2.10.0. I'm closing this ticket since it's resolved but please feel free to reopen if you find otherwise. Thanks!

PR #4919, opened 2025-06-16
Fix: removed scope from token request as per RFC standard

nonoratoDEV:
# Description

This relates to bug #4388 and removes the `scope` parameter from the token request. Currently these requests may be rejected by standard compliant identity providers. Significantly limiting the utility of the authorization feature in Bruno.

### DIFF
--- a/packages/bruno-electron/src/utils/oauth2.js
+++ b/packages/bruno-electron/src/utils/oauth2.js
@@ -58,7 +58,6 @@ const getOAuth2TokenUsingAuthorizationCode = async ({ request, collectionUid, fo
     clientId,
     clientSecret,
     callbackUrl,
-    scope,
     pkce,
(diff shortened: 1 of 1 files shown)


naman-bruno:
Hi @nonoratoDEV, thanks for your contribution!

Removing the scope parameter entirely from the oauth request could lead to issues, as it's an optional but sometimes required field depending on the provider. To avoid potential problems for other users, we’ve instead updated the logic to exclude the scope parameter only when it’s empty.

Closing this PR.

nonoratoDEV:
@naman-bruno After thinking about this a bit I don't think that will be a working solution. The scope parameter is optional for the /authorize call but shouldn't be required for the /token exchange call. For my IDP which is rfc compliant the /authorize call requires scope but the /token request will fail if it's included.

Issue #4991, opened 2025-06-26
Postman ignores empty "auth apikey", bruno fails "header name must be a non-empty string".


toehser:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.6.1 daily 6/25

### Operating System

Linux

### Describe the bug

Some of our postman collections have *both* auth section api key and it not set and *also* a header value {{api-key}} from the environment, in the .bru file it ends up like:

headers {
  x-fnma-access-token: {{x-FFFFFFFFF-access-token}}
  x-fnma-api-key: {{x-FFFFFFFF-api-key}}
  Content-Type: application/json
  ServiceName: AWS
}

auth:apikey {
  key: 
  value: 
  placement: header
}

coming out of a working postman collection.
Bruno fails with:
header name must be a non-empty string
It would be nice if Bruno could also not fail :)

### .bru file to reproduce the bug

headers {
  x-fnma-access-token: {{x-FFFFFFFFF-access-token}}
  x-fnma-api-key: {{x-FFFFFFFF-api-key}}
  Content-Type: application/json
  ServiceName: AWS
}

auth:apikey {
  key: 
  value: 
  placement: header
}

### Screenshots/Live demo link

It is obviously not what was intended by the author but it works in postman and fails when imported to bruno.

toehser:
I tested the pr code and it works - please merge into 2.7.0 :)

PR #5007, opened 2025-06-30
fix: Ignore empty header on Auth API Key(Header) to prevent sending request error

wibaek:
# Description

Empty header in Auth API Key is ignored to prevent sending request error. (https://github.com/usebruno/bruno/issues/4991)

Before:
image

After:
image

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/prepare-request.js
+++ b/packages/bruno-electron/src/ipc/network/prepare-request.js
@@ -61,10 +61,9 @@ const setAuthHeaders = (axiosRequest, request, collectionRoot) => {
         break;
       case 'apikey':
         const apiKeyAuth = get(collectionAuth, 'apikey');
+        if (apiKeyAuth.key.length === 0) break;
         if (apiKeyAuth.placement === 'header') {
-          if (apiKeyAuth.key.length > 0) {
-            axiosRequest.headers[apiKeyAuth.key] = apiKeyAuth.value;
-          }
+          axiosRequest.headers[apiKeyAuth.key] = apiKeyAuth.value;
         } else if (apiKeyAuth.placement === 'queryparams') {
           // If the API key authentication is set and its placement is 'queryparams', add it to the axios request object. This will be used in the configureRequest function to append the API key to the query parameters of the request URL.
           axiosRequest.apiKeyAuthValueForQueryParams = apiKeyAuth;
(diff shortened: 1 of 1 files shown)


toehser:
I have verified this locally and am hoping it gets merged into 2.7.0 :)

toehser:
Looks like it didn't make it into 2.7.0?  The release notes looked like it did but the pr looks like it didn't (along with testing :))

maintainer-bruno (approved):

sanish-bruno (approved):

packages/bruno-electron/src/ipc/network/prepare-request.js
```diff
@@ -248,7 +250,9 @@ const setAuthHeaders = (axiosRequest, request, collectionRoot) => {
       case 'apikey':
         const apiKeyAuth = get(request, 'auth.apikey');
         if (apiKeyAuth.placement === 'header') {
-          axiosRequest.headers[apiKeyAuth.key] = apiKeyAuth.value;
+          if (apiKeyAuth.key.length > 0) {
+            axiosRequest.headers[apiKeyAuth.key] = apiKeyAuth.value;
+          }
         } else if (apiKeyAuth.placement === 'queryparams') {
```
naman-bruno, line 254: Thanks for this fix! This is a good improvement to prevent empty API keys from being set as headers.
I noticed that the same validation logic should probably be applied to the query parameters case as well. Currently, when `apiKeyAuth.placement === 'queryparams'` we're still unconditionally assigning:
```
axiosRequest.apiKeyAuthValueForQueryParams = apiKeyAuth;
```
An empty API key in query parameters could potentially cause similar issues (like malformed URLs with empty parameter names). Would it make sense to add the same if `(apiKeyAuth.key.length > 0)` check before setting apiKeyAuthValueForQueryParams in both the collection-level and request-level auth sections?
This would ensure consistent behavior regardless of whether the API key is configured for headers or query parameters.
What do you think?

wibaek: Good point — yes, adding the same key.length > 0 check for query parameters makes sense. I’ll apply it to both collection and request-level logic for consistency. Thanks for catching that!

wibaek:
```
const apiKeyAuth = get(request, 'auth.apikey');
        if (apiKeyAuth.key.length === 0) break;
        if (apiKeyAuth.placement === 'header') {
          axiosRequest.headers[apiKeyAuth.key] = apiKeyAuth.value;
        } else if (apiKeyAuth.placement === 'queryparams') {
          // If the API key authentication is set and its placement is 'queryparams', add it to the axios request object. This will be used in the configureRequest function to append the API key to the query parameters of the request URL.
          axiosRequest.apiKeyAuthValueForQueryParams = apiKeyAuth;
        }
        break;
```

I’ve updated the code as above, so that empty API keys are now consistently excluded from both headers and query parameters.

Before:
image

After:
image

naman-bruno (approved):

maintainer-bruno (approved):

PR #5009, opened 2025-06-30
fix: handle undefined title in collection name and improve error handling

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-1218)


sanish-bruno:
@sanjaikumar-bruno can you add test cases for this?

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -364,8 +364,7 @@ export const parseOpenApiCollection = (data) => {
         return;
       }
 
-      // TODO what if info.title not defined?
-      brunoCollection.name = collectionData.info.title;
+      brunoCollection.name = collectionData.info.title || 'Untitled Collection';
```
sanish-bruno: can we use null coalescing operator? 
brunoCollection.name = collectionData.info.title ?? 'Untitled Collection';

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -445,8 +444,11 @@ export const openApiToBruno = (openApiSpecification) => {
     const validatedCollection = validateSchema(hydratedCollection);
     return validatedCollection
   } catch (err) {
-    console.error(err);
-    throw new Error('Import collection failed');
+    if (err instanceof Error) {
+      throw new Error(err.message);
+    } else {
+      throw new Error('Unknown error');
```
sanish-bruno, line 452: I think we can be more specific about this error

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -364,8 +364,7 @@ export const parseOpenApiCollection = (data) => {
         return;
       }
 
-      // TODO what if info.title not defined?
-      brunoCollection.name = collectionData.info.title;
+      brunoCollection.name = collectionData.info.title ?? 'Untitled Collection';
```
maintainer-bruno: Any reason to use Null coalescing operator here? 
I think title must be a string and we need not accept any other falsey values. So logical OR operator makes more sense here.

packages/bruno-converters/tests/openapi/openapi-to-bruno/openapi-to-bruno.spec.js
```diff
@@ -7,6 +7,47 @@ describe('openapi-collection', () => {
 
     expect(brunoCollection).toMatchObject(expectedOutput);
   });
+
+  it('should use info.title if present', () => {
+    const openApiWithTitle = `
+openapi: "3.0.0"
+info:
+  version: "1.0.0"
```
maintainer-bruno: Please use single quotes for string values in js files

packages/bruno-converters/src/openapi/openapi-to-bruno.js
sanish-bruno: @maintainer-bruno Initially we were using || operator. then moved on to null coalescing cause of the case where an `empty string` is given as title,  Was bit confused on whether we should change the title to "Untitiled Collection" or keep the empty string as it is.

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -445,8 +447,12 @@ export const openApiToBruno = (openApiSpecification) => {
     const validatedCollection = validateSchema(hydratedCollection);
     return validatedCollection
   } catch (err) {
+    if (!(err instanceof Error)) {
+      console.error(err);
```
sanish-bruno: @sanjaikumar-bruno we don't have to have 2 consoles within the catch block, we should ideally have one

packages/bruno-converters/tests/openapi/openapi-to-bruno/openapi-to-bruno.spec.js
```diff
@@ -7,6 +7,87 @@ describe('openapi-collection', () => {
...
+        '200':
+          description: 'OK'
+servers:
+  - url: 'https://example.com'
+`;
+    const result = openApiToBruno(openApiWithTitle);
+    expect(result.name).toBe('My API');
+  });
+
+  it('defaults to Untitled Collection if info.title is an empty string or only whitespace', () => {
```
sanish-bruno: @sanjaikumar-bruno we probably need to split this into two test cases where we consider an empty string and another for testing whitespace. tests have to be as dumb as possible. empty string and and string with white space can behave in different ways depending on how the code is being handled. we shouldn't be writing tests to match the implementation, rather it should be opposite.

maintainer-bruno (approved):

naman-bruno (approved):

PR #5018, opened 2025-07-01
feat: introduce `res.getSize()` helper (header/body/total)

pooja-bruno:
# Description

Adds a new helper to the scripting runtime that mirrors Postman’s `pm.response.size()`

[JIRA](https://usebruno.atlassian.net/browse/BRU-1032)

### Contribution Checklist:

https://github.com/user-attachments/assets/3d71ec45-56e8-47c7-9062-5c607cb10075

### DIFF
--- a/packages/bruno-js/src/bruno-response.js
+++ b/packages/bruno-js/src/bruno-response.js
@@ -89,6 +89,11 @@ class BrunoResponse {
     const headerSize = Buffer.byteLength(headerLines.join('\r\n'));
 
     return { header: headerSize, body: bodySize, total: headerSize + bodySize };
+    
+  }
+
+  getDataBuffer() {
+    return this.res ? this.res.dataBuffer : null;
   }
 }
 
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -1191,6 +1191,7 @@ const registerNetworkIpc = (mainWindow) => {
                 error.response.responseTime = error.response.headers.get('request-duration');
                 error.response.headers.delete('request-duration');
                 error.response.data = data;
+                error.response.dataBuffer = dataBuffer;
 
                 timeEnd = Date.now();
                 response = {


packages/bruno-tests/collection/scripting/api/res/getSize.bru
```diff
@@ -0,0 +1,27 @@
...
+  console.log(res.getSize())
+}
+
+tests {
+  test("test body size", function() {
+    const bodySize = res.getSize().body;
+    expect(bodySize === 1024).to.be.true;
+  });
+  
+  test("test header size", function() {
```
sanish-bruno, line 27: @pooja-bruno Any reason why we didn't add test case for total similar to body and header?

```js
test("test total size", function() {
  const bodySize = res.getSize().total;
  console.log(bodySize)
  expect(bodySize === 1263).to.be.true;
});
```

packages/bruno-js/src/bruno-response.js
```diff
@@ -51,6 +51,36 @@ class BrunoResponse {
...
+        Array.isArray(value)
+          ? value.map((v) => `${key}: ${v}`)
+          : [`${key}: ${value}`]
+      ),
+      '',
+      ''
+    ];
+    const headerSize = Buffer.byteLength(headerLines.join('\r\n'));
+
+    return { header: headerSize, body: bodySize, total: headerSize + bodySize };
```
sanish-bruno, line 91: I guess we were having `headers` instead of `header` as per the Jira, plural although postman just 'header', can you confirm once with @anusree-bruno @sreelakshmi-bruno

packages/bruno-tests/collection/scripting/api/res/getSize.bru
sanish-bruno: @pooja-bruno the tests are success while we run them individually, both in safe mode and developer mode, but when i run the tests through the runner is seems to be failing, especially cause headers seems to have a different value, can you look into the issue on possibly, what might be causing it, or considering an alternate way of testing it? 

probably it inherits some extra headers?

{header: 262, body: 1024, total: 1286}

packages/bruno-js/src/bruno-response.js
pooja-bruno: Confirmed with @sreelakshmi-bruno — using `header` is also fine. So we’re good to proceed with `header`.

packages/bruno-converters/tests/postman/postman-to-bruno/postman-translations/postman-response.spec.js
```diff
@@ -9,6 +9,11 @@ describe('postmanTranslations - response commands', () => {
       const responseJson = pm.response.json();
       const responseStatus = pm.response.status;
       const responseHeaders = pm.response.headers;
+      const responseSize = pm.response.size();
```
sanish-bruno, line 12: @pooja-bruno Just FYI, probably can consider adding transpiler test cases for different scenarios, This can also be a separate PR as this is not the core thing we are addressing here, not a mandatory thing

Example [gist](https://gist.github.com/sanish-bruno/19a154c95e28d558f2cd645d03227919)

packages/bruno-js/src/bruno-response.js
```diff
@@ -51,6 +51,36 @@ class BrunoResponse {
     this.res.data = clonedData;
     this.body = clonedData;
   }
+
+  getSize() {
+    if (!this.res) return { header: 0, body: 0, total: 0 };
+
+    let bodySize = 0;
+    // Prefer raw dataBuffer if available
+    if (Buffer.isBuffer(this.res.dataBuffer)) {
```
Pragadesh-45: here, we are not handling the case where `this.res.data` is `null` or `undefined`. we should handle that explicitly.
* consider adding a check for `null` or `undefined` values, and set `bodySize` to 0 or return early if that's the case.

packages/bruno-tests/collection/scripting/api/res/getSize.bru
```diff
@@ -0,0 +1,32 @@
...
```
Pragadesh-45: @pooja-bruno, can we use our own endpoint for testing, which is https://www.httpfaker.org/api/random/json?size=1mb?

packages/bruno-js/src/bruno-response.js
```diff
@@ -55,16 +55,14 @@ class BrunoResponse {
   getSize() {
     if (!this.res) return { header: 0, body: 0, total: 0 };
```
sanish-bruno: @pooja-bruno we can enclose the return in curly braces rather than it being inline

sanish-bruno (changes requested):

packages/bruno-js/src/bruno-response.js
```diff
@@ -51,6 +51,36 @@ class BrunoResponse {
...
+        Array.isArray(value)
+          ? value.map((v) => `${key}: ${v}`)
+          : [`${key}: ${value}`]
+      ),
+      '',
+      ''
+    ];
+    const headerSize = Buffer.byteLength(headerLines.join('\r\n'));
+
+    return { header: headerSize, body: bodySize, total: headerSize + bodySize };
```
Pragadesh-45: In some servers, there will be an actual transmitted size, which will be the response size. So, we can use it and fall back to `JSON.stringify` as a last resort, since it is memory expensive.

```js
  getSize() {
    if (!this.res) {
      return { header: 0, body: 0, total: 0 };
    }

    const { data, dataBuffer, headers } = this.res;
    let bodySize = 0;

    // Use raw received bytes
    if (Buffer.isBuffer(dataBuffer)) {
      bodySize = dataBuffer.length;
    } else {
      // Use server-reported Content-Length
      const contentLength = headers && (headers['content-length'] || headers['Content-Length']);
      if (contentLength && !isNaN(contentLength)) {
        bodySize = parseInt(contentLength, 10);
      } else if (data != null) {
        // Manual calculation
        const raw = typeof data === 'string' ? data : JSON.stringify(data);
        bodySize = Buffer.byteLength(raw);
      }
    }

    const headerLines = [
      `HTTP/1.1 ${this.res.status} ${this.res.statusText}`,
      ...Object.entries(this.res.headers || {}).flatMap(([key, value]) =>
        Array.isArray(value)
          ? value.map((v) => `${key}: ${v}`)
          : [`${key}: ${value}`]
      ),
      '',
      ''
    ];
    const headerSize = Buffer.byteLength(headerLines.join('\r\n'));

    return { header: headerSize, body: bodySize, total: headerSize + bodySize };
  }
```

packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/response.test.js
```diff
@@ -553,4 +553,35 @@ describe('Response Translation', () => {
         expect(res.getBody()).to.equal({"status": "ok"});
         `);
     });
+
+    // --- getSize translations ---------------------------
+    it('should translate pm.response.size()', () => {
+        const code = 'const sz = pm.response.size();';
+        const translatedCode = translateCode(code);
+        expect(translatedCode).toBe('const sz = res.getSize();');
+    });
```
Pragadesh-45: can we change `sz` to `size` as the variable name since and `respSz` to `responseSize`

Pragadesh44-Bruno (dismissed):

maintainer-bruno (approved):

naman-bruno (approved):

naman-bruno (approved):

PR #5048, opened 2025-07-06
Support for Odata style path params

tester22:
# Description

THe odata request format adds path parameters in url that isn't added directly after the slash: https://www.odata.org/documentation/odata-version-2-0/uri-conventions/

This PR adds support for this by using a regex to parse path params from the url.
image

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/utils/url/index.js
+++ b/packages/bruno-app/src/utils/url/index.js
@@ -49,19 +49,24 @@ export const parsePathParams = (url) => {
     paths = uri.split('/');
   }
 
-  paths = paths.reduce((acc, path) => {
-    if (path !== '' && path[0] === ':') {
-      let name = path.slice(1, path.length);
-      if (name) {
-        let isExist = find(acc, (path) => path.name === name);
-        if (!isExist) {
-          acc.push({ name: path.slice(1, path.length), value: '' });
+  // Enhanced: also match :param inside parentheses and/or quotes
+  const paramRegex = /[:](\w+)/g;
+  const foundParams = new Set();
+  paths.forEach((segment) => {
+    let match;
+    while ((match = paramRegex.exec(segment))) {
+      if (match[1]) {
+        // Clean up: remove trailing quotes/parentheses if present
+        let name = match[1].replace(/[')"`]+$/, '');
+        // Remove leading quotes/parentheses if present
+        name = name.replace(/^[('"`]+/, '');
+        if (name && !foundParams.has(name)) {
+          foundParams.add(name);
         }
       }
     }
-    return acc;
-  }, []);
-  return paths;
+  });
+  return Array.from(foundParams).map((name) => ({ name, value: '' }));
 };
 
 export const stringifyQueryParams = (params) => {
@@ -128,10 +133,20 @@ export const interpolateUrlPathParams = (url, params) => {
     return pathname
       .split('/')
       .map((segment) => {
-        if (segment.startsWith(':')) {
-          const pathParamName = segment.slice(1);
-          const pathParam = params.find((p) => p?.name === pathParamName && p?.type === 'path');
-          return pathParam ? pathParam.value : segment;
+        // Enhanced: also match :param inside parentheses and/or quotes
+        let match;
+        const regex = /[:](\w+)/g;
+        while ((match = regex.exec(segment))) {
+          if (match[1]) {
+            // Clean up: remove trailing quotes/parentheses if present
+            let name = match[1].replace(/[')"`]+$/, '');
+            // Remove leading quotes/parentheses if present
+            name = name.replace(/^[('"`]+/, '');
+            if (name) {
+              const pathParam = params.find((p) => p?.name === name && p?.type === 'path');
+              return pathParam ? pathParam.value : segment;
+            }
+          }
         }
         return segment;
       })
--- a/packages/bruno-app/src/utils/url/index.spec.js
+++ b/packages/bruno-app/src/utils/url/index.spec.js
@@ -111,6 +111,12 @@ describe('Url Utils - parsePathParams', () => {
       { name: 'postId', value: '' }
     ]);
   });
+
+  it('should parse path param inside parentheses and quotes', () => {
+    const params = parsePathParams("https://example.com/ExchangeRates(':ExchangeRateOID')");
+    expect(params).toEqual([{ name: 'ExchangeRateOID', value: '' }]);
+  });
+
 });
 
 describe('Url Utils - splitOnFirst', () => {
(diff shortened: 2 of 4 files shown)


tester22:
@helloanoop any news on this?
This causing problems for us since we need to add variables directly to the url instead of using parameters.

helloanoop:
Hey @tester22 

At first glance, supporting OData-style path parameters is definitely something we want to do. I’ve set aside time in the first half of October for the team to review this and work towards getting it merged.

packages/bruno-electron/src/ipc/network/interpolate-vars.js
```diff
@@ -133,12 +133,17 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
...
-          return '/' + path;
+        const paramRegex = /[:](\w+)/g;
+        const matches = path.match(paramRegex);
+        if (matches) {
+          const paramName = matches[0].slice(1); // Remove the : prefix
+          const existingPathParam = request.pathParams.find(param => param.name === paramName);
+          if (!existingPathParam) {
+            return '/' + path;
+          }
+          return '/' + path.replace(':' + paramName, existingPathParam.value);
```
sid-bruno: Can you also add tests for the same in `packages/bruno-electron/tests/network/interpolate-vars.spec.js` , I realise you've added some at the top but would be nice to have some specific to the package as well

packages/bruno-app/src/utils/url/index.js
```diff
@@ -128,10 +133,20 @@ export const interpolateUrlPathParams = (url, params) => {
...
+          if (match[1]) {
+            // Clean up: remove trailing quotes/parentheses if present
+            let name = match[1].replace(/[')"`]+$/, '');
+            // Remove leading quotes/parentheses if present
+            name = name.replace(/^[('"`]+/, '');
+            if (name) {
+              const pathParam = params.find((p) => p?.name === name && p?.type === 'path');
+              return pathParam ? pathParam.value : segment;
+            }
+          }
```
sid-bruno, line 105: A few things 
- `const regex = ...` can be moved up to avoid getting defined for every segment 
-  the interpreter executes the first `exec` which is costlier than `startsWith` and `regex.test` , can you change that so the regex is tested and the segment is returned immediately

Bench for point 2
https://jsbm.dev/9OSZgAL6ppIOk

packages/bruno-app/src/utils/url/index.spec.js
```diff
@@ -111,6 +111,12 @@ describe('Url Utils - parsePathParams', () => {
       { name: 'postId', value: '' }
     ]);
   });
+
+  it('should parse path param inside parentheses and quotes', () => {
+    const params = parsePathParams("https://example.com/ExchangeRates(':ExchangeRateOID')");
+    expect(params).toEqual([{ name: 'ExchangeRateOID', value: '' }]);
+  });
+
```
sid-bruno, line 107: Can we add a few more just to cover a few more cases 

```suggestion
  it('should parse path param inside parentheses and quotes', () => {
    const params = parsePathParams("https://example.com/ExchangeRates(':ExchangeRateOID')");
    expect(params).toEqual([{ name: 'ExchangeRateOID', value: '' }]);
  });
  
  it('should parse path param inside parentheses and no quotes', () => {
    const params = parsePathParams("https://example.com/ExchangeRates(:ExchangeRateOID)");
    expect(params).toEqual([{ name: 'ExchangeRateOID', value: '' }]);
  });

  it('should parse multiple path params inside parentheses', () => {
    const params = parsePathParams("https://example.com/Exchange(:ExchangeId)/ExchangeRates(:ExchangeRateOID)");
    expect(params).toEqual([{ name: 'ExchangeId', value: '' },{ name: 'ExchangeRateOID', value: '' }]);
  });

  it('should parse mix and match of normal and param inside parentheses', () => {
    const params = parsePathParams("https://example.com/Exchange(:ExchangeId)/:key");
    expect(params).toEqual([{ name: 'ExchangeId', value: '' },{ name: 'key', value: '' }]);
  });
```

packages/bruno-cli/src/runner/interpolate-vars.js
```diff
@@ -119,12 +119,17 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
       .split('/')
       .filter((path) => path !== '')
       .map((path) => {
-        if (path[0] !== ':') {
-          return '/' + path;
+        const paramRegex = /[:](\w+)/g;
```
sid-bruno: Regex can be moved up instead of being defined in each loop for each path

packages/bruno-electron/src/ipc/network/interpolate-vars.js
sid-bruno: Also same as the other comments, regex can be moved up in the logic

sid-bruno:
Also, can update the PR with the latest code from `main` ?

there were tests added to check a few cases, it'd help us if those tests were also run as a part of the merge instead of being run post merge

tester22:
@sid-bruno I think all your comments are resolved now.

sid-bruno:
@tester22, thanks! 

This seems to have gotten missed https://github.com/usebruno/bruno/pull/5048#discussion_r2362104246, the interpolation has it's own edge cases so if we have the tests there as well with values being passed and tested for replacement, that would add to the confidence of the implementation for later

tester22:
> @tester22, thanks!
> 
> This seems to have gotten missed [#5048 (comment)](https://github.com/usebruno/bruno/pull/5048#discussion_r2362104246), the interpolation has it's own edge cases so if we have the tests there as well with values being passed and tested for replacement, that would add to the confidence of the implementation for later

Added now

sid-bruno:
> > @tester22, thanks!
> > This seems to have gotten missed [#5048 (comment)](https://github.com/usebruno/bruno/pull/5048#discussion_r2362104246), the interpolation has it's own edge cases so if we have the tests there as well with values being passed and tested for replacement, that would add to the confidence of the implementation for later
> 
> Added now

Thanks for additional tests but I meant adding them to the `network/interpolate-vars.spec.js` , the one's you've added are helpful for extracting the path params but there aren't any added to test them being replaced in the final url which are a part of `packages/bruno-electron/tests/network/interpolate-vars.spec.js` , let me know if you want me to help write tests for that

tester22:
Ok. It might be faster if you make the changes.

sid-bruno:
> Ok. It might be faster if you make the changes.

Cool, if you see a `allow edits from maintainers` checkbox somewhere on the PR, please do enable it

bijin-bruno:
Merged to feature branch for fine tuning the PR changes
@sid-bruno @tester22

Issue #5078, opened 2025-07-11
[Feature Request] Include request URLs in search matching


Sergey27-dev:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] blocks me from using Bruno
- [x] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### Describe the feature you want to add, and how it would change your usage of Bruno

Hi,

I'm in the process of moving from Postman to Bruno, and I noticed that currently there's no way to search requests by URL.

In Postman, I often rely on searching by parts of the request URL to quickly find what I need, especially in large collections. I’d like to request this feature in Bruno — to make it possible for the search function to also match against the request URL, not just the name.

Thanks for the great tool!

BasharSamara1:
This is an important feature, but there is already a workaround for it — pressing `**Ctrl+K**` opens a search that matches both request names and URLs. It might be worth documenting this more visibly for users who are not aware of it.

PR #5079, opened 2025-07-11
feat: expand search to include request URL

Sergey27-dev:
# Description

This PR extends the request search functionality by including the request URL in the search criteria. Previously, only the request name was used to match against the search text. With this change, users will be able to find requests by matching either the name or the URL, which makes it easier to locate requests in large collections, especially when request names are not descriptive.

### Contribution Checklist:

- [x] **Create an issue and link to the pull request.  
#5078**

<details>
<summary>
Screenshots
</summary>
image

image
</details>

### DIFF
--- a/packages/bruno-app/src/utils/collections/search.js
+++ b/packages/bruno-app/src/utils/collections/search.js
@@ -3,7 +3,12 @@ import filter from 'lodash/filter';
 import find from 'lodash/find';
 
 export const doesRequestMatchSearchText = (request, searchText = '') => {
-  return request?.name?.toLowerCase().includes(searchText.toLowerCase());
+  const searchItems = [
+    request?.name ?? '',
+    request?.request?.url ?? '',
+  ];
+
+  return searchItems.join('').toLowerCase().includes(searchText.toLowerCase());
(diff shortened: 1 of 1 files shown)


helloanoop:
Hey @Sergey27-dev 

Thanks for taking the time to work on this.

One challenge with the current approach is that when multiple requests share the same URL, the result list becomes cluttered with matches. The existing UX doesn't clearly indicate whether a match was due to the request name or the request URL, which can be confusing.

We're planning a broader overhaul of the search experience to make it more aligned with how VS Code handles search.

In the meantime, we're also working on [#4983](https://github.com/usebruno/bruno/pull/4983) to introduce a quick search (accessible via cmd+k). Could you take a look and share your feedback?

Sergey27-dev:
Hi @helloanoop, thanks for the feedback!

Got it - makes sense to hold off given the upcoming search refactor. The quick search functionality in #4983 does look like a more comprehensive solution, especially with the cmd+k shortcut. Happy to close this PR to avoid duplicate work.

Let me know if I can help with the new search implementation - would be glad to contribute!

PR #5147, opened 2025-07-18
Expose bruno-runner as runtime library

anuragagarwal561994:
# Description

Currently bruno can only be used as cli, when running in CI / CD this means we will have to start the node process, parse the grammer and load everytime which can be wasteful.

In our setup basically we have written code over the existing newman library where we get collections to run from a queue and run it over the entire stack. This is possible because we are using newman as a library, but bruno doesn't provide such functionality and we have to start a new node process everytime. 

The idea here is to separate the code for running bruno collections from cli and allow the users to use both as per their desire or even to write custom logic if they intend to.

The API is more or less like `newman` so folks have familiarity, there are 2 APIs:
* a method which returns result as promise when the collection run is complete
* an event emitter where one can listen events from bruno runner when a collection is started, done or when an item is being processed

We tried this in our CI / CD pipeline and it surely does perform better than starting new node process.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-cli/LIBRARY_USAGE.md
+++ b/packages/bruno-cli/LIBRARY_USAGE.md
@@ -1,0 +1,446 @@
+# Bruno Runner Library
+
+The Bruno Runner has been refactored to provide a clean library interface that can be called externally by other applications multiple times. This allows you to programmatically execute Bruno collections from your own Node.js applications.
+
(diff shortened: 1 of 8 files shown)


altaiezior:
@helloanoop @lohit-bruno can we check this change?

Issue #2160, opened 2024-04-24
Environment variables are not saved to the disk when using bru.setProcessEnv()


gcharondkt:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

1. Create an environment
2. Run the request
3. Open the variable : the "key" is there
4. Quit Bruno and reopen it
5. There is no "key" anymore

### .bru file to reproduce the bug

meta {
  name: test
  type: http
  seq: 2
}

get {
  url: http://google.fr
  body: none
  auth: none
}

tests {
  bru.setEnvVar("key", "value");
}

### Screenshots/Live demo link

N/A

busy-panda:
It should be "bru.setEnvVar" in the title (not bru.setProcessEnv) I guess.

Indeed, it does not persist the change in environment variables, I guess it should.

Daniel-Byrne:
Is there any update on this? I have the same issue and it renders Bruno practically useless for my needs. 

I am using `v1.19.0` on Mac Sonoma 14.5.

helloanoop:
I am not sure if its a good idea to have `bru.setEnvVar("secret", token)` to be saved to the env bru file.
A common use case of people updating the env vars involves fetching a token and saving it to the env var.

When it comes to managing tokens, we recommend to use `bru.setVar()`

@Daniel-Byrne @busy-panda Can you share some use case examples where we need to have the `bru.setEnvVar()` write changes to the file?

We can consider an option to pass to like - `bru.setEnvVar(name, value, {psersist: true})`

alesbolka:
Here's my use case: 
I maintain the account system for our company and because certain things happen only on account creation, I made a collection to simplify account creation. The "automated account creation" request has a pre-request script that generates a unique suffix for an email and upon success, sets the email as an environment variable as the last generated account. That email can then be used by other requests in the collection to check things. As sometimes development of new features can take a longer time, it's nice being able to go "this was the last account I created" and persist that, as you continue working on whatever comes after account creation. 
We use this a lot as we have separate teams working on separate features and each team would like to maintain the collections used for their own systems.

I'd be happy with a way to manually call something like "bru.saveEnv" from within the pre/post request scripts to store the data. 

The collection variables don't even update visually when setting them from within a script, let alone persist (maybe the latter part is intentional), but that's a separate issue.

busy-panda:
[...]

PR #5172, opened 2025-07-23
feat: add persistent environment variable handling in IPC events and Bru class

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-772)

Fixes: https://github.com/usebruno/bruno/issues/2160


packages/bruno-js/src/bru.js
```diff
@@ -74,12 +75,20 @@ class Bru {
...
+  setEnvVar(key, value, options = {}) {
     if (!key) {
       throw new Error('Creating a env variable without specifying a name is not allowed.');
     }
 
     this.envVariables[key] = value;
+
+    if (options.persist) {
+      this.persistentEnvVariables[key] = value
+    } else {
```
sanish-bruno, line 137: @sanjaikumar-bruno why are we deleting the variable if persist is not provided?

packages/bruno-app/src/providers/App/useIpcEvents.js
```diff
@@ -112,6 +113,42 @@ const useIpcEvents = () => {
...
 
+    const removePersistentEnvVariablesUpdateListener = ipcRenderer.on('main:persistent-env-variables-update', (val) => {
+      const { persistentEnvVariables, collectionUid, collection } = val;
+      if (persistentEnvVariables && collection) {
+        const environmentUid = collection.activeEnvironmentUid;
+        if (environmentUid) {
+          const environment = collection.environments?.find(env => env.uid === environmentUid);
+          let existingVars = Array.isArray(environment?.variables) ? environment.variables : [];
+
+          let newVars = persistentEnvVariables;
```
sanish-bruno: it will keep the object reference, why are we having another variable here?

packages/bruno-app/src/providers/App/useIpcEvents.js
```diff
@@ -112,6 +113,42 @@ const useIpcEvents = () => {
...
+              name,
+              value,
+              type: 'text',
+              enabled: true,
+              secret: false
+            }));
+          }
+
+          const mergedVars = [...existingVars];
+          newVars.forEach(newVar => {
```
sanish-bruno: currently we can have duplicate values within environment, why are we limitting that  when  it comes to scripts,

packages/bruno-js/src/runtime/script-runtime.js
```diff
@@ -122,6 +122,7 @@ class ScriptRuntime {
         request,
         envVariables: cleanJson(envVariables),
         runtimeVariables: cleanJson(runtimeVariables),
+        persistentEnvVariables: bru.persistentEnvVariables,
```
sanish-bruno, line 125: @sanjaikumar-bruno don't we need to provide the same for test runtime?

sanish-bruno (changes requested):

packages/bruno-app/src/providers/App/useIpcEvents.js
sanish-bruno: other tools currently replaces the value, if there is any env with same name, rather than creating a duplicate entry. can you research how environment setter functions work in other tools?

sanish-bruno (changes requested):

packages/bruno-js/src/bru.js
```diff
@@ -74,12 +76,20 @@ class Bru {
...
+      if (this.persistentEnvVariables[key]) {
+        delete this.persistentEnvVariables[key];
+      }
+    }
```
Pragadesh-45: Do we need to add `options?.persist` incase of persist don't exist

packages/bruno-js/tests/runtime.spec.js
```diff
@@ -175,4 +177,68 @@ describe('runtime', () => {
       });
     });
   });
+
+  describe('persistentEnvVariables integration', () => {
```
sanish-bruno, line 181: @sanjaikumar-bruno can we also add a test cases for CLI run as a bru file, which is checking the persistence between requests

sanish-bruno (approved):

packages/bruno-js/tests/runtime.spec.js
sanish-bruno: @sanjaikumar-bruno can we also add a test cases for CLI run as a bru file, which is checking the persistence between requests

sanish-bruno (changes requested):

sanish-bruno:
@sanjaikumar-bruno add intellisense for codemirror

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -467,6 +467,12 @@ const registerNetworkIpc = (mainWindow) => {
         collectionUid
       });
 
+      mainWindow.webContents.send('main:persistent-env-variables-update', {
+        persistentEnvVariables: scriptResult.persistentEnvVariables,
+        collectionUid,
+        collection
```
sanish-bruno: @sanjaikumar-bruno we can remove collection from prams being sent from all the instances of `.send('main:persistent-env-variables-update'` , as we don't intend to use it eventually, Even if there is a need, we can always get the collection from collectionUid

packages/bruno-js/tests/runtime.spec.js
```diff
@@ -175,4 +177,153 @@ describe('runtime', () => {
...
```
sanish-bruno: @sanjaikumar-bruno the test suite is incomplete as we are not actually looking for persistence, we are just looking for it's existence within persistVariables. The. same variables were also present envVariables, even in the previous logic were didn't have any logic related to persistence, if we check for the variable within envVariables, it will always succeed, here the code just checks another object for respective values, Practically both are same,  the existence of persistentVariables is just aid in the process of persisting, tests shouldn't be caring about the it all, infact, tests should be agnostic of the implementation details. It is better to add playwright test cases for this feature. Only playwright testcase can deterministically fail, if there are any failure.

sanish-bruno: We can have this test suite to check, when persist is true, are we returning error for values other than string, are string values going through the function properly? are these variables are available within envVariables or not etc. we shouldn't be checking their presence within persistent variables

sanish-bruno (changes requested):

packages/bruno-app/src/utils/codemirror/autocomplete.js
```diff
@@ -60,7 +60,9 @@ const STATIC_API_HINTS = {
     'bru.getEnvVar(key)',
     'bru.getFolderVar(key)',
     'bru.getCollectionVar(key)',
-    'bru.setEnvVar(key,value)',
+    'bru.setEnvVar(key, value, options)',
+    'bru.setEnvVar(key, value, { persist: true })',
+    'bru.setEnvVar(key, value, { persist: false })',
```
bijin-bruno: I think we don't have to provide all the options available in autocomplete. Just the below would be enough
```
'bru.setEnvVar(key, value)'
'bru.setEnvVar(key, value, options)'
```

naman-bruno (approved):

naman-bruno (approved):

PR #5178, opened 2025-07-23
feat: Implement key escaping for Bruno query parameters

### LINKED ISSUE #2810: Params Query Bug

### LINKED ISSUE #2878: space in any param name is blocking the loading of the request file in GUI

### LINKED ISSUE #3037: query parameters containing a colon (:) are not parsed correctly when changed

Pragadesh-45:
# Description

- Replaced quoting with escaping for query parameter keys containing special characters: ':'\, '\"', '{', '}' or space.
- Refactored bruToJson and jsonToBru to consistently escape keys with special characters.
- Introduced escapeKey function to handle key escaping.
- .bru files now support storing keys with escaped characters, including the double-quote (e.g., \\\"`).
- Updated test fixtures to reflect the new escaping logic.

Fixes #3037
Fixes #2810
Fixes #2878

### Contribution Checklist:

### DIFF
--- a/packages/bruno-lang/v2/src/bruToJson.js
+++ b/packages/bruno-lang/v2/src/bruToJson.js
@@ -40,7 +40,9 @@ const grammar = ohm.grammar(`Bru {
   stnl = st | nl
   tagend = nl "}"
   optionalnl = ~tagend nl
-  keychar = ~(tagend | st | nl | ":") any
+  keychar = esc_keychar | unesc_keychar
+  esc_keychar = esc_char (":" | "\\x22" | "{" | "}" | " " | esc_char)
+  unesc_keychar = ~(tagend | st | nl | ":" | "\\x22" | "{" | "}" | " " | esc_char) any
   valuechar = ~(nl | tagend) any
 
    // Multiline text block surrounded by '''
@@ -51,8 +53,10 @@ const grammar = ohm.grammar(`Bru {
   dictionary = st* "{" pairlist? tagend
   pairlist = optionalnl* pair (~tagend stnl* pair)* (~tagend space)*
   pair = st* key st* ":" st* value st*
-  key = keychar*
-  value = list | multilinetextblock | valuechar*
+  disable_char = "~"
+  esc_char = "\\\\"
+  key = keychar+
+  value = multilinetextblock | valuechar*
 
   // Dictionary for Assert Block
   assertdictionary = st* "{" assertpairlist? tagend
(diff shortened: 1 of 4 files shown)


Pragadesh-45:
Hey @pietrygamat, Could you please review this PR?

pietrygamat:
@Pragadesh-45 has the team decided that the problem should be solved by means of escaping vs quoting, or is this PR just a PoC? I don't immediately see any code issues in this PR to report, but I have an opinion on the usability side of things  

First, the string full of escape sequences is worse to read, copy and paste. The bru format is IDE and git friendly, so quick copy/paste directly in the bru file instead of doing so through Bruno UI seems a legitimate workflow, now negatively affected - as user must know which characters are special and need escaping (and the list is not obvious - why `"` but not `'`, why `{}` but not `[]`, why space but not URI-special characters: `=%?&`, etc). And it gets extra confusing, becasue escaping is only for keys, but not for the same characters in values. 

With quoting we only need a very short list of escape characters: the quote itself and the escape character - it is self evident why they need escaping, and the rest can be provided as is.


And secondly: this change breaks previously working collections. It is one thing, something like:
```
my param: my value
```
that never worked before - so users will gladly manualy update their bru files to have it fixed. But now users with file:
```
{param}: {value}
```
cannot open their previously correct collection - with no word of warning why (except deep in log). Maybe some sort of schema upgrade is required as part of this PR?

Pragadesh-45:
@pietrygamat, thanks for looking at the PR. Yes, we haven't decided on a single approach, and this PR is a PoC. Personally, I thought the escaping logic was good and wanted you to have a look. I understand the confusion regarding escaping keys alone. Thanks for the perspective; I'll have a discussion with the team and let you know.

stake25:
I was watching this PR and was excited to see where this was going as this was an issue my teams at work are having. Is this going to be changed in another issue or handled some other way?

Pragadesh-45:
@stake25 Yes, this PR was closed because we finalized the approach of quoting instead of escaping logic. The new solution is currently under review here: https://github.com/usebruno/bruno/pull/5332

stake25:
@Pragadesh-45 Thankyou will watch this instead now. Look forward to these changes being made

Issue #5181, opened 2025-07-24
Folder Expansion Required to Display Collections


gopi898:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [x] annoying
- [x] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.8.0

### Operating System

Windows 11

### Describe the bug

We have recently started using Bruno for API collection management and execution. We've encountered a consistent issue where the application initially reports "0 collections" for a folder, even when collections are present within it. Crucially, we've observed that if the folder is not expanded, attempts to execute the collection ("Run Collection" feature) fail or don't recognize any APIs to execute. The collections only become visible and available for execution after the containing folder is explicitly expanded in the Bruno UI

### .bru file to reproduce the bug

1. Open Bruno.
2. Navigate to a workspace containing folders with collections nested inside.
3. Observe a folder that is currently collapsed and contains collections.
4. Notice that the collection count for this collapsed folder (if displayed) or the general collection view might indicate 0 collections within it.
5. Attempt to execute the collection within this collapsed folder (e.g., right-click on the folder and select "Run").
6. Observe that the execution fails, reports no collections, or does not recognize the APIs within the folder for execution.
7. Now, click on the collapsed folder to expand it.
8. Observe that the nested collections are now correctly displayed and accessible for execution.
9. Attempt to execute the collection again from the now expanded folder. Observe that it now runs as expected.

**_Expected Behavior_**
When a folder contains collections, Bruno should accurately display the presence of these collections (e.g., by showing a non-zero count or an indicator that collections exist) even when the folder is in a collapsed state. The collections should be recognized and accessible for operations like collection execution regardless of the folder's expanded/collapsed state in the UI.

**_Actual Behavio_r**
Collections within a folder are only recognized and displayed after the folder has been manually expanded. Until then, they appear as if they do not exist, leading to confusion and hindering efficient workflow, especially with large numbers of collections or deeply nested structures.

### Screenshots/Live demo link

https://github.com/user-attachments/assets/af0db902-9c5e-4942-bfe9-624a617c5453

anusree-bruno:
similar https://github.com/usebruno/bruno/issues/4945

PR #5138, opened 2025-07-18
Fix: Loading state while collection mount

naman-bruno:
### Description

This PR addresses two issues:

- Empty export files and runners without requests when accessed via the context menu without mounting the collection.
- Consistent loading behavior during collection mounting.

**Screenshots (when collection is loading)**

Screenshot 2025-07-18 at 1 23 18 PM

Screenshot 2025-07-18 at 1 32 59 PM

### Contribution Checklist:

- [ ] **The pull request only addresses one issue or adds one feature.**
- [ ] **The pull request does not introduce any breaking changes**
- [ ] **I have added screenshots or gifs to help explain the change if applicable.**
- [ ] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ ] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

PR #5191, opened 2025-07-25
fix: to run collection a left click was needed

ernestorbemx:
# Description

As explained on #5181, to run a collection it is needed to collapse or open the collection from the sidebar. This happens because when a collection is opened its items are not loaded by default. It can stay the same but ensuring the collection is mounted when the "Run" item is clicked in the dropdown menu.

The solution looks like this in the UI:

bruno run collection

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
@@ -51,6 +51,7 @@ const Collection = ({ collection, searchText }) => {
   });
 
   const handleRun = () => {
+    ensureCollectionIsMounted();
     dispatch(
       addTab({
         uid: uuid(),


Pragadesh-45 (approved):

naman-bruno:
Hi @ernestorbemx, thank you for your contribution!

We’re aware of this issue and have already created a PR to address it: #5138. While your solution initiates collection loading when the "Run" button is clicked, it doesn’t fully resolve the problem. The same issue also occurs in the Share modal—if the collection isn't fully loaded and a user attempts to export it, they’ll end up with an empty result.

Even when loading starts immediately, large collections can take time to fully load. Without a proper loading state, users may end up running or exporting incomplete collection.

To address this comprehensively, PR #5138 introduces a robust loading mechanism that ensures the collection is fully loaded before proceeding. It also adds proper loading states to both the Runner and Share modal.

Therefore, I’ll be closing this PR. Your feedback on that PR would be very welcome!
Thanks again for your effort and contribution!

PR #5216, opened 2025-07-29
fix: test only flag in cli to inclue pre and post test

pooja-bruno:
# Description

This PR fixes the `--tests-only` flag in Bruno CLI to properly include requests that contain `test()` function calls in pre-request and post-response scripts.

[JIRA](https://usebruno.atlassian.net/browse/BRU-1241)

### Contribution Checklist:

### DIFF
--- a/packages/bruno-cli/src/utils/common.js
+++ b/packages/bruno-cli/src/utils/common.js
@@ -42,9 +42,8 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) => 
   return { data, dataBuffer };
 };
 
-
 module.exports = {
   lpad,
   rpad,
-  parseDataFromResponse,
+  parseDataFromResponse
 };


packages/bruno-cli/src/commands/run.js
```diff
@@ -467,10 +468,17 @@ const handler = async function (argv) {
     requestItems = getCallStack(resolvedPaths, collection, { recursive });
 
     if (testsOnly) {
-      requestItems = requestItems.filter((iter) => {
-        const requestHasTests = iter.request?.tests;
-        const requestHasActiveAsserts = iter.request?.assertions.some((x) => x.enabled) || false;
-        return requestHasTests || requestHasActiveAsserts;
+      requestItems = requestItems.filter((item) => {
+        const requestHasTests = item.request?.tests;
```
sanish-bruno: @pooja-bruno It is possible that the tests script doesn't contain any  tests, although we don't the check currently, can we introduce that check now? Let me know your thoughts!

packages/bruno-cli/src/commands/run.js
```diff
@@ -577,10 +585,10 @@ const handler = async function (argv) {
       // bail if option is set and there is a failure
       if (bail) {
         const requestFailure = result?.error && !result?.skipped;
-        const testFailure = result?.testResults?.find((iter) => iter.status === 'fail');
-        const assertionFailure = result?.assertionResults?.find((iter) => iter.status === 'fail');
-        const preRequestTestFailure = result?.preRequestTestResults?.find((iter) => iter.status === 'fail');
-        const postResponseTestFailure = result?.postResponseTestResults?.find((iter) => iter.status === 'fail');
+        const testFailure = result?.testResults?.find((item) => item.status === 'fail');
```
sanish-bruno: @pooja-bruno It is better to keep the changes limited, as the changes are not necessary for the implementation. This is fine for now!

packages/bruno-cli/src/utils/common.js
```diff
@@ -42,8 +42,49 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) =>
...
+    .replace(/'(?:[^'\\]|\\.)*'/g, "''")  // Remove single-quoted strings
+    .replace(/`(?:[^`\\]|\\.)*`/g, '``'); // Remove template literals
+  
+  // Look for standalone test() calls (not object method calls like obj.test())
+  // Find all test( occurrences and check they're not preceded by dots
+  let hasValidTest = false;
+  let searchFrom = 0;
+  
+  while (true) {
+    const index = cleanScript.indexOf('test', searchFrom);
```
sanish-bruno: we can also define tests with `it`

packages/bruno-cli/src/utils/common.js
```diff
@@ -42,8 +42,49 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) =>
   return { data, dataBuffer };
 };
 
+// Check for meaningful test() calls (not commented out or in strings)
+const hasValidTests = (script) => {
```
sanish-bruno: @pooja-bruno can we check the performance issue we may face if we introduce ast to the mix, we can use acorn, which is much faster compared to the ast parser we are currently using for traslation, here's a [gist ](https://gist.github.com/sanish-bruno/b5a15520a7150e5d2104ad18df5d4cf9)of acorn implementation for the same

sanish-bruno: can we test ast approach here? how it affect performance etc, our current ast parser, might not be the right one as it is slow and inefficient, here's the implemetation with acorn, which should be performant, checkout this [gist](https://gist.github.com/sanish-bruno/b5a15520a7150e5d2104ad18df5d4cf9)
here, regex will be used as a fallback.

sanish-bruno (approved):

packages/bruno-cli/src/utils/common.js
maintainer-bruno: This method is not very commonly used and hence common.js doesn't seem right place for this method. Let's create another util file `/util/request.js` for this.

Rename `hasValidTests` => `hasExecutableTestInScript` / `hasActiveTestInScript`, since the term valid is slightly misleading with syntax or some other validations.

packages/bruno-cli/tests/utils/common.spec.js
maintainer-bruno, line 1: Nicely done on the unit tests! They made the functionality much easier to follow.

maintainer-bruno (changes requested):

naman-bruno (approved):

maintainer-bruno (approved):

Issue #5231, opened 2025-07-31
JS API req object has getPathParams() function


Skrivoo:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] blocks me from using Bruno
- [x] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### Describe the feature you want to add, and how it would change your usage of Bruno

In a prescript it is possible to get every path parameters as an array of objects like
http://localhost:1234/customers/:customerId/something

`req.getPathParams()`

with this parameters like :customerId can be resolved in prescript. It is useful in some cases for example to have an escher auth (https://escherauth.io/) where the url needs to be compared in server side with the one witch the request contains and for this, every parameter and envVar must be resolved.

Skrivoo:
Hi @anusree-bruno ! any plans to merge the linked PR please? Thank you!

anusree-bruno:
similar as https://github.com/usebruno/bruno/issues/5677. I am closing this as a duplicate, Please check that issue for any updates on this issue.

PR #5235, opened 2025-07-31
feat: js api supports get path params

Skrivoo:
solves #5231 , #5677


sid-bruno:
Merging it in internally to fix conflicts and make any needed polishing changes

Issue #5171, opened 2025-07-23
OAuth2 client_credentials grant type - client_secret became mandatory


coxinol:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [x] this feature was working in a previous version but is broken in the current release.

[Jira](https://usebruno.atlassian.net/browse/BRU-1425)
### Bruno version

2.8.0

### Operating System

Windows 10 22H2

### Describe the bug

I am using OAuth 2.0 authentication type on a folder and leaving the client secret blank as it is not defined in my organization.
The folder contains requests with Auth configured as Inherit
When I run the requests, the OAuth 2.0 request is not triggered at all.
If I set a dummy client secret, the OAuth2 request is triggered, but fails because of a wrong secret.
So it seems that the client_secret became mandatory, wheras in the last version I used (2.4.0) it was not.

### Screenshots/Live demo link

Folder OAuth2 settings :
<img width="717" height="638" alt="Image" src="https://github.com/user-attachments/assets/635ac1a9-4b56-4b3b-ab3c-ae8284716a67" />

Bruno 2.8.0 - Request Timeline :
<img width="1689" height="346" alt="Image" src="https://github.com/user-attachments/assets/1eab29b1-baaf-4044-a7e4-281cd7cbc87d" />

Bruno 2.4.0 - Request Timeline :
<img width="1688" height="385" alt="Image" src="https://github.com/user-attachments/assets/b58e2bbe-b912-4d24-af11-1a612613c849" />

helloanoop:
This has been resolved in the latest version. Closing issue.

PR #5259, opened 2025-08-05
Fix: OAuth2 Token Scope Handling and Optional Client Secret

naman-bruno:
### Description

This PR addresses the following issues: #5171 and #4388.

Fixes included:
**Scope parameter handling:** The scope parameter was incorrectly required during the /token exchange call.
**Client secret requirement:** The client_secret is mandatory for OAuth2 flow, which is incorrect.

[Jira](https://usebruno.atlassian.net/browse/BRU-1425)

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/utils/oauth2.js
+++ b/packages/bruno-electron/src/utils/oauth2.js
@@ -259,15 +259,13 @@ const getOAuth2TokenUsingAuthorizationCode = async ({ request, collectionUid, fo
     redirect_uri: callbackUrl,
     client_id: clientId,
   };
-  if (clientSecret && credentialsPlacement !== "basic_auth_header") {
+  if (clientSecret && clientSecret.trim() !== '' && credentialsPlacement !== "basic_auth_header") {
     data.client_secret = clientSecret;
   }
   if (pkce) {
     data['code_verifier'] = codeVerifier;
   }
-  if (scope && scope.trim() !== '') {
-    data.scope = scope;
-  }
+
   axiosRequestConfig.data = qs.stringify(data);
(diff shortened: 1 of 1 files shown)


packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -427,14 +416,14 @@ const getOAuth2TokenUsingClientCredentials = async ({ request, collectionUid, fo
     'content-type': 'application/x-www-form-urlencoded',
     'Accept': 'application/json',
   };
-  if (credentialsPlacement === "basic_auth_header") {
+  if (credentialsPlacement === "basic_auth_header" && clientSecret && clientSecret.trim() !== '') {
     axiosRequestConfig.headers['Authorization'] = `Basic ${Buffer.from(`${clientId}:${clientSecret}`).toString('base64')}`;
```
aarondandy, line 420: I'm not sure if this is the right PR for such a change, but it was mentioned to me recently that the clientid value for `Authorization: Basic` should be form URL encoded prior to base64 encoding the full parameter value to handle cases where the clientid may contain a `:` value.

https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

Hat tip to @kevinchalet for pointing that out to me.

mlkauppila:
This is great! We are eagerly awaiting this fix :)

packages/bruno-electron/src/utils/oauth2.js
naman-bruno: Thanks for pointing that out! This issue is actually addressed in this PR: #5348 
Let me know if you have any further thoughts on the implementation.

PR #5260, opened 2025-08-05
feat: persist cookies in app

pooja-bruno:
# Description

Added cookies persist logic in app

[JIRA](https://usebruno.atlassian.net/browse/BRU-882)

### Contribution Checklist:

### DIFF
--- a/e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
+++ b/e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
@@ -1,14 +1,12 @@
 import { test, expect } from '../../playwright';
 
 test('should persist cookies across app restarts', async ({ createTmpDir, launchElectronApp }) => {
+  test.setTimeout(2 * 60 * 1000);
   // Create a temporary user-data directory so we control where the cookies store file is written.
   const userDataPath = await createTmpDir('cookie-persistence');
 
   const app1 = await launchElectronApp({ userDataPath });
   const page1 = await app1.firstWindow();
-
-
-
   await page1.waitForSelector('[data-trigger="cookies"]');
 
   // Open Cookies modal via the status-bar button.
--- a/e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
+++ b/e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
@@ -1,28 +1,13 @@
 import { test, expect } from '../../playwright';
-import { _electron as electron } from 'playwright';
-import * as path from 'path';
 
-
-test('should persist cookies across app restarts', async ({ createTmpDir }) => {
-  // Resolve the path to the Electron application we want to launch.
-  const electronAppPath = path.join(__dirname, '../../packages/bruno-electron');
-
-  // Create a temporary (but deterministic for this test) user-data directory so
-  // we control where the cookies store file is written.
+test('should persist cookies across app restarts', async ({ createTmpDir, launchElectronApp }) => {
+  // Create a temporary user-data directory so we control where the cookies store file is written.
   const userDataPath = await createTmpDir('cookie-persistence');
 
-  const launchApp = () =>
-    electron.launch({
-      args: [electronAppPath],
-      env: {
-        ...process.env,
-        ELECTRON_USER_DATA_PATH: userDataPath
-      }
-    });
-
-
-  const app1 = await launchApp();
+  const app1 = await launchElectronApp({ userDataPath });
   const page1 = await app1.firstWindow();
+
+
 
   await page1.waitForSelector('[data-trigger="cookies"]');
 
@@ -49,7 +34,7 @@ test('should persist cookies across app restarts', async ({ createTmpDir }) => {
   await app1.close();
 
   // Second launch – verify the cookie was persisted and re-loaded
-  const app2 = await launchApp();
+  const app2 = await launchElectronApp({ userDataPath });
   const page2 = await app2.firstWindow();
 
   // Open the Cookies modal again.
--- a/e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
+++ b/e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
@@ -1,0 +1,64 @@
+import { test, expect } from '../../playwright';
+import { _electron as electron } from 'playwright';
+import * as path from 'path';
+
+
+test('should persist cookies across app restarts', async ({ createTmpDir }) => {
+  // Resolve the path to the Electron application we want to launch.
+  const electronAppPath = path.join(__dirname, '../../packages/bruno-electron');
+
+  // Create a temporary (but deterministic for this test) user-data directory so
+  // we control where the cookies store file is written.
+  const userDataPath = await createTmpDir('cookie-persistence');
+
+  const launchApp = () =>
+    electron.launch({
+      args: [electronAppPath],
+      env: {
+        ...process.env,
+        ELECTRON_USER_DATA_PATH: userDataPath
+      }
+    });
+
+
+  const app1 = await launchApp();
+  const page1 = await app1.firstWindow();
+
+  await page1.waitForSelector('[data-trigger="cookies"]');
+
+  // Open Cookies modal via the status-bar button.
+  await page1.click('[data-trigger="cookies"]');
+
+  // When no cookies are present the modal shows a centred "Add Cookie" button.
+  await page1.getByRole('button', { name: /Add Cookie/i }).click();
+
+  // Fill out the form.
+  await page1.fill('input[name="domain"]', 'example.com');
+  await page1.fill('input[name="path"]', '/');
+  await page1.fill('input[name="key"]', 'session');
+  await page1.fill('input[name="value"]', 'abc123');
+
+  await page1.getByRole('button', { name: 'Save' }).click();
+
+  await expect(page1.getByText('example.com')).toBeVisible();
+
+  await page1.waitForTimeout(500);
+
+  await app1.close();
+
+  // Second launch – verify the cookie was persisted and re-loaded
+  const app2 = await launchApp();
+  const page2 = await app2.firstWindow();
+
+  // Open the Cookies modal again.
+  await page2.waitForSelector('[data-trigger="cookies"]');
+  await page2.click('[data-trigger="cookies"]');
+
+  // The domain we added earlier should still be present.
+  await expect(page2.getByText('example.com')).toBeVisible();
+
+  // Allow some time for manual observation / renderer stabilization.
+  await page2.waitForTimeout(3000);
+
+  await app2.close();
+});
--- a/e2e-tests/002-cookies-tests/002-corrupted-passkey.spec.ts
+++ b/e2e-tests/002-cookies-tests/002-corrupted-passkey.spec.ts
@@ -1,25 +1,14 @@
 import { test, expect } from '../../playwright';
-import { _electron as electron } from 'playwright';
 import * as path from 'path';
 import * as fs from 'fs/promises';
 
-test('should handle corrupted passkey and still display saved cookie list', async ({ createTmpDir }) => {
-  const electronAppPath = path.join(__dirname, '../../packages/bruno-electron');
-
+test('should handle corrupted passkey and still display saved cookie list', async ({ createTmpDir, launchElectronApp }) => {
   const userDataPath = await createTmpDir('corrupted-passkey');
 
+  const app1 = await launchElectronApp({ userDataPath });
 
-  const launchApp = () =>
-    electron.launch({
-      args: [electronAppPath],
-      env: {
-        ...process.env,
-        ELECTRON_USER_DATA_PATH: userDataPath
-      }
-    });
 
   // 1. First run – add a cookie via the UI so `cookies.json` is created.
-  const app1 = await launchApp();
   const page1 = await app1.firstWindow();
 
   await page1.waitForSelector('[data-trigger="cookies"]');
@@ -47,7 +36,7 @@ test('should handle corrupted passkey and still display saved cookie list', asyn
   await fs.writeFile(cookiesFilePath, JSON.stringify(cookiesJson, null, 2));
 
   // 3. Second run – Bruno should recover and still list the cookie domain
-  const app2 = await launchApp();
+  const app2 = await launchElectronApp({ userDataPath });
   const page2 = await app2.firstWindow();
 
   await page2.waitForSelector('[data-trigger="cookies"]');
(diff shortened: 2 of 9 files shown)


packages/bruno-electron/src/ipc/collection.js
```diff
@@ -889,45 +890,58 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
     }
   });
 
-  ipcMain.handle('renderer:delete-cookies-for-domain', async (event, domain) => {
-    try {
-      await deleteCookiesForDomain(domain);
 
+  const updateCookiesAndNotify = async () => {
```
sanish-bruno, line 894: @pooja-bruno does this function needs a try catch block as all the places it is being used is already within a try catch? we are eventually throwing the error and will be catched by the parent catch block!

packages/bruno-electron/src/utils/encryption.js
```diff
@@ -29,6 +29,33 @@ function deriveKeyAndIv(password, keyLength, ivLength) {
   return { key, iv };
 }
 
+function aes256EncryptWithPasskey(data, passkey) {
```
sanish-bruno: Do we have to create another function just to handle pass key? I thought we will just be adding a new param called pass key to existing functions and will be using that to encrypt/decrypt instead of `machineId`, have you faced any issues with that approach?

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,184 @@
...
+      if (!this.passkey) {
+        console.warn('Failed to decrypt passkey, falling back to unencrypted cookies');
+      }
+    } catch (err) {
+      console.warn('Failed to initialize encryption, falling back to unencrypted cookies:', err);
+      this.passkey = null;
+    }
+  }
+
+  encryptCookieValue(value) {
```
sanish-bruno: these functions `encryptCookieValue` and `decryptCookieValue`  just host logic for encryption/decryption not necessarily something specific to cookieValue, I think, guard clauses related handling we have here, probably should be handled by the `encryptString` , `decryptString` functions, what needs to be returned after encrpytion should be upto the encryption module, I guess, the follwing should be enough

```js
 const decryptedCookie = {
            ...cookie,
            value: decryptString(cookie.value, this.passkey)
          };
```

packages/bruno-electron/src/utils/encryption.js
pooja-bruno: You’re right having a separate pair of helpers was unnecessary. I have updated the code to use same function for passkey.

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,167 @@
...
+            value: ''
+          });
+        }
+      });
+    });
+
+    return decryptedCookies;
+  }
+
+  setCookies(cookies) {
```
sanish-bruno, line 90: @pooja-bruno one of the concern we had early was that how many times should we actually trigger a disc write, if we do a disc write on every update, that might be a bit overwhelming and un-necessary, can we have an interval based writing to the disc or a debouncing logic of writing to the disc?

sanish-bruno (changes requested):

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -32,6 +32,8 @@ const { getProcessEnvVars } = require('../../store/process-env');
 const { getBrunoConfig } = require('../../store/bruno-config');
 const Oauth2Store = require('../../store/oauth2');
 const { isRequestTagsIncluded } = require('@usebruno/common');
+const { cookiesStore } = require('../../store/cookies');
+
```
sanish-bruno, line 36: remove the extra line

e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { test, expect } from '../../playwright';
+
+test('should persist cookies across app restarts', async ({ createTmpDir, launchElectronApp }) => {
+  // Create a temporary user-data directory so we control where the cookies store file is written.
+  const userDataPath = await createTmpDir('cookie-persistence');
+
+  const app1 = await launchElectronApp({ userDataPath });
+  const page1 = await app1.firstWindow();
+
```
sanish-bruno: remove unncessary gaps within the file

sanish-bruno: Add time-out to the test

e2e-tests/002-cookies-tests/002-corrupted-passkey.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { test, expect } from '../../playwright';
+import * as path from 'path';
+import * as fs from 'fs/promises';
+
+test('should handle corrupted passkey and still display saved cookie list', async ({ createTmpDir, launchElectronApp }) => {
+  const userDataPath = await createTmpDir('corrupted-passkey');
```
sanish-bruno, line 7: add timeout to the test, also remove unnecessary new lines

playwright/index.ts
```diff
@@ -37,8 +37,13 @@ export const test = baseTest.extend<
   launchElectronApp: [
     async ({ playwright, createTmpDir }, use, workerInfo) => {
       const apps: ElectronApplication[] = [];
-      await use(async ({ initUserDataPath } = {}) => {
-        const userDataPath = await createTmpDir('electron-userdata');
+      await use(async ({ initUserDataPath, userDataPath: providedUserDataPath } = {}) => {
```
sanish-bruno, line 40: @pooja-bruno what was the need for having userDataPath? can't we use the initUserDataPath intead? you can use folder `init-user-data` to and can have preferences.json within it, like in the case of other test cases, should ideally work with newly added logic here, this will also remove the need to create a additional directory from test cases, i believe.

pooja-bruno: userDataPath points Electron to a live folder that keeps the cookie; initUserDataPath is just a one-time template and would start each launch fresh, breaking the persistence check.

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,193 @@
...
+        const passkey = this.generatePasskey();
+        encryptedPasskey = encryptString(passkey);
+        if (!encryptedPasskey) {
+          console.warn('Failed to encrypt new passkey, falling back to unencrypted cookies');
+          this.passkey = null;
+          return;
+        }
+        this.store.set('encryptedPasskey', encryptedPasskey);
+      }
+      this.passkey = decryptString(encryptedPasskey);
```
sanish-bruno: can you initialize all the concerned variables with repective default values within the constructor? like `this.passkey`

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,193 @@
...
+        encryptedPasskey: null,
+        cookies: {}
+      }
+    });
+
+    this.initializeEncryption();
+
+  }
+
+  generatePasskey() {
```
sanish-bruno: @pooja-bruno some of the functions like generatePasskey are related to the implementation, rather the the methods, we want to expose, we are already doing _saveTimer, i guess javascript recently introduced `#saveTimer`  like syntax to differentiate private methods and function, it is possible to do it here?

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,193 @@
...
+const { cookieJar } = cookiesModule;
+const { Cookie } = require('tough-cookie');
+const { createCookieString } = cookiesModule;
+const crypto = require('crypto');
+const { encryptString, decryptString } = require('../utils/encryption');
+
+
+class CookiesStore {
+  constructor() {
+    this._saveTimer = null;
```
sanish-bruno: can be renamed to _saveTimerId for better understanding

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,193 @@
...
+    } catch (err) {
+      console.warn('Failed to save cookie jar:', err);
+    } finally {
+      this._debounceStart = null;
+    }
+  }
+
+  saveCookieJar(immediate = false) {
+    // Debounced write to avoid excessive disk I/O during rapid request bursts
+    const DEBOUNCE_MS = 1000;
```
sanish-bruno: @pooja-bruno can be moved to top. along with initialization of variables, also can be increased to 5 - 10 seconds.

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -889,12 +890,21 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedCollection
...
   });
 
+
+  const updateCookiesAndNotify = async () => {
+    const domainsWithCookies = await getDomainsWithCookies();
+    mainWindow.webContents.send(
+      'main:cookies-update',
+      safeParseJSON(safeStringifyJSON(domainsWithCookies))
+    );
+    cookiesStore.saveCookieJar(true);
```
sanish-bruno: @pooja-bruno we are are actually forcing immediate update to the on every update, which undercut the purpose of having a a timer? I guess, immediate save should be called only when user triggers the close of the app

playwright/index.ts
```diff
@@ -37,8 +37,13 @@ export const test = baseTest.extend<
...
+        const userDataPath = providedUserDataPath || (await createTmpDir('electron-userdata'));
+
+        // Ensure dir exists when caller supplies their own path
+        if (providedUserDataPath) {
+          await fs.promises.mkdir(userDataPath, { recursive: true });
```
sanish-bruno, line 45: @pooja-bruno it is possible to use `createTmpDir` here, and it will help us understand how `fs.promises.mkdir` and  and `fs.promises.mkdtemp` behaves. probably creating a temp directory, will be better, as we don't need that directory post this test. probably it might be used in post-test cleanup. It will be better if we can understand it more.

sanish-bruno (changes requested):

playwright/index.ts
pooja-bruno: Using `createTmpDir()` inside launchElectronApp would give us a brand-new folder on every call, so the second launch wouldn’t see the cookies written by the first one. We only need to ensure the path we already created in the test exists, so the simple `mkdir(..., { recursive: true })` is the safest choice here

sanish-bruno: cool

e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  await page1.fill('input[name="key"]', 'session');
+  await page1.fill('input[name="value"]', 'abc123');
+  await page1.check('input[name="secure"]');
+  await page1.check('input[name="httpOnly"]');
+
+  await page1.getByRole('button', { name: 'Save' }).click();
+
+  await expect(page1.getByText('example.com')).toBeVisible();
+
+  await page1.waitForTimeout(500);
```
bijin-bruno, line 30: Why do we need these timeouts? (line 30 & 45)

e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { test, expect } from '../../playwright';
+
+test('should persist cookies across app restarts', async ({ createTmpDir, launchElectronApp }) => {
+  test.setTimeout(2 * 60 * 1000);
```
bijin-bruno, line 4: Is there a need to increase the timeout in each of these tests?

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,198 @@
...
+const crypto = require('crypto');
+const { encryptString, decryptString } = require('../utils/encryption');
+
+const DEBOUNCE_MS = 5000; // Debounce duration (ms) for persisting cookie jar
+
+
+class CookiesStore {
+  #saveTimerId = null;
+  #debounceStart = null; // Track first debounce time
+  #passkey = null;
```
bijin-bruno, line 13: Any reason why the class variable name starts with #? This doesn’t seem consistent with our usual naming pattern

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,198 @@
...
+const { cookies: cookiesModule } = require('@usebruno/common');
+const { cookieJar } = cookiesModule;
+const { Cookie } = require('tough-cookie');
+const { createCookieString } = cookiesModule;
+const crypto = require('crypto');
+const { encryptString, decryptString } = require('../utils/encryption');
+
+const DEBOUNCE_MS = 5000; // Debounce duration (ms) for persisting cookie jar
+
+
```
bijin-bruno: Please remove the unwanted linebreaks

playwright/index.ts
```diff
@@ -67,10 +72,10 @@ export const test = baseTest.extend<
         });
 
         const { workerIndex } = workerInfo;
-        app.process().stdout.on('data', (data) => {
+        app.process()?.stdout?.on('data', (data) => {
```
bijin-bruno, line 75: Do we need optional chaining here?

packages/bruno-electron/src/store/cookies.js
```diff
@@ -0,0 +1,198 @@
...
+    Object.values(cookieStore).forEach(domainCookies => {
+      if (!Array.isArray(domainCookies)) return;
+
+      domainCookies.forEach(cookie => {
+        try {
+
+          // Create cookie with decrypted value
+          const decryptedCookie = {
+            ...cookie,
+            value: decryptString(cookie.value, this.#passkey)
```
bijin-bruno, line 73: Note: There have been recent changes to how errors are softened. Please rebase and verify whether we need to use the new decryptStringSafe method. I believe we are already handling the errors here.

bijin-bruno (changes requested):

playwright/index.ts
pooja-bruno: I added optional chaining to resolve some TypeScript errors that were occurring.

e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
pooja-bruno: The app closes too quickly, so we can’t check whether the cookies are still present. That’s why I added this.

packages/bruno-electron/src/store/cookies.js
pooja-bruno: The # prefix is used to define private class fields in JavaScript. Since this variable is not intended to be accessed outside the class, I marked it as private.

e2e-tests/002-cookies-tests/001-cookie-persistence.spec.ts
pooja-bruno: Added this to stop any hang after 2 min instead of running indefinitely.

packages/bruno-electron/src/store/cookies.js
pooja-bruno: I don't think we need to use that

Issue #5274, opened 2025-08-06
Use a better marking color for search


gesslinger:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] blocks me from using Bruno
- [x] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### Describe the feature you want to add, and how it would change your usage of Bruno

Change the color of highlighting words while search. Yellow text marking is more common and better visible, then the light grey.

<img width="612" height="726" alt="Image" src="https://github.com/user-attachments/assets/8ecfcc34-bbf7-42cc-825f-5a08e3f7ea51" />

anusree-bruno:
Improved search is now available in latest release, please checkout.

https://github.com/user-attachments/assets/881234b5-52bf-40d0-9417-905b55c13d61

gesslinger:
Awesome! Thanks for this enhancement!

Issue #4871, opened 2025-06-10
Search function in Body is Case-sensitive


EQ72077:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.0.1

### Operating System

Windows 11

### Describe the bug

This is more on the 'Search' function in the Body. Here's what I found:

if all characters are in correct capitalization, it will show the result and proceed to the occurrence
 
**if one character is in wrong capitalization, it will show the result but will not proceed to the occurrence**
 
if all characters are not in correct capitalization, it it will show the result and proceed to the occurrence

-- so more of, if only one letter is in wrong capitalization, it will not proceed to the occurrence but will still show the result.

### Screenshots/Live demo link

![Image](https://github.com/user-attachments/assets/f2556b51-b6e2-489c-b056-ec7297cd6d02)

![Image](https://github.com/user-attachments/assets/199dbc13-1aa7-46fd-a7eb-484e88855fb1)

kishansinghifs1:
Can I work on this .

EQ72077:
> Can I work on this .

Yes please!

ernestbrixx:
Is @kishansinghifs1 still working on this or can I work on it?

EQ72077:
> Is [@kishansinghifs1](https://github.com/kishansinghifs1) still working on this or can I work on it?

no, nobody worked on this issue.

EQ72077:
> > Is [@kishansinghifs1](https://github.com/kishansinghifs1) still working on this or can I work on it?
> 
> no, nobody worked on this issue.

Yes @Ernest2026  please work on this as @kishansinghifs1 have not updated

vox91:
What would be the expected result ?

The CodeMirror search module (addon) seems to be case sensitive, that is why it not proceed to the next occurence, while the counter isn't.

anusree-bruno:
Improved search is now available in latest release, please checkout.

https://github.com/user-attachments/assets/881234b5-52bf-40d0-9417-905b55c13d61

Issue #4659, opened 2025-05-13
Search result counter doesn't work with regular expressions


discapes:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.2.0

### Operating System

macOS

### Describe the bug

When searching the body of a request with a regular expression (for example /foo/), even though CodeMirror finds matches and is able to highlight them and jump through them, the "N results" dialog just shows "0 results"

tags: regex regexp

### Screenshots/Live demo link

![Image](https://github.com/user-attachments/assets/7ee82898-2b06-43f0-a43f-794e4768a575)

anusree-bruno:
Improved search is now available in latest release, please checkout.

https://github.com/user-attachments/assets/881234b5-52bf-40d0-9417-905b55c13d61

Issue #4395, opened 2025-04-01
Search text option response section is not user friendly


keshavakn:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] blocks me from using Bruno
- [x] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### Describe the feature you want to add, and how it would change your usage of Bruno

Hi Team,

Search option plays a vital role for our day to day executions. It needs to be user friendly when looking for any particular response. Attached an image something like this.  

![Image](https://github.com/user-attachments/assets/f7ef4a8c-b479-4a7a-b85e-75bed8e5cdd3)

anusree-bruno:
Similar https://github.com/usebruno/bruno/issues/3702

anusree-bruno:
We have improved our search to filter by case, word, regex and also added arrows for navigating up and down, please checkout our latest release.

PR #5278, opened 2025-08-07
Feat/editor custom search

anusree-bruno:
This PR implements a custom search functionality for the code editor with enhanced visual highlighting.

**Changes Made**

- Custom Search Component: Added a new search bar with regex, case-sensitive, and whole word search options
- Search Navigation: Implemented next/previous match navigation with keyboard shortcuts (Enter/Shift+Enter)
- Enhanced Highlighting: Updated search match highlighting to use yellow background colors for better visibility

**Technical Details**

- Created custom search component with advanced options
- Added keyboard shortcuts (Ctrl+F/Cmd+F to open, Escape to close)
- Integrated with existing CodeMirror editor

**Issues**

- https://github.com/usebruno/bruno/issues/5274
- https://github.com/usebruno/bruno/issues/4871
- https://github.com/usebruno/bruno/issues/4659
- https://github.com/usebruno/bruno/issues/4395
- https://github.com/usebruno/bruno/issues/3702
- https://github.com/usebruno/bruno/issues/1726

**Demo**

https://github.com/user-attachments/assets/56b7c734-a8be-4ce5-9069-85fa6976d4a5

- [x] The pull request only addresses one issue or adds one feature.
- [x]  The pull request does not introduce any breaking changes
- [x]  I have added screenshots or gifs to help explain the change if applicable.
- [x]  I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).
- [x]  Create an issue and link to the pull request.


bfmags:
+1  Thanks for submitting this! We definitely need it, and it looks brilliant 👍

packages/bruno-app/src/components/CodeEditor/CustomSearch/index.js
```diff
@@ -0,0 +1,200 @@
...
+  const [caseSensitive, setCaseSensitive] = useState(false);
+  const [wholeWord, setWholeWord] = useState(false);
+  const [matchIndex, setMatchIndex] = useState(0);
+  const [matchCount, setMatchCount] = useState(0);
+
+  const searchMarks = useRef([]);
+  const searchLineHighlight = useRef(null);
+  const searchMatches = useRef([]);
+
+  const doSearch = useCallback(debounce((newIndex = 0) => {
```
sid-bruno: I don't think `debounce` would actually work this way, you might want to try to debounce the value updates instead of the callback itself.

a demo showing that it fires for every character whether I pause or not

https://github.com/user-attachments/assets/f0baf1c3-36d0-4328-8c2b-b3ee996a4494

packages/bruno-app/src/components/CodeEditor/CustomSearch/index.js
```diff
@@ -0,0 +1,200 @@
...
+      } else {
+        searchLineHighlight.current = null;
+      }
+
+      setMatchCount(matches.length);
+      setMatchIndex(matchIndex);
+      if (matches.length) {
+        editor.scrollIntoView(matches[matchIndex].from, 100);
+        editor.setSelection(matches[matchIndex].from, matches[matchIndex].to);
+      }
```
sid-bruno: the 2nd `if` block isn't dependent on the result from the previous one, any other reason to split it into 2 blocks?

packages/bruno-app/src/components/CodeEditor/CustomSearch/index.js
```diff
@@ -0,0 +1,200 @@
...
+      searchMatches.current = [];
+    }
+  }, 150), [searchText, regex, caseSensitive, wholeWord, editor]);
+
+  useEffect(() => {
+    if (visible) setSearchBarVisible(true);
+  }, [visible]);
+
+  const handleSearchBarClose = useCallback(() => {
+    setSearchBarVisible(false);
```
sid-bruno: Assuming the `searchBarVisible` state is added to allow the CustomSearch to close the search bar, we are already lifting the event up into the `onClose` function, which also sets the same state, I think we can remove the `useEffect` here, I think it's redundant, let me know if it was intentional

packages/bruno-app/src/components/CodeEditor/CustomSearch/index.js
```diff
@@ -0,0 +1,200 @@
...
+    setMatchIndex(prev);
+    doSearch(prev);
+  }, [matchIndex, doSearch]);
+
+  // Effect to trigger search when search options change
+  useEffect(() => {
+    if (searchText) {
+      doSearch(0);
+    }
+  }, [regex, caseSensitive, wholeWord, doSearch, searchText, searchBarVisible]);
```
sid-bruno: Recheck if this is needed since you have an event handler on each change re-running the search, this might cause a delayed 2nd search

tests/request/tests/custom-search/custom-search.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+test.describe('Custom Search Functionality in Scripts Tab', () => {
+  test('should open search box when Cmd+F or Ctrl+F is pressed in scripts tab', async ({ pageWithUserData: page }) => {
+    await page.getByTitle('custom-search').click();
+    
+    try {
+      await page.getByLabel('Safe Mode').check({ timeout: 5000 });
+      await page.getByRole('button', { name: 'Save' }).click();
+    } catch (e) {
+    }
```
sid-bruno: Can offload this to be done automatically by using `collection-security.json`

refer to `tests/environments/api-setEnvVar/init-user-data/collection-security.json` for an example

sid-bruno:
Also, not sure if we want to change how this is implemented but currently for each `doSearch` the entire list of matches is recomputed, a more efficient way might be to compute the search results using a `useMemo` everytime one of the options change and just use that as the source for finding next and previous values. 

Example: 
Line 144 to 149 in `packages/bruno-app/src/components/CodeEditor/CustomSearch/index.js` 
Triggers a `doSearch` everytime which recalculates the query and runs it again, so if you had 100 results, all are re-computed every time

helloanoop:
@anusree-bruno Can we make the search container absolute and not push the line down ?

image

packages/bruno-app/src/components/CodeEditor/CustomSearch/index.js
```diff
@@ -0,0 +1,211 @@
...
+      setMatchCount(0);
+      setMatchIndex(0);
+      searchMatches.current = [];
+    }
+
+  }, [searchText, regex, caseSensitive, wholeWord, editor, memoizedMatches]);
+
+  const debounceSearch = useMemo(() => debounce((index = 0) => {
+    doSearch(index);
+  }, 150), [doSearch]);
```
sid-bruno: https://github.com/user-attachments/assets/32c1d1a8-ee61-418f-9955-a5e5d0b171de

This won't work either since the `doSearch` function will be redefined every time matches or searchText change, so you have a new debounce call, an easy way to check is by changing the delay from 150 to 1000 and checking if it matches the whole word or each letter. 

You can instead do something like this 

```js

// to be defined outside the component (wherever we save custom hooks)
function useDebounce(value, delay) {
  const [debouncedValue, setDebouncedValue] = useState(value);

  useEffect(() => {
    const handler = setTimeout(() => {
      setDebouncedValue(value);
    }, delay);

    return () => {
      clearTimeout(handler);
    };
  }, [value, delay]);

  return debouncedValue;
}

// then use it like so 

const debouncedSearchText = useDebounce(searchText, 150)

useEffect(()=>{
  doSearch(0)
},[debouncedSearchText, doSearch])
```

tests/request/tests/custom-search/custom-search.spec.ts
anusree-bruno: Done

anusree-bruno:
> @anusree-bruno Can we make the search container absolute and not push the line down ?
> 
> image

Done

sid-bruno (approved):
LGTM

Issue #5258, opened 2025-08-05
Problem with Digest authentication with query params


ganesh-bruno:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.8.0

### Operating System

MAc OS

### Describe the bug

When I call an API without query params with Digest authentication, Bruni works well.

If I add params, my server sends a 400 error. I don’t have this error code with other tools.

### .bru file to reproduce the bug

NA

### Screenshots/Live demo link

NA

PR #5317, opened 2025-08-13
Fix: resolve URL and Search string for proper handling in digest auth interceptor

Pragadesh-45:
Fixes: #5258 
[JIRA](https://usebruno.atlassian.net/browse/BRU-1401)

# Description

Updated the digest authentication interceptor to correctly build the full URL with the proper search string, which will also include the digest authentication without leaving the query parameters behind

### Contribution Checklist:


packages/bruno-requests/src/auth/digestauth-helper.js
```diff
@@ -89,9 +89,16 @@ export function addDigestInterceptor(axiosInstance, request) {
...
 
-        const uri = new URL(request.url, request.baseURL || 'http://localhost').pathname; // Handle relative URLs
+        // Build full URL from the original request (may include query params and baseURL)
+        const resolvedUrl = new URL(
+          originalRequest.url || request.url,
+          originalRequest.baseURL || request.baseURL || 'http://localhost'
+        );
+        const uri = `${resolvedUrl.pathname}${resolvedUrl.search}`;
+        // Used 'GET' as default method to avoid missing method error
+        const method = (originalRequest.method || request.method || 'GET').toUpperCase();
```
sanish-bruno, line 99: @Pragadesh44-Bruno previously we were not uppercasing the method! what will be the impact of this, is it expected that the method always has to be uppercase, and we are just making sure that it will be the case always?

sanish-bruno (changes requested):

packages/bruno-requests/src/auth/digestauth-helper.js
Pragadesh-45: This has been added for the MD5 to work properly, since the MD5 hashing algorithm is case-sensitive and it is only correct if passed as the standard

pooja-bruno (approved):

bijin-bruno (approved):

PR #5328, opened 2025-08-14
Update Bruno's Age 🎉

Pragadesh-45:
# Description

Update Bruno's Age

### DIFF
--- a/packages/bruno-common/src/interpolate/index.spec.ts
+++ b/packages/bruno-common/src/interpolate/index.spec.ts
@@ -6,13 +6,13 @@ describe('interpolate', () => {
     const inputObject = {
       'user.name': 'Bruno',
       user: {
-        age: 4
+        age: 6
       }
     };
 
     const result = interpolate(inputString, inputObject);
 
-    expect(result).toBe('Hello, my name is Bruno and I am 4 years old');
+    expect(result).toBe('Hello, my name is Bruno and I am 6 years old');
   });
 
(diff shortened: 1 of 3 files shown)


packages/bruno-common/src/interpolate/index.spec.ts
```diff
@@ -6,13 +6,13 @@ describe('interpolate', () => {
     const inputObject = {
       'user.name': 'Bruno',
       user: {
-        age: 4
+        age: 6
```
bijin-bruno: Let's use a constant at top level and see if we can dynamically set value using birthdate

bijin-bruno (changes requested):

packages/bruno-common/src/interpolate/index.spec.ts
helloanoop: @bijin-bruno Lets make an exception here :)

helloanoop:
@Pragadesh-45 Can you fix the failing test?

bijin-bruno (approved):

packages/bruno-tests/collection/string interpolation/runtime vars.bru
```diff
@@ -38,18 +38,56 @@ assert {
 }
 
 script:pre-request {
+  const brunoBirthDate = new Date('2019-08-08');
+
+  const calculateAgeFromBirthDate = (birthDate = brunoBirthDate) => {
```
bijin-bruno, line 43: I think this method can be moved to a helper script file inside the collection

PR #5354, opened 2025-08-18
fix: openapi auth import in bruno

pooja-bruno:
# Description

### Contribution Checklist:

### DIFF
--- a/packages/bruno-converters/src/openapi/openapi-to-bruno.js
+++ b/packages/bruno-converters/src/openapi/openapi-to-bruno.js
@@ -154,16 +154,15 @@ const transformOpenapiRequestItem = (request) => {
           uid: uuid(),
           name: auth.name,
           value: '{{apiKey}}',
-          description: 'Authentication header',
+          description: auth.description || '',
           enabled: true
         });
       } else if (auth.in === 'query') {
-        // Add as query param
         brunoRequestItem.request.params.push({
           uid: uuid(),
           name: auth.name,
           value: '{{apiKey}}',
-          description: 'Authentication query param',
+          description: auth.description || '',
           enabled: true,
           type: 'query'
         });


packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -108,13 +110,16 @@ const transformOpenapiRequestItem = (request) => {
...
+    brunoRequestItem.request.auth.mode = 'inherit';
+    return brunoRequestItem;
+  }
+
+  let auth = null;
   if (_operationObject.security && _operationObject.security.length > 0) {
-    let schemeName = Object.keys(_operationObject.security[0])[0];
+    const schemeName = Object.keys(_operationObject.security[0])[0];
     auth = request.global.security.getScheme(schemeName);
-  } else if (request.global.security.supported.length > 0) {
```
sanish-bruno, line 116: @pooja-bruno is this change intentional? currently we are defaulting to apiSpec.security when security schemas are not defined, why are we no longer doing it?

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -129,14 +134,88 @@ const transformOpenapiRequestItem = (request) => {
...
+          description: 'Authentication query param',
+          enabled: true,
+          type: 'query'
+        });
+      }
+    } else if (auth.type === 'oauth2') {
+      // Determine flow (grant type)
+      let flows = auth.flows || {};
+      let grantType = 'client_credentials';
+      if (flows.authorizationCode) {
```
sanish-bruno, line 174: It is possible that the the flows object can contain multiple grantType details, as shown below
https://swagger.io/docs/specification/v3_0/authentication/oauth2/#security-scheme-examples

```yml
components:
  securitySchemes:
    oAuth2:
      type: oauth2
      description: For more information, see https://developers.getbase.com/docs/rest/articles/oauth2/requests
      flows:
        implicit:
          authorizationUrl: https://api.getbase.com/oauth2/authorize
          scopes:
            read: Grant read-only access to all your data except for the account and user info
            write: Grant write-only access to all your data except for the account and user info
            profile: Grant read-only access to the account and user info only
        password:
          tokenUrl: https://api.getbase.com/oauth2/token
          scopes:
            read: Grant read-only access to all your data except for the account and user info
            write: Grant write-only access to all your data except for the account and user info
            profile: Grant read-only access to the account and user info only
```
I believe with the current logic, in the case we have multiple oauth flows, it will always use the first one recognized according to `if - else if` statements. another client seems to be taking the first one out of the flows defined. It would be nice if we can do something better here, Is there any key the can help us identify what type of grant_type is being used for a particular request, when there are multiple types of flows defined, it will be great if can find one. current flow is good enough though, i wouldn't stress much to find the distinguishing key!

sanish-bruno (changes requested):

packages/bruno-converters/src/openapi/openapi-to-bruno.js
sanish-bruno: If we need this, i guess we can add test cases to verify this behaviour

pooja-bruno: Yes, I removed it because if security is empty, the auth mode will now default to inherit. I’ll also create a dedicated test case for this.

packages/bruno-converters/src/openapi/openapi-to-bruno.js
pooja-bruno: Based on my research, I don’t think there’s a way to determine which auth a request is using, so we’ll need to go with this logic for now.

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -129,14 +134,88 @@ const transformOpenapiRequestItem = (request) => {
...
+      };
+      brunoRequestItem.request.auth.mode = 'apikey';
+      brunoRequestItem.request.auth.apikey = apikeyConfig;
+
+      if (auth.in === 'header' || auth.in === 'cookie') {
+        brunoRequestItem.request.headers.push({
+          uid: uuid(),
+          name: auth.name,
+          value: '{{apiKey}}',
+          description: 'Authentication header',
```
sanish-bruno: @pooja-bruno do we have the concept of description for headers(also applicable to param descriptions too) within bruno? ,I don't have much clue why we were adding description even before. If open api supports header, param descriptions, I guess we should be passing the original description instead of `Authentication header` or `Authentication query param`

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -108,13 +110,16 @@ const transformOpenapiRequestItem = (request) => {
     }
   });
 
-  let auth;
-  // allow operation override
+  // Handle explicit no-auth case where security: [] on the operation
+  if (Array.isArray(_operationObject.security) && _operationObject.security.length === 0) {
+    brunoRequestItem.request.auth.mode = 'inherit';
```
JustinMegaport, line 115: I'm a bit late to the party, but I am curious to know why we default to `inherit` auth when an operation specifies `security: []`. 

My understanding is that an empty array passed into the `security` field of an operation declares that the operation has no authentication requirements (and can be used to override the global / default security scheme).

Setting the auth mode in this scenario to `inherit` instead of "no auth" has the same outcome as having no security property on the operation's spec. I believe this to be the incorrect desired behaviour.

PR #5373, opened 2025-08-20
fix(cookie-store): defer encryption setup to prevent early macOS ‘Chr…

pooja-bruno:
# Description

Delay (defer) cookie-store encryption setup so the Keychain prompt appears after the app is fully initialised, fixing the unwanted “Chromium Safe Storage” dialog on macOS.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/package.json
+++ b/packages/bruno-electron/package.json
@@ -65,6 +65,7 @@
     "nanoid": "3.3.8",
     "qs": "^6.11.0",
     "socks-proxy-agent": "^8.0.2",
+    "tough-cookie": "^6.0.0",
     "uuid": "^9.0.0",
     "yup": "^0.32.11"
   },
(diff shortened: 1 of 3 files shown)


packages/bruno-electron/package.json
```diff
@@ -65,6 +65,7 @@
     "nanoid": "3.3.8",
     "qs": "^6.11.0",
     "socks-proxy-agent": "^8.0.2",
+    "tough-cookie": "^6.0.0",
```
sanish-bruno, line 68: @pooja-bruno do we need to add tough cookie within bruno-electon, i believe cookie related stuff is part of bruno-common, probably adding it here is redundant, unless we are directly using tough cookie within bruno-electron?

packages/bruno-common/src/cookies/index.ts
```diff
@@ -14,9 +14,7 @@ const addCookieToJar = (setCookieHeader: string, requestUrl: string): void => {
 };
 
 const getCookiesForUrl = (url: string) => {
-  return cookieJar.getCookiesSync(url, {
-    secure: isPotentiallyTrustworthyOrigin(url)
```
sanish-bruno, line 18: @pooja-bruno i guess we introduced this function cause cookies behaved differently within localhost and other kinds domain,  although the function is tested well for it's behaviour, the application of the same is not tested. and this function added few months ago to a particular issue, this can cause a regression. can you investigate on why we added it earlier, and what can break if we remove this.

packages/bruno-electron/package.json
pooja-bruno: we are using it inside electron

PR #5404, opened 2025-08-22
fix: environment persistence and UI

sanjaikumar-bruno:
https://github.com/user-attachments/assets/374019d0-c4ac-4299-9477-7332a98c8318

### DIFF
--- a/packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
+++ b/packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
@@ -41,7 +41,8 @@ import {
   initRunRequestEvent,
   updateRunnerConfiguration as _updateRunnerConfiguration,
   updateActiveConnections,
-  saveRequest as _saveRequest
+  saveRequest as _saveRequest,
+  saveEnvironment as _saveEnvironment
 } from './index';
 
 import { each } from 'lodash';
@@ -1167,8 +1168,12 @@ export const copyEnvironment = (name, baseEnvUid, collectionUid) => (dispatch, g
     const sanitizedName = sanitizeName(name);
 
     const { ipcRenderer } = window;
+
+    // strip "ephemeral" metadata
+    const variablesToCopy = (baseEnv.variables || []).filter((v) => !v.ephemeral).map(({ ephemeral, ...rest }) => rest);
+
     ipcRenderer
-      .invoke('renderer:create-environment', collection.pathname, sanitizedName, baseEnv.variables)
+      .invoke('renderer:create-environment', collection.pathname, sanitizedName, variablesToCopy)
       .then(
         dispatch(
           updateLastAction({
@@ -1249,12 +1254,21 @@ export const saveEnvironment = (variables, environmentUid, collectionUid) => (di
       return reject(new Error('Environment not found'));
     }
 
-    environment.variables = variables;
+    // Filter out ephemerals unless user explicitly chose to persist them
+    // The modal Save should remove ephemeral vars from the environment file and Redux state
+    const persisted = (variables || []).filter(v => !v.ephemeral).map(({ ephemeral, ...rest }) => rest);
+    environment.variables = persisted;
 
     const { ipcRenderer } = window;
+    const envForValidation = cloneDeep(environment);
+
     environmentSchema
       .validate(environment)
-      .then(() => ipcRenderer.invoke('renderer:save-environment', collection.pathname, environment))
+      .then(() => ipcRenderer.invoke('renderer:save-environment', collection.pathname, envForValidation))
+      .then(() => {
+        // Update Redux to reflect removal of ephemerals so watcher doesn't reattach them
+        dispatch(_saveEnvironment({ variables: persisted, environmentUid, collectionUid }));
+      })
       .then(resolve)
       .catch(reject);
   });
@@ -1311,12 +1325,17 @@ export const mergeAndPersistEnvironment =
         }
       });
 
-      environment.variables = merged;
+      // Save only non-ephemeral vars, or ephemerals explicitly persisted this run
+      const persistedNames = new Set(Object.keys(persistentEnvVariables));
+      const environmentToSave = cloneDeep(environment);
+      environmentToSave.variables = merged
+        .filter((v) => !v.ephemeral || persistedNames.has(v.name))
+        .map(({ ephemeral, ...rest }) => rest);
 
       const { ipcRenderer } = window;
       environmentSchema
-        .validate(environment)
-        .then(() => ipcRenderer.invoke('renderer:save-environment', collection.pathname, environment))
+        .validate(environmentToSave)
+        .then(() => ipcRenderer.invoke('renderer:save-environment', collection.pathname, environmentToSave))
         .then(resolve)
         .catch(reject);
     });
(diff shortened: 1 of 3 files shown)


packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -1249,12 +1258,30 @@ export const saveEnvironment = (variables, environmentUid, collectionUid) => (di
...
     const { ipcRenderer } = window;
+    const envForValidation = cloneDeep(environment);
+
     environmentSchema
       .validate(environment)
-      .then(() => ipcRenderer.invoke('renderer:save-environment', collection.pathname, environment))
+      .then(() => ipcRenderer.invoke('renderer:save-environment', collection.pathname, envForValidation))
+      .then(() => {
+        // Update Redux to reflect removal of ephemerals so watcher doesn't reattach them
+        dispatch(_saveEnvironment({ variables: persisted, environmentUid, collectionUid }));
```
sanish-bruno, line 1281: @sanjaikumar-bruno do we have to explicitly dispatch _saveEnvironment after invoking renderer:save-environment? , how does `renderer:create-environment` all work currently?

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -41,7 +41,8 @@ import {
   initRunRequestEvent,
   updateRunnerConfiguration as _updateRunnerConfiguration,
   updateActiveConnections,
-  saveRequest as _saveRequest
```
sanish-bruno, line 44: @sanjaikumar-bruno as discussed earlier, we need to have playwright test-case to verify the behaviour

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -1249,12 +1258,30 @@ export const saveEnvironment = (variables, environmentUid, collectionUid) => (di
       return reject(new Error('Environment not found'));
     }
 
-    environment.variables = variables;
+    // Filter out ephemerals unless user explicitly chose to persist them
+    // The modal Save should remove ephemeral vars from the environment file and Redux state
+    const persisted = (variables || [])
```
sanish-bruno: can we create utility function and use them within the filter and map, for better readability?

```js
const isPersistable = (variable) => 
  !variable.ephemeral || (variable.ephemeral && variable.persistedValue !== undefined);

const transformVariable = (variable) => {
  if (variable.ephemeral && variable.persistedValue !== undefined) {
    const { ephemeral, persistedValue, ...rest } = variable;
    return { ...rest, value: persistedValue };
  }
  const { ephemeral, persistedValue, ...rest } = variable;
  return rest;
};

const persisted = (variables || [])
  .filter(isPersistable)
  .map(transformVariable);

environment.variables = persisted;
```

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -1311,12 +1338,24 @@ export const mergeAndPersistEnvironment =
         }
       });
 
-      environment.variables = merged;
+      // Save only non-ephemeral vars, or ephemerals explicitly persisted this run
+      const persistedNames = new Set(Object.keys(persistentEnvVariables));
```
sanish-bruno, line 1340: can we do the similar approach here too? moving them the logic into a function and using them within filter and map, considering readbility!

packages/bruno-js/src/bru.js
```diff
@@ -125,6 +125,13 @@ class Bru {
       throw new Error('Creating a env variable without specifying a name is not allowed.');
     }
 
+    if (variableNameRegex.test(key) === false) {
+      throw new Error(
+        `Variable name: "${key}" contains invalid characters!` +
```
sanish-bruno: @sanjaikumar-bruno why can't we make the entire error message a template string, rather than using `string concat`?

packages/bruno-js/tests/setEnvVar.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    /* requestVariables */ {},
+    /* globalEnvironmentVariables */ {},
+    /* oauth2CredentialVariables */ {},
+    /* iterationDetails */ {},
+    /* collectionName */ 'Test'
+  );
+
+  test('updates envVariables and does not mark persistent when persist=false', () => {
+    const bru = makeBru();
+    bru.setEnvVar('non_persist', 'value', { persist: false });
```
sanish-bruno, line 24: create another test where we are not all passing the options object!

packages/bruno-js/tests/setEnvVar.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(bru.envVariables.non_persist).toBe('value');
+    expect(bru.persistentEnvVariables.non_persist).toBeUndefined();
+  });
+
+  test('updates envVariables and tracks persistent when persist=true (string only)', () => {
+    const bru = makeBru();
+    bru.setEnvVar('persist_me', 'value', { persist: true });
```
sanish-bruno, line 31: @sanjaikumar-bruno add test cases when there is an environment already existing with the same name but we are changing the value!

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -2275,7 +2289,14 @@ export const collectionsSlice = createSlice({
         const existingEnv = collection.environments.find((e) => e.uid === environment.uid);
 
         if (existingEnv) {
+          const prevEphemerals = (existingEnv.variables || []).filter((v) => v.ephemeral);
```
sanish-bruno, line 2292: @sanjaikumar-bruno can you explain the idea behind this logic? what is the purpose of `collectionAddEnvFileEvent`?
We are getting the `previous Ephemeral values` , what is the data we receive on environment.variables(It doesn't include any ephemeral values, are these read from the file?).

packages/bruno-app/src/utils/environments.js
```diff
@@ -0,0 +1,41 @@
...
+export const isPersistableEnvVar = (v) => {
+  return !v?.ephemeral || v?.persistedValue !== undefined;
+};
+
+/*
+ Transform a UI/runtime var into the shape we write to disk.
+ - strip ephemeral/persistedValue flags
+ - for ephemerals with persistedValue, use that as the saved value
+*/
+export const toPersistedEnvVar = (v) => {
```
sanish-bruno: rename to `transformPersistedEnvVar` ?

packages/bruno-app/src/utils/environments.js
```diff
@@ -0,0 +1,41 @@
...
+export const isPersistableEnvVarForMerge = (persistedNames) => (v) => {
+  return !v?.ephemeral || v?.persistedValue !== undefined || (v?.name && persistedNames.has(v.name));
+};
+
+/*
+ Merge transform:
+ - for ephemerals not explicitly persisted this run, write their persistedValue
+ - strip ephemeral/persistedValue flags in all cases
+*/
+export const toPersistedEnvVarForMerge = (persistedNames) => (v) => {
```
sanish-bruno: rename to `transformPersistedEnvVarForMerge` ?

sanish-bruno (changes requested):

bijin-bruno (approved):

lohit-bruno:
~ `bru.setEnvVar` with `persist` option set to `true`, adds env data with the `persistedValue` and `ephemeral` props which affects the bruno collection export/import functionality.

~ if a new env value is added using the `bru.setEnvVar` with the `persist` option set to `true`, modifying these values and trying to save them using the env configure modal deletes the values instead of updating them.

can you check this ? @sanjaikumar-bruno

sanish-bruno (approved):

Issue #3320, opened 2024-10-17
Move requests between collections


nikischin:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

[Jira](https://usebruno.atlassian.net/browse/BRU-1399)

### Describe the feature you want to add

I would like to move requests/folders between collections using the UI. Currently it is only possible to move requests/folders within a collection.

### Mockups or Images of the feature

![Mockup](https://github.com/user-attachments/assets/181194ce-a92e-46d1-b592-05600625e140)

TheRealRyanReynolds:
Thanks @nikischin! Is this capability something you use today in Postman? Would love to hear a bit more about your use case.

nikischin:
Ah no it has nothing to do with Postman. It's just some times you don't exactly know how to split your collections.

For me I wanted to create a collection per service and also a separate collection to cover all API test cases (including negative Tests). So because I needed all my sample data which is saved in local variables from one service in another one I decided to move both services in one Collection instead of two.

In future when I will have a separate collection which cover all tests including negative tests I will want to duplicate request within one collection in order to move it to the other one. I can do this via file system luckily, however it would be handy/intuitive doing it via the UI.

TheRealRyanReynolds:
Thanks @nikischin , that's really helpful. I know moving collections like that is something available in other tools, which is why I asked. 

We'll need to add this to our planning. In the spirit of transparency, I'm not sure it's something we'll get to in 2024. Right now the team is heads-down on crossing off blocking issues. Quality of life enhancements are really important to us and we totally get it, but need a little bit of time before we'll be able to shift.

nikischin:
@TheRealRyanReynolds yeah sure no worries! I just wanted it to be reported but it does not have the highest priority for me.

#2397 though would be a real game changer for me, as it would introduce many new opportunities of how to use bruno to test complex applications, so if this is something you could upvote, I would be really happy! 😃

kbublitz:
Have my upvote. I just imported some collections and wanted to reorganize them, so the only way was to move the bru files in the file explorer. Luckily Bruno has this nice file based approach, so it's not a big deal doing so.
It is not a daily thing I do, but it does happen from time to time that I want to move (or mostly duplicate) requests to another collection.
I totally get that there are other issues that are more important.

ElementalBear:
[...]

PR #5410, opened 2025-08-23
feat: move requests between collections

jayakrishnancn:
# Description

Move requests between collections.
Move requests/folders between collections using drag and drop. Currently it is only possible to move requests/folders within a collection.
while moving, if the target collection or folder has the same request/folder, then rename the new dropped folder/request and move it to the new target

resolves https://github.com/usebruno/bruno/issues/3320 

https://github.com/user-attachments/assets/776c64cc-d2b4-4628-8a9f-6f5eb792c313

### Contribution Checklist:


helloanoop:
Thank you, @jayakrishnancn!
We’ve scheduled this for review in the first half of September.

sanish-bruno:
Hi @jayakrishnancn, Thanks a lot for working on this. It works well when it comes to moving requests/folders across the collections. I was testing the PR on my local and found a major issue within the PR, , for some reason none of the `grpc requests` are not copied over. I have added sample grpc collection export json, screen recording  for your reference. 

On top of this - I guess we will have to expand the playwright test-cases to accomodate this scenario
Also i think it will better to have a drop indication border for collection too, similar to folder, currently only a line appear, give us an impression that it being dropped in between the collection

I guess we can add similar 
[Grpcbin-sample.json](https://github.com/user-attachments/files/22133110/Grpcbin-sample.json)

https://github.com/user-attachments/assets/854a4e7d-b48f-4332-bdbd-8c249e396aa4

jayakrishnancn:
Hi @sanish-bruno,  
great catch!  
This bug wasn't introduced in this PR; it's an existing issue.  
https://github.com/usebruno/bruno/blob/e0b3b1a/packages/bruno-electron/src/ipc/collection.js#L702  

However, I've included a fix for it in my latest commit. I'm reusing the "clone-folder," where the 'grpc-request' option was missing. Ideally, all requests should be cloned, in my opinion, and shouldn't be limited to just a few types. Since I'm not sure about the impact of removing the `.includes ` condition, I won't remove it. I'll just add the grpc option to it.  

If you suggest removing it, I can do that through another bugfix PR.  

Before my change (existing main branch code): 
image

Cloning a gRPC request within a collection wasn't functioning properly in main branch code: 

https://github.com/user-attachments/assets/4e0681d0-9392-467b-9c6b-4e3472d97f85

after the chagnes: 
Both cloning folders within a collection and moving folders across collections using gRPC requests are functioning as expected after my changes.

https://github.com/user-attachments/assets/fcb682c8-18e8-4e7e-a8f1-e9c0dc1bfc4f

jayakrishnancn:
> On top of this - I guess we will have to expand the playwright test-cases to accomodate this scenario

can you clarify on this? I have already added few cases in the test files
this covers drag and drop of both requests and folders.

- e2e-tests/003-collection-management-tests/001-cross-collection-drag-drop-request.spec.ts 
- e2e-tests/003-collection-management-tests/001-cross-collection-drag-drop-request.spec.ts

sanishchirayath1:
> > On top of this - I guess we will have to expand the playwright test-cases to accomodate this scenario
> 
> can you clarify on this? I have already added few cases in the test files
> this covers drag and drop of both requests and folders.
> 
> - e2e-tests/003-collection-management-tests/001-cross-collection-drag-drop-request.spec.ts 
> - e2e-tests/003-collection-management-tests/001-cross-collection-drag-drop-request.spec.ts

Scenarios where we have grpc, http, graphql  requests within a folder and all of them are moved successfully

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -120,14 +125,25 @@ const CollectionItem = ({ item, collectionUid, collectionPathname, searchText })
...
+      const { uid: draggedItemUid, sourceCollectionUid } = draggedItem;
   
       if (draggedItemUid === targetItemUid) return;
   
       const dropType = determineDropType(monitor);
       if (!dropType) return;
 
-      await dispatch(handleCollectionItemDrop({ targetItem: item, draggedItem, dropType, collectionUid }))
+      // Handle cross-collection drops differently
+      if (sourceCollectionUid !== collectionUid) {
```
sanish-bruno: @jayakrishnancn do we need the check here? we can just pass the 'sourceCollectoionUid' and handle it from `handleCollectionItemDrop` function?

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
```diff
@@ -155,11 +155,22 @@ const Collection = ({ collection, searchText }) => {
...
   
   const [{ isOver }, drop] = useDrop({
-    accept: ["collection", `collection-item-${collection.uid}`],
+    accept: ["collection", "collection-item"],
     drop: (draggedItem, monitor) => {
       const itemType = monitor.getItemType();
       if (isCollectionItem(itemType)) {
-        dispatch(handleCollectionItemDrop({ targetItem: collection, draggedItem, dropType: 'inside', collectionUid: collection.uid }))
+        // Handle cross-collection drops
+        if (draggedItem.sourceCollectionUid && draggedItem.sourceCollectionUid !== collection.uid) {
```
sanish-bruno: do we need the check here? we can just pass the 'sourceCollectoionUid' and handle it from `handleCollectionItemDrop` function if it is `undefined`, same as above?

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -92,10 +92,15 @@ const CollectionItem = ({ item, collectionUid, collectionPathname, searchText })
 
   const canItemBeDropped = ({ draggedItem, targetItem, dropType }) => {
     const { uid: targetItemUid, pathname: targetItemPathname } = targetItem;
-    const { uid: draggedItemUid, pathname: draggedItemPathname } = draggedItem;
+    const { uid: draggedItemUid, pathname: draggedItemPathname, sourceCollectionUid } = draggedItem;
 
     if (draggedItemUid === targetItemUid) return false;
 
+    // For cross-collection moves, we allow the drop
+    if (sourceCollectionUid !== collectionUid) {
```
sanish-bruno, line 100: @jayakrishnancn I guess we can add this logic inside `calculateDraggedItemNewPathname` function, were we will return a newPathname according to the collection pathname.

Don't have the clarity one when it will happen but what if `sourceCollectionId` is undefined, the function will return `true` leading into some error/failure along the line.

What are your thought on this?

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -713,9 +713,22 @@ export const moveItem =
...
 
 export const handleCollectionItemDrop =
-  ({ targetItem, draggedItem, dropType, collectionUid }) =>
+  ({ targetItem, draggedItem, dropType, collectionUid, sourceCollectionUid }) =>
   (dispatch, getState) => {
     const state = getState();
+    
+    // Handle cross-collection moves
+    if (sourceCollectionUid && sourceCollectionUid !== collectionUid) {
+      return dispatch(handleCrossCollectionItemMove({
```
sanish-bruno: @jayakrishnancn Just out of curiosity, do we need a extra `handleCrossCollectionItemMove` function?  is it possible to make little modifications to the existing logic to handle this?
For example: 
we can generate targetItemDirectory (L 735) , draggedItemDirectory (L 737) from collectionUid , sourceCollectionUid respectively. 

Have you tried that approach before?
I was thinking of hassle of maintaining two different functions that almost handle similar functionality. Let me know if you think it is better to keep them seperate.

sanish-bruno:
> Hi @sanish-bruno, great catch! This bug wasn't introduced in this PR; it's an existing issue. https://github.com/usebruno/bruno/blob/e0b3b1a/packages/bruno-electron/src/ipc/collection.js#L702
> 
> However, I've included a fix for it in my latest commit. I'm reusing the "clone-folder," where the 'grpc-request' option was missing. Ideally, all requests should be cloned, in my opinion, and shouldn't be limited to just a few types. Since I'm not sure about the impact of removing the `.includes ` condition, I won't remove it. I'll just add the grpc option to it.
> 
> If you suggest removing it, I can do that through another bugfix PR.
> 
> Before my change (existing main branch code): image
> 
> Cloning a gRPC request within a collection wasn't functioning properly in main branch code:
> 
>  before.mov 
> after the chagnes: Both cloning folders within a collection and moving folders across collections using gRPC requests are functioning as expected after my changes.
> 
>  Screen.Recording.2025-09-04.at.1.48.11.PM.mov

Yup make sense, thanks for fixing it! we will be keeping the `.includes` for now. But yes removing the check is something we can think of. not right now.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
jayakrishnancn: As far as I know, every item being dragged should have a unique ID. While being undefined is valid in isolation, sourceCollectionUid is set during drag initialization. The drag system ensures this property when dragging begins.

> I guess we can add this logic inside calculateDraggedItemNewPathname function, were we will return a newPathname according to the collection pathname.


Personally, I’m not in favor of moving this logic to the `calculateDraggedItemNewPathname` method, as its purpose is to return the pathname, not to determine if it’s a cross-collection.
Additionally, `calculateDraggedItemNewPathname` is also utilized in `handleReorderInSameLocation`. I'm unsure about the impact of modifying it.

 if you still believe moving it makes sense, we can make the changes

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
jayakrishnancn: At first, I thought of adding features like renaming through generateUniqueFilename() when duplicate request names exist in a collection/folders, since I usually work with collectionswhere this happens. However, I didn’t want to modify the current code. 

After thinking it through, you’re right—reusing the method without renaming is better. In the case of duplicates, we can show an error toast. Later, if needed, we can introduce the renaming feature in a different PR if required

jayakrishnancn:
Hi @sanish-bruno,  
Please re-review the PR.  

I have updated the UX as well. While dropping, instead of showing a line, it now displays a complete border, similar to a folder drop. Additionally, in case of a duplicate request during dropping, it will now show the default error message.  

 

https://github.com/user-attachments/assets/134aca95-060c-43d6-88ca-df3081cea40a

jayakrishnancn:
> Scenarios where we have grpc, http, graphql requests within a folder and all of them are moved successfully


For some reason, I couldn't find the gRPC option under "new request," so I skipped writing the scenario where the folder includes a gRPC request..  

image

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
jayakrishnancn: Makes sense. Removed the guard 'if' condition.

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
jayakrishnancn: Made adjustments.

sanish-bruno:
@jayakrishnancn currently gRPC is in beta, you can enable it from preferences -> beta -> enable grpc, we just hidden the entry points until it is out of beta https://docs.usebruno.com/send-requests/grpc/overview, but you can ignore the grpc-requests for now! we can write a playwright test case once it is out of beta

sanish-bruno:
@jayakrishnancn i rebased, squashed commits moved the test files to a new folder as per the main branch and raised a new PR #5525 Thanks for working on this! We will try to get it merged for upcoming release that will ideally happen in 2 weeks time.

helloanoop:
@jayakrishnancn We made some changes on top of your PR and have [merged](https://github.com/usebruno/bruno/pull/5525) it.

Thank you!

PR #5507, opened 2025-09-04
fix(common): avoid double encoding urls params. Fixes #5380.

### LINKED ISSUE #5380: Query parameters values are double encoded after adding a path parameter

### This bug is:

- [ ] annoying

### Bruno version

2.9.1

### Operating System

Microsoft Windows 11 Pro 24H2 1000.26100.197.0

### Describe the bug

After adding a path parameter to a request the query parameters values are double encoded.

E.g. After changing from `https://httpbin.io/anything/foo/?baz=hello world` to `https://httpbin.io/anything/:foo/?baz=hello world` the actual value sent for the `baz` parameter goes from `hello%20world` to `hello%2520world`.

This breaks requests expecting specific values (e.g. OData requests) and could lead to destructive unintended side effects, more or less subtle depending on how the value is used by the server.

Removing the path parameter from a broken request fixes it. Adding it back breaks it again.

Disabling the "URL Encoding" setting in the request's settings removes the behavior, but then any special characters used no longer get encoded.

```
meta {
  name: Test
  type: http
  seq: 1
}

get {
  url: https://httpbin.io/anything/:foo/?baz=hello world
  body: none
  auth: inherit
}

params:query {
  baz: hello world
}

params:path {
  foo: bar
}

settings {
  encodeUrl: true
}
```

Image

barelyhuman:
Fixes #5380.

Decode and Re-encode urls to avoid double encoding

# Description

The PR adds in relevant code fix to avoid a second re-encode

### Contribution Checklist:

### DIFF
--- a/packages/bruno-common/src/utils/url/index.ts
+++ b/packages/bruno-common/src/utils/url/index.ts
@@ -66,7 +66,7 @@ const encodeUrl = (url: string): string => {
     return url;
   }
 
-  const queryParams = parseQueryParams(queryString.join('?'), { decode: false });
+  const queryParams = parseQueryParams(queryString.join('?'), { decode: true });
   // Parse and re-encode query parameters
   const encodedQueryString = buildQueryString(queryParams, { encode: true });
 


barelyhuman:
There's 2 unit tests that dedicatedly test if the URLs are re-encoded, before I change their output, I'd like to know if they were intentional. 


https://github.com/usebruno/bruno/blob/1902329226f5936a88418f1bafc41bd7cd06713a/packages/bruno-common/src/utils/url/index.spec.ts#L100-L110

barelyhuman:
> Also, please keep existing formatting/indentation unless any code change is required to keep the PR changes minimal

My bad, missed that the formatting went through, fixed

packages/bruno-common/src/utils/url/index.ts
```diff
@@ -66,7 +66,7 @@ const encodeUrl = (url: string): string => {
     return url;
   }
 
-  const queryParams = parseQueryParams(queryString.join('?'), { decode: false });
+  const queryParams = parseQueryParams(queryString.join('?'), { decode: true });
```
bijin-bruno: We should not decode URL parameters that are already encoded. Instead, Bruno will apply encoding again, which may result in double encoding. This is the expected behavior unless encoding is explicitly turned off.

barelyhuman: discarding all changes from the pr not the reason for the issue

packages/bruno-electron/src/ipc/network/interpolate-vars.js
```diff
@@ -126,7 +137,7 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
 
   if (request?.pathParams?.length) {
     let url = request.url;
-
+    const urlSearchRaw = getRawQueryString(request.url)
```
bijin-bruno, line 134: As per line 141, I think there may be cases where we get url without protocol. That causes failures when passed to `new URL('<urlWithoutProtocol>')`. Since url validation is already handled at line 146, can we just simplify this method like below
```
const getRawQueryString = (url) => {
  const index = url.indexOf('?');
  return index !== -1 ? url.slice(index) : '';
};
```

packages/bruno-electron/src/ipc/network/interpolate-vars.js
```diff
@@ -152,7 +163,7 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
       .join('');
 
     const trailingSlash = url.pathname.endsWith('/') ? '/' : '';
-    request.url = url.origin + urlPathnameInterpolatedWithPathParams + trailingSlash + url.search;
+    request.url = url.origin + urlPathnameInterpolatedWithPathParams + trailingSlash + `${urlSearchRaw}`;
```
bijin-bruno: I think we can simplify this by using plain string concatenation instead of template literal interpolation in just one part.

packages/bruno-electron/tests/network/interpolate-vars.spec.js
```diff
@@ -54,6 +54,91 @@ describe('interpolate-vars: interpolateVars', () => {
...
+            {
+              type: 'path',
+              name: 'param',
+              value: 'foobar'
+            }
+          ]
+        };
+
+        const result = interpolateVars(request, null, null, null);
+        expect(result.url).toMatchInlineSnapshot(`"http://example.com/foobar/?search=hello world"`);
```
bijin-bruno: Any reason for using `toMatchInlineSnapshot` here? Since this is a simple string assertion, `toBe('<stringValue>')` might be clearer and more straightforward.

barelyhuman: Just prefer snapshot testing so used to typing that, no hard opinion on it's usage

packages/bruno-electron/src/ipc/network/interpolate-vars.js
barelyhuman: If the url can be without protocol then yes, makes sense, I'll add that to the test cases as well so it's easier to make sure that's tested later

barelyhuman: Wrote it with the validation since the existing code did `new URL` so assumed we'd always have a valid url at this point

barelyhuman: Oh nvm, I see what you mean, we add the protocol later, make sense.

Issue #5512, opened 2025-09-05
Base64 encoding in OAuth2 Basic header is incorrect


theholy7:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.10.0

### Operating System

macOS 15.6.1

### Describe the bug

I am migrating Postman collections to Bruno.

In my Postman collection, when using oauth2 with `Add credentials to: Basic Auth Header`, the client id and secret for a specific service I am using get correctly encoded as `base64 ( "{client_id}:{client_secret}" )`. When I decode this value, I get the original values back.

In Bruno it seems like the values are being encoded incorrectly, because I don't get the original value.

Here is a short excerpt from my secret, where I noticed the error.
Decoded Postman value: `ucgYUhbc=`
Decoded Bruno value: `ucgYUhbc%3D`

Thus the `/token` request fails, and I can't authenticate.

Luckily the service also accepts the client ID and secret in the body. I can do `Add credentials to: Request Body` in Bruno, and the values there are correctly set and encoded.

### Screenshots/Live demo link

When using Basic Auth Header:

<img width="281" height="67" alt="Image" src="https://github.com/user-attachments/assets/62ecd80a-1bf9-4b15-a1ba-a2fa2bcaee01" />

When using request body:

<img width="328" height="63" alt="Image" src="https://github.com/user-attachments/assets/e1a82b46-8f86-4c85-ac0e-085158b3c4e5" />

pietrygamat:
Broken in #5348? @naman-bruno , why do `encodeURIComponent` before base64?

theholy7:
The basic authentication standard (which I think applies here, let me know if wrong) says:

> To receive authorization, the client sends the userid and password,
   separated by a single colon (":") character, within a base64 [[7](https://datatracker.ietf.org/doc/html/rfc2617#ref-7)]
   encoded string in the credentials.

[Reference](https://datatracker.ietf.org/doc/html/rfc2617#section-2)

I think that the OAuth2 standard obeys this.

pietrygamat:
[...]

Issue #4668, opened 2025-05-14
OAuth 2.0, client credential fail because client_id value is added in body


zhang928160849:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

V2.2.0

### Operating System

macos

### Describe the bug

![Image](https://github.com/user-attachments/assets/1c7a7e4d-f9e6-4364-91c8-7dd54e0e063b)

client_id id is in body which is incorrect, remove it and keep grant_type only that's the right way to fetch token.
currently, it would be treated as 400 bad request

![Image](https://github.com/user-attachments/assets/cce9f4ff-420d-4312-b208-e6021b9dbad4)
pleas remove it from body.

### Screenshots/Live demo link

![Image](https://github.com/user-attachments/assets/e016c465-bdde-4383-8a19-af2549fdf6a6)

![Image](https://github.com/user-attachments/assets/8a10646e-085d-4939-b9b6-36af216abcd3)

anusree-bruno:
@zhang928160849 Can you check in your oauth2 settings you have selected Basic Auth Header? I'll check the RFC and see if we can default it to Basic Auth Header for Client Credentials Authentication. Let me know if this resolves the issue

![Image](https://github.com/user-attachments/assets/09b6b8de-41ce-4af8-a8bb-d715f535f759)

anusree-bruno:
similar https://github.com/usebruno/bruno/issues/4388

bsteimer:
Would it be possible to make client_id a non required parameter and only send it if set? Makes bruno unusable for me at the moment.

helloanoop:
We've dedicated time to look into this in our current sprint.

sreelakshmi-bruno:
Hi @zhang928160849, the issue where the client ID was being sent in the request body instead of the headers (when specified to be added to headers) has been fixed and included in the September release. Closing this ticket as the issue is resolved, please feel free to reopen it if you notice any recurrence.

PR #5521, opened 2025-09-08
fix(oauth2): always send client id if set

shyim:
# Description

image

I am using Password Grant using OAuth. Since the last Update this has been broken as when the credential placement is header the client_id is not send anymore.

We require a client_id for password grant, so I changed it to send it always when it's set.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/utils/oauth2.js
+++ b/packages/bruno-electron/src/utils/oauth2.js
@@ -259,10 +259,10 @@ const getOAuth2TokenUsingAuthorizationCode = async ({ request, collectionUid, fo
     code: authorizationCode,
     redirect_uri: callbackUrl,
   };
-  if (credentialsPlacement !== "basic_auth_header") {
+  if (clientId && clientId.trim() !== '') {
     data.client_id = clientId;
   }
-  if (clientSecret && clientSecret.trim() !== '' && credentialsPlacement !== "basic_auth_header") {
+  if (clientSecret && clientSecret.trim() !== '') {
     data.client_secret = clientSecret;
   }
   if (pkce) {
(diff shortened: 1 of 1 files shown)


packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -259,10 +259,10 @@ const getOAuth2TokenUsingAuthorizationCode = async ({ request, collectionUid, fo
     code: authorizationCode,
     redirect_uri: callbackUrl,
   };
-  if (credentialsPlacement !== "basic_auth_header") {
+  if (clientId && clientId.trim() !== '') {
     data.client_id = clientId;
   }
```
pietrygamat, line 264: This is not the correct approach. As per  [RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3) the `clientId` is
> REQUIRED, if the client is not authenticating with the
> authorization server as described in [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1).

which naturally also means it is NOT required when the client DOES use authentication. 

This change may fix some usecases (publc clients with no secret), but will break others (#4668). I guess the condition for including client_id in the body should depend on the existence of client secret, because only if the secret is missing, the `client_id` body parameter is REQUIRED, and when it's present, it's up to the selected `credentialsPlacement`.

pietrygamat (changes requested):

packages/bruno-electron/src/utils/oauth2.js
shyim: Would it be fine when I do that only for password credentials? Does where I had the error. I just "unified" all to the same :)

pietrygamat: @shyim to my understanding, it's the same for all grant types.
Does your usecase require sending client_id in both request body AND authentication header? If so, this reveals another corner case that needs to be covered, but most likely you should be fine with just switching credential placement to Request body, no?

sreelakshmi-bruno: Hi @shyim, since the proposed approach isn’t RFC-compliant, we won’t be able to proceed with your PR. Could you please check out this [comment](https://github.com/usebruno/bruno/issues/5512#issuecomment-3454811990) and see if the suggested workaround addresses your use case?

Thanks @pietrygamat, appreciate your contribution!

shyim: Hey,

sorry was on long vacation. I need this just for the password credentials grant. Is that fine there and I would revert this?

PR #5536, opened 2025-09-09
feat: Add sample collection and first launch hook.

sanjaikumar-bruno:


packages/bruno-app/src/utils/common/platform.js
```diff
@@ -45,3 +45,33 @@ export const getAppInstallDate = () => {
   const date = new Date(dateString);
   return date;
 };
+
+export const isFirstLaunch = () => {
```
sanish-bruno: @sanjaikumar-bruno as discussed before we don't have to use localstorage to check if bruno has launched before, what is preventing us from using preferences.json

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,120 @@
...
+      console.error('Failed to handle first launch:', error);
+      // Still mark as launched to prevent retry on next startup
+      preferencesUtil.markAsLaunched();
+    }
+  }
+
+  /**
+   * Get the default location for collections
+   */
+  async getDefaultCollectionLocation() {
```
sanish-bruno: @sanjaikumar-bruno we will have to check if  we really need the if conditions. It is highly unlikely that electron will return a path that doesn't exist. It would rather throw an error it the app.getPath cannot find the concerned path. will you be able to check the exact behaviour on how app.getPath works and  write the function according to it.

But with this logic, desktop will never get used if app.getPath throws an error.

packages/bruno-electron/src/utils/collection-import.js
```diff
@@ -0,0 +1,116 @@
+const fs = require('fs');
```
sanish-bruno: @sanjaikumar-bruno why do we have a new file? what is preventing us from using existing functions

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /**
+   * Import the sample collection
+   */
+  async importSampleCollection(collectionLocation) {
+    try {
+      // Load the sample collection from assets
+      const possiblePaths = [
+        path.join(__dirname, '../../../bruno-app/src/assets/sample-collection.json'),
```
sanish-bruno: @sanjaikumar-bruno why aren't we able to definitively tell where the sample collection resides? we can host them in bruno-electron itself right. Did you face any issues with the approach?

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,120 @@
...
+const fsPromises = require('fs').promises;
+const path = require('path');
+const { app } = require('electron');
+const { preferencesUtil } = require('../store/preferences');
+const { importCollection } = require('../utils/collection-import');
+
+/**
+ * Handles first launch logic including sample collection import
+ */
+class FirstLaunchHandler {
```
sanish-bruno: can you check if we need this class, How import collection is handled in main? How when import collection called how it updated lasOpenedCollections from preferences.json? If import function internally handles all of those. then probably we don't need to have seperate logic.

sanish-bruno (changes requested):

packages/bruno-electron/src/app/first-launch.js
sanjaikumar-bruno: The `renderer:import-collection` IPC handler updates the lastOpenedCollections. We are centralizing the collection creation logic for reuse within `FirstLaunchHandler`. As a result, we will modify `renderer:import-collection` to utilize this shared function, as confirmed by Anoop.

packages/bruno-electron/src/utils/collection-import.js
sanjaikumar-bruno: The existing collection creation logic is inside the `renderer:import-collection` IPC handler, which can only be called from the renderer process. Since we need to create collections from the main process (during first launch), we have two options:

1. **Complex mocking**: Mock the IPC communication to make the main process think it's the renderer
2. **Extract shared logic**: Move the core collection creation logic to a utility that both can use

We chose option 2 because it's cleaner and avoids complex mocking. The `collection-import.js` utility contains the same logic as the IPC handler, just extracted into a reusable function.

This way, both the IPC handler and the first-launch logic can use the same collection creation code without duplication or complex workarounds.

packages/bruno-electron/src/app/first-launch.js
sanjaikumar-bruno: We have a similar centralization of the collection creation logic inside the CLI,

https://github.com/usebruno/bruno/blob/2864ddaa728c5fce045182a6fbac853db97d6c85/packages/bruno-cli/src/utils/collection.js#L403

packages/bruno-electron/src/app/first-launch.js
sanjaikumar-bruno: Updated the logic to catch the error.

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,108 @@
...
+   */
+  async handleFirstLaunch() {
+    try {
+      // Check if this is the first launch
+      if (!preferencesUtil.isFirstLaunch()) {
+        console.log('FirstLaunchHandler: Not first launch, exiting');
+        return;
+      }
+
+      // Get the default collection location
```
sanish-bruno: @sanjaikumar-bruno can remove un necessary comments whereever the function name itself communicates well

tests/onboarding/first-launch.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { test, expect } from '../../playwright';
```
sanish-bruno: @sanjaikumar-bruno i guess we will have to rename the file to something else probably 'default-collection'

packages/bruno-electron/src/utils/collection-import.js
```diff
@@ -0,0 +1,117 @@
...
+const { sanitizeName, createDirectory, writeFile, safeWriteFileSync, getCollectionStats } = require('./filesystem');
+const { generateUidBasedOnHash, stringifyJson } = require('./common');
+const { stringifyRequestViaWorker, stringifyCollection, stringifyEnvironment, stringifyFolder } = require('@usebruno/filestore');
+
+/**
+ * Import a collection - shared logic used by both IPC handler and first-launch
+ */
+async function importCollection(collection, collectionLocation, mainWindow, lastOpenedCollections, uniqueFolderName = null) {
+  // Use provided unique folder name or generate one with timestamp
+  let folderName = uniqueFolderName ? sanitizeName(uniqueFolderName) : sanitizeName(`collectionName ${Date.now()}`);
```
sanish-bruno: use the collection.name itself as before, 
```js
let folderName = uniqueFolderName ? sanitizeName(uniqueFolderName) : sanitizeName(collection.name)`)
```

sanish-bruno:
@sanjaikumar-bruno I guess, so many playwright test cases are failing cause of  default collection changes, we will have to update them

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,104 @@
+const fs = require('fs');
```
helloanoop: `first-launch` is an odd name for a file. Please rename as `onboarding.js`

You don't need to expose a class to be instantiated. A function 

```javascript
module.exports = onboardUser
```
should be fine

The rest of the functions can be local functions.

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  async handleFirstLaunch() {
+    try {
+      if (!preferencesUtil.isFirstLaunch()) {
+        console.log('FirstLaunchHandler: Not first launch, exiting');
+        return;
+      }
+
+      const collections = await this.lastOpenedCollections.getAll();
```
helloanoop: logic seems redundant

packages/bruno-electron/src/index.js
```diff
@@ -178,6 +179,15 @@ app.on('ready', async () => {
         return safeParseJSON(safeStringifyJSON(_));
       })]);
     }
+    
+    // Handle first launch logic
```
helloanoop: let the onboardUser() call handle the try catch.

```javascript
// handle onboarding
onboardUser();
```

packages/bruno-electron/src/index.js
```diff
@@ -35,6 +35,7 @@ const registerGlobalEnvironmentsIpc = require('./ipc/global-environments');
 const { safeParseJSON, safeStringifyJSON } = require('./utils/common');
 const { getDomainsWithCookies } = require('./utils/cookies');
 const { cookiesStore } = require('./store/cookies');
+const FirstLaunchHandler = require('./app/first-launch');
```
helloanoop: Please have this as 
```javascript
const onboardUser  = require('./app/onboarding'); 
```

packages/bruno-electron/src/store/preferences.js
```diff
@@ -83,6 +86,9 @@ const preferencesSchema = Yup.object().shape({
   beta: Yup.object({
     grpc: Yup.boolean(),
     nodevm: Yup.boolean()
+  }),
```
helloanoop, line 89: schema looks good, no further changes needed

tests/onboarding/first-launch.spec.ts
helloanoop: Yes, please rename this as `sample-collection.spec.js`

tests/onboarding/first-launch.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+test.describe('First Launch Onboarding', () => {
```
helloanoop: Should be just
```javascript
test.describe('Onboarding', () => {
```

tests/onboarding/first-launch.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    const newApp = await reuseOrLaunchElectronApp({ userDataPath });
+    const newPage = await newApp.firstWindow();
+    await newPage.waitForTimeout(1000);
+  
+    // Sample collection should not appear since it's no longer first launch
+    const sampleCollections = newPage.locator("#sidebar-collection-name").filter({ hasText: 'Sample API Collection' });
+    await expect(sampleCollections).toHaveCount(0);
+  });
+
+  test('should not create sample collection if user has already opened a collection (old users who already have a collection opened)', async ({ pageWithUserData: page }) => {
```
helloanoop: Good job on writing comprehensive tests

packages/bruno-electron/src/app/first-launch.js
helloanoop: I realised after seeing the tests that this logic is valid for people with existing collections. I would explain it with a comment

```javascript
async onboardUser() {
  try {
    if (!preferencesUtil.isFirstLaunch()) {
      return;
    }

    // Onboarding was added later;
    // if a collection already exists, user is old → skip onboarding
    const collections = await this.lastOpenedCollections.getAll();
    if (collections.length > 0) {
      preferencesUtil.markAsLaunched();
      return;
    }

    const collectionLocation = await this.getDefaultCollectionLocation();
    await this.importSampleCollection(collectionLocation);
    preferencesUtil.markAsLaunched();
  } catch (error) {
    console.error('Failed to handle first launch:', error);
    // Still mark as launched to prevent retry on next startup
    preferencesUtil.markAsLaunched();
  }
}
```

packages/bruno-electron/src/app/first-launch.js
```diff
@@ -0,0 +1,104 @@
...
+        preferencesUtil.markAsLaunched();
+        return;
+      }
+
+      const collectionLocation = await this.getDefaultCollectionLocation();
+      await this.importSampleCollection(collectionLocation);
+      preferencesUtil.markAsLaunched();
+    } catch (error) {
+      console.error('Failed to handle first launch:', error);
+      // Still mark as launched to prevent retry on next startup
```
helloanoop: appreciate the comment here explaining why we still go ahead and `markAsLaunched()`

bijin-bruno (approved):

PR #5573, opened 2025-09-16
Feature: support import paths for gRPC

### LINKED ISSUE #5520: Multiple proto Import Paths for gRPC

### This feature

- [x] blocks me from using Bruno
- [ ] would improve my quality of life in Bruno
- [ ] is something I've never seen an API client do before

### **Problem Description**

When working with gRPC APIs, `.proto` files often have dependencies on files located in different directories. The standard `protoc` command-line tool handles this with the `-I` or `--proto_path` flag, which allows for specifying multiple import paths.

Currently, Bruno does not appear to support this functionality. This makes it **impossible** to import `.proto` files with dependencies that are outside of the current directory. This effectively prevents users from working with gRPC projects where `.proto` files are not located in a single, centralized folder. Other popular tools like **Postman already support this feature, which many developers consider a standard, basic functionality.**

### **Proposed Solution**

I propose expanding Bruno's gRPC functionality to allow users to define **multiple import paths for `.proto` files**. These paths should be configurable either within the settings of a gRPC request or at the project/collection level.

### **Benefits**

  - **Improved Compatibility**: Bruno would better support the standard `protoc` conventions, making integration with existing projects much smoother.
  - **Greater Flexibility**: Users would be able to manage complex project structures with distributed `.proto` files more effectively, without having to create redundant copies.
  - **Enhanced User Experience**: It would remove the need to manually copy all `.proto` files into a single directory, which can lead to redundant files and errors.

### **Example Implementation**

The user interface could provide a simple way to add a list of directory paths. The configuration could look like this in the `bruno.json` file:

```json
{
  "grpc": {
    "protoPaths": [
      "/path/to/my/protos",
      "/path/to/my/common/protos",
      "/path/to/my/vendor/protos"
    ]
  }
}
```

_No response_

sanish-bruno:
fixes #5520
[Jira](https://usebruno.atlassian.net/browse/BRU-1842)
# Description
Currently, Bruno fails to resolve import paths from another protofile unless they are absolute. This requires users to place all files in the same folder as the root protofile, which is inconvenient.

- Added support for import paths
- updated protofile management ui within colection settings
- updated query url dropdown ui/ux
- Protofiles , import paths added at request level gets automatically added to collection settings
- Added playwright test cases for protofile management
- Added unit test cases for migration script

https://github.com/user-attachments/assets/1b3dae70-b3b0-47fc-9845-9b55472c261d

### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/CollectionSettings/Presets/index.js
+++ b/packages/bruno-app/src/components/CollectionSettings/Presets/index.js
@@ -5,11 +5,9 @@ import StyledWrapper from './StyledWrapper';
 import toast from 'react-hot-toast';
 import { updateBrunoConfig } from 'providers/ReduxStore/slices/collections/actions';
 import cloneDeep from 'lodash/cloneDeep';
-import { useBetaFeature, BETA_FEATURES } from 'utils/beta-features';
 
 const PresetsSettings = ({ collection }) => {
   const dispatch = useDispatch();
-  const isGrpcEnabled = useBetaFeature(BETA_FEATURES.GRPC);
   const {
     brunoConfig: { presets: presets = {} }
   } = collection;
@@ -17,15 +15,10 @@ const PresetsSettings = ({ collection }) => {
   const formik = useFormik({
     enableReinitialize: true,
     initialValues: {
-      requestType: presets.requestType === 'grpc' && !isGrpcEnabled ? 'http' : presets.requestType || 'http',
+      requestType: presets.requestType || 'http',
       requestUrl: presets.requestUrl || ''
     },
     onSubmit: (newPresets) => {
-      // If gRPC is disabled but the preset is set to grpc, change it to http
-      if (!isGrpcEnabled && newPresets.requestType === 'grpc') {
-        newPresets.requestType = 'http';
-      }
-
       const brunoConfig = cloneDeep(collection.brunoConfig);
       brunoConfig.presets = newPresets;
       dispatch(updateBrunoConfig(brunoConfig, collection.uid));
@@ -70,22 +63,18 @@ const PresetsSettings = ({ collection }) => {
               GraphQL
             </label>
 
-            {isGrpcEnabled && (
-              <>
-                <input
-                  id="grpc"
-                  className="ml-4 cursor-pointer"
-                  type="radio"
-                  name="requestType"
-                  onChange={formik.handleChange}
-                  value="grpc"
-                  checked={formik.values.requestType === 'grpc'}
-                />
-                <label htmlFor="grpc" className="ml-1 cursor-pointer select-none">
-                  gRPC
-                </label>
-              </>
-            )}
+            <input
+              id="grpc"
+              className="ml-4 cursor-pointer"
+              type="radio"
+              name="requestType"
+              onChange={formik.handleChange}
+              value="grpc"
+              checked={formik.values.requestType === 'grpc'}
+            />
+            <label htmlFor="grpc" className="ml-1 cursor-pointer select-none">
+              gRPC
+            </label>
           </div>
         </div>
         <div className="mb-3 flex items-center">
--- a/packages/bruno-app/src/components/CollectionSettings/index.js
+++ b/packages/bruno-app/src/components/CollectionSettings/index.js
@@ -18,11 +18,9 @@ import StyledWrapper from './StyledWrapper';
 import Vars from './Vars/index';
 import StatusDot from 'components/StatusDot';
 import Overview from './Overview/index';
-import { useBetaFeature, BETA_FEATURES } from 'utils/beta-features';
 
 const CollectionSettings = ({ collection }) => {
   const dispatch = useDispatch();
-  const isGrpcEnabled = useBetaFeature(BETA_FEATURES.GRPC);
   const tab = collection.settingsSelectedTab;
   const setTab = (tab) => {
     dispatch(
@@ -174,12 +172,10 @@ const CollectionSettings = ({ collection }) => {
           Client Certificates
           {clientCertConfig.length > 0 && <StatusDot />}
         </div>
-        {isGrpcEnabled && (
-          <div className={getTabClassname('protobuf')} role="tab" onClick={() => setTab('protobuf')}>
-            Protobuf
-            {protobufConfig.protoFiles && protobufConfig.protoFiles.length > 0 && <StatusDot />}
-          </div>
-        )}
+        <div className={getTabClassname('protobuf')} role="tab" onClick={() => setTab('protobuf')}>
+          Protobuf
+          {protobufConfig.protoFiles && protobufConfig.protoFiles.length > 0 && <StatusDot />}
+        </div>
       </div>
       <section className="mt-4 h-full overflow-auto">{getTabPanel(tab)}</section>
     </StyledWrapper>
--- a/packages/bruno-app/src/components/Preferences/Beta/index.js
+++ b/packages/bruno-app/src/components/Preferences/Beta/index.js
@@ -11,11 +11,6 @@ import get from 'lodash/get';
 // Beta features configuration
 const BETA_FEATURES = [
   {
-    id: 'grpc',
-    label: 'gRPC Support',
-    description: 'Enable gRPC request support for making gRPC calls to services'
-  },
-  {
     id: 'nodevm',
     label: 'Node VM Runtime',
     description: 'Enable Node VM runtime for JavaScript execution in Developer Mode'
@@ -103,16 +98,6 @@ const Beta = ({ close }) => {
                 <label className="block ml-2 select-none font-medium" htmlFor={feature.id}>
                   {feature.label}
                 </label>
-                {feature.id === 'grpc' && (
-                  <a 
-                    href="https://github.com/usebruno/bruno/discussions/5447" 
-                    target="_blank" 
-                    rel="noopener noreferrer" 
-                    className="ml-2 text-xs text-blue-500 hover:text-blue-600 underline"
-                  >
-                    Share feedback
-                  </a>
-                )}
               </div>
               <div className="beta-feature-description ml-6 text-xs text-gray-500 dark:text-gray-400">
                 {feature.description}
--- a/packages/bruno-app/src/components/ResponsePane/ResponseSave/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/ResponseSave/index.js
@@ -11,7 +11,7 @@ const ResponseSave = ({ item }) => {
   const saveResponseToFile = () => {
     return new Promise((resolve, reject) => {
       ipcRenderer
-        .invoke('renderer:save-response-to-file', response, item?.requestSent?.url)
+        .invoke('renderer:save-response-to-file', response, item?.requestSent?.url, item.pathname)
         .then(resolve)
         .catch((err) => {
           toast.error(get(err, 'error.message') || 'Something went wrong!');
--- a/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
+++ b/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
@@ -21,14 +21,12 @@ import Help from 'components/Help';
 import StyledWrapper from './StyledWrapper';
 import SingleLineEditor from 'components/SingleLineEditor/index';
 import { useTheme } from 'styled-components';
-import { useBetaFeature, BETA_FEATURES } from 'utils/beta-features';
 
 const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
   const dispatch = useDispatch();
   const inputRef = useRef();
 
   const storedTheme = useTheme();
-  const isGrpcEnabled = useBetaFeature(BETA_FEATURES.GRPC);
 
   const collection = useSelector((state) => state.collections.collections?.find((c) => c.uid === collectionUid));
   const {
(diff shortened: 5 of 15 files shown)


helloanoop:
@sanish-bruno When I select the import path as the collection itself, the path is being written as the absolute path. I believe in this case, it should be just store as "." ?

image

helloanoop:
@sanish-bruno Can we include in this PR to rename `gRPC` tab as `Protobuf`
And the same changes to bruno.json while keeping backwards compatible to pick up the config from older `bruno.json`

helloanoop:
Alternate UX approach to think about

image

packages/bruno-app/src/components/CollectionSettings/Protobuf/index.js
```diff
@@ -0,0 +1,484 @@
...
+  IconFolder,
+  IconPlus
+} from '@tabler/icons';
+import { getRelativePath, getBasename, getDirPath } from 'utils/common/path';
+import { Tooltip } from 'react-tooltip';
+import { existsSync, resolvePath, browseDirectory, isDirectory } from '../../../utils/filesystem';
+
+const ProtobufSettings = ({ collection }) => {
+  const dispatch = useDispatch();
+  // Merge grpc and protobuf configs for backward compatibility
```
helloanoop: @sanish-bruno This component should not know anything about backwards compatibility. This should be done at the electron layer.

packages/bruno-app/src/components/Dropdown/index.js
```diff
@@ -2,7 +2,12 @@ import React from 'react';
 import Tippy from '@tippyjs/react';
 import StyledWrapper from './StyledWrapper';
 
-const Dropdown = ({ icon, children, onCreate, placement, transparent, ...props }) => {
+const Dropdown = ({ icon, children, onCreate, placement, transparent, visible, ...props }) => {
+  // When in controlled mode (visible prop is provided), don't use trigger prop
+  const tippyProps = visible !== undefined 
+    ? { ...props, visible, interactive: true, appendTo: "parent" }
+    : { ...props, trigger: "click", interactive: true, appendTo: "parent" };
```
bijin-bruno: Please use single quotes for string values in js files

packages/bruno-app/src/components/RequestPane/GrpcQueryUrl/Tabs/ProtoFilesTab/index.js
```diff
@@ -0,0 +1,116 @@
...
+  setShowProtoDropdown
+}) => {
+  return (
+    <>
+      {collectionProtoFiles && collectionProtoFiles.length > 0 && (
+        <div className="px-3 py-2">
+          <div className="flex items-center justify-between mb-1">
+            <div className="text-xs text-neutral-500">From Collection Settings</div>
+            <button 
+              onClick={(e) => {
```
bijin-bruno: Extract inline onClick into a handler for reuse and separation of concerns

The current inline click logic mixes templating with behavior. Let’s move it into a named handler so the template stays declarative and the logic is reusable/testable.

packages/bruno-app/src/utils/common/path.js
```diff
@@ -9,21 +9,29 @@ const isWindowsOS = () => {
...
-  const parts = filePath.split(path.sep);
-  return parts[parts.length - 1];
+
+  const resolvedPath = brunoPath.resolve(fromPath, relativePath);
+  const basename = brunoPath.basename(resolvedPath);
+
+  return basename;
 };
 
 const getDirPath = (filePath) => {
```
bijin-bruno: Probably a similar fix as that of getBasename is required here?

packages/bruno-app/src/utils/network/index.js
```diff
@@ -139,20 +142,31 @@ export const endGrpcStream = async (requestId) => {
...
+
+    if(collection) {
+      const config = cloneDeep(collection.brunoConfig);
+
+      if(config.protobuf && config.protobuf.importPaths) {
+        // Use Promise.all to wait for all resolvePath calls to complete
+        const enabledImportPaths = config.protobuf.importPaths.filter(importPath => importPath.enabled);
+        importPaths = await Promise.all(
+          enabledImportPaths.map(async (importPath) => {
+            return await resolvePath(importPath.path, collection.pathname);
```
bijin-bruno: await might not be required here since we await Promise.all(?

packages/bruno-electron/src/ipc/filesystem.js
```diff
@@ -48,6 +49,15 @@ const registerFilesystemIpc = (mainWindow) => {
       return relativePath;
     }
   });
+
+  // is directory
+  ipcMain.handle('renderer:is-directory', async (_, pathname) => {
+    try {
+      return isDirectory(pathname);
```
bijin-bruno: Please remove obvious comments

try catch is not required as isDirectory doesn't throw any errors

tests/protobuf/collection/bruno.json
bijin-bruno, line 1: rename Grpcbin to collection for consistency and readability

packages/bruno-app/src/components/Dropdown/index.js
sanish-bruno: done

packages/bruno-app/src/utils/common/path.js
sanish-bruno: removed the function as it is no longer used

packages/bruno-app/src/utils/filesystem.js
```diff
@@ -30,4 +30,27 @@ export const resolvePath = async (relativePath, basePath) => {
...
+    return [];
+  }
+};
+
+/**
+ * Check if a path is a directory
+ * @param {string} dirPath - The directory path to check
+ * @returns {Promise<boolean>} - True if path is a directory, false otherwise
+ */
+export const isDirectory = async dirPath => {
```
helloanoop: I feel this function shouldn’t fail silently. Returning false on error can mislead callers into thinking the path isn’t a directory when the check failed.

packages/bruno-app/src/utils/filesystem.js
```diff
@@ -30,4 +30,27 @@ export const resolvePath = async (relativePath, basePath) => {
...
-}; 
\ No newline at end of file
+};
+
+export const browseDirectory = async pathname => {
+  try {
+    return await window?.ipcRenderer?.invoke('renderer:browse-directory', pathname);
+  } catch (error) {
+    console.error('Error browsing directory:', error);
+    return [];
```
helloanoop: Is there a reason we want to make this code fail silently ? 
This code can cause issues similar to https://github.com/usebruno/bruno/pull/5573/files#r2373268402

packages/bruno-app/src/utils/common/path.js
```diff
@@ -9,32 +9,34 @@ const isWindowsOS = () => {
 
 const brunoPath = isWindowsOS() ? path.win32 : path.posix;
 
-const getRelativePath = (absolutePath, collectionPath) => {
+const getRelativePath = (toPath, fromPath) => {
```
helloanoop: Can we maintain the same order as the [path lib](https://nodejs.org/api/path.html#pathrelativefrom-to)

```
const getRelativePath = (fromPath, toPath) => {
```

Also given that the arg names are a bit confusing (imo), I'd suggest adding a jsdoc for people to quickly grasp what is happening in this function
```
/**
 * Get a relative path from one location to another.
 *
 * This function attempts to compute the relative path between two given paths
 *
 * @param {string} fromPath - The starting path.
 * @param {string} toPath - The target path.
 * @returns {string} The relative path from `fromPath` to `toPath`, `"."` if both are the same,
 *                   or `toPath` if resolution fails.
 *
 * @example
 * // Assuming current dir: /users/john/projects
 * getRelativePath('/users/john/projects', '/users/john/projects/app');
 * // → "app"
 *
 * @example
 * getRelativePath('/users/john/projects', '/users/john/projects');
 * // → "."
 *
 * @example
 * getRelativePath('/users/john/projects', '/users/john/docs/readme.md');
 * // → "../docs/readme.md"
 */
```

packages/bruno-app/src/components/RequestPane/GrpcQueryUrl/Tabs/TabNavigation/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+const TabNavigation = ({ activeTab, onTabChange, collectionProtoFiles, collectionImportPaths }) => {
```
helloanoop, line 4: We should use `StyledWrapper` approach for the styling the colors similar to how we have in rest of the app.

packages/bruno-app/src/components/RequestPane/GrpcQueryUrl/GrpcurlModal/index.js
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      handleCancel={onClose}
+      title={(
+        <div className="flex items-center gap-2">
+          <span>Generate gRPCurl Command</span>
+          <span className="px-2 py-0.5 text-xs font-medium bg-indigo-100 text-indigo-700 dark:bg-indigo-900 dark:text-indigo-300 rounded">BETA</span>
```
helloanoop: We should use `StyledWrapper` to apply colors.
This is keeping inline with the approach followed in rest of the app and allows us in the future to support more themes by adding new theme files with different colors.

packages/bruno-app/src/components/RequestPane/GrpcQueryUrl/index.js
```diff
@@ -168,10 +115,32 @@ const GrpcQueryUrl = ({ item, collection, handleRun }) => {
     fetchCollectionProtoFilesExistence();
   }, [fileExists]);
 
+  useEffect(() => {
```
helloanoop: Not sure if this logic should be here in the front-end. Can we have electron layer compute whether a file is present or not and make it available as an additional property -- ex: `exists` -- in the brunoConfig when it is sent.

We just need to ensure that while we save the config back, we should strip the `exists` property before we save it.

packages/bruno-app/src/components/RequestPane/GrpcQueryUrl/Tabs/TabNavigation/index.js
sid-bruno: Still needs to be addressed

sid-bruno:
Other than the Color token convention, the rest looks good to me. 

I'm a little concerned about the Windows path resolutions, can we make sure we run a manual test if possible on the window machine?

sid-bruno (dismissed):

sid-bruno:
`packages/bruno-app/src/utils/common/path.js` 

We seem to be using path.win32 and path.posix to manage the base path used for the rest of the relative paths, can we try it out on windows and if it doesn't work as expected, normalise the paths using a posixify util that normalises the path into posix format and use that instead. just a safety net, not a mandatory change 

```js

/** @param {string} str */
function posixify(str) {
	return str.replace(/\\/g, '/');
}
```

tests/protobuf/manage-protofile.spec.ts
```diff
@@ -0,0 +1,141 @@
+import { test, expect } from '../../playwright';
+
+test.describe('manage protofile', () => {
+  test('protofiles, import paths from bruno.json are visible in the protobuf settings', async ({ pageWithUserData: page }) => {
+    await page.locator('#sidebar-collection-name').filter({ hasText: 'Grpcbin' }).click();
+
+    await page.getByRole('tab', { name: 'Protobuf' }).click();
+
+    // Wait for protobuf settings to load
+    await page.locator('[data-test-id="protobuf-proto-files-section"]').waitFor();
```
bijin-bruno: attribute name should be `data-testid` and then we could use the playwright locator method `getbyTestId()` which makes the code cleaner

tests/protobuf/manage-protofile.spec.ts
```diff
@@ -0,0 +1,141 @@
+import { test, expect } from '../../playwright';
+
+test.describe('manage protofile', () => {
```
bijin-bruno, line 4: let's add an afterAll hook to use closeAllCollections step from utils. This provides a fresh app for the next test suite

bijin-bruno (changes requested):

sanish-bruno:
> `packages/bruno-app/src/utils/common/path.js`
> 
> We seem to be using path.win32 and path.posix to manage the base path used for the rest of the relative paths, can we try it out on windows and if it doesn't work as expected, normalise the paths using a posixify util that normalises the path into posix format and use that instead. just a safety net, not a mandatory change
> 
> ```js
> /** @param {string} str */
> function posixify(str) {
> 	return str.replace(/\\/g, '/');
> }
> ```

@sid-bruno Yes, tested it on windows, working without any problem but paths/files added form windows will have different slashes, will add the fix.
image (1)

tests/protobuf/manage-protofile.spec.ts
sanish-bruno: i couldn't find `closeAllCollections` step from utils. are using is some where? @bijin-bruno can you clarify

sanish-bruno:
@sid-bruno the same is happening for client certificates too. client certificates stores the relative path within the bruno json file, stored relative path is windows one. we will have to do the same there too.

sid-bruno:
LGTM

bijin-bruno (approved):

PR #5591, opened 2025-09-18
fix: filter out internal content-type headers for no body requests in axiosinstance

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-1784)
Fixes: #1693

**The Problem:**
When you select "No Body" for a POST/PUT/PATCH request in Bruno, the timeline was showing `content-type: false` in the headers section, even though this header wasn't actually being sent to the server. As Bruno internally sets `content-type: false` to tell Axios not to send a Content-Type header when there's no body. But this internal flag was being captured in the timeline before Axios had a chance to remove it.

**The Fix:**
We filtered out the `content-type: false` header from the timeline in two places:
1. The detailed step-by-step timeline (axios interceptor)
2. The main request timeline (collection timeline)

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -616,10 +616,11 @@ const registerNetworkIpc = (mainWindow) => {
 
       // Remove false Content-Type header (used to stop axios from auto-setting it); no Content-Type was actually set or sent.
       const headersSent = { ...request.headers };
-      const contentTypeKey = Object.keys(headersSent).find((key) => key.toLowerCase() === 'content-type');
-      if (contentTypeKey && headersSent[contentTypeKey] === false) {
-        delete headersSent[contentTypeKey];
-      }
+      Object.keys(headersSent).forEach((key) => {
+        if (key.toLowerCase() === 'content-type' && headersSent[key] === false) {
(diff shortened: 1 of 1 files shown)


sanish-bruno:
I guess we will have to check this on windows as told within the issue
https://github.com/usebruno/bruno/issues/1693#issuecomment-3322960094

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -610,10 +610,17 @@ const registerNetworkIpc = (mainWindow) => {
       );
 
       const { data: requestData, dataBuffer: requestDataBuffer } = parseDataFromRequest(request);
+      
+      // Filter out Bruno's internal content-type: false for no body requests from timeline
+      const filteredHeaders = { ...request.headers };
+      if (filteredHeaders['content-type'] === false) {
```
bijin-bruno: 1. It might be safer to make this check case-insensitive to avoid potential mismatches.
2. rename `filteredHeaders` to `headersSent`
3. Also please update all the comments for clarity
```
// Remove false Content-Type header (used to stop axios from auto-setting it); no Content-Type was actually set or sent.
```

bijin-bruno (changes requested):

PR #5616, opened 2025-09-23
feat: Enhance EnvironmentVariables component with read-only support for non-string values

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-716)

**Fix:**
- Allow mixed types in environment variable schema (`value: Yup.mixed().nullable()`)
- Non-string values are read-only in UI with warning tooltip
- String values remain fully editable
- Maintains backward compatibility with existing string-only environments

https://github.com/user-attachments/assets/6b429158-d931-46ad-9951-1ef0556cb94a

### DIFF
--- a/packages/bruno-app/src/components/GlobalEnvironments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
+++ b/packages/bruno-app/src/components/GlobalEnvironments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
@@ -1,6 +1,6 @@
 import React, { useRef, useEffect } from 'react';
 import cloneDeep from 'lodash/cloneDeep';
-import { IconTrash, IconAlertCircle } from '@tabler/icons';
+import { IconTrash, IconAlertCircle, IconInfoCircle } from '@tabler/icons';
 import { useTheme } from 'providers/Theme';
 import { useDispatch, useSelector } from 'react-redux';
 import MultiLineEditor from 'components/MultiLineEditor/index';
@@ -41,7 +41,7 @@ const EnvironmentVariables = ({ environment, setIsModified, originalEnvironmentV
         secret: Yup.boolean(),
         type: Yup.string(),
         uid: Yup.string(),
-        value: Yup.string().trim().nullable()
+        value: Yup.mixed().nullable()
       })
     ),
     onSubmit: (values) => {
@@ -160,9 +160,24 @@ const EnvironmentVariables = ({ environment, setIsModified, originalEnvironmentV
                       name={`${index}.value`}
                       value={variable.value}
                       isSecret={variable.secret}
+                      readOnly={typeof variable.value !== 'string'}
                       onChange={(newValue) => formik.setFieldValue(`${index}.value`, newValue, true)}
                     />
                   </div>
+                  {typeof variable.value !== 'string' && (
+                    <span className="ml-2 flex items-center">
+                      <IconInfoCircle
+                        id={`${variable.name}-disabled-info-icon`}
+                        className="text-muted"
+                        size={16}
+                      />
+                      <Tooltip
+                        anchorId={`${variable.name}-disabled-info-icon`}
+                        content="Non-string values set via scripts are read-only and can only be updated through scripts."
+                        place="top"
+                      />
+                    </span>
+                  )}
                 </td>
                 <td className="text-center">
                   <input
--- a/packages/bruno-app/src/components/MultiLineEditor/index.js
+++ b/packages/bruno-app/src/components/MultiLineEditor/index.js
@@ -35,6 +35,7 @@ class MultiLineEditor extends Component {
       brunoVarInfo: {
         variables
       },
+      readOnly: this.props.readOnly ? 'nocursor' : false,
       tabindex: 0,
       extraKeys: {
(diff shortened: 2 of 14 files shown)


helloanoop:
@sanjaikumar-bruno please work with @sreelakshmi-bruno to finalise the popover description. The current one is not communicating the essence of why we are not able to edit

packages/bruno-app/src/providers/ReduxStore/slices/global-environments.js
```diff
@@ -144,8 +144,16 @@ export const saveGlobalEnvironment = ({ variables, environmentUid }) => (dispatc
...
 
     const { ipcRenderer } = window;
+    // Validate a copy that stringifies values to satisfy schema, but persist original types
+    const envToValidate = {
+      ...environment,
+      variables: (variables || []).map(v => ({
+        ...v,
+        value: v?.value == null ? null : String(v.value),
+      })),
+    };
```
bijin-bruno: Is there an impact on collection environments if we update the schema to support non-string values? 
If not, we can just go with that as we are eventually planning to support non-string. 
if yes, a better approach would be to just override the schema like the example given below.

```
// Base schema
const baseSchema = yup.object({
  name: yup.string().required("Name is required"),
  email: yup.string().email("Invalid email").required(),
  age: yup.number().min(18, "Must be 18 or older"),
});

// Override just the `age` field
const extendedSchema = baseSchema.shape({
  age: yup.number().min(21, "Must be 21 or older"),
});
```

bijin-bruno (changes requested):

packages/bruno-app/src/components/GlobalEnvironments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
```diff
@@ -160,9 +161,21 @@ const EnvironmentVariables = ({ environment, setIsModified, originalEnvironmentV
                       name={`${index}.value`}
                       value={variable.value}
                       isSecret={variable.secret}
+                      readOnly={typeof variable.value !== 'string'}
                       onChange={(newValue) => formik.setFieldValue(`${index}.value`, newValue, true)}
                     />
                   </div>
+                  {typeof variable.value !== 'string' && (
+                    <span className="ml-2 flex items-center">
+                      <ToolHint
```
bijin-bruno: Please use Tooltip instead of ToolHint for UI consistency. 
Please add a playwright test case to validate the field is disabled and the info tooltip is displayed.

packages/bruno-app/src/components/GlobalEnvironments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
```diff
@@ -160,9 +161,21 @@ const EnvironmentVariables = ({ environment, setIsModified, originalEnvironmentV
...
+                        toolhintId={`data-${variable.uid}`}
```
bijin-bruno: Any particular reason for prefixing the id with `data-`? If not, I’d suggest using `<variable-name>-disabled-info-tooltip instead`. This keeps the naming consistent and also makes it easier to use as a Playwright element locator.

bijin-bruno (changes requested):

packages/bruno-app/src/components/MultiLineEditor/index.js
```diff
@@ -35,6 +35,7 @@ class MultiLineEditor extends Component {
       brunoVarInfo: {
         variables
       },
+      readOnly: this.props.readOnly ? 'nocursor' : false,
```
sanish-bruno, line 38: @sanjaikumar-bruno why are we passing no-cursor for readOnly flag? it can accept booleans as well as strings?, i am seeing  `&.read-only {}` styles. i guess that is  adding the cursor: not-allowed style. not necessarily this config, can you check this!

sanjaikumar-bruno: Yes, the readOnly flag can accept both boolean and string values. The `&.read-only {}` style applies the not-allowed cursor, but the input remains focusable. To make the input non-focusable, we need to add the readOnly flag with the 'nocursor' option.

packages/bruno-app/src/components/GlobalEnvironments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
```diff
@@ -160,9 +160,24 @@ const EnvironmentVariables = ({ environment, setIsModified, originalEnvironmentV
...
                       value={variable.value}
                       isSecret={variable.secret}
+                      readOnly={typeof variable.value !== 'string'}
                       onChange={(newValue) => formik.setFieldValue(`${index}.value`, newValue, true)}
                     />
                   </div>
+                  {typeof variable.value !== 'string' && (
+                    <span className="ml-2 flex items-center">
+                      <IconInfoCircle
+                        id={`${variable.name}-disabled-info-tooltip`}
```
bijin-bruno:
```suggestion
                        data-testid={`${variable.name}-disabled-info-icon`}
```

Issue #5340, opened 2025-08-15
Large file upload crashes Bruno


RumianR:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.7.0

### Operating System

Windows 11 23H2 Build 22631.5624

### Describe the bug

Hello,

Whenever I try to upload a larger mp4 file of 367mb, Bruno crashes/closes. It works fine for smaller file uploads. I recently transitioned from Postman to this tool and this is a huge limitation as I need to currently test file uploads for one of my new service. Any help would be appreciated. Thanks! 

<img width="1031" height="382" alt="Image" src="https://github.com/user-attachments/assets/1f2c304a-f92a-4562-96bb-cf75af35938e" />

<--- Last few GCs --->

[10524:0000094C00160000]   123776 ms: Mark-Compact 139.7 (157.3) -> 127.1 (145.6) MB, pooled: 0 MB, 51.77 / 0.00 ms  (+ 0.2 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 338 ms) (average mu = 0.999, current mu [10524:0000094C00160000]   126281 ms: Mark-Compact 415.1 (433.6) -> 232.4 (252.9) MB, pooled: 0 MB, 15.72 / 0.00 ms  (+ 0.1 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 1447 ms) (average mu = 0.998, current mu

<--- JS stacktrace --->

FATAL ERROR: invalid table size Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 00007FF7D6DE166A uv_cpumask_size+56906
 2: 00007FF7D6D5A43B node::OnFatalError+747
 3: 00007FF7DA0083A4 v8::Function::NewInstance+868
 4: 00007FF7DA008313 v8::Function::NewInstance+723
 5: 00007FF7DA0E73C7 v8::CppHeap::CollectGarbageInYoungGenerationForTesting+70967
 6: 00007FF7D86B1EC5 v8::internal::StrongRootAllocatorBase::deallocate_impl+1428853
 7: 00007FF7D866009E v8::internal::StrongRootAllocatorBase::deallocate_impl+1093454
[...]

PR #5637, opened 2025-09-25
fix: improve file upload handling in prepare-request to use streaming

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-1773)
Fixes: https://github.com/usebruno/bruno/issues/5340

Bruno crashes with a "JavaScript heap out of memory" error when uploading large files (e.g., 367MB MP4 files) in file body mode. The issue occurs because the entire file is loaded into memory using `fs.readFile()`.

### Fix:
- **Small files (≤100MB)**: Use `fs.readFileSync()` for better performance
- **Large files (>100MB)**: Use `fs.createReadStream()` for streaming upload to prevent memory overflow

### DIFF
--- a/packages/bruno-cli/src/runner/prepare-request.js
+++ b/packages/bruno-cli/src/runner/prepare-request.js
@@ -314,7 +314,7 @@ const prepareRequest = async (item = {}, collection = {}) => {
         }
 
         try {
-          // Large files (>20MB) can cause "JavaScript heap out of memory" errors when loaded entirely into memory.
+          // Large files can cause "JavaScript heap out of memory" errors when loaded entirely into memory.
           if (isLargeFile(filePath, STREAMING_FILE_SIZE_THRESHOLD)) {
             // For large files: Use streaming to avoid memory issues
             axiosRequest.data = fs.createReadStream(filePath);
--- a/packages/bruno-cli/src/runner/prepare-request.js
+++ b/packages/bruno-cli/src/runner/prepare-request.js
@@ -4,10 +4,13 @@ const filter = require('lodash/filter');
 const find = require('lodash/find');
 const decomment = require('decomment');
 const crypto = require('node:crypto');
-const fs = require('node:fs/promises');
+const fs = require('node:fs');
 const { mergeHeaders, mergeScripts, mergeVars, mergeAuth, getTreePathFromCollectionToItem } = require('../utils/collection');
 const { buildFormUrlEncodedPayload } = require('../utils/form-data');
 const path = require('node:path');
+const { isLargeFile } = require('../utils/filesystem');
+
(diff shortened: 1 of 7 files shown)


packages/bruno-electron/src/ipc/network/prepare-request.js
```diff
@@ -2,6 +2,7 @@ const { get, each, filter, find } = require('lodash');
 const decomment = require('decomment');
 const crypto = require('node:crypto');
 const fs = require('node:fs/promises');
+const fsSync = require('node:fs');
```
bijin-bruno:
```suggestion
const fs = require('node:fs');
```

bijin-bruno (commented):
@sanjaikumar-bruno We need to handle this in the cli as well. We now have file upload support in cli

sanish-bruno:
@sanjaikumar-bruno  few unit tests regarding prepare request is failing can you look into it?

packages/bruno-cli/src/runner/prepare-request.js
```diff
@@ -311,8 +311,18 @@ const prepareRequest = async (item = {}, collection = {}) => {
...
 
         try {
-          const fileContent = await fs.readFile(filePath);
-          axiosRequest.data = fileContent;
+          // Check file size to determine the best approach for file upload
+          // Large files (>100MB) can cause "JavaScript heap out of memory" errors
+          // when loaded entirely into memory, so we use streaming for them
+          const stat = fs.statSync(filePath);
+
+          if (stat.size > 100 * 1024 * 1024) { // 100MB threshold
```
bijin-bruno: 1. Please create a constant at the top to define the threshold size, say, `STREAMING_FILE_SIZE_THRESHOLD_BYTES`
2. Let's create a util function in `src/utils/filesystem.js` to check the filesize with default 10MB threshold
```
/**
 * Checks if a file is larger than a given threshold.
 * @param {string} filePath - The path to the file.
 * @param {number} threshold - The threshold in bytes. Default is 10MB.
 * @returns {boolean} True if the file is larger than the threshold, false otherwise.
 */
const isLargeFile = (filePath, threshold = 10 * 1024 * 1024) => {
  if (!isFile(filePath)) {
    throw new Error(`File ${filePath} is not a file`);
  }

  const size = fs.statSync(filePath).size;
  
  return size > threshold;
}
```

bijin-bruno (approved):

StefanFanaru:
Hello, great to see these improvements! My Bruno was so easy to crash while working with files of 20+ MB. After 3-4 requests it was becoming unresponsive.
My workflow includes interaction with Azure SAS URLs to upload files to Azure Blob Storage.
In the more recent versions (v2.13.2) for example, I can no longer make these requests with Bruno because it adds a "Transfer-Encoding" header to the request which is rejected by Azure.
Is there any chance it's linked to these changes? Can I do a workardound?
image

Issue #3690, opened 2024-12-23
Support global environment for CLI


moveurbody:
### I have checked the following:

- [X] I've searched existing issues and found nothing related to my issue.

### Describe the feature you want to add

bruno-cli can support a global environment or specify an environment Json file.

e.g.`bru run --env-file {file_path}`

https://learning.postman.com/docs/collections/using-newman-cli/newman-options/#setup-options

### Mockups or Images of the feature

e.g.`bru run --env-file {file_path}`

https://learning.postman.com/docs/collections/using-newman-cli/newman-options/#setup-options

sreelakshmi-bruno:
Hi @moveurbody, thanks for the suggestion! Adding it to our development roadmap.

mesopotamiaMD:
Dear Bruno developlement, I need the mentioned functionality also, as:

`bru run --env-file ../../global_environment.json`

**global_environment.json** is already created via the UI, but BrunoCLI (bru run) doesn't support it. 

The global environments are needed because Collections can have subfolders, and REST Calls (scripts) can use created/returned variables of other calls globally. The variable exchange between two subfolders of a collection is possible via UI and not via BrunoCLI. 

In short, please implement/complete the functionality considering these points. 

Happy to get feedback soon. regards!

mesopotamiaMD:
Dear @sreelakshmi-bruno, is there an update on this issue? Or do you have a workaround that can help too?

anusree-bruno:
We are actively working on adding support for JSON environment files and this is expected to go in our next release.

mesopotamiaMD:
Dear @anusree-bruno, thank you for the update. When do you expect the change will be available? If you have a pre-release version, I am glad to make some tests and give feedback to you. Best regards

sreelakshmi-bruno:
Hi @mesopotamiaMD , supporting JSON files in the CLI has been delayed since we are working on some other crucial tasks. Because of this reason, I cannot give you an ETA on this but we'll update here when we pick it up.


> The global environments are needed because Collections can have subfolders, and REST Calls (scripts) can use created/returned variables of other calls globally. The variable exchange between two subfolders of a collection is possible via UI and not via BrunoCLI.

I agree, Bruno CLI currently lacks global environment support, which creates a gap between UI and CLI functionality for variable exchange between subfolders.

Current Solutions:

You can achieve cross-subfolder variable exchange using Bruno's scripting functions:

**Runtime Variables:**
```javascript
[...]

PR #5660, opened 2025-09-29
feat: enhance json environment file support in bruno-cli

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-1287)
Closes: #3690

This PR enables the Bruno CLI to accept JSON environment files via the `--env-file` parameter, bridging the gap between UI-only global environments and CLI-based workflows.

Accepts single environment objects with this structure:
```json
{
  "name": "My Environment",
  "variables": [
    { "name": "host", "value": "https://api.example.com", "enabled": true }
  ]
}
```

### DIFF
--- a/packages/bruno-cli/tests/utils/parse-environment-json.spec.js
+++ b/packages/bruno-cli/tests/utils/parse-environment-json.spec.js
@@ -1,5 +1,6 @@
 const { describe, it, expect } = require('@jest/globals');
-const { parseEnvironmentJson, getEnvVars } = require('../../src/utils/bru');
+const { getEnvVars } = require('../../src/utils/bru');
+const { parseEnvironmentJson } = require('../../src/utils/environment');
 
 describe('parseEnvironmentJson', () => {
   it('normalizes single environment object', () => {
@@ -7,7 +8,7 @@ describe('parseEnvironmentJson', () => {
       name: 'My Env',
       variables: [
         { name: 'host', value: 'https://www.httpfaker.org' },
-        { name: 'token', value: 'abc', enabled: false }
+        { name: 'token', value: 'abc', enabled: false, secret: true }
       ]
     };
     const env = parseEnvironmentJson(input);
@@ -20,6 +21,7 @@ describe('parseEnvironmentJson', () => {
       secret: false
     });
     expect(env.variables[1].enabled).toBe(false);
+    expect(env.variables[1].secret).toBe(true);
 
     const vars = getEnvVars(env);
     expect(vars).toEqual({ host: 'https://www.httpfaker.org' });


packages/bruno-cli/tests/utils/parse-environment-json.spec.js
```diff
@@ -0,0 +1,30 @@
+const { describe, it, expect } = require('@jest/globals');
```
sanish-bruno, line 1: @sanjaikumar-bruno can you add a playwright test case too?

packages/bruno-cli/src/commands/run.js
```diff
@@ -346,10 +346,28 @@ const handler = async function (argv) {
...
+        try {
+          envJsonContent = fs.readFileSync(envFilePath, 'utf8');
+          // Allow comments-free JSON only; if needed in future, introduce decomment here
+          const parsed = JSON.parse(envJsonContent);
+          const normalizedEnv = parseEnvironmentJson(parsed);
+          envVars = getEnvVars(normalizedEnv);
+          envVars.__name__ = path.basename(envFilePath, '.json');
+        } catch (err) {
+          console.error(chalk.red(`Failed to parse Environment JSON: ${err.message}`));
+          process.exit(constants.EXIT_STATUS.ERROR_INVALID_JSON);
```
sanish-bruno: @sanjaikumar-bruno does this error exists within constants?

sanish-bruno (changes requested):

tests/runner/cli-json-env-file.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  test('CLI: Run with non-existent JSON environment file should fail', async () => {
+    const command = `node ../../../../packages/bruno-cli/bin/bru.js run --env-file non-existent.json`;
+
+    try {
+      execSync(command, { stdio: 'pipe' });
+      // If we get here, the command succeeded when it should have failed
+      expect(true).toBe(false);
+    } catch (error) {
+      // Expected to fail with non-zero exit code
+      expect(error.status).not.toBe(constants.EXIT_STATUS.ERROR_ENV_NOT_FOUND);
```
sanish-bruno: the expect should be checking for  `expect(error.status).toBe(constants.EXIT_STATUS.ERROR_INVALID_JSON)`.

tests/runner/cli-json-env-file.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+  });
+
+  test('CLI: Run with invalid JSON environment file should fail', async () => {
```
sanish-bruno: can we also write a test case where we use one of the variables from the json within the requests and request succeeds. In the truest sense that what should be testing!

packages/bruno-cli/tests/utils/parse-environment-json.spec.js
```diff
@@ -0,0 +1,30 @@
...
+    expect(env.variables[0]).toEqual({
+      name: 'host',
+      value: 'https://www.httpfaker.org',
+      type: undefined,
+      enabled: undefined,
+      secret: undefined
+    });
+
+    const vars = getEnvVars(env);
+    expect(Object.keys(vars)).toHaveLength(0);
```
sanish-bruno: @sanjaikumar-bruno why are we  expecting for length 0 here, it should be 2 right?

packages/bruno-cli/src/utils/bru.js
```diff
@@ -139,9 +139,35 @@ const getOptions = () => {
...
+  }
+
+  const normalized = {
+    name: parsed.name,
+    variables: (parsed.variables || [])
+      .filter(Boolean)
+      .map((variable) => ({
+        name: variable.name,
+        value: variable.value,
+        type: variable.type,
```
sanish-bruno: i guess the type, enabled, secret can have a fallback to 'text', true,false respectively. not necessarily undefined , secret can alway be false too

packages/bruno-cli/tests/utils/parse-environment-json.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        { name: 'host', value: 'https://www.httpfaker.org' },
+        { name: 'token', value: 'abc', enabled: false }
+      ]
+    };
+    const env = parseEnvironmentJson(input);
+    expect(Array.isArray(env.variables)).toBe(true);
+    expect(env.variables[0]).toEqual({
+      name: 'host',
+      value: 'https://www.httpfaker.org',
+      type: undefined,
```
sanish-bruno: may need to re-write the test case once we have changes in the `arseEnvironmentJson` please also add a scenario where these keys are specified within json file and parse them accurately. secret i believe should always be false i think, no matter what people pass from json

packages/bruno-cli/src/commands/run.js
```diff
@@ -346,10 +346,28 @@ const handler = async function (argv) {
...
+      if (ext === '.json') {
+        // Parse Bruno schema JSON environment
+        let envJsonContent;
+        try {
+          envJsonContent = fs.readFileSync(envFilePath, 'utf8');
+          // Allow comments-free JSON only; if needed in future, introduce decomment here
+          const parsed = JSON.parse(envJsonContent);
+          const normalizedEnv = parseEnvironmentJson(parsed);
+          envVars = getEnvVars(normalizedEnv);
+          envVars.__name__ = path.basename(envFilePath, '.json');
```
sanish-bruno: what are we using this for '__name__'  ? unlike in the case of bru files, the name is taken from the file name, the json files have their env name within the json file, should we use that?

sanish-bruno (changes requested):

packages/bruno-cli/tests/utils/parse-environment-json.spec.js
sanjaikumar-bruno: Because [getEnvVars](https://github.com/usebruno/bruno/blob/857a159bfabe5ec4329431f42b44e9a5c88628f1/packages/bruno-cli/src/utils/bru.js#L121-L135) returns only variables with `enabled === true`. In this test, `host` has no `enabled` and `token` has `enabled: false`, so both are filtered out leading to length 0.

tests/runner/cli-json-env-file/cli-json-env-file.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+import { execSync } from 'child_process';
+import fs from 'fs';
+import path from 'path';
+import constants from '../../../packages/bruno-cli/src/constants.js';
+
+test.describe('CLI JSON Environment File Support', () => {
+  const collectionPath = path.resolve(__dirname, 'collection');
+  const BRU = 'node ../../../../packages/bruno-cli/bin/bru.js';
+
+  const runFrom = (cwd: string, args: string): number => {
```
sanish-bruno, line 14: @sanjaikumar-bruno can you explain the test briefly as a comment. just wanted to understand what is happening within the test, you can tell it is emulating `bru run`

packages/bruno-cli/src/utils/bru.js
```diff
@@ -139,9 +139,35 @@ const getOptions = () => {
...
+
+  const normalized = {
+    name: parsed.name,
+    variables: (parsed.variables || [])
+      .filter(Boolean)
+      .map((variable) => ({
+        name: variable.name,
+        value: variable.value,
+        type: variable.type || 'text',
+        enabled: variable.enabled !== false,
```
sanish-bruno: just fallback to false will do right? `Boolean(variable.enabled || false)` , i think current approach may convert undefined and null values to true. i guess, unless explicitly false, in all remaining cases, it will return true

packages/bruno-cli/src/constants.js
```diff
@@ -23,6 +23,8 @@ const EXIT_STATUS = {
   ERROR_INCORRECT_ENV_OVERRIDE: 8,
   // Invalid output format requested
   ERROR_INCORRECT_OUTPUT_FORMAT: 9,
+  // Invalid JSON format
+  ERROR_INVALID_JSON: 10,
```
sanish-bruno: @sanjaikumar-bruno don't know whether we should call it ERROR_INVALID_JSON, then we will have to write errors for each type of file, that seems unnecessary.  I guess ERROR_INVALID_FILE should be good enough

sanish-bruno (changes requested):

tests/runner/cli-json-env-file/collection/env.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "CLI JSON Env",
+  "variables": [
+    { "name": "baseUrl", "value": "https://httpbin.org", "enabled": true }
```
sanish-bruno: @sanjaikumar-bruno can we use echobruno?

sanish-bruno (changes requested):

packages/bruno-cli/tests/utils/parse-environment-json.spec.js
```diff
@@ -0,0 +1,59 @@
...
+  it('throws on invalid shape', () => {
+    expect(() => parseEnvironmentJson({ name: 'x' })).toThrow(/Invalid environment JSON/i);
+  });
+
+  it('respects explicit fields and preserves secret flag', () => {
+    const input = {
+      name: 'My Env',
+      variables: [
+        { name: 'one', value: '1', type: 'text', enabled: true, secret: true },
+        { name: 'two', value: '2', type: 'file', enabled: false, secret: true }
```
sanishchirayath1: chnage on of them to false, we need both the cases, also undefined, null keys

packages/bruno-cli/tests/utils/parse-environment-json.spec.js
```diff
@@ -0,0 +1,59 @@
+const { describe, it, expect } = require('@jest/globals');
+const { parseEnvironmentJson, getEnvVars } = require('../../src/utils/bru');
+
+describe('parseEnvironmentJson', () => {
+  it('normalizes single environment object', () => {
+    const input = {
+      name: 'My Env',
+      variables: [
+        { name: 'host', value: 'https://www.httpfaker.org' },
+        { name: 'token', value: 'abc', enabled: false }
```
sanishchirayath1: can add an expect statement to check enabled!

sanishchirayath1 (changes requested):

packages/bruno-cli/src/utils/bru.js
```diff
@@ -139,9 +139,35 @@ const getOptions = () => {
   return options;
 };
 
+/**
+ * Parse a Bruno JSON environment object and normalize variables
+ * Accepts only single environment object: { name?, uid?, variables: [...] }
+ */
+const parseEnvironmentJson = (parsed = {}) => {
```
bijin-bruno: Please move this method to `packages/bruno-cli/src/utils/environment.js`

PR #5672, opened 2025-09-30
feat: add redirect and timeout in request settings

pooja-bruno:
# Description

## ✨ Features Added

- **Automatically Follow Redirects toggle** 
- **Max Redirects setting** 
- **Timeout setting**

[JIRA](https://usebruno.atlassian.net/browse/BRU-1899)

### Contribution Checklist:

image

### DIFF
--- a/packages/bruno-lang/v2/tests/settings/settings.spec.js
+++ b/packages/bruno-lang/v2/tests/settings/settings.spec.js
@@ -43,18 +43,6 @@ describe('Settings Conversion Tests', () => {
   });
 
   describe('round-trip conversion', () => {
-    it('should maintain data integrity through BRU -> JSON -> BRU conversion', () => {
-      const originalBru = fs.readFileSync(path.join(fixturesDir, 'settings-all-options.bru'), 'utf8');
-
-      // Convert BRU to JSON
-      const json = bruToJson(originalBru);
-
-      // Convert JSON back to BRU
-      const convertedBru = jsonToBru(json);
-
-      expect(convertedBru).toEqual(originalBru);
-    });
-
     it('should maintain data integrity through JSON -> BRU -> JSON conversion', () => {
       const originalJson = require(path.join(fixturesDir, 'settings-all-options.json'));
 


tests/request/settings/no-redirects-settings.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    // Verify default value from .bru file (5)
+    await expect(maxRedirectsInput).toHaveValue('5');
+
+    // Test Follow Redirects toggle - should be unchecked
+    const followRedirectsToggle = page.locator('[data-testid="follow-redirects-toggle"]');
+    await expect(followRedirectsToggle).toBeVisible();
+    await expect(followRedirectsToggle).not.toBeChecked();
+
+    // Send the request - should stop at first redirect (302) without following
+    await page.locator('[data-testid="send-arrow-icon"]').click();
```
bijin-bruno: Please use `getByTestId()` for cleanliness

tests/request/settings/no-redirects.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    await expect(followRedirectsToggle).toBeChecked();
+
+    // Send request again - now should follow redirects and get 200
+    await page.locator('[data-testid="send-arrow-icon"]').click();
+    await expect(page.locator('[data-testid="response-status-code"]')).toContainText('200', { timeout: 15000 });
+  });
+
+  test.afterEach(async ({ pageWithUserData: page }) => {
+    // Close the single open tab
+    await page.locator('.close-icon-container').click();
```
bijin-bruno, line 47: can we use the closeAllCollections step from the utils? That would take user back to welcome / home page

packages/bruno-app/src/components/SettingsInput/index.js
bijin-bruno, line 1: I think this could be turned into a generic, reusable pure component and placed under our ui folder. That way, it can also be reused in places like collection settings.

Additionally, the handler logic should live outside the component. The parent should decide what action to perform on the callbacks. For example, handleKeydown would be better placed in the parent component rather than inside this one.

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -192,7 +202,8 @@ const configureRequest = async (
     addDigestInterceptor(axiosInstance, request);
   }
 
-  request.timeout = preferencesUtil.getRequestTimeout();
+  // Get timeout from request settings, fallback to global preference
+  request.timeout = request.settings?.timeout || preferencesUtil.getRequestTimeout();
```
bijin-bruno: With this we cannot set request timeout as 0 since it is a falsey value and would take the value from preferences. Is this an expected behavior?

pooja-bruno: Yes, that’s expected.

tests/request/settings/no-redirects.spec.ts
pooja-bruno: This utility function isn’t available in this branch. I’ll create a separate PR for it

packages/bruno-schema/src/collections/index.js
```diff
@@ -432,7 +432,10 @@ const itemSchema = Yup.object({
     })
   }),
   settings: Yup.object({
-    encodeUrl: Yup.boolean().nullable()
+    encodeUrl: Yup.boolean().nullable(),
+    followRedirects: Yup.boolean().nullable(),
+    maxRedirects: Yup.number().min(0).max(50).nullable(),
+    timeout: Yup.number().min(0).max(300000).nullable()
```
helloanoop: @pooja-bruno @bijin-bruno I am not sure if we want to set the max constraint here. Very likely that someone will raise a bug saying that they cannot save settings

I would recommend keeping the min 0 constraint and leaving it the user to max out as they wish

packages/bruno-lang/v2/src/bruToJson.js
```diff
@@ -409,9 +409,34 @@ const sem = grammar.createSemantics().addAttribute('ast', {
...
+      }
+    }
+
     return {
       settings: {
-        encodeUrl: typeof settings.encodeUrl === 'boolean' ? settings.encodeUrl : settings.encodeUrl === 'true'
+        encodeUrl: typeof settings.encodeUrl === 'boolean' ? settings.encodeUrl : settings.encodeUrl === 'true',
+        followRedirects: parsedSettings.followRedirects,
+        maxRedirects: parsedSettings.maxRedirects,
+        timeout: parsedSettings.timeout
```
helloanoop: Please add a test in `bruno-lang` package that tests the bruToJson and jsonToBru of request files have these settings.

You can refer recently written tests for [custom-methods](https://github.com/usebruno/bruno/tree/main/packages/bruno-lang/v2/tests/custom-methods)

tests/request/settings/collection/max-redirects.bru
```diff
@@ -0,0 +1,17 @@
+meta {
```
helloanoop, line 1: I don't think `-test` suffix is needed. It's inferred.

tests/request/settings/max-redirects.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { test, expect } from '../../../playwright';
```
helloanoop, line 1: I don't think `-settings` suffix is needed. It's inferred since its already inside `tests/request/settings` folder

image

tests/request/settings/no-redirects.spec.ts
bijin-bruno: @pooja-bruno The utility is available on main branch. Let's rebase and update this code.

pooja-bruno: I have added it

PR #5698, opened 2025-10-03
fix: Show active global environment in config modal

pooja-bruno:
# Description

Fixes an issue where the global environment configuration modal always opened with the first environment selected instead of the currently active environment.

[JIRA](https://usebruno.atlassian.net/browse/BRU-1899)

### Contribution Checklist:

Before:
https://github.com/user-attachments/assets/f13da276-67c0-4ef6-ae28-680b1208a240

After:
https://github.com/user-attachments/assets/4648db7e-929f-4f95-88f4-79b442b5b7bf


packages/bruno-app/src/components/ToolHint/index.js
```diff
@@ -11,7 +11,8 @@ const ToolHint = ({
   place = 'top',
   offset,
   theme = null,
-  className = ''
+  className = '',
```
sanish-bruno, line 16: @pooja-bruno why do we have these chnages, i guess we are using tooltip from react tooltip, not ToolHint component

packages/bruno-app/src/components/Environments/EnvironmentSelector/EnvironmentListContent/index.js
```diff
@@ -22,12 +23,26 @@ const EnvironmentListContent = ({
...
+                  id={`env-tooltip-${env.uid}`}
                   className={`dropdown-item ${env.uid === activeEnvironmentUid ? 'active' : ''}`}
                   onClick={() => onEnvironmentSelect(env)}
+                  data-tooltip-content={env.name}
+                  data-tooltip-place="right"
                 >
                   <span className="max-w-32 truncate no-wrap">{env.name}</span>
                 </div>
               ))}
+              <Tooltip
```
sanish-bruno: I guess we are doing two things in the PR

tests/environments/global-env-config-selection/global-env-config-selection.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { test, expect } from '../../../playwright';
+
+test.describe('Global Environment Configuration Selection Tests', () => {
+  test('should open global environment config with currently active environment selected', async ({
+    pageWithUserData: page
+  }) => {
```
sanish-bruno, line 12: @pooja-bruno use closeCollections(). function from test/utils to close the collection, it is in the latest code

sanish-bruno: @pooja-bruno can you also do the same  for  collection envs?

sanish-bruno (changes requested):

bijin-bruno:
@pooja-bruno We should add tooltip for the toolbar labels too
Screenshot 2025-10-06 at 2 48 06 PM

packages/bruno-app/src/components/ToolHint/index.js
```diff
@@ -28,18 +30,25 @@ const ToolHint = ({
     color: toolhintTextColor
   };
 
+  const tooltipProps_final = anchorSelect
```
bijin-bruno: rename to use toolhintProps instead of tooltipProps

packages/bruno-app/src/components/ToolHint/index.js
```diff
@@ -28,18 +30,25 @@ const ToolHint = ({
...
-          anchorId={toolhintId}
-          content={text}
+          {...tooltipProps_final}
+          content={anchorSelect ? undefined : text}
           className="toolhint"
           offset={offset}
           place={place}
           noArrow={true}
           style={combinedToolhintStyle}
+          {...tooltipProps}
```
bijin-bruno: toolhintFinalProps should already be merged with any other props.
if this is just for one prop (`positionStrategy`), we can explicitly use that prop as we did for all existing ones like `place`

tests/environments/global-env-config-selection/init-user-data/global-environments.json
bijin-bruno, line 1: Please use 2 spaces intentation

packages/bruno-app/src/components/ToolHint/index.js
bijin-bruno: @pooja-bruno, Resurfacing this comment in case it was overlooked.

bijin-bruno (changes requested):
@pooja-bruno, I think it would be better to display the tooltip only when the text exceeds the width or a certain limit.
We could also add a small delay (at least 2 seconds) before showing the tooltip.

PR #4903, opened 2025-06-14
Import Insomnia Collection with folder level variables and also Import collection level environments

hippasus:
# Description

There are three enhancements for importing insomnia collection in the request:

- Import Insomnia folder level variables
- Do variable normalization for request url/body/headers/variables
- Import Insomnia collection level environments

### Contribution Checklist:

- [x ] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [ ] **I have added screenshots or gifs to help explain the change if applicable.**
- [x ] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ ] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

Issue #1764, opened 2024-03-12
Missing Insomnia Environment Import


cmcconomyfwig:
The Insomnia environment format is incredibly simple -- just a top level json key/value set.

I wrote this python script in case it helps anyone else:

``` python
from pathlib import Path
import json

# expects contents to be,
#  { 'envname': {k:v, k2:v2}} etc.
envs = json.loads(Path('./insomnia_envs.json').read_text())

for env, values in envs.items():
    this_env = {
        "name": env,
        "values": [
            {
                "key": key,
                "value": value,
                "type" : "secret" if "secret" in key or "key" in key else "default",
                "enabled": True
            } for (key, value) in values.items()
        ]
    }
    Path(f'./postman.env.{env}.json').write_text(json.dumps(this_env, indent=2))
    print(f"Wrote ./postman.env.{env}.json")
```

Constructing the required `./insomnia_envs.json` is simple; it should look as follows:
``` json
{
"env1": {
  "key1": "env1_val1",
  "key2": "env1_val2"
}, 
"env2": {
  "key1": "env2_val1",
  "key2": "env2_val2"
}
}
```

etc.

Hope this helps!

What I would suggest is a capability in Bruno to open up a textbox and just paste in a k/v json as an environment.

skysbsb:
```
# this script will get only the environments from the Insomnia.json file
# and save them in a new file called environments.json
import json
with open('Insomnia.json') as f:
    data = json.load(f)
    # grab only the resources that are _type = "environment"
    environments = [resource for resource in data['resources'] if resource['_type'] == 'environment']
    # save the environments in a new file called environments.json
    # format:
    # {
    # "env1": {
    #   "key1": "env1_val1",
    #   "key2": "env1_val2"
    # }, 
    # "env2": {
    #   "key1": "env2_val1",
    #   "key2": "env2_val2"
    # }
    # }
    environments_dict = {}
    for env in environments:
        env_name = env['name']
        env_data = env['data']
        environments_dict[env_name] = env_data 
    # save pretty
    with open('environments.json', 'w') as outfile:
        json.dump(environments_dict, outfile, indent=4)
        
    print('environments.json created')
```

Cicolas:
Hi, I'm new at Bruno, and I'm enjoying it. But recently I tried to import all my old project endpoints from insomnia to Bruno, and sadly the environments imports are not working. I saw this PR (#3634), that seems to add this support, any news on this?

LunaticoCR:
[...]

Issue #472, opened 2023-10-09
[Migrating to Bruno] Missing / not imported environments when importing collection from Insomnia


F0rce:
First of all, I'd like to congratulate Bruno on his recent success. I have had Bruno on my radar for quite some time, but had not found the time to play with it. With the recent "announcements/requirements" of a certain application (no names needed), I finally decided to migrate to Bruno and everything except the environments has been a completely flawless experience.

Let me explain the problem I have:
In Insomnia we use environments for the different urls of our api (test local/staging/prod). All of them have different urls and different API keys

<img width="1186" alt="Bildschirmfoto 2023-10-09 um 10 34 56" src="https://github.com/usebruno/bruno/assets/60381251/d6befe08-f010-4951-b7f7-58a55fd913f8">

Even though I checked the "export environments" box when exporting the collection from Insomnia, they do not seem to be imported into Bruno. 

<img width="1285" alt="Bildschirmfoto 2023-10-09 um 10 58 04" src="https://github.com/usebruno/bruno/assets/60381251/22446ba3-d040-44b6-a6d3-19fb4a872590">

Is this a "bug" or has this feature not yet been implemented?

If wanted I could take a look at an implementation (I would need some help to get started).

Thanks again for making such a great client <3

bijin-bruno:
This feature has been delivered via https://github.com/usebruno/bruno/pull/5716 and is now available starting from Bruno v2.14.0.

PR #5716, opened 2025-10-06
feat: Import Insomnia environments

sanjaikumar-bruno:
[JIRA](https://usebruno.atlassian.net/browse/BRU-232)
Fixes: https://github.com/usebruno/bruno/issues/1764 
Fixes: https://github.com/usebruno/bruno/issues/472
Closes: https://github.com/usebruno/bruno/pull/4903

### Summary
- Add Insomnia environments import (v4 and v5).
- Generate Bruno environments from:
  - v4: workspace base env + sub envs (nearest-base shallow merge).
  - v5: base env + subEnvironments (base flattened + shallow override).
- Flatten nested env data into JSON-path-like keys (e.g., nested.user.name, items[0].id).
- Extract env logic into `insomnia/env-utils.js`.
- Add tests for v4/v5 merging and key flattening.

### DIFF
--- a/tests/import/insomnia/import-insomnia-v4-environments.spec.ts
+++ b/tests/import/insomnia/import-insomnia-v4-environments.spec.ts
@@ -39,7 +39,7 @@ test.describe('Import Insomnia v4 Collection - Environment Import', () => {
       await page.locator('#collection-location').fill(await createTmpDir('insomnia-v4-env-test'));
       await page.getByRole('button', { name: 'Import', exact: true }).click();
 
-      await expect(page.getByText('Test API Collection v4 with Environments')).toBeVisible();
+      await expect(page.locator('#sidebar-collection-name').getByText('Test API Collection v4 with Environments')).toBeVisible();
 
       await openCollectionAndAcceptSandbox(page, 'Test API Collection v4 with Environments', 'safe');
     });
@@ -79,11 +79,9 @@ test.describe('Import Insomnia v4 Collection - Environment Import', () => {
       await expect(v4BaseUrlInput).toBeVisible();
       await expect(v4AuthTokenInput).toBeVisible();
 
-      const v4BaseUrlRow = v4BaseUrlInput.locator('xpath=ancestor::tr');
-      const v4AuthTokenRow = v4AuthTokenInput.locator('xpath=ancestor::tr');
       // Assert: Top-level string values are preserved exactly as in the source
-      await expect(v4BaseUrlRow.locator('.CodeMirror-line').first()).toHaveText('https://api.example.com');
-      await expect(v4AuthTokenRow.locator('.CodeMirror-line').first()).toHaveText('your_auth_token_here');
+      await expect(page.getByTestId('env-var-row-baseUrl').locator('.CodeMirror-line').first()).toHaveText('https://api.example.com');
+      await expect(page.getByTestId('env-var-row-authToken').locator('.CodeMirror-line').first()).toHaveText('your_auth_token_here');
 
       // **Assertion 2: Nested Object Flattening**
       // Verifies that nested objects are flattened to dot-notation keys (e.g., user.name, user.id)
@@ -92,12 +90,10 @@ test.describe('Import Insomnia v4 Collection - Environment Import', () => {
       await expect(v4UserNameInput).toBeVisible();
       await expect(v4UserIdInput).toBeVisible();
 
-      const v4UserNameRow = v4UserNameInput.locator('xpath=ancestor::tr');
-      const v4UserIdRow = v4UserIdInput.locator('xpath=ancestor::tr');
       // Assert: Nested object properties are accessible via dot notation
-      await expect(v4UserNameRow.locator('.CodeMirror-line').first()).toHaveText('admin');
+      await expect(page.getByTestId('env-var-row-user.name').locator('.CodeMirror-line').first()).toHaveText('admin');
       // Assert: Numeric values are converted to strings and preserved
-      await expect(v4UserIdRow.locator('.CodeMirror-line').first()).toHaveText('123');
+      await expect(page.getByTestId('env-var-row-user.id').locator('.CodeMirror-line').first()).toHaveText('123');
 
       // **Assertion 3: Array Flattening**
       // Verifies that arrays are flattened using JavaScript-style square bracket notation (e.g., user.roles[0], user.roles[1])
@@ -106,11 +102,9 @@ test.describe('Import Insomnia v4 Collection - Environment Import', () => {
       await expect(v4UserRoles0Input).toBeVisible();
       await expect(v4UserRoles1Input).toBeVisible();
 
-      const v4UserRoles0Row = v4UserRoles0Input.locator('xpath=ancestor::tr');
-      const v4UserRoles1Row = v4UserRoles1Input.locator('xpath=ancestor::tr');
       // Assert: Array elements are accessible via JavaScript-style square bracket notation
-      await expect(v4UserRoles0Row.locator('.CodeMirror-line').first()).toHaveText('admin');
-      await expect(v4UserRoles1Row.locator('.CodeMirror-line').first()).toHaveText('user');
+      await expect(page.getByTestId('env-var-row-user.roles[0]').locator('.CodeMirror-line').first()).toHaveText('admin');
+      await expect(page.getByTestId('env-var-row-user.roles[1]').locator('.CodeMirror-line').first()).toHaveText('user');
     });
 
     await test.step('Test Staging Environment - verify merging with base', async () => {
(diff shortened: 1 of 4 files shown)


packages/bruno-converters/src/insomnia/env-utils.js
```diff
@@ -0,0 +1,99 @@
+import { uuid } from '../common';
+import { flattenObject } from '../utils/flatten';
+
+/**
+ * Converts an Insomnia environment node into a Bruno environment using JSON-path-like keys.
+ * - Flattens env.data to dot-notation keys; values are converted to strings.
+ */
+export const toBrunoEnv = (env, index = 0) => {
+  const variables = [];
+  const flat = flattenObject(env?.data || {});
```
sanish-bruno: could be named better 'envData', 'envObj', 'flattenedEnv' ?

packages/bruno-converters/src/insomnia/env-utils.js
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Builds Bruno environments from Insomnia v5 environments.
+ * - Expects a single object (base env) with optional subEnvironments.
+ * - Creates one env for base and one env per sub using flattened, shallow-merged keys.
+ */
+export const buildV5Environments = (environmentsNode) => {
+  if (!environmentsNode || typeof environmentsNode !== 'object') return [];
+
+  const baseEnv = environmentsNode;
```
sanish-bruno: @sanjaikumar-bruno Do we need to create a variable reference here, it will be anyway referencing the same object under the hood, can any changes we make to baseEnv will be available to environmentNode too, can consider renaming the param itself to `baseEnv`!

sanish-bruno: we can use cloneDeep in order to avoid making any unintentional changes to env object that is being passed! please also check the performance of the same

packages/bruno-converters/src/insomnia/env-utils.js
```diff
@@ -0,0 +1,99 @@
...
+  const result = [];
+  // include base as standalone
+  result.push(toBrunoEnv(baseEnv, 0));
+
+  const subs = Array.isArray(baseEnv.subEnvironments) ? baseEnv.subEnvironments : [];
+  const baseFlat = flattenObject(baseEnv?.data || {});
+  subs.forEach((sub, i) => {
+    const subFlat = flattenObject(sub?.data || {});
+    const mergedFlat = shallowMergeFlat(baseFlat, subFlat);
+    result.push(toBrunoEnv({ name: sub?.name || `Environment ${i + 1}`, data: mergedFlat }, i + 1));
```
sanish-bruno: looks like we already have the fallback for the name is  written within `toBrunoEnv`

packages/bruno-converters/src/insomnia/env-utils.js
```diff
@@ -0,0 +1,99 @@
...
+ * - Expects a single object (base env) with optional subEnvironments.
+ * - Creates one env for base and one env per sub using flattened, shallow-merged keys.
+ */
+export const buildV5Environments = (environmentsNode) => {
+  if (!environmentsNode || typeof environmentsNode !== 'object') return [];
+
+  const baseEnv = environmentsNode;
+  const result = [];
+  // include base as standalone
+  result.push(toBrunoEnv(baseEnv, 0));
```
sanish-bruno: can remove the 0 as the function will take by default will have 0

packages/bruno-converters/src/insomnia/env-utils.js
```diff
@@ -0,0 +1,99 @@
...
+  // base envs
+  const baseEnvs = allEnvResources.filter(isBaseEnv);
+  baseEnvs.forEach((base, idx) => {
+    result.push(toBrunoEnv(base, idx));
+  });
+
+  // sub envs
+  const subEnvs = allEnvResources.filter((e) => !isBaseEnv(e));
+  subEnvs.forEach((sub, idx) => {
+    const base = getNearestBase(sub);
```
sanish-bruno: i guess we don't have to find the base for each iteration, if we only have one base env to care for, just find the base env, and use it for every thing else, if in the case insomnia supports multiple base envs, then add a comment so that we can remember this case

packages/bruno-converters/src/insomnia/insomnia-to-bruno.js
```diff
@@ -287,6 +288,9 @@ const parseInsomniaCollection = (data) => {
     }
 
     brunoCollection.items = createFolderStructure(requestsAndFolders, insomniaCollection._id);
+
+    // Build environments from resources
+    brunoCollection.environments = buildV4Environments(insomniaResources, insomniaCollection._id);
```
sanish-bruno, line 293: @sanjaikumar-bruno check if `insomniaCollection._id` is the workspace id, just check for json , yaml exports of insomnia we have the workspace id, and is not `undefined` in the case f=of v4 imports, cause we are using it to find the baseEnv

sanjaikumar-bruno: The workspace id is present in the v4 json.

packages/bruno-converters/src/insomnia/insomnia-to-bruno.js
```diff
@@ -2,13 +2,14 @@ import each from 'lodash/each';
...
 import jsyaml from 'js-yaml';
 import { validateSchema, transformItemsInCollection, hydrateSeqInCollection, uuid } from '../common';
+import { buildV5Environments, buildV4Environments } from './env-utils';
 
 const parseGraphQL = (text) => {
   try {
     const graphql = JSON.parse(text);
 
     return {
-      query: graphql.query,
```
sanish-bruno, line 11: @sanjaikumar-bruno why do we need these changes, is it related to the import envs? we are using the `normalizeVariables` at many places. what was the need of it?

sanish-bruno (changes requested):

packages/bruno-converters/src/insomnia/env-utils.js
```diff
@@ -0,0 +1,90 @@
...
+    variables
+  };
+};
+
+/**
+ * Shallowly merges two flattened env data objects.
+ * - Keys in override replace keys in base.
+ * - No recursive merging.
+ */
+const shallowMergeFlat = (baseFlat = {}, overrideFlat = {}) => ({ ...baseFlat, ...overrideFlat });
```
sanjaikumar-bruno, line 34: We are avoiding lodash merge because we only need a shallow merge, not a deep one.

tests/import/insomnia/import-insomnia-v4-environments.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+      await page.locator('#import-collection-loader').waitFor({ state: 'hidden' });
+
+      const locationModal = page.getByTestId('import-collection-location-modal');
+      await expect(locationModal.getByText('Test API Collection v4 with Environments')).toBeVisible();
+
+      await page.locator('#collection-location').fill(await createTmpDir('insomnia-v4-env-test'));
+      await page.getByRole('button', { name: 'Import', exact: true }).click();
+
+      await expect(page.getByText('Test API Collection v4 with Environments')).toBeVisible();
```
sanish-bruno: @sanjaikumar-bruno Going forward we will be storing the common locators like selecting collection, folder, requests etc in locator.ts, we will try to follow the structure going forward ref: https://github.com/usebruno/bruno/pull/5800, it is okay if we don't follow it here, something to keep in mind for future

If possible you can create a section for environment and move fe selectors there, the selectors can then be used in very test that involves environment. ensure that the the selectors are specific. 
for the table rows we can consider adding dynamic test ids `env-table-row-${index}` from the dom, so that we don't have to use `xpath:ancestor:tr` like selectors. I think test ids will bring more predictability to testing

PR for reference https://github.com/usebruno/bruno/pull/5816

sanish-bruno: we could make the locator more specific by adding the id
```
page.locator('#sidebar-collection-name').getByText('Test API Collection v4 with Environments'))
```

tests/import/insomnia/import-insomnia-v4-environments.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+      await expect(v4AuthTokenRow.locator('.CodeMirror-line').first()).toHaveText('your_auth_token_here');
+
+      // **Assertion 2: Nested Object Flattening**
+      // Verifies that nested objects are flattened to dot-notation keys (e.g., user.name, user.id)
+      const v4UserNameInput = page.locator('input[value="user.name"]');
+      const v4UserIdInput = page.locator('input[value="user.id"]');
+      await expect(v4UserNameInput).toBeVisible();
+      await expect(v4UserIdInput).toBeVisible();
+
+      const v4UserNameRow = v4UserNameInput.locator('xpath=ancestor::tr');
```
sanish-bruno: @sanjaikumar-bruno It is a bit unclear, what does this selector does. I guess it selects the parent tr element!

tests/import/insomnia/import-insomnia-v4-environments.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+      const v4UserIdRow = v4UserIdInput.locator('xpath=ancestor::tr');
```
sanish-bruno: @sanjaikumar-bruno I feel like it is better to go from parent to children, that gives structured flow to the tests. from parent to children. It seems like we are getting the children first, user that to find the parent, then use the parent to find other things within in. This make the process a bit confusing. This is applicable to every test here

PR #5750, opened 2025-10-09
moved custom search to components folder

anusree-bruno:
moved custom search to components folder to use search across different editors


helloanoop:
@anusree-bruno I would recommend naming the component as `CodeMirrorSearch` as it is a component that is a specific to Codemirror

sid-bruno:
Can we validate if the tests are specific to this branch or are also failing on `main` if not failing on `main` then please make sure you update the current branch so we can be sure nothing broke because of the PR even though the PR is just re-arranging imports

Issue #5770, opened 2025-10-13
Variables created with setEnvVar + { persist: true } disappear in cloned environment


ganesh-bruno:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

[Jira](https://usebruno.atlassian.net/browse/BRU-1968)

### Bruno version

2.12.0

### Operating System

Mac OS

### Describe the bug

Environment variables created via script using `bru.setEnvVar("key", "bruno", { persist: true })` are set correctly in the current environment, but when that environment is cloned, the variable disappears in the cloned copy.

### .bru file to reproduce the bug

```js
meta {
  name: echo-bru
  type: http
  seq: 3
}

post {
  url: https://echo.usebruno.com
  body: json
  auth: none
}

body:json {
  {
    "name":"usebruno"
  }
}

script:pre-request {
  bru.setEnvVar("key", "bruno", {persist:true})
}

settings {
  encodeUrl: true
}


```

### Screenshots/Live demo link

https://github.com/user-attachments/assets/6aa1225b-d04e-4079-a12a-de3c6fe788fb

bijin-bruno:
Issue has been fixed via PR https://github.com/usebruno/bruno/pull/5783.
This is available from Bruno v2.14.0 🎉

PR #5783, opened 2025-10-14
feat: enhance environment variable persistence handling

sanjaikumar-bruno:
JIRA: [BRU-1961](https://usebruno.atlassian.net/browse/BRU-1961), [BRU-1968](https://usebruno.atlassian.net/browse/BRU-1968)
Fixes: https://github.com/usebruno/bruno/issues/5770 too

### Issue
When using `bru.setEnvVar` with `{ persist: true }` in multiple requests within the same collection, only the last variable gets persisted to the environment file. Previous persistent variables are lost.

**Steps to reproduce:**
1. Create two requests and a collection environment
2. In first request: `bru.setEnvVar("key1", "value1", { persist: true })`
3. In second request: `bru.setEnvVar("key2", "value2", { persist: true })`
4. Check environment file - only `key2: value2` is present, `key1: value1` is missing

#### Root Cause
The issue was in the Redux store's `scriptEnvironmentUpdateEvent` reducer. When `bru.setEnvVar` was called with `{ persist: true }`, the system correctly tracked persistent variables, but the Redux store was always marking variables as `ephemeral: true`, regardless of the persist option.

#### Solution
- Updated `main:script-environment-update` communication to include `persistentEnvVariables` data.
- Modified `scriptEnvironmentUpdateEvent` to check if variables are in `persistentEnvVariables` and mark them as non-ephemeral accordingly.
- Enhanced `mergeAndPersistEnvironment` logic to preserve all existing non-ephemeral variables when adding new persistent ones.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -9,8 +9,8 @@ if (process.env.CI) {
 export default defineConfig({
   fullyParallel: false,
   forbidOnly: !!process.env.CI,
-  retries: process.env.CI ? 1 : 0,
-  workers: process.env.CI ? undefined : 1,
+  retries: process.env.CI ? 2 : 0,
+  workers: process.env.CI ? undefined : 3,
   reporter,
 
   use: {


tests/environments/api-setEnvVar/multiple-persist-vars.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { test, expect } from '../../../playwright';
+
+test.describe.serial('bru.setEnvVar multiple persistent variables', () => {
+  test.setTimeout(2 * 10 * 1000);
+
+  test('should persist multiple environment variables from different requests', async ({ pageWithUserData: page }) => {
+    await test.step('Select collection', async () => {
+      await page.locator('#sidebar-collection-name').click();
```
sanish-bruno, line 36: @sanjaikumar-bruno please also filter it by collection's name, just so that it never resolves to 2 elements and the test fail. it is always better to be more specific

tests/environments/api-setEnvVar/multiple-persist-vars.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+      await page.getByText('Run', { exact: true }).click();
+
+      // Click Run button in the modal
+      await page.getByRole('button', { name: 'Run', exact: true }).click();
+
+      // Wait for the folder to finish running
+      await page.getByRole('button', { name: 'Run Again' }).waitFor({ timeout: 30000 });
+    });
+
+    await test.step('Verify both environment variables are set in UI', async () => {
```
sanish-bruno, line 65: @sanjaikumar-bruno please add a clean up step were we will delete newly added variables, so that we can reset Stage.bru to it's old state.

tests/environments/api-setEnvVar/multiple-persist-vars.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { test, expect } from '../../../playwright';
```
sanish-bruno, line 1: @sanjaikumar-bruno this test will pass irrespective of whether we have enabled persist: true or not. cause even temporary variables are also visible on the UI. We have to check if the values are written to the bru file successfully

sanish-bruno (changes requested):

Issue #5810, opened 2025-10-16
In Bruno version 2.13.1, the method of req.setBody is used to report an error.


patelflow:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.13.1

### Operating System

Windows 10 19042.870

### Describe the bug

When I upgraded from version 2.12.0 to version 2.13.1, I got an error in using the req.setBody method in the pre-request.

Script
```
const data = {
  user_id: 123
}
req.setBody(data);
```

Response

> Pre-Request Script Error
> params is not iterable

### Screenshots/Live demo link

https://github.com/user-attachments/assets/282c20e9-f02f-4fbd-909b-c82cd158356e

helloanoop:
Thanks for reporting . We are looking into it right now.

sreelakshmi-bruno:
Hi @patelflow , to confirm, does this issue happen only when the body/content type is set to `form-url-encoded`?

bijin-bruno:
Hey @patelflow, thanks for reporting this!
We’ve addressed the issue in the PR https://github.com/usebruno/bruno/pull/5811.
The fix will be included in the upcoming patch release v2.13.2, which is expected to ship later tonight (IST), October 17, 2025.

Issue #5806, opened 2025-10-15
New Content-Type "fix" is breaking a GET call with a content-type Header


JohnnyTripod:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [x] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.13.0

### Operating System

macOS 15.7.1

### Describe the bug

For v2.13.0, there is a bug fix entitled "Don't set Content-Type on requests without a body". Instead, if a request with no body DOES have a Content-Type Header, instead of ignoring it, Bruno throws an error:

Pre-Request Script Error
params is not iterable

### Screenshots/Live demo link

Request:
▼
 Headers
(1)
> Content-Type:application/x-www-form-urlencoded
▼
 Body
No Body found

Response:
Error 00
▼
 Headers
No Headers found
▼
 Body
No Body found

helloanoop:
Thanks for reporting this. We are looking into it right now.

bijin-bruno:
Hey @JohnnyTripod  , thanks for reporting this!
We’ve addressed the issue in the PR https://github.com/usebruno/bruno/pull/5811.
The fix will be included in the upcoming patch release v2.13.2, which is expected to ship later tonight (IST), October 17, 2025.

PR #5811, opened 2025-10-16
fix: harden type checks for `buildFormUrlEncodedPayload`

barelyhuman:
# Description

Fixes: [#5806](https://github.com/usebruno/bruno/issues/5806)
Fixes: #5810

Adds in stricter type checking for off cases where the user's hand define headers as compared to them being defined by the UI

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/utils/form-data.js
+++ b/packages/bruno-electron/src/utils/form-data.js
@@ -13,7 +13,7 @@ const buildFormUrlEncodedPayload = (params) => {
   const resultParams = new URLSearchParams();
   for (const param of params) {
     // Invalid items are ignored
-    if (typeof param != 'object') continue;
+    if (typeof param !== 'object') continue;
     if (!('name' in param)) continue;
     resultParams.append(param.name, param.value ?? '');
   }


helloanoop:
@barelyhuman Can we add a test case in bruno testbench ?

barelyhuman:
> @barelyhuman Can we add a test case in bruno testbench ?

Already added cases specific to invalid types, don't really need one on the playwright setup

packages/bruno-electron/src/utils/form-data.js
```diff
@@ -8,8 +8,14 @@ const path = require('path');
  * @returns {string} Returns a order respecting standard compliant string of form encoded values
  */
 const buildFormUrlEncodedPayload = (params) => {
```
bijin-bruno, line 10:
```
      let requestObj = [
        { name: 'item' },
        { name: 'item', value: 3 }
      ];

      buildFormUrlEncodedPayload(requestObj);
```
current result = `'item=undefined&item=3'`
expected result = `// 'item=&item=3'`

When a key is present but its value is missing, the form-url-encoded string should use an empty value rather than "undefined"

barelyhuman: Nice catch, done

Issue #5805, opened 2025-10-15
Colon in request url causing failure


WillHoffmanFS:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

[Jira](https://usebruno.atlassian.net/browse/BRU-2020)

### Bruno version

2.13

### Operating System

MacOS 15.7.1 Apple Silicon

### Describe the bug

When I make a request that contains a colon in the url I get an error that shows part of the request URL is missing.
My request looks similar to this
  /start/1:2:AHLS-HASD/form
The error that comes back says
"java.lang.IllegalArgumentException: 1:AHLS-HASD is not valid"
Notice the "2:" is missing in the error message.  
If I modify the request by putting and extra 1: the the request works. ( i.e.  /start/1:1:2:AHLS-HASD/form)
This was working up through version 2.12 of Bruno

### .bru file to reproduce the bug

This was working up through version 2.12 of Bruno

### Screenshots/Live demo link

<img width="602" height="110" alt="Image" src="https://github.com/user-attachments/assets/fea6949b-d223-4572-ad81-c4f23d104bd4" />

sreelakshmi-bruno:
Hi, thanks for reporting. We're looking into this.

bijin-bruno:
Hey @WillHoffmanFS , thanks for reporting this!
We’ve addressed the issue in the PR https://github.com/usebruno/bruno/pull/5812.
The fix will be included in the upcoming patch release v2.13.2, which is expected to ship later tonight (IST), October 17, 2025.

PR #5812, opened 2025-10-16
fix: enhance URL parameter parsing and interpolation logic

Pragadesh-45:
fixes: #5805

# Description

- Improved handling of traditional path parameters and OData-style parameters in the URL utilities.
- Added regression tests to ensure embedded colons are not treated as path parameters.

### Contribution Checklist:

Before: 
image
After:
image

### DIFF
--- a/packages/bruno-app/src/utils/url/index.js
+++ b/packages/bruno-app/src/utils/url/index.js
@@ -41,24 +41,27 @@ export const parsePathParams = (url) => {
       if (name && !foundParams.has(name)) {
         foundParams.add(name);
       }
-    } else {
-      // for OData-style parameters (parameters inside parentheses)
-      // Check if segment matches valid OData syntax:
-      // 1. EntitySet('key') or EntitySet(key)
-      // 2. EntitySet(Key1=value1,Key2=value2)
-      // 3. Function(param=value)
-      if (/^[A-Za-z0-9_.-]+\([^)]*\)$/.test(segment)) {
-        const paramRegex = /[:](\w+)/g;
-        let match;
-        while ((match = paramRegex.exec(segment))) {
-          if (match[1]) {
-            let name = match[1].replace(/[')"`]+$/, '');
-            name = name.replace(/^[('"`]+/, '');
-            if (name && !foundParams.has(name)) {
-              foundParams.add(name);
-            }
-          }
-        }
+      return;
+    }
+
+    // for OData-style parameters (parameters inside parentheses)
+    // Check if segment matches valid OData syntax:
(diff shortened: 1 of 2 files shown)


sid-bruno:
There should be a test case that's checking `param='value'` somewhere in the odata tests, that is invalid and should be changed to use the query params with variables interpolation so that might need checking before we merge this in

packages/bruno-app/src/utils/url/index.js
```diff
@@ -33,18 +33,27 @@ export const parsePathParams = (url) => {
...
-          foundParams.add(name);
+    // traditional path parameters
+    if (segment.startsWith(':')) {
+      const name = segment.slice(1);
+      if (name && !foundParams.has(name)) {
+        foundParams.add(name);
+      }
+    } else {
+      // for OData-style parameters (parameters inside parentheses and quotes)
+      if (segment.includes('(') || segment.includes(')') || segment.includes('\'') || segment.includes('"') || segment.includes('`')) {
```
sid-bruno: `segment.includes('\'')` can you refer me to the part in the spec where this would be possible?

packages/bruno-app/src/utils/url/index.js
```diff
@@ -33,18 +33,27 @@ export const parsePathParams = (url) => {
...
-          foundParams.add(name);
+    // traditional path parameters
+    if (segment.startsWith(':')) {
+      const name = segment.slice(1);
+      if (name && !foundParams.has(name)) {
+        foundParams.add(name);
+      }
+    } else {
+      // for OData-style parameters (parameters inside parentheses)
+      if (segment.includes('(') || segment.includes(')')) {
```
bijin-bruno: Above check matches any word which contains '(' and ')' in any order.
I think we need to check if the segment matches to a valid odata syntax by using a regex like `^[A-Za-z0-9_.-]+\([^)]*\)$`

bijin-bruno (changes requested):

packages/bruno-app/src/utils/url/index.js
Pragadesh-45: yes, I just checked and updated the code from the spec. Thanks

packages/bruno-app/src/utils/url/index.spec.js
```diff
@@ -100,10 +96,6 @@ describe('Url Utils - parsePathParams', () => {
     expect(params).toEqual([{ name: 'categoryId', value: '' }]);
   });
 
-  it('should parse OData action with complex parameters', () => {
```
bijin-bruno, line 103: Any reason to remove these test cases? I think these are still valid cases

PR #6064, opened 2025-11-11
feat: add variable interpolation support for WebSocket requests

chirag-bruno:
# Description

- Add WebSocket body interpolation in interpolateVars function
- Interpolate URL, headers, and all messages in request.body.ws array with full variable context
- Refactor sendWsRequest to use main process preparation (removes duplication)
- Add mode property to wsRequest object for proper request type detection
- Ensure consistent variable precedence matching HTTP/gRPC requests
- Centralise all interpolation logic in main process via prepareWsRequest

This PR addresses this [JIRA](https://usebruno.atlassian.net/browse/BRU-2044) ticket.

### Contribution Checklist:

### DIFF
--- a/packages/bruno-electron/src/ipc/network/ws-event-handlers.js
+++ b/packages/bruno-electron/src/ipc/network/ws-event-handlers.js
@@ -284,7 +284,7 @@ const registerWsEventHandlers = (window) => {
       try {
         const itemCopy = cloneDeep(item);
         const preparedRequest = await prepareWsRequest(itemCopy, collection, environment, runtimeVariables, {});
-        
+
         // If messageContent is provided, find and queue that specific message (interpolated)
         // Otherwise, queue all messages
         if (messageContent !== undefined && messageContent !== null) {
@@ -309,7 +309,7 @@ const registerWsEventHandlers = (window) => {
               });
           }
         }
-        
+
         return { success: true };
       } catch (error) {
         console.error('Error queuing WebSocket message:', error);


packages/bruno-electron/src/ipc/network/ws-event-handlers.js
```diff
@@ -65,6 +67,7 @@ const prepareWsRequest = async (item, collection, environment, runtimeVariables,
 
   let wsRequest = {
     uid: item.uid,
+    mode: request.body.mode,
```
sid-bruno, line 70: Not needed, `mode` for `ws` is always `ws`

packages/bruno-electron/src/ipc/network/ws-event-handlers.js
```diff
@@ -286,6 +289,29 @@ const registerWsEventHandlers = (window) => {
     }
   });
 
+  // Prepare and queue WebSocket messages with proper variable interpolation
+  ipcMain.handle('renderer:ws:prepare-and-queue-messages',
```
sid-bruno: Doesn't need a new handler, modify the original `rendere:queue-message` handle with the needed interpolation methods

sid-bruno:
Add tests in playwright for this as well

packages/bruno-electron/src/ipc/network/ws-event-handlers.js
chirag-bruno: This is used in the interpolate-vars.js to parse JSON. I am adding it here to keep it in sync with gRPC's check implementation.

packages/bruno-electron/src/ipc/network/ws-event-handlers.js
chirag-bruno: The `__renderer:ws:queue-message__` channel calls the `queueMessage` function from the `wsClient`. However, this function only queues the request—it does not handle query preparation. This means that using the available IPC channel (`__renderer:ws:queue-message__`) doesn’t meet our requirement, since we need to both prepare and queue the request.

Additionally, this channel is invoked by `__queueWsMessage()__`, but that function is not currently called by any handlers, making it effectively unused. In other words, the current implementation of the `__renderer:ws:queue-message__` channel isn’t being utilized anywhere, and the parts of the code where we actually need it have specific prerequisites.

* For setting up a WebSocket connection: `prepare → establish connection → queue message`
* For sending WebSocket messages: `prepare → get available connection → queue message`

Please let me know if I’ve missed anything or if you see any potential improvements.

packages/bruno-electron/src/ipc/network/ws-event-handlers.js
sid-bruno: nvm, needed for mode detection for interpolation specific logic

Issue #6365, opened 2025-12-09
Bruno does generate code if URL contains a parameter referring a collection variable


tools400:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.15.1

### Operating System

Windows 11 Enterprise / 10.0.22631 Build 22631

### Describe the bug

Bruno does not generate cUrl code if the URL contains a parameter `:language` that refers to an environment variable `{{current_language}}`.

### .bru file to reproduce the bug

[Bruno Collection.bru.txt](https://github.com/user-attachments/files/24054346/Bruno.Collection.bru.txt)
[Bug Report.bru.txt](https://github.com/user-attachments/files/24054347/Bug.Report.bru.txt)

### Screenshots/Live demo link

Bruno Request:
<img width="1295" height="951" alt="Image" src="https://github.com/user-attachments/assets/118c6c4a-5284-4fc1-bfc9-f01fbe2ed04b" />

Generate cUrl Error:
<img width="801" height="548" alt="Image" src="https://github.com/user-attachments/assets/f8d0c43c-0d98-40dc-adb1-c6f5f73595ce" />

Pragadesh44-Bruno:
@tools400 Thanks for reporting this. I can verify the issue, and we have started working on it

PR #6394, opened 2025-12-12
feat: enable variable interpolation in URL field for code generation

Pragadesh-45:
fixes: #6365 
### Description

This PR introduces support for variable interpolation in the URL field during code generation. This allows dynamic values to be inserted into URLs, enabling more flexible and customizable code generation processes. With this feature, users can now define variables in the URL field that get replaced with actual values at runtime

#### Contribution Checklist:

Before:
image
image

After:
image

## Summary by CodeRabbit

* **New Features**
  * Path parameters in URLs are now interpolated from collection variables and then normalized/encoded, so endpoints reflect substituted values.
* **Bug Fixes**
  * Added input validation and early-exit guards for URL interpolation to better handle missing or invalid inputs.
* **Tests**
  * Added tests covering path-param interpolation, variable substitution flow, and multiple-parameter scenarios.

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
@@ -2,7 +2,7 @@ import { buildHarRequest } from 'utils/codegenerator/har';
 import { getAuthHeaders } from 'utils/codegenerator/auth';
 import { getAllVariables, getTreePathFromCollectionToItem, mergeHeaders } from 'utils/collections/index';
 import { interpolateHeaders, interpolateBody } from './interpolation';
-import { interpolateUrl } from 'utils/url/index';
+import { interpolateUrl, normalizeAndEncodeUrl } from 'utils/url/index';
 import { get } from 'lodash';
 
 const generateSnippet = ({ language, item, collection, shouldInterpolate = false }) => {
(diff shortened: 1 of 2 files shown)


packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
```diff
@@ -24,8 +25,15 @@ const generateSnippet = ({ language, item, collection, shouldInterpolate = false
       headers = [...headers, ...authHeaders];
     }
 
-    // Interpolate headers and body if needed
+    // Interpolate headers, body and path params if needed
     if (shouldInterpolate) {
+      // Interpolate path parameters
+      request.url = interpolateUrlPathParams(
```
bijin-bruno: Per current implementation,
1. I think the `interpolateUrlPathParams` shouldn't be done here since it is expected to be resolved already from the parent component `GenerateCodeItem`.
2. Considering the above fact, variable interpolation shouldn't happen inside `interpolateUrlPathParams` since it should respect the setting called `shouldInterpolate`. `interpolateUrlPathParams` should only interpolate the path params.
3. I think we should revert other changes in this PR and just do interpolateUrl here.
```
request.url = interpolateUrl(request.url, variables);
```

packages/bruno-app/src/utils/url/index.js
```diff
@@ -99,6 +99,20 @@ export const interpolateUrl = ({ url, variables }) => {
 };
 
 export const interpolateUrlPathParams = (url, params) => {
+  if (!url || typeof url !== 'string') {
+    return url;
+  }
+
+  // If there are no params or no enabled path params, return the URL as-is
+  if (!params || !params.length) {
```
bijin-bruno, line 137: Though this is not the right place to do the url encoding, it seems like this method `interpolateUrlPathParams` is also doing the encoding job at the code block 156 to 171 as below.
```js
  let uri;

  if (!url.startsWith('http://') && !url.startsWith('https://')) {
    url = `http://${url}`;
  }

  try {
    uri = new URL(url);
  } catch (error) {
    // if the URL is invalid, return the URL as is
    return url;
  }

  const basePath = getInterpolatedBasePath(uri.pathname, params);

  return `${uri.origin}${basePath}${uri?.search || ''}`;
```
So without moving the encoding task out of this method, we cannot place these safety checks here since the encoding will break when there is no params included.

Issue #5658, opened 2025-09-29
Unable to retrieve cookies on redirect in CLI / runner


ahamayeah59:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.12.0

### Operating System

Windows 11

### Describe the bug

Hi there,

Trouble retrieving cookies from a redirect response (302) but only with CLI (and runner).

I have 3 request : 

1. GET an uri to obtain an saml_request
2. POST to another URI with saml_request in the body and obtain an ID as response 302. (I need to store the cookies here). I had to req.setMaxRedirects(0) to make the request run correctly
3. GET to the same URI with another endpoint using the stored ID (and obviously i have to use the stored cookies here)

When i launch each request manually i have no trouble but using runner or CLI request 3 throw me a 403. 

In request 2 I use the bru.cookies.jar to manipulate cookies and to show them in console. When launch manually it retrieve the cookies correctly :

<img width="847" height="232" alt="Image" src="https://github.com/user-attachments/assets/97c5df12-5888-40f1-94a2-cf6f1293f27d" />

When launch via runner or CLI :

<img width="742" height="238" alt="Image" src="https://github.com/user-attachments/assets/6776abd5-8f61-4e9c-9c52-4dfd39905c19" />

I read several issue like https://github.com/usebruno/bruno/issues/4942  or https://github.com/usebruno/bruno/issues/1023  but it don't seem to work

Thanks for the help !

### Screenshots/Live demo link

<img width="847" height="232" alt="Image" src="https://github.com/user-attachments/assets/97c5df12-5888-40f1-94a2-cf6f1293f27d" />

<img width="742" height="238" alt="Image" src="https://github.com/user-attachments/assets/6776abd5-8f61-4e9c-9c52-4dfd39905c19" />

ahamayeah59:
Any news about this bug ? Thanks

berichar-renaissance:
I am also seeing this issue.
Following https://github.com/usebruno/bruno/issues/1332 I am able to reproduce it using the same bit.ly link used as an example in that issue.
Here is a video showing me reproducing the issue with that link. If I run the requests manually it works, and the cookie is set. If I run it via the runner, the cookie is not set.

[...]

PR #6094, opened 2025-11-13
Save cookies on redirect response

JeroenVinke:
# Description

When the response is a redirect (e.g. 302) the set-cookie headers were not processed. 

This problem happened in the collection runner and the CLI.

Let me know if the change is too naive

Are there any automated tests I can add?

Fixes #5658 

### Contribution Checklist:

- [X] **The pull request only addresses one issue or adds one feature.**
- [X] **The pull request does not introduce any breaking changes**
- [X] **I have added screenshots or gifs to help explain the change if applicable.**
- [X] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [X] **Create an issue and link to the pull request.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cookie management for error responses. When cookie storage is enabled, the application now properly persists cookies received from failed HTTP requests. This enhancement ensures consistent session handling across both successful and unsuccessful API calls, providing more reliable authentication and session state management for users working with HTTP error scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5387, opened 2025-08-21
fix: store redirect cookies under initial request domain

pooja-bruno:
# Description

Fix incorrect cookie domain assignment on HTTP redirects.

[JIRA](https://usebruno.atlassian.net/browse/BRU-1426)

[Issue](https://github.com/usebruno/bruno/issues/5268)

### Contribution Checklist:

- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

https://github.com/user-attachments/assets/91867f2e-330e-4863-9b82-23516da96c35

PR #6413, opened 2025-12-15
Revert "Save cookies on redirect response"

bijin-bruno:
Reverts usebruno/bruno#6094 in favour of https://github.com/usebruno/bruno/pull/5387

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated cookie persistence handling to prevent automatic cookie saving when HTTP error responses occur. Cookies are now only stored for successful requests, ensuring cleaner session management and more predictable behavior when working with APIs.

### DIFF
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -546,10 +546,6 @@ const runSingleRequest = async function (
         err.response.dataBuffer = dataBuffer;
         response = err.response;
 
-        if (!options.disableCookies) {
-          saveCookies(request.url, err.response.headers);
-        }
-
         // Prevents the duration on leaking to the actual result
         responseTime = response.headers.get('request-duration');
         response.headers.delete('request-duration');
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -1417,10 +1417,6 @@ const registerNetworkIpc = (mainWindow) => {
                 error.response.data = data;
                 error.response.dataBuffer = dataBuffer;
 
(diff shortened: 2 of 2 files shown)


JeroenVinke:
@sanish-bruno can issue https://github.com/usebruno/bruno/issues/5658 be reopened? I have looked into this a bit further based on the feedback you gave me. Thanks for that, it was very helpful! 

It seems that the problem occurs when followredirect = false (which internally means setMaxRedirects(0)). When the server responds with redirect and set-cookie headers, the request is reported as an error, and the cookies are not stored.

When running a single request from electron, the cookies ARE stored after the redirect, but not when running from the CLI or running a folder from electron. So there is a difference in behavior.

The core issue I think is that followredirect is not implemented correctly. Currently when followdirect = false, the result is reported as an error which is not true

image

https://github.com/usebruno/bruno/blob/639c8e573f3086c7e7ab478444058469a962a7fc/packages/bruno-electron/src/ipc/network/axios-instance.js#L280C1-L288C12

I believe the request should be reported as an success when followredirect = false, and the cookies should be stored.

Do you agree with my assessment, and if so, how do you suggest we get this fixed? Thank you for your help

sanishchirayath1:
@JeroenVinke Thanks for looking further into the issue, I guess we will have to re-open the issue. We are occupied with v3 release. If you strongly feel that 'follow redirects' is the core issue, You can create an Issue/PR. we will look into the issue! But do expect a delay! 

We understand that the bug is very annoying. We are extremely sorry that we are unable to prioritise right now. But I do believe, we can get this merged post the holidays! 

Thanks for understanding!

ahamayeah59:
No trouble if you don't have time, but please don't close the issue https://github.com/usebruno/bruno/issues/5658 because it's not done

PR #6288, opened 2025-12-03
fix: openapi body import

pooja-bruno:
### Description

Fix OpenAPI 3 import to handle requestBody `$ref` pointing to `#/components/requestBodies/...` where the referenced schema has properties.

[JIRA](https://usebruno.atlassian.net/browse/BRU-1792)
issue: https://github.com/usebruno/bruno/issues/4808

#### Contribution Checklist:

- [x] **I've used AI significantly to create this pull request**
- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [ ] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Broadened detection of object-like OpenAPI request bodies: schemas with properties (even without explicit "object" type) now populate JSON payloads, form-urlencoded fields, and multipart fields. Arrays with object-like items and numeric defaults (integer/number) are handled more inclusively.

* **Tests**
  * Added tests covering JSON, form-urlencoded, multipart bodies, numeric defaults, nested objects in arrays, and nested property population.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6429, opened 2025-12-17
fix: OpenAPI import fails when securitySchemes are not defined

sanjaikumar-bruno:
[BRU-2347](https://usebruno.atlassian.net/browse/BRU-2347)

OpenAPI import fails with "e.global.security.getScheme is not a function" when importing specs that don't define securitySchemes.

### Cause  
The recent PR #6288 fixed a bug in the `getSecurity` function:

**Before:** `Object.keys(securitySchemes) === 0` (always false since arrays ≠ numbers)
**After:** `Object.keys(securitySchemes).length === 0` (correctly checks for empty)

However, when no `securitySchemes` are defined, the function returns `{ supported: [] }` without a `getScheme` method, causing the error when other code tries to call `getScheme()`.

### Fix
Ensure `getScheme` method always exists, returning `null` when no security schemes are defined:
- When `securitySchemes` is empty: `getScheme: () => null`
- When `securitySchemes` exists: `getScheme: (name) => securitySchemes[name] || null`

## Summary by CodeRabbit

* **Improvements**
  * Enhanced OpenAPI import to better handle security schemes, including bearer tokens, basic auth, API keys, and OAuth2 configurations.
  * Improved robustness when importing OpenAPI specifications with or without defined security schemes.

### DIFF
--- a/packages/bruno-converters/src/openapi/openapi-to-bruno.js
+++ b/packages/bruno-converters/src/openapi/openapi-to-bruno.js
@@ -925,24 +925,18 @@ const getDefaultUrl = (serverObject) => {
 
 const getSecurity = (apiSpec) => {
   let defaultSchemes = apiSpec.security || [];
-
   let securitySchemes = get(apiSpec, 'components.securitySchemes', {});
-  if (Object.keys(securitySchemes).length === 0) {
-    return {
-      supported: [],
-      getScheme: () => null
-    };
-  }
+
+  const hasSchemes = Object.keys(securitySchemes).length > 0;
 
   return {
-    supported: defaultSchemes.map((scheme) => {
-      var schemeName = Object.keys(scheme)[0];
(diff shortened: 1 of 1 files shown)


packages/bruno-converters/src/openapi/openapi-to-bruno.js
bijin-bruno, line 932: To avoid the maintainance overhead and have consistent return, let's remove the early return and manage a singe return as below
```
const hasSchemes = Object.keys(securitySchemes).length > 0;

return {
  supported: hasSchemes
    ? defaultSchemes
        .map((scheme) => securitySchemes[Object.keys(scheme)[0]])
        .filter(Boolean)
    : [],
  schemes: securitySchemes,
  getScheme: (schemeName) => securitySchemes[schemeName]
};
```

Issue #6513, opened 2025-12-26
SSE sequence is lost in the ipc layer


sid-bruno:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.15.x

### Operating System

*

### Describe the bug

When working with super fast SSE responses, the chunks might get re-ordered since they all might have the same timestamp and the delay between electron's ipc and the client might cause the chunks to get re-ordered.

Sequencing for this needs to be hardened at the electron level so that the client can sort it based on the hardened sequence. This impacts both Websocket and SSE in fast incoming messages. 

### Screenshots/Live demo link

<img width="1470" height="919" alt="Image" src="https://github.com/user-attachments/assets/d5244fd3-a2b7-4062-8b14-35e72fb8b377" />

mr-weed-88:
@sid-bruno raised a PR for the fix please have a look if it is what you are looking for
 
[fix: sse sequence in ipc layer #6569](https://github.com/usebruno/bruno/pull/6569)

sid-bruno:
Thanks, i'll review in the morning 

> @sid-bruno raised a PR for the fix please have a look if it is what you are looking for
>  
> [fix: sse sequence in ipc layer #6569](https://github.com/usebruno/bruno/pull/6569)

PR #6503, opened 2025-12-24
fix: websocket message scroll

pooja-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2395)

#### Contribution Checklist:

- [x] **I've used AI significantly to create this pull request**
- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [x] **Create an issue and link to the pull request.**

https://github.com/user-attachments/assets/f4eaab5f-74af-467a-9211-366ede7df30c

## Summary by CodeRabbit

* **New Features**
  * Added virtualized rendering for WebSocket message lists, improving performance when handling large numbers of messages.

* **Improvements**
  * Enhanced auto-scroll behavior with intelligent detection of user scrolling patterns.
  * Improved message expansion and collapse functionality for better control over message visibility.

* **Changes**
  * Removed sort order functionality from WebSocket response pane.
  * Updated styling for tab action icons with improved hover states and theme compatibility.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6569, opened 2025-12-30
fix: add a hard sequence in SSE and WS requests

ganja-smoker-8888:
#### Contribution Checklist:

Screenshot 2025-12-30 203549
Screenshot 2025-12-30 201742
Screenshot 2025-12-30 201618

This PR addresses issue : [#6513](https://github.com/usebruno/bruno/issues/6513)

When working with super fast SSE responses, the chunks might get re-ordered since they all might have the same timestamp and the delay between electron's ipc and the client might cause the chunks to get re-ordered.
Sequencing for this needs to be hardened at the electron level so that the client can sort it based on the hardened sequence. This impacts both Websocket and SSE in fast incoming messages.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ordering and focus logic for streamed messages so timeline and response panels show correct, stable sequence and highlighted items.

* **New Features**
  * Responses and timeline entries include explicit sequence and timestamp metadata for reliable chronological display.
  * Streaming events now preserve per-request sequence numbers, improving event history accuracy and rendering stability.
  * Right-side response controls (tabs/format selector) are hidden for streaming responses to avoid irrelevant controls.

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/index.js
@@ -225,7 +225,7 @@ const ResponsePane = ({ item, collection }) => {
           onClick={() => setShowScriptErrorCard(true)}
         />
       )}
-      {focusedTab?.responsePaneTab === 'response' && item?.response ? (
+      {focusedTab?.responsePaneTab === 'response' && item?.response && !(item.response?.stream ?? false) ? (
         <>
           {/* Result View Tabs (Visualizations + Response Format) */}
           <div className="result-view-tabs">


packages/bruno-electron/src/ipc/network/index.js
sid-bruno, line 1: Can we just have the sequence be incremented and sent, the rest adds a delay between the actual receive and the flush  that the user might not expect when they are testing their connection and APIs (we want to keep it as quick as possible to visually show the responses), the back pressure management and flushing logic can be removed, the UI hangs up because of how many items we render which is solved in another [PR](http://github.com/usebruno/bruno/pull/6503) with list virtualizations. 

Also we want to do the same hard seq logic in websockets as well.

mr-weed-88: Do you want me to remove  "Normalize line endings: CRLF / CR -> LF logic" as well because while testing i was getting these blank responses

Screenshot 2025-12-31 123506

mr-weed-88: https://github.com/user-attachments/assets/be60d58b-15f9-49f2-a0ad-5f373a6bb41d


https://github.com/user-attachments/assets/2cea77e1-f2c3-4dc6-85e2-42c7163859dc

Test responses for fast SSE

sid-bruno: The line ending normalisation should be okay, remove the flushing and back pressure stuff

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -1011,19 +1012,105 @@ const registerNetworkIpc = (mainWindow) => {
...
+        let id = null;
+        let eventType = null;
+
+        for (const line of block.split('\n')) {
+          if (!line) continue;
+          if (line.startsWith(':')) continue;
+
+          if (line.startsWith('data:')) dataLines.push(line.slice(5).trimStart());
+          else if (line.startsWith('id:')) id = line.slice(3).trimStart();
+          else if (line.startsWith('event:')) eventType = line.slice(6).trimStart();
```
sid-bruno: Don't manipulate the output, keep the prefixes, remember that it's a platform to test the API not to use or correct it.

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -1011,19 +1012,105 @@ const registerNetworkIpc = (mainWindow) => {
...
+
+          const block = raw.slice(0, cut);
+          raw = raw.slice(cut);
 
-        mainWindow.webContents.send('main:http-stream-end', { collectionUid, itemUid: item.uid });
-        deleteCancelToken(response.cancelTokenUid);
+          if (raw.startsWith('\r\n\r\n')) raw = raw.slice(4);
+          else if (raw.startsWith('\n\n')) raw = raw.slice(2);
+
+          parseSseBlock(block);
```
sid-bruno: Don't need the chunks to be split based on new lines, the stream server is already deciding on it's own chunking logic.

packages/bruno-electron/src/ipc/network/index.js
sid-bruno: And I understand the tick is doing the flush but you are recreating a ticker which isn't needed since the `data` object might already come from a nodejs stream object which already handles backpressure and that isn't needed as it's already slow because of the IPC, I don't want it to be even slower. 

Ref: https://github.com/axios/axios/blob/7373fbff24cd92ce650d99ff6f7fe08c2e2a0a04/lib/adapters/http.js#L397

mr-weed-88:
websockets sequencing test

https://github.com/user-attachments/assets/18996568-0cc3-4bc1-b4c9-e602fb5fe437

mr-weed-88:
websockets sequencing test with respect to each request

https://github.com/user-attachments/assets/38de6a77-a56d-440e-9493-212300838aa3

packages/bruno-electron/src/ipc/network/index.js
mr-weed-88: removed flushing and back pressure, added hard sequencing for both SSE and Websockets

packages/bruno-app/src/utils/network/ws-event-listeners.js
sid-bruno, line 1: These are on the React layer, the sequence would've been lost at this point, we needed this on the electron side of things.

mr-weed-88:
@sid-bruno i had bit problem in understanding the websockets can you please on confirm are you trying to say sequencing should be handled in bruno\packages\bruno-requests\src\ws\ws-client.js. i did try adding seq here in #addConnection with 

this.#addConnection(requestId, collectionUid, wsConnection, seq: 0);

but latter moved to reduxStore and function to increase sequence. It hard in understanding class based {} code block so is it somewhere else i need to look would help in understanding repo better

sid-bruno:
> @sid-bruno i had bit problem in understanding the websockets can you please on confirm are you trying to say sequencing should be handled in bruno\packages\bruno-requests\src\ws\ws-client.js. i did try adding seq here in #addConnection with
> 
> this.#addConnection(requestId, collectionUid, wsConnection, seq: 0);
> 
> but latter moved to reduxStore and function to increase sequence. It hard in understanding class based {} code block so is it somewhere else i need to look would help in understanding repo better

Yeah, it'll need a little work in ws since all connections are in one client, i'll handle it, i'm also making changes to remove additional stuff that AI might have done so I'll take up the PR since I need it ready before the next release

Hope you don't mind, if you'd like you can browse for other issues, the PR will still be merged with your authorship

PR #6776, opened 2026-01-11
fix:prevent JS hint leak on Ctrl+Space and show allowed root hints

### LINKED ISSUE #5871: In Script tab, CTRL+Space kb shortcut flashes the autocomplete popup

### This bug is:

- [x] annoying

### Bruno version

2.13.2

### Operating System

Windows 11 25H2 26200.6901, 25H2 26200.6899

### Describe the bug

1. Open Bruno
2. (Create/)Open any collection or request
3. Navigate to Script tab
4. Focus on either Pre Request or Post Response fields
5. Press CTRL+Space
The autocomplete window shows up for a split second, but immediately disappears.
I'd expect that it'd stay open, so that I can easily check the currently available public objects.

I have checked that this happens on a fresh install of Windows 11, where the only other two softwares that were installed were scoop.sh and Firefox (from Scoop). I have also installed Bruno from scoop, but the version seems to be up to date, and using the Installer/portable versions downloaded directly from this repo doesn't change the behaviour.
Additionally I have tested back to Bruno 2.9.1 x64 portable release. It seems like the issue was present back then as well.

For me this is a required feature, and made me instantly look for other software. I cannot rely on trial-and-error to figure out which objects are available.

_No response_

Happens in Developer and Safe mode, too.

https://github.com/user-attachments/assets/5c41b6d2-d94e-41d3-96c1-fe86e780953d

Abhinandan-pixel:
#### Root cause
In the Script editor, pressing Ctrl + Space when the cursor is at an empty position triggers autocomplete without a valid word context. Since the hint provider returns null in this case, CodeMirror immediately closes the completion popup, causing it to briefly flash and disappear.

#### Fix
This change explicitly handles Ctrl + Space when the editor is empty by showing root-level hints (API objects), while preserving the existing keyup-based autocomplete behavior for all other cases.
Additionally, context detection was improved to allow progressive discovery of API hints (r → re → req).

- Fixes #5871 
#### Contribution Checklist:

#### Ctrl+Space now shows root hints without flashing 
fix

## Summary by CodeRabbit

* **New Features**
  - Broader autocomplete context detection for more relevant API suggestions
  - Added an action to surface top-level completions on demand via keyboard shortcuts

* **Bug Fixes**
  - Refined autocomplete closing behavior so the completion panel only closes when appropriate based on cursor context

### DIFF
--- a/packages/bruno-app/src/utils/codemirror/autocomplete.spec.js
+++ b/packages/bruno-app/src/utils/codemirror/autocomplete.spec.js
@@ -482,7 +482,7 @@ describe('Bruno Autocomplete', () => {
         mockedCodemirror.state.completionActive = mockCompletion;
 
         mockedCodemirror.getCursor.mockReturnValue({ line: 0, ch: 0 });
-        mockedCodemirror.getLine.mockReturnValue('   ');
+        mockedCodemirror.getLine.mockReturnValue('req.bodyy');
         mockedCodemirror.getRange.mockReturnValue('');
 
         const mockEvent = { key: 'a' };
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -8,7 +8,7 @@
 import React from 'react';
 import { isEqual, escapeRegExp } from 'lodash';
 import { defineCodeMirrorBrunoVariablesMode } from 'utils/common/codemirror';
-import { setupAutoComplete } from 'utils/codemirror/autocomplete';
+import { setupAutoComplete, showRootHints } from 'utils/codemirror/autocomplete';
 import StyledWrapper from './StyledWrapper';
 import * as jsonlint from '@prantlf/jsonlint';
 import { JSHINT } from 'jshint';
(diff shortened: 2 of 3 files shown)


packages/bruno-app/src/utils/codemirror/autocomplete.spec.js
```diff
@@ -482,7 +482,7 @@ describe('Bruno Autocomplete', () => {
         mockedCodemirror.state.completionActive = mockCompletion;
 
         mockedCodemirror.getCursor.mockReturnValue({ line: 0, ch: 0 });
-        mockedCodemirror.getLine.mockReturnValue('   ');
+        mockedCodemirror.getLine.mockReturnValue('req.bodyy');
```
edomato-e2e, line 485: Seems there's a typo here, don't know how it could affect the feature (maybe the mock returns the wrong value?).

Abhinandan-pixel: An empty value (' ') is no longer valid because I changed the close condition to:

if (cm.state.completionActive && wordInfo) {
  cm.state.completionActive.close();
}

When the editor is empty, wordInfo is null. If we allowed the popup to close in that case, the root hints shown on Ctrl + Space would immediately be closed by the keyUp handler (which runs after the Ctrl + Space handler). That’s exactly the bug this PR is fixing.

So the behavior is now:
- Empty editor + Ctrl + Space → show root hints and do not close them
- User types some text (e.g. req.bodyy) with no matching hints → wordInfo exists, so we do close the popup

The test was updated to reflect this new logic by using a non-matching word instead of an empty value, ensuring we only close the popup when the user has actually typed something.

PR #6868, opened 2026-01-21
Fix/ephemeral env vars disappearing

sanjaikumar-bruno:
[BRU-2444](https://usebruno.atlassian.net/browse/BRU-2444)

**Issue**
Ephemeral variables created via `bru.setEnvVar()` disappeared from the environment modal when:
1. A user added a new variable in the modal and saved, or
2. The environment file was updated by the file watcher.
These variables are script-created and not persisted to disk, but should remain visible in the UI.

**Cause**
Two reducers were overwriting the environment variables array:
1. `saveEnvironment` — replaced all variables with only persisted ones, dropping ephemerals.
2. `collectionAddEnvFileEvent` — reloaded from file and only restored ephemerals that existed on disk, losing script-only ephemerals.

**Solution**
**Fix-1: `saveEnvironment` reducer**
- Preserves ephemeral variables that aren't in the new persisted set.
- Merges persisted variables with existing ephemerals instead of replacing.

**Fix-2: `collectionAddEnvFileEvent` reducer**
- Re-adds purely ephemeral variables (script-created, not on disk) after file updates.
- Ensures script-only ephemerals remain visible after file watcher updates.


packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -255,7 +255,13 @@ export const collectionsSlice = createSlice({
         const environment = findEnvironmentInCollection(collection, environmentUid);
 
         if (environment) {
-          environment.variables = variables;
+          // Preserve ephemeral variables that aren't in the new persisted set
+          const newVarNames = new Set(variables.map((v) => v.name));
```
sanish-bruno, line 259: @sanjaikumar-bruno I think we are currently supporting variables with the same name within the environments, how this code will behave in that case?

sanish-bruno:
@coderabbitai review

Issue #6913, opened 2026-01-23
Searching in JSON is very slow and is causing a memory leak


dennheer:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

3.02

### Operating System

Windows 11 10.0.26200 Build 26200

### Describe the bug

When searching in a JSON request, the search is very slow and causes the app to be inresponsive. When looking in Windows TaskManager, I see that the memory used by the application is increasing rapidly from a few hundred MB to several GB's. 

### .bru file to reproduce the bug

[upsertClaim.bru.txt](https://github.com/user-attachments/files/24820743/upsertClaim.bru.txt)

### Screenshots/Live demo link

Trying to search in the request : Typed PolicyNumber before becoming inresponsive.

<img width="740" height="754" alt="Image" src="https://github.com/user-attachments/assets/c574d004-bbb3-4c95-a712-9cbd3a69a0d6" />

Taskmanager : fast increasing memory usage 

<img width="1105" height="222" alt="Image" src="https://github.com/user-attachments/assets/ec8edf25-d240-4308-88fd-b21789e12a12" />

Keeps increasing rapidly. Application is still inresponsive
<img width="1073" height="225" alt="Image" src="https://github.com/user-attachments/assets/9a694350-5df6-495f-9dca-ceb4d4f6f8c1" />

PR #6920, opened 2026-01-25
perf: Improve search performance in code editor

Its-treason:
Added cache key to reduce duplicate searches over the complete text. Added incremental update logic, to not update every marked entry when clicking "next" or "previous". Limited the maximum results to 99,999 to prevent long loading times.

Fixes: https://github.com/usebruno/bruno/issues/6913
[JIRA](https://usebruno.atlassian.net/browse/BRU-2773)

### Before

https://github.com/user-attachments/assets/010f7c0f-204f-4ebc-b702-953299bff22a

### After

https://github.com/user-attachments/assets/cf29dd9e-0b8f-4b67-af30-119339648cec

## Summary by CodeRabbit

* **New Features**
  * Adds search result caching for faster repeated queries and navigation.

* **Performance**
  * Reuses cached matches to reduce work during navigation, improving responsiveness.
  * Input debounce increased for steadier searching.

* **Bug Fixes**
  * Resets search state on close or error; current match reliably highlighted, scrolled into view, and Next/Prev navigation consistent.

* **Refactor**
  * Simplifies synchronization between input, visibility, and cached results.


packages/bruno-app/src/components/CodeMirrorSearch/index.js
```diff
@@ -8,6 +8,44 @@ function escapeRegExp(string) {
...
+    return out;
+  } catch (e) {
+    console.error('Search error:', e);
+    return [];
+  }
+}
+
+function createCacheKey(editor, searchText, regex, caseSensitive, wholeWord) {
+  return `${editor.getValue().length}-${searchText}-${regex}-${caseSensitive}-${wholeWord}`;
+}
```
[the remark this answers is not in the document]
Its-treason: `changeGeneration` must still be implemented, it's always 1. And just using the whole text as key is too much memory coping for my liking. The size will be different in 99.9% percent of cases

packages/bruno-app/src/components/CodeMirrorSearch/index.js
```diff
@@ -8,6 +8,44 @@ function escapeRegExp(string) {
...
+    }
+    return out;
+  } catch (e) {
+    console.error('Search error:', e);
+    return [];
+  }
+}
+
+function createCacheKey(editor, searchText, regex, caseSensitive, wholeWord) {
+  return `${editor.getValue().length}-${searchText}-${regex}-${caseSensitive}-${wholeWord}`;
```
sid-bruno: I'd change the delimiter to be something like `⇴` , as the chances of something conflict with a unicode symbol is rare compared to `-`,  while there's no immediate use for it right now splitting the key would be ambiguous

Any another special symbol works

sid-bruno:
rest LGTM.

packages/bruno-app/src/components/CodeMirrorSearch/index.js
Its-treason: I updated the delimeter in the newest version

Its-treason:
@sid-bruno Sorry for the late reply, I did not see the ping in my notifications. I fixed the merge conflict, but couldn't test the newest changes, because of an error on the main branch:

grafik

And after dismissing that & sending a request:

grafik

I made sure the request works on Bruno version 3.1.1

sid-bruno:
@Its-treason , he could be an issue with your `pnpm` setup where it's resolving the package registry version instead of the local one and if it's resolving the local one can you run a `pnpm -r run build` for all the projects to see if it's an issue with that. 

I can see the function in `@usebruno/common` , can you check that and validate if you see it in the source files of `packages/bruno-common` . 

If yes, the above mentioned build should help you, if no then you might need to rebase changes from `main` 

--- 

If you are using a separate fork that uses `npm` only, then just run `npm run setup` and it should take care of building everything for you

Its-treason:
@sid-bruno Thanks, the `npm run setup` fixed the problems. I tested the PR (everything still works) and updated the branch

PR #6929, opened 2026-01-26
fix: enable SSL session caching for faster consecutive requests

### LINKED ISSUE #5574: Bruno is not caching SSL sessions, leading to slow consecutive requests

### This bug is:

- [ ] annoying

### Bruno version

2.11.0

### Operating System

Windows 10/Windows 11

### Describe the bug

💥 When making multiple consecutive requests to the same endpoint, Bruno does not appear to be caching the SSL session. This results in the SSL handshake being performed for every request, which significantly increases the total response time.

In my test, the first request took over 700 ms, with the SSL handshake alone accounting for about 450 ms. The second request, performed immediately after the first, also performed a full SSL handshake, leading to a similar overall response time.

For comparison, Postman caches the SSL session on the second request, reducing the overall time to around 300 ms.

Steps to Reproduce

Make a GET request to any secure endpoint, for example, https://api.nationalize.io/?name=nathaniel.

Observe the full response time, including the SSL handshake duration in the timeline.

Immediately make the exact same request again without any changes.

Analyze the response time of the second request.

**Expected behavior**

The SSL session should be cached after the first request. The second and subsequent requests should reuse the existing session, skipping the SSL handshake and resulting in a much faster response time.

**Actual behavior**

The SSL handshake is performed for every request, regardless of a previous successful connection. This adds a significant delay to consecutive requests.

curl --request GET \
  --url 'https://api.nationalize.io/?name=nathaniel'

Bruno
Image
Image

Postman
Image
Image
Image

[JIRA](https://usebruno.atlassian.net/browse/BRU-1876)

kxbnb:
## Summary

Fixes the SSL session caching issue where consecutive requests to the same endpoint performed a full TLS handshake every time, adding ~450ms overhead per request.

## Problem

Bruno was creating a new `https.Agent` for every request. Since Node.js agents cache TLS sessions internally, creating fresh agents meant no session reuse was possible. Postman and other clients reuse agents, which is why their second request is much faster.

## Solution

- Added an agent cache keyed by TLS configuration (certificates, proxy settings, SSL options)
- Agents are now reused when their configuration matches
- TLS session tickets are cached and reused automatically by Node.js

## Changes

- `packages/bruno-electron/src/utils/proxy-util.js`:
  - Added `agentCache` Map for storing reusable agents
  - Added `getAgentCacheKey()` to generate cache keys from TLS options
  - Added `getOrCreateAgent()` to retrieve or create cached agents
  - Updated `setupProxyAgents()` to use cached agents
  - Added `clearAgentCache()` and `getAgentCacheSize()` for testing/debugging

## Testing

1. Make a GET request to any HTTPS endpoint (e.g., `https://api.nationalize.io/?name=nathaniel`)
2. Note the response time (~700ms with ~450ms SSL handshake)
3. Make the same request again immediately
4. Second request should now be significantly faster (SSL session reused)

## Backward Compatibility

- No breaking changes to existing functionality
- Timeline logging preserved (moved to setup phase)
- Proxy and SSL validation behavior unchanged

Fixes #5574

## Summary by CodeRabbit

* **Chores**
  * Implemented SSL session reuse caching to optimize network connections and reduce overhead for repeated requests.
  * Improved proxy and connection handling with agent reuse and keep‑alive to enhance performance and stability.
  * Added connection diagnostics and proxy usage logging to surface validation and proxy activity for troubleshooting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ bruno.iml
 
 # Development plan files
 CLAUDE.md
+AGENTS.md
 *.plan.md
 
 # packages dist
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,7 +1,0 @@
-{
-  "trailingComma": "none",
-  "tabWidth": 2,
-  "semi": true,
-  "singleQuote": true,
-  "printWidth": 120
-}
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -66,7 +66,16 @@ Remember, these rules are here to make our codebase harmonious. If something doe
 
 - Use styled component's theme prop to manage CSS colors and not CSS variables when in the context of a styled component or any react component using the styled component 
 - Styled Components are used as wrappers to define both self and children components style, tailwind classes are used specifically for layout based styles. 
-- Styled Component CSS might also change layout but tailwind classes shouldn't define colors. 
+- Styled Component CSS might also change layout but tailwind classes shouldn't define colors.
+- MUST: Prefer custom hooks for business logic, data fetching, and side-effects.
+- MUST: Avoid `useEffect` unless absolutely needed. Prefer derived state, event handlers.
+- SHOULD: Memoize only when necessary (`useMemo`/`useCallback`), and prefer moving logic into hooks first.
+- MUST: Do not use namespace access for hooks in app code (e.g., `React.useCallback`, `React.useMemo`, `React.useState`). Import hooks directly.
+  - Correct: `import { useCallback, useMemo, useState } from "react";`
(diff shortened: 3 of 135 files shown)


kxbnb:
Thanks for the review! Addressed both items:

1. **Passphrase in cache key** - Now included (hashed like other secrets) so agents with different passphrases aren't incorrectly reused.

2. **Bounded cache with LRU eviction** - Added `MAX_AGENT_CACHE_SIZE = 100` with LRU-style eviction. On cache hit, entries are moved to the end; on insert when full, the oldest entry is evicted.

Also fixed the consistent `node:` prefix for the crypto import.

PR #6975, opened 2026-01-29
Performance/file parse and mount

chirag-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2407)

This PR addresses performance issues encountered when mounting large collections.

In the previous implementation, every file in a collection went through an expensive processing pipeline:

```
chokidar addFile event
→ Classify file
→ Parse file
→ Notify UI
→ Re-render UI
```

For large collections (e.g. ~3000 requests), this resulted in thousands of rapid re-renders, causing the UI to become sluggish and temporarily unresponsive until all files were processed.

#### What’s changed

**1. Introduced a caching layer**

We now cache parsed files and only re-parse them when necessary:

```
chokidar addFile event
→ Classify file
→ Check cache
   → If cache hit and mtime matches, return cached result
   → Else, parse file and update cache
```

This ensures that unchanged files are served directly from cache, significantly reducing redundant parsing work.

**2. Batched UI updates**

Previously, each file triggered its own UI re-render. In the new implementation, UI updates are performed in batches, triggering a single re-render per batch instead of per file.

This greatly reduces renderer load and results in a much smoother and more responsive UI, especially for large collections.

Overall, these changes dramatically improve collection mount performance and user experience without altering existing behavior.

#### Contribution Checklist:

* [x] **I've used AI significantly to create this pull request**
* [x] **The pull request only addresses one issue or adds one feature.**
* [x] **The pull request does not introduce any breaking changes**
* [ ] **I have added screenshots or gifs to help explain the change if applicable.**
* [x] **I have read the [[contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md)](https://github.com/usebruno/bruno/blob/main/contributing.md).**
* [ ] **Create an issue and link to the pull request.**

JIRA: https://usebruno.atlassian.net/browse/BRU-2407

## Summary by CodeRabbit

* **New Features**
  * Added cache management settings in Preferences with visibility into cached items and cache version
  * Ability to purge cache directly from application preferences

* **Performance**
  * Improved collection discovery speed with file parsing cache and batched updates
  * Enhanced loading state indicators across the application for better user feedback

### DIFF
--- a/packages/bruno-electron/tests/app/collection-tree-batcher.spec.js
+++ b/packages/bruno-electron/tests/app/collection-tree-batcher.spec.js
@@ -6,7 +6,7 @@ const createMockWindow = (id = 1) => {
   return {
     id,
     isDestroyed: jest.fn(() => false),
-    on: jest.fn((event, callback) => {
+    once: jest.fn((event, callback) => {
       listeners[event] = callback;
     }),
     emit: (event) => {


packages/bruno-electron/src/store/parsed-file-cache.js
```diff
@@ -0,0 +1,398 @@
...
+      const prefix = `${collectionPath}\0${dirPath}`;
+      const keysToDelete = [];
+
+      // Find all keys that start with this directory prefix
+      for (const { key } of this.db.getRange({ start: prefix })) {
+        // Check if key still starts with our prefix (getRange might go beyond)
+        if (!key.startsWith(prefix)) {
+          break;
+        }
+        keysToDelete.push(key);
```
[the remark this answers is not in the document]
chirag-bruno: Done

packages/bruno-electron/src/app/collection-watcher.js
```diff
@@ -335,61 +338,71 @@ const add = async (win, pathname, collectionUid, collectionPath, useWorkerThread
       }
     };
 
-    const fileStats = fs.statSync(pathname);
-    let content = fs.readFileSync(pathname, 'utf8');
+    const batchAggregator = getAggregator(win);
```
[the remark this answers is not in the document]
chirag-bruno: Done

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -2681,6 +2681,149 @@ export const collectionsSlice = createSlice({
...
+                examples: file.data.examples,
+                filename: file.meta.name,
+                pathname: file.meta.pathname,
+                draft: null,
+                partial: file.partial,
+                loading: file.loading,
+                size: file.size,
+                error: file.error
+              });
+            }
```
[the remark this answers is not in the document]
chirag-bruno: Done

packages/bruno-app/src/store/parsedFileCache.js
```diff
@@ -0,0 +1,187 @@
...
+          store.createIndex('parsedAt', 'parsedAt');
+        }
+      }
+    });
+  }
+  return dbPromise;
+};
+
+const generateKey = (collectionPath, filePath) => {
+  return `${collectionPath}\n${filePath}`;
```
sid-bruno: Use a more unique separator to avoid `\n` get confudes with a windows path where `\newFile` would also be a part of the key causing an invalid split

```suggestion
  return `${collectionPath}↝${filePath}`;
```

packages/bruno-electron/src/app/collection-tree-batcher.js
sid-bruno, line 1: 1. Tests 
2. Since it's so tied to the file tree update, I'd rename it from `batch-aggregator` to a more relevant name in both usage and the file name

packages/bruno-electron/src/store/parsed-file-cache-idb.js
sid-bruno, line 1: Not sure if we want to just error out and forget about the cache, the user might end up with a 30 day long cache while the latest collection that's being used is not being cached because IndexedDB might become full. 

I'd rather have is as a LRU instead of a linear cache like it is right now. Open to discussing

packages/bruno-app/src/store/parsedFileCache.js
cchirag: Done

packages/bruno-electron/src/store/parsed-file-cache-idb.js
chirag-bruno: IndexedDB quotas are typically 50-60% of disk space (hundreds of MB to GB). For small parsed file entries, hitting the limit is rare. We already provide a 'Clear Cache' option. Instead of LRU complexity (which adds writes on every read), we could:             
-  Reduce default TTL from 30 to 15 days                                                                                            
 - Add a configurable cache duration setting in the Cache UI                                                                        
                                                                                                                                      
  This keeps the implementation simple while giving users control. Thoughts?

packages/bruno-app/src/store/parsedFileCache.js
```diff
@@ -0,0 +1,187 @@
...
+        if (cursor.value.filePath.startsWith(dirPath)) {
+          await cursor.delete();
+        }
+        cursor = await cursor.continue();
+      }
+      await tx.done;
+    } catch (error) {
+      console.error('ParsedFileCacheStore: Error invalidating directory cache:', error);
+    }
+  },
```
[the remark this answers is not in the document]
sid-bruno: @coderabbitai, the cache is only going to have top level collections, and invalidations will be for the same, so this should be fine

Issue #7006, opened 2026-02-01
`JSON.Stringify` in `bru.setVar` in GraphQL variable acting weirdly


arxeiss:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

UI 3.0.2 & CLI 3.0.3

### Operating System

Ubuntu 24.04

### Describe the bug

We have Pre request script, where we build dynamically JSON string, which is then passed into GraphQL variable. We use `JSON.Stringify` to convert JS object into string, which is stored into variable.

However, there is inconsistency between Bruno UI and `bru` CLI. While we have to escape `"` with `'\\"'` for CLI, we need double escaping `'\\\\\\"'` for Bruno UI to achieve same thing. Otherwise getting cryptic message from UI.

See bru file and screenshots

### .bru file to reproduce the bug

```bru
meta {
  name: gqltest
  type: graphql
  seq: 19
}

post {
  url: https://echo.free.beeceptor.com
  body: graphql
  auth: inherit
}

body:graphql {
  mutation store(
        $my_json: String!
    ) {
        store(
            input: {
                source: $my_json
            }
        ) {
            id
        }
    }
}

body:graphql:vars {
  {
    "my_json": "{{my_json}}"
  }
}

script:pre-request {
  const testData = {
    a: [1,2,3],
    b: {
      c: "test",
      d: "another value"
    }
  };
  
  // Single escaping
  let cv = JSON.stringify(testData).replace(/"/g, '\\"');
  // Double escaping
  let cv = JSON.stringify(testData).replace(/"/g, '\\\\\\"');
  console.log(cv);
  bru.setVar("my_json", cv)
}

settings {
  encodeUrl: true
  timeout: 0
}
```

### Screenshots/Live demo link

P.S. For testing I'm using https://echo.free.beeceptor.com which just returns whatever we sent to them.

Bruno UI with **single escaping**:
<img width="1535" height="770" alt="Image" src="https://github.com/user-attachments/assets/8280d402-7977-42e4-9622-226cb1b490af" />

Bruno UI with **double escaping**:
<img width="1529" height="766" alt="Image" src="https://github.com/user-attachments/assets/424e58ab-b337-4505-981d-56fb17adc8be" />

Bruno CLI with **single escaping** 
[...]

PR #7041, opened 2026-02-04
fix: graphQL variables escaping inconsistency between UI and CLI

sanjaikumar-bruno:
Fixes https://github.com/usebruno/bruno/issues/7006

### DIFF
--- a/packages/bruno-common/src/interpolate/index.ts
+++ b/packages/bruno-common/src/interpolate/index.ts
@@ -75,6 +75,7 @@ const interpolate = (
   }
 
   const { escapeJSONStrings } = options;
+  const escapeJson = escapeJSONStrings ?? false;
 
   const preparedStr = prepareMock(str, escapeJSONStrings ?? false);
 
@@ -83,7 +84,7 @@ const interpolate = (
   }
   // process the object with the mock data functions
   const preparedObj = prepareMockObj(obj, escapeJSONStrings ?? false);
-  return replace(preparedStr, preparedObj, new Set(), new Map(), { escapeJSONStrings: escapeJSONStrings ?? false });
+  return replace(preparedStr, preparedObj, new Set(), new Map(), { escapeJSONStrings: escapeJson });
 };
 
 const replace = (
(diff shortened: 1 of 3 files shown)


packages/bruno-cli/src/runner/prepare-request.js
```diff
@@ -371,7 +371,7 @@ const prepareRequest = async (item = {}, collection = {}) => {
   if (request.body.mode === 'graphql') {
     const graphqlQuery = {
       query: get(request, 'body.graphql.query'),
-      variables: JSON.parse(decomment(get(request, 'body.graphql.variables') || '{}'))
```
sanish-bruno, line 374: @sanjaikumar-bruno add unit test cases to verify the behaviour within bruno-cli as well as in the APP

sanjaikumar-bruno: Updated the PR with tests.

packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -286,28 +238,33 @@ const runSingleRequest = async function (
...
       }
+
+      preRequestTestResults = result?.results || [];
     }
 
     // interpolate variables inside request
     interpolateVars(request, envVariables, runtimeVariables, processEnvVars);
 
+    // parse GraphQL variables after interpolation
+    if (request.mode === 'graphql') {
```
sanish-bruno, line 264: @sanjaikumar-bruno do we really need to pass the request.mode to this file? we removed `JSON.parse` step from `packages/bruno-cli/src/runner/prepare-request.js` to be placed here? did we really needed changes within these files?

packages/bruno-common/src/interpolate/index.ts
```diff
@@ -110,15 +113,26 @@ const replace = (
...
+        const result = replace(replacement, obj, visited, results, options);
         results.set(match, result);
 
         matchFound = true;
         return result;
       }
 
       visited.add(match);
-      const result = replacement !== undefined ? replacement : match;
+      let result = replacement !== undefined ? replacement : match;
```
sanish-bruno, line 124: @sanjaikumar-bruno considering single responsibility principle, probably we might shouldn't be doing `escapeJSONString` here, the `replace` function should only have on duty, replace the any variable with it's value. we may need to find some other place to do the `escapeJSONString` part

sanjaikumar-bruno:
Raised a new PR with updated implementation,

https://github.com/usebruno/bruno/pull/7049

PR #7049, opened 2026-02-05
fix: graphQL variables interpolation consistency (UI and CLI)

sanjaikumar-bruno:
[BRU-2664](https://usebruno.atlassian.net/browse/BRU-2664)
Fixes https://github.com/usebruno/bruno/issues/7006

### DIFF
--- a/packages/bruno-cli/src/runner/interpolate-vars.js
+++ b/packages/bruno-cli/src/runner/interpolate-vars.js
@@ -79,7 +79,9 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
   if (!isGraphqlRequest) {
     if (contentType.includes('json') && !Buffer.isBuffer(request.data)) {
       if (typeof request.data === 'string') {
-        request.data = _interpolate(request.data, { escapeJSONStrings: true });
+        if (request?.data?.length) {
+          request.data = _interpolate(request.data, { escapeJSONStrings: true });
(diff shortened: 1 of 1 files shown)


packages/bruno-cli/src/runner/interpolate-vars.js
```diff
@@ -67,8 +67,21 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
   });
 
   const contentType = getContentType(request.headers);
+  const isGraphqlRequest = request.mode === 'graphql';
+
+  // GraphQL: interpolate the variables string (and query) only; do not stringify the whole body.
+  // So {{var}} sits in one JSON string (the variables text). Scripts use single escape (e.g. .replace(/"/g, '\\"')).
+  // Full-body stringify would nest that inside the outer "variables" value → two layers → double escape.
+  if (isGraphqlRequest && request.data != null && typeof request.data === 'object' && typeof request.data.variables === 'string') {
```
bijin-bruno: 1. Please remove the irrelevant comments above
2. Any reason to have `&& typeof request.data.variables === 'string'` check here?
- What is the case variables is not a string for a graphql request at this point?
- We should be able to process and send gql queries without variables
3. Also verify the value of content-type to make sure it wont enter any blocks below

bijin-bruno (changes requested):

PR #7091, opened 2026-02-10
fix: openapi content level example

pooja-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2667)

Fix openapi content level example

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Request bodies now derive examples from content-level examples when schema-level examples are absent, improving generated request payloads across JSON, XML, text, form, multipart, SPARQL and other media types.

* **Tests**
  * Expanded test coverage validating example priority rules and content-level example handling across many media types and request body formats.

### DIFF
--- a/packages/bruno-converters/tests/openapi/openapi-to-bruno/openapi-body.spec.js
+++ b/packages/bruno-converters/tests/openapi/openapi-to-bruno/openapi-body.spec.js
@@ -908,42 +908,224 @@ paths:
     // Should produce an empty array
     expect(request.request.body.json).toBe('[]');
   });
+});
 
-  it('should use content-level examples for text/plain requestBody', () => {
-    const openApiSpec = `
+describe('content-level example vs examples priority', () => {
+  it('should prefer singular example over examples (plural) and fall back to examples when example is absent', () => {
+    const spec = `
 openapi: "3.1.0"
 info:
   version: "1.0.0"
-  title: "Text Body Examples Test"
+  title: "Test"
 servers:
   - url: "https://api.example.com"
 paths:
+  /both:
+    post:
+      summary: "Both example and examples"
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                name:
+                  type: string
+            example:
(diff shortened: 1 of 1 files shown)


packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -347,6 +347,12 @@ const BODY_TYPE_HANDLERS = [
   }
 ];
 
+const getContentLevelExample = (bodyContent) => {
+  if (bodyContent.example !== undefined) return bodyContent.example;
+  const firstExample = Object.values(bodyContent.examples ?? {})[0];
+  return firstExample?.value;
+};
+
```
sid-bruno, line 355: What if both `example` and `examples` are there and are we only importing one?

packages/bruno-converters/tests/openapi/openapi-to-bruno/openapi-body.spec.js
sid-bruno, line 1: I'm sure there's more possible iterations of this, can you spend some time adding in tests also for multiple response types with multiple examples for different response codes, if the test for that is already there then okay but if not, please add a few

packages/bruno-converters/src/openapi/openapi-to-bruno.js
pooja-bruno: Yes, when both `example` and `examples` are present, we only import one — the singular `example` takes priority. This follows the OpenAPI spec behavior and matches how most tools interpret it.

The priority chain is:
1. `schema.example` — highest priority, if present content-level is skipped entirely
2. Content-level `example` (singular) — wins over `examples` (plural)
3. Content-level `examples` (plural, first entry) — used as fallback when singular is absent

Added a test that covers all three cases to confirm this behavior.

pooja-bruno: Yes, when both `example` and `examples` are present, we only import one `example` it take priority over `examples`.

Issue #7074, opened 2026-02-06
Reordered collection not saved


robert-petta-alida:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

3.0.2

### Operating System

Windows 11

### Describe the bug

**Steps to reproduce**

1. Drag-and-drop any collection to change its position.
2. Close Bruno
3. Reopen Bruno

**Expected result** 
The collection's new position is preserved.

**Actual result** 
The collection's new position is not preserved, as if the change never happened.

### Screenshots/Live demo link

![Image](https://github.com/user-attachments/assets/401674a5-9e3e-4a31-be78-98b14c27ebe7)

robert-petta-alida:
As of version `3.1.4`, this issue persists.

MWDidata:
I just made the same ticket and the bot refered me to this ticket.
https://github.com/usebruno/bruno/issues/7453

I did just find a workaround:
Open folder %Appdata%\Roaming\bruno\default-workspace and change the workspace.yml file to the correct order

onacit:
Seems not fixed with 3.2.0.

PR #7093, opened 2026-02-10
fix: collection reorder not persisting after restart

sanjaikumar-bruno:
[BRU-2746](https://usebruno.atlassian.net/browse/BRU-2746)
Fixes: https://github.com/usebruno/bruno/issues/7074

### DIFF
--- a/packages/bruno-electron/src/utils/workspace-config.js
+++ b/packages/bruno-electron/src/utils/workspace-config.js
@@ -419,8 +419,10 @@ const reorderWorkspaceCollections = async (workspacePath, collectionPaths) => {
     const matched = new Set();
 
     for (const absolutePath of collectionPaths) {
-      const targetPath = path.normalize(absolutePath);
-      const entry = existing.find((c) => getNormalizedAbsoluteCollectionPath(workspacePath, c) === targetPath);
+      const targetPath = posixifyPath(path.normalize(absolutePath));
+      const entry = existing.find(
+        (c) => posixifyPath(getNormalizedAbsoluteCollectionPath(workspacePath, c)) === targetPath
+      );
       if (entry && !matched.has(entry)) {
         inNewOrder.push(entry);
         matched.add(entry);


packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -2681,11 +2681,38 @@ export const importCollectionFromZip = (zipFilePath, collectionLocation) => asyn
   return collectionPath;
 };
 
+/**
+ * Updates Redux collection order and persists it to the active workspace's workspace.yml.
+ */
 export const moveCollectionAndPersist
   = ({ draggedItem, targetItem }) =>
     (dispatch, getState) => {
       dispatch(moveCollection({ draggedItem, targetItem }));
```
sanish-bruno: @sanjaikumar-bruno we have a `moveCollection` function , what does it do, i believe it updates the redux store,similar to how `saveCollectionSecurityConfig` is done we should be dispathing it within the then block, the catch block should reject with error. else like we have currently, if anything fails in the renderer invocation, the ui state and the file state will be entirely different

packages/bruno-electron/tests/utils/workspace-config.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    workspacePath = fs.mkdtempSync(path.join(os.tmpdir(), 'bruno-ws-'));
+  });
+
+  afterEach(() => {
+    fs.rmSync(workspacePath, { recursive: true, force: true });
+  });
+
+  test('reorders collections to match given path list', async () => {
+    writeWorkspaceYml([
+      collection('API', 'api'),
```
sanish-bruno: @sanjaikumar-bruno if possible make the path-segments look like paths, for better readability

sanish-bruno (changes requested):

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -2726,11 +2726,43 @@ export const importCollectionFromZip = (zipFilePath, collectionLocation) => asyn
...
     (dispatch, getState) => {
-      dispatch(moveCollection({ draggedItem, targetItem }));
-      return Promise.resolve();
+      const state = getState();
+      const activeWorkspace = state.workspaces.workspaces.find(
+        (w) => w.uid === state.workspaces.activeWorkspaceUid
+      );
+      if (!activeWorkspace?.pathname || !activeWorkspace.collections?.length) {
+        return Promise.resolve();
+      }
```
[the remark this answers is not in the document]
sanjaikumar-bruno: It’s not a real issue. Every workspace (including the default one) gets a pathname when it’s opened, so you never have “has collections but no pathname.” The current guard is fine; no change needed.

packages/bruno-electron/src/utils/workspace-config.js
```diff
@@ -391,6 +397,41 @@ const removeCollectionFromWorkspace = async (workspacePath, collectionPath) => {
...
 };
 
+/**
+ * Reorders the collections array in the workspace's workspace.yml to match the given path list.
+ * Entries not in the list are appended at the end.
+ * @param {string} workspacePath - Absolute path to the workspace directory
+ * @param {string[]} collectionPaths - Absolute collection pathnames in the desired order
+ */
+const reorderWorkspaceCollections = async (workspacePath, collectionPaths) => {
+  if (!Array.isArray(collectionPaths)) {
```
sanish-bruno, line 410: @sanjaikumar-bruno Rebase the branch from main, recently we had some changes done within workspace collections, we introduced posixifying the path so that the paths can work across different platforms. most likely we do not have to make any changes, and the received paths by the function will already be normalized, just make sure we are inline with the behaviour.

bijin-bruno (approved):

PR #7170, opened 2026-02-17
feat: enhance API spec export with environment variables support

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-1246)

- Improve OpenAPI export to preserve server variables, multi-environment servers, operation-level server overrides, path parameter values, and request body examples
- Improve OpenAPI import to support path-item parameter inheritance, operation-level servers as request variable overrides, and better parameter value extraction

###  Export 

- Export {{var}} template URLs as OpenAPI server variables with {var} syntax and default values
- Export each Bruno environment (with a baseUrl) as a separate servers[] entry with description
- Export request-level vars.req baseUrl overrides as operation-level servers in the spec
- Fix path params being duplicated as query params — now filters by param.type
- Export path parameter example values from the params array
- Preserve JSON/XML request body values as schema.example for round-trip

### Import

- Add extractServerVars() helper to convert OpenAPI server variables to Bruno {{var}} syntax
- Refactor environment creation to use extractServerVars(), preserving all server template variables (not just baseUrl)
- Add mergeParams() to inherit path-item level parameters into operations per OpenAPI spec
- Import operation-level servers as request.vars.req overrides (with path-item fallback)
- Add getParameterExampleValue() to extract values from param.example, param.examples, schema.default, schema.example, schema.enum, schema.examples, schema.minimum
- Add getSchemaPropertyExampleValue() to extract values from schema properties in request bodies
- Populate parameter and schema property values during import (previously always empty)
- Support querystring as alias for query parameter type

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Export API specs with multi-environment servers, operation-level server overrides, and server URL template variable support.

* **Improvements**
  * OpenAPI conversion now merges path-item and operation parameters, avoids duplicate path params, preserves request/body examples, and populates representative parameter and per-operation server variables for richer exports.

* **Tests**
  * Extensive test coverage for server-variable extraction, multi-environment exports, path-parameter merging, and example/value resolution.

### DIFF
--- a/packages/bruno-app/src/utils/exporters/openapi-spec.js
+++ b/packages/bruno-app/src/utils/exporters/openapi-spec.js
@@ -3,7 +3,7 @@ import { interpolate } from '@usebruno/common';
 import { isValidUrl } from 'utils/url/index';
 const xml2js = require('xml2js');
 
-export const exportApiSpec = ({ variables, items, name }) => {
+export const exportApiSpec = ({ variables, items, name, environments }) => {
   items = items.filter((item) => !['grpc-request'].includes(item.type));
 
   const components = {
@@ -22,11 +22,60 @@ export const exportApiSpec = ({ variables, items, name }) => {
     });
   };
 
-  const addUrlToServersList = (url) => {
-    if (!servers?.find((s) => s?.url === url)) {
-      servers.push({ url });
+  const templateVarRegex = /\{\{([^}]+)\}\}/g;
+
+  // Build an OpenAPI server entry from a URL (supports {{var}} templates)
+  const buildServerEntry = (url, vars, description) => {
+    const cleanedUrl = url.endsWith('/') ? url.slice(0, -1) : url;
+    const matches = [...cleanedUrl.matchAll(templateVarRegex)];
+    const entry = {};
+
+    if (matches.length > 0) {
+      entry.url = cleanedUrl.replace(templateVarRegex, '{$1}');
+      const serverVariables = {};
+
+      // each match m is an array where m[0] is the full match (e.g. {{protocol}}) and m[1] is the capture group (e.g. protocol).
+      matches.map((m) => m[1]).forEach((varName) => {
+        if (vars[varName] !== undefined) {
+          serverVariables[varName] = { default: String(vars[varName]) };
+        }
+      });
+      if (Object.keys(serverVariables).length > 0) {
+        entry.variables = serverVariables;
+      }
+    } else {
+      entry.url = cleanedUrl;
     }
+
+    if (description) entry.description = description;
+    return entry;
   };
+
+  // Collect all baseUrl sources: collection variables + each environment
+  // Each source becomes a self-contained server entry in the OpenAPI spec.
+  // On import, each server entry maps to a separate Bruno environment.
+  const baseUrlSources = [];
+
+  // Add collection-level baseUrl if present
+  const collectionBaseUrl = variables?.baseUrl || '';
+  if (collectionBaseUrl) {
+    baseUrlSources.push({ baseUrl: collectionBaseUrl, description: 'Base Server', vars: variables });
+  }
+
+  // Add each environment that defines its own baseUrl
+  if (environments && environments.length > 0) {
+    for (const env of environments) {
+      const envVars = getEnabledVarsAsObject(env.variables);
+      if (envVars.baseUrl) {
+        baseUrlSources.push({ baseUrl: envVars.baseUrl, description: env.name, vars: envVars });
+      }
+    }
+  }
+
+  // Build root server entries
+  for (const source of baseUrlSources) {
+    servers.push(buildServerEntry(source.baseUrl, source.vars, source.description));
+  }
 
   const extractTagFromDepth = (item) => {
     const { pathname, depth } = item;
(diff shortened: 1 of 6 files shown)


sanish-bruno (commented):
## Automated Review

No issues found. All checks passed. ✅

---
*Generated by pr-review.sh*

sanish-bruno (commented):
## Manual Code Review

Found **2 bugs** and **6 suggestions** across the export and import changes. See inline comments.

packages/bruno-app/src/utils/exporters/openapi-spec.js
```diff
@@ -347,7 +451,8 @@ export const exportApiSpec = ({ variables, items, name }) => {
       return {
         url,
         method: method.toLowerCase(),
-        data: pathBody
+        data: pathBody,
+        operationLevelServer
```
sanish-bruno, line 458: **Bug (Major, pre-existing)**: The `multipartForm` (line 279), `formUrlEncoded` (line 300), and `text` (line 321) switch cases above use `return` instead of `break`, which exits the `.map()` callback and produces `undefined` items in `_items`. They also lack `break` statements, so they fall through to the next case.

Now that `operationLevelServer` flows through this same `.map()` → `.reduce()` pipeline, any `undefined` item will crash in the reduce on `item?.url`.

Suggested fix: replace `return` with `break` in the early-exit guards, and add `break` at the end of each case body.

```diff
 case 'multipartForm':
-  if (!body?.multipartForm) return;
+  if (!body?.multipartForm) break;
   // ...
+  break;
 case 'formUrlEncoded':
-  if (!body?.formUrlEncoded) return;
+  if (!body?.formUrlEncoded) break;
   // ...
+  break;
 case 'text':
-  if (!body?.text) return;
+  if (!body?.text) break;
   // ...
+  break;
```

packages/bruno-app/src/utils/exporters/openapi-spec.js
```diff
@@ -41,14 +90,48 @@ export const exportApiSpec = ({ variables, items, name }) => {
...
+  // then known baseUrl sources. Falls back to full resolution for unknown URLs.
+  const resolveRequestUrl = (rawUrl, requestVars) => {
+    // Request has a baseUrl override in vars.req — export as operation-level server
+    if (rawUrl.startsWith('{{baseUrl}}') && requestVars) {
+      const baseUrlOverride = requestVars.find((v) => v.name === 'baseUrl' && v.enabled);
+      if (baseUrlOverride) {
+        const reqVarsMap = {};
+        requestVars.filter((v) => v.enabled).forEach((v) => { reqVarsMap[v.name] = v.value; });
+        const path = rawUrl.slice('{{baseUrl}}'.length) || '/';
+        return { url: interpolate(path, variables), operationLevelServer: buildServerEntry(baseUrlOverride.value, reqVarsMap) };
```
sanish-bruno: **Bug**: `interpolate(path, variables)` will consume path parameter placeholders if `variables` contains keys matching path param names.

For example, if the URL is `{{baseUrl}}/users/{{userId}}` and `variables` has `userId: '123'`, the exported OpenAPI path becomes `/users/123` instead of `/users/{userId}`, losing the path parameter.

This also applies to lines 111, 118, and 127.

Suggested fix: don't interpolate the path with collection variables (the path should preserve template placeholders for OpenAPI output), or interpolate with a restricted set that excludes known path param names.

packages/bruno-app/src/utils/exporters/openapi-spec.js
```diff
@@ -22,12 +22,61 @@ export const exportApiSpec = ({ variables, items, name }) => {
...
+      const envVars = getEnabledVarsAsObject(env.variables);
+      if (envVars.baseUrl) {
+        baseUrlSources.push({ baseUrl: envVars.baseUrl, description: env.name, vars: envVars });
+      }
+    }
+  }
+
+  // Build root server entries
+  for (const source of baseUrlSources) {
+    servers.push(buildServerEntry(source.baseUrl, source.vars, source.description));
```
sanish-bruno, line 76: **Design**: `baseUrlSources` allows duplicate entries when collection and environment define the same `baseUrl` (confirmed intentional by tests). However, duplicate `servers[]` entries are unusual in OpenAPI specs and may confuse consumers. Consider deduplicating by URL, or adding a brief comment here explaining the intentional trade-off for round-trip fidelity.

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -1220,45 +1306,58 @@ export const parseOpenApiCollection = (data, options = {}) => {
 
     // Create environments based on the servers
     servers.forEach((server, index) => {
-      let baseUrl = getDefaultUrl(server);
-      let environmentName = server.description ? server.description : `Environment ${index + 1}`;
+      let environmentName = server.name || server.description || `Environment ${index + 1}`;
```
sanish-bruno, line 1305: **Note**: `server.name` is not part of the OpenAPI 3.0 or 3.1 Server Object specification (only `url`, `description`, and `variables` are defined). This is fine as a progressive/defensive measure, but worth a comment noting it's a non-standard extension so future maintainers don't assume it's always present in valid specs.

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -567,21 +611,40 @@ const transformOpenapiRequestItem = (request, usedNames = new Set(), options = {
...
       // Expand object schema properties into individual parameters
+      const schemaExample = param.schema.example || {};
+
       each(param.schema.properties, (prop, propName) => {
         const isRequired = Array.isArray(param.schema.required) && param.schema.required.includes(propName);
 
-        if (param.in === 'query') {
+        const propValue = getSchemaPropertyExampleValue(prop, propName, schemaExample);
+
+        if (param.in === 'query' || param.in === 'querystring') {
```
sanish-bruno, line 639: **Note**: `querystring` is not a valid value for the `in` field in the OpenAPI spec (valid values are `query`, `path`, `header`, `cookie`). This is a useful defensive measure for malformed specs, but deserves a brief comment so it's clear this is an intentional workaround, not a spec-backed value.

packages/bruno-app/src/utils/exporters/openapi-spec.js
```diff
@@ -22,12 +22,61 @@ export const exportApiSpec = ({ variables, items, name }) => {
...
+  // Add collection-level baseUrl if present
+  const collectionBaseUrl = variables?.baseUrl || '';
+  if (collectionBaseUrl) {
+    baseUrlSources.push({ baseUrl: collectionBaseUrl, description: 'Base Server', vars: variables });
+  }
+
+  // Add each environment that defines its own baseUrl
+  if (environments && environments.length > 0) {
+    for (const env of environments) {
+      const envVars = getEnabledVarsAsObject(env.variables);
```
sanish-bruno, line 67: **Nit**: `getEnabledVarsAsObject` is called here but defined at line 511 as a `const` arrow function. This works at runtime (the call is inside `exportApiSpec` which runs later), but unlike `function` declarations, `const` arrow functions are not hoisted. If someone later moves this code outside the closure, it will break with a ReferenceError. Consider moving the helper definition above this call, or converting it to a `function` declaration.

packages/bruno-app/src/utils/exporters/openapi-spec.js
```diff
@@ -22,12 +22,61 @@ export const exportApiSpec = ({ variables, items, name }) => {
...
+    const cleanedUrl = url.endsWith('/') ? url.slice(0, -1) : url;
+    const matches = [...cleanedUrl.matchAll(templateVarRegex)];
+    const entry = {};
+
+    if (matches.length > 0) {
+      entry.url = cleanedUrl.replace(templateVarRegex, '{$1}');
+      const serverVariables = {};
+
+      // each match m is an array where m[0] is the full match (e.g. {{protocol}}) and m[1] is the capture group (e.g. protocol).
+      matches.map((m) => m[1]).forEach((varName) => {
```
sanish-bruno: **Nit**: `.map((m) => m[1]).forEach(...)` creates an unnecessary intermediate array. Could simplify to:

```js
matches.forEach((m) => {
  const varName = m[1];
  if (vars[varName] !== undefined) {
    serverVariables[varName] = { default: String(vars[varName]) };
  }
});
```

packages/bruno-converters/src/openapi/openapi-to-bruno.js
```diff
@@ -567,21 +611,40 @@ const transformOpenapiRequestItem = (request, usedNames = new Set(), options = {
     }
   };
 
+  // If the operation has its own servers, override baseUrl via request vars
+  if (request.servers && request.servers.length > 0) {
+    const serverVarPairs = extractServerVars(request.servers[0]);
```
sanish-bruno, line 612: **Note**: Only the first operation-level server (`request.servers[0]`) is used. Additional servers are silently ignored. This is a reasonable simplification, but worth a brief comment so it's clear this is intentional — e.g. `// Use first server only; additional operation-level servers are not supported`.

packages/bruno-app/src/utils/exporters/openapi-spec.js
```diff
@@ -397,6 +507,17 @@ const generateInfoSection = (name) => {
...
+// Convert env variable array to { name: value } object (only enabled vars)
+const getEnabledVarsAsObject = (variables = []) => {
+  const result = {};
+  variables.forEach((v) => {
+    if (v.name && v.enabled) {
+      result[v.name] = v.value;
+    }
+  });
+  return result;
+};
```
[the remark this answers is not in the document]
abhishek-bruno: Not a valid issue — intentionally not applying this change. The current condition v.name && v.enabled is correct. Adding a v.value truthiness check would silently drop legitimate falsy values like 0, "", and false

packages/bruno-app/src/utils/exporters/openapi-spec.js
abhishek-bruno: It's by design — each environment becomes a separate server entry so the round-trip (export → import) preserves all Bruno environments.

packages/bruno-converters/src/openapi/openapi-to-bruno.js
abhishek-bruno: server.name is part of the OpenAPI 3.2.0 spec. It was added as an optional field for uniquely identifying a server. Please refer: https://spec.openapis.org/oas/v3.2.0.html#server-object

packages/bruno-converters/src/openapi/openapi-to-bruno.js
abhishek-bruno: querystring is a valid value in the OpenAPI 3.2.0 spec. It's a new addition — it treats the entire query string as a single value. Please refer: https://spec.openapis.org/oas/v3.2.0.html#parameter-object

bijin-bruno (dismissed):

bijin-bruno (approved):

PR #7187, opened 2026-02-18
Enable encodeUrl setting to control URL encoding in generated snippets

### LINKED ISSUE #5756: Generate code does not respect "Automatically encode query parameters in the URL" setting and will always encode

### This bug is:

- [x] annoying

### Bruno version

2.12.0

### Operating System

Windows 11

[Jira](https://usebruno.atlassian.net/browse/BRU-2000)

### Describe the bug

Executing a normal request respects the "Automatically encode query parameters in the URL" setting but the generate code view (accessed through the button in the address field) does not. The encoding will replace `=` in my query parameter with `%3D` which is not accepted by the receiving server.

For example:
https://domain.com/path?query1=someEncryptedCode==&query2=test

Will be replaced in the generate code view with:
https://domain.com/path?query1=someEncryptedCode%3D%3D&query2=test

I have seen it work correctly in this bruno version but I can not reproduce it working.

Could not upload a .bru file. This is the content:
```
meta {
  name: Bug generate code
  type: http
  seq: 3
}

get {
  url: https://domain.com/path?query1=someEncryptedCode==&query2=test
  body: none
  auth: inherit
}

params:query {
  query1: someEncryptedCode==
  query2: test
}

settings {
  encodeUrl: false
}

```

Request url:
Image

Send request url normal. Not a real domain. The 404 is expected:
Image

Generate code url is unexpectedly urlEncoded (`=` -> `%3D`):
Image

Setting is turned of:
Image

sanish-bruno:
## Summary

[Jira](https://usebruno.atlassian.net/browse/BRU-2000)
Fixes #5756 — Generate Code now respects the `encodeUrl` setting on requests.

Previously, the "Generate Code" feature always produced encoded query parameters (e.g., `token=abc123%3D%3D`) regardless of the request's `encodeUrl` setting. This happened because HTTPSnippet internally encodes all URLs and has no API to disable this behavior.

### What changed

**`GenerateCodeItem/index.js`**
- Changed `rawUrl` computation from using `decodeURI(finalUrl)` to using `interpolatedUrl` (pre-`new URL()` version)
- Preserves user's exact encoding: if they typed `%20`, it stays `%20`; if they typed a space, it stays a space
- Path params (`:id`) are resolved via simple string replacement to avoid unwanted encoding side-effects
- Supports both traditional (`:id`) and OData-style (`Entity(:id)`) path parameters

**`snippet-generator.js`**
- When `encodeUrl` is not explicitly `true`, replaces HTTPSnippet's encoded path+query with the raw version
- For HTTP raw target, spaces are always encoded as `%20` (RFC 7230 compliance: `METHOD <request-target> HTTP-version` delimiter)
- Other special chars (`=`, `:`, `/`) remain unencoded as they're safe in the request line

**`snippet-generator.spec.js`**
- Added 12 test cases covering:
  - Equals signs in query values (`?token=abc123==`) — preserved unencoded when `encodeUrl` is false
  - Email with `+` and `@` (`?email=test+alias@example.com`)
  - Redirect URLs with colons and slashes (`?redirect=https://other.com/callback`)
  - Comma-separated values and time format (`?tags=a,b,c&time=10:30`)
  - Spaces in pathname/query with `encodeUrl: false` — preserved
  - User-typed `%20` preservation (not decoded to literal space)
  - **New:** HTTP target with spaces re-encoded to `%20` (RFC compliance)
  - **New:** OData-style paths with parenthesized params (e.g., `Products(123)`)
  - Encoding preserved when `encodeUrl: true`
  - Default behavior (no settings) preserves raw URL
  - Draft settings override base settings
  - Path-only targets (python `http.client`) are handled correctly

### How it works

1. `GenerateCodeItem` computes two URL versions:
   - `finalUrl` from `interpolateUrlPathParams()` (uses `new URL()`, fully encoded)
   - `rawUrl` from `interpolatedUrl` + manual path param replacement (preserves user encoding)

2. In `snippet-generator`:
   - Generate snippet normally via HTTPSnippet (which encodes the URL)
   - If `encodeUrl !== true`, replace HTTPSnippet's encoded path+query with the raw version
   - For HTTP target, re-encode spaces to `%20` (RFC 7230 requirement)

3. Works across all targets because the path is always present in the output — either in the full URL (curl, fetch) or standalone (python, raw HTTP)

## Test plan

- [x] Run `npm run test --workspace=packages/bruno-app -- --testPathPattern=snippet-generator` — all 40 tests pass
- [ ] Create a request with URL `https://example.com/api?token=abc123==`
  - [ ] Set `encodeUrl: false` in request Settings tab
  - [ ] Open Generate Code → verify curl shows `token=abc123==` (not `%3D%3D`)
  - [ ] Set `encodeUrl: true` → verify curl shows `token=abc123%3D%3D`
- [ ] Create a request with URL `https://example.com/api?token=abc%20123` (user typed `%20`)
  - [ ] Set `encodeUrl: false` → verify snippet shows `%20` (not decoded to space)
- [ ] Test HTTP raw target with spaces in query → verify spaces are always `%20` (RFC 7230 compliance)
- [ ] Test OData URL like `https://example.com/odata/Products(:id)` with path param replacement
- [ ] Test with python http.client target → verify path shows unencoded values
- [ ] Test with python-requests target → verify params dict has decoded values either way

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Configurable URL encoding for generated code snippets so users can choose whether path/query characters are preserved or encoded.
  * Generated snippets now include a preserved raw-URL variant that reflects the original request after variable resolution.

* **Bug Fixes**
  * Better handling of spaces, reserved characters, fragments and pre-encoded values to avoid unintended re-encoding.

* **Tests**
  * Added comprehensive tests covering encode-url behaviors, overrides, and client-specific cases.


packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
```diff
@@ -79,6 +95,17 @@ const generateSnippet = ({ language, item, collection, shouldInterpolate = false
...
+    // Respect encodeUrl setting: when not explicitly true, replace HTTPSnippet's encoded URL with the raw URL
+    // encodeUrl defaults to false in the UI when undefined/null
+    const settings = item.draft ? get(item, 'draft.settings') : get(item, 'settings');
+    if (settings?.encodeUrl !== true) {
+      const rawUrl = request.url;
+      const encodedUrl = getEncodedUrl(rawUrl);
+      if (encodedUrl !== rawUrl) {
+        result = result.replaceAll(encodedUrl, rawUrl);
+      }
+    }
```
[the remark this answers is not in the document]
sanish-bruno: The concern is that the encoded path (e.g., /api?token=abc123%3D%3D) could match in headers or body        
  sections, not just the URL position.                                                                       
                                                                                                             
  I don't think the regex approach is the right fix here. Here's why:                                        
                                                                                                             
  1. Raw HTTP has no quotes around the path: GET /api?token=abc123%3D%3D HTTP/1.1 — a quoted-token regex like
   /(["'])\1/` would miss this entirely.                                                                     
  2. Some shell output is unquoted too — curl's quote() function skips quoting when the URL matches
  /^[a-z0-9-_/.@%^=:]+$/i, which encoded URLs with % characters do match.
  3. The quoting varies wildly across 39+ targets — single quotes (curl), double quotes (python, java, go),
  no quotes (raw HTTP), template literals (some JS targets). A regex trying to handle all of these would be
  more fragile than replaceAll.
  4. The false positive scenario is practically unrealistic — for a false match, the user's header value or
  request body would need to contain the exact string /api?token=abc123%3D%3D. That's their own URL's encoded
   path appearing verbatim in their own body. Even if it did happen (e.g., a Referer header), replacing it
  would be the correct behavior since the intent is "don't encode my URLs."

  The current replaceAll is simpler, works across all targets regardless of quoting style, and the false
  positive risk is theoretical rather than practical. I'd recommend keeping it as-is.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
```diff
@@ -4,6 +4,21 @@ import { getAllVariables, getTreePathFromCollectionToItem, mergeHeaders } from '
...
+// the path as a substring, and path-only targets (http.client, raw HTTP) use it directly.
+const getEncodedPath = (rawUrl) => {
+  const parsed = parse(rawUrl, true, true);
+  if (!parsed.query || Object.keys(parsed.query).length === 0) {
+    return null;
+  }
+  const search = stringify(parsed.query);
+  const encodedPath = search ? `${parsed.pathname}?${search}` : parsed.pathname;
+  return { encodedPath, rawPath: parsed.path };
+};
```
gopu-bruno: @sanish-bruno, Found these consistent issues:

**Spaces stay encoded even when encodeUrl is false**

For example: https://example.com/api?search=hello world

Even with `encodeUrl: false`, the generated snippet still contains %20 instead of preserving the space as written.

**Hash fragment lost in URLs with #**

For example: https://example.com/api?token=abc==#section

With `encodeUrl: false`, the #section fragment is not preserved correctly and gets lost in the generated snippet.

sanish-bruno: Fragment is stripped — correct per RFC 3986 §3.5: user agents MUST NOT include the fragment
    in the HTTP request target sent to the origin server (though fragments can still appear in
    user-facing URLs, SPA routing, and are inherited across redirects per RFC 9110 §10.2.2).
    https://datatracker.ietf.org/doc/html/rfc3986#section-3.5
    https://datatracker.ietf.org/doc/html/rfc9110#section-10.2.2

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
```diff
@@ -1,5 +1,13 @@
 import { getAuthHeaders } from 'utils/codegenerator/auth';
 
+jest.mock('@usebruno/common/utils', () => {
+  const actual = jest.requireActual('@usebruno/common/utils');
+  return {
+    ...actual,
+    encodeUrl: actual.encodeUrl
```
bijin-bruno: Is this necessary?

gopu-bruno (approved):

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
```diff
@@ -79,6 +82,38 @@ const generateSnippet = ({ language, item, collection, shouldInterpolate = false
...
+      // The HTTP raw target (http/http1.1) uses the request line format:
+      //   METHOD <request-target> HTTP-version
+      // Spaces delimit these fields, so a literal space in the request-target
+      // would be parsed as the end of the URI (RFC 7230 §3.1.1).
+      if (language.target === 'http') {
+        desiredPath = desiredPath.replace(/ /g, '%20');
+      }
+    }
+
+    if (httpSnippetPath !== desiredPath) {
```
bijin-bruno: If the URL is just https://example.com (no path), parsed.pathname would be /. Doing replaceAll('/', desiredPath) would be catastrophic. The current guard if (httpSnippetPath !== desiredPath) helps, but only if desiredPath also equals /. Consider adding an explicit guard for very short paths.

```
if (httpSnippetPath !== desiredPath && httpSnippetPath?.length > 1) {
```

Issue #7176, opened 2026-02-17
Response not updating in Raw/JSON view (no preview) when using Authorization Bearer + multipart/form-data


Mansourkia:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

3.1.1

### Operating System

Windows 11

### Description

When sending a request with **Authorization: Bearer** header and **multipart/form-data** body,
the response viewer shows stale data in **Raw / JSON view**.

If the same request is sent again after the server response has changed,
the previous response is still displayed.

This issue does **not** occur when Preview mode is enabled.

### Environment

- Back-end: Laravel 12
- Bruno version: 3.1.1
- OS: Windows
- Auth: Authorization Bearer
- Body type: multipart/form-data
- Response view: Raw / JSON (Preview disabled)

### Steps to Reproduce

1. Create a request with:
   - `Authorization: Bearer <token>`
   - Body type: `multipart/form-data`
2. Ensure response view is set to **Raw / JSON** (Preview is disabled)
3. Send the request
4. Change the server-side response for the same request
5. Send the request again
6. Observe that the response still shows the previous data

### Expected Behavior

The response view should update immediately after each request
and always display the latest server response.

### Actual Behavior

The response in Raw / JSON view is not updated until:

- Switching to another request tab and back, or
- Enabling Preview mode (when available)

### Notes

- Network request is executed correctly
- Server returns updated data
- Preview mode always displays the latest response
- This appears to be a UI re-render/state update issue in the Raw/JSON response viewer,
  possibly related to multipart/form-data requests with Authorization headers.
- The request sends Cache-Control header

### .bru file to reproduce the bug

[Get Artist.yml](https://github.com/user-attachments/files/25369768/Get.Artist.yml)

### Screenshots/Live demo link

https://github.com/user-attachments/assets/53327454-ac1c-4914-9d1c-6451cb08a361

Its-treason:
I encountered this at work today. After 30 minutes of trying to fix my deployment, I finally figured out that Bruno was the problem. Very frustrating.

sanjaikumar-bruno:
The issue is reproducible. We're working on a fix.

PR #7218, opened 2026-02-19
fix: response viewer not updating when focused (read-only editors)

sanjaikumar-bruno:
[BRU-2747](https://usebruno.atlassian.net/browse/BRU-2747)
Fixes: https://github.com/usebruno/bruno/issues/7176

### DIFF
--- a/tests/utils/page/actions.ts
+++ b/tests/utils/page/actions.ts
@@ -265,9 +265,9 @@ const createRequest = async (
       await page.locator('.bruno-modal .method-selector').click();
       const isStandardMethod = STANDARD_HTTP_METHODS.includes(method.toUpperCase());
       if (isStandardMethod) {
-        await page.locator('.dropdown-item').filter({ hasText: method }).click();
+        await locators.modal.newRequestMethodOption(method).click();
       } else {
-        await page.locator('.dropdown-item').filter({ hasText: '+ Add Custom' }).click();
+        await locators.modal.newRequestMethodOption('add-custom').click();
         await page.locator('.bruno-modal .method-selector input').fill(method);
         await page.keyboard.press('Enter');
       }
--- a/packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/index.js
+++ b/packages/bruno-app/src/components/RequestPane/QueryUrl/HttpMethodSelector/index.js
@@ -179,6 +179,7 @@ const HttpMethodSelector = ({ method = DEFAULT_METHOD, onMethodSelect, showCaret
           items={menuItems}
           placement="bottom-start"
           selectedItemId={selectedItemId}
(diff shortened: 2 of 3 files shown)


tests/request/response-pane-update-when-focused.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      await bodyEditor.click();
+      await page.locator('.request-pane textarea').last().fill('{"run": 2}');
+      await page.waitForTimeout(300);
+    });
+
+    await test.step('Click inside response pane (Raw/JSON editor) to give it focus', async () => {
+      const responseEditor = page.locator('[data-testid="response-preview-container"] .CodeMirror').first();
+      await responseEditor.waitFor({ state: 'visible', timeout: 5000 });
+      await responseEditor.click();
+      await page.waitForTimeout(200);
```
sanish-bruno: @sanjaikumar-bruno , let's double check if we really need `await page.waitForTimeout(200);`, if the tests are consistently passing without them we do not have to keep them. I believe, you have tested this with the main branch and the tests were consistently failing.

tests/request/response-pane-update-when-focused.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    await test.step('Press Cmd+Enter / Ctrl+Enter to re-send request', async () => {
+      await page.keyboard.press(runShortcut);
+      await page.getByTestId('response-status-code').waitFor({ state: 'visible', timeout: 15000 });
```
sanish-bruno: @sanjaikumar-bruno can you double check if we have locators already defined in locators
`const locators = buildCommonLocators(page);`, this comment is also applicable for `bodymode selector` etc whereever we are using `page.locator`. we should should beaccessing components through locators, whenever possible, you can also add a `locator` if you find it ti be missing in the `common locators` file

sanish-bruno (changes requested):

sanish-bruno (approved):

tests/utils/page/actions.ts
```diff
@@ -258,6 +261,19 @@ const createRequest = async (
     await locators.dropdown.item('New Request').click();
     await page.getByPlaceholder('Request Name').fill(requestName);
 
+    if (method) {
+      await page.locator('.bruno-modal .method-selector').click();
+      const isStandardMethod = STANDARD_HTTP_METHODS.includes(method.toUpperCase());
+      if (isStandardMethod) {
+        await page.locator('.dropdown-item').filter({ hasText: method }).click();
```
bijin-bruno: `page.locator('.dropdown-item')` is too generic. It would be better to add and use a test-id here

bijin-bruno (approved):

Issue #7258, opened 2026-02-23
Incorrect Collection/Request appears.


baek-sang:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [x] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

3.1.3

### Operating System

Windows 11

### Describe the bug

1. WorkspaceA -> CollectionA -> send RequestA
2. Change WorkspaceB
3. WorkspaceB -> CollectionB -> send RequestB
4. Close tab RequestB
5. RequestA is displayed.

I have switched to WorkspaceB, so RequestA should not be visible because it belongs to WorkspaceA.

### Screenshots/Live demo link

https://github.com/user-attachments/assets/2e0c62b7-658e-40fa-88c4-ac20fbc0ab53

sanjaikumar-bruno:
The issue is reproducible. We're investigating the cause

PR #7259, opened 2026-02-23
fix: wrong workspace request shown after closing tab

sanjaikumar-bruno:
[BRU-2753](https://usebruno.atlassian.net/browse/BRU-2753)
Fixes: https://github.com/usebruno/bruno/issues/7258

Cause:
Tabs are global. When the active tab was closed, the tabs reducer chose the next active tab as “last tab in the list” and did not consider the workspace, so it could pick a tab from another workspace.

Fix:
After closing tabs (in the `closeTabs` thunk and after `closeAllCollectionTabs` in the remove-collection flow), we run `ensureActiveTabInCurrentWorkspace`: if the new active tab is not in the current workspace, we focus an in-workspace tab (another request or the workspace overview), creating the overview tab if needed.


packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -3136,6 +3157,9 @@ export const closeTabs = ({ tabUids }) => async (dispatch, getState) => {
   // Close the tabs first
   await dispatch(_closeTabs({ tabUids }));
 
+  // After close, the reducer may have set active tab to one from another workspace. Ensure it belongs to this workspace: prefer any open in-workspace tab, then workspace overview if none.
+  await dispatch(ensureActiveTabInCurrentWorkspace());
```
sanish-bruno, line 3163: **Bug (potential race)**: `ensureActiveTabInCurrentWorkspace` reads state via `getState()` synchronously, but `_closeTabs` is a Redux reducer (synchronous), so the state should be updated by the time this runs. However, `await dispatch(...)` on a synchronous action is a no-op await — the `await` here is misleading since `_closeTabs` isn't async. This is harmless but could confuse readers into thinking there's an async dependency.

More importantly: if `_closeTabs` is ever refactored to be async (e.g., middleware intercepts), this code would still work. So this is fine, but consider adding a comment clarifying the dispatch is synchronous.

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -2353,6 +2354,7 @@ export const removeCollection = (collectionUid) => (dispatch, getState) => {
       .then((remainingWorkspaces) => {
         // Close tabs for this collection
         dispatch(closeAllCollectionTabs({ collectionUid }));
+        dispatch(ensureActiveTabInCurrentWorkspace());
```
sanish-bruno: **Suggestion**: `ensureActiveTabInCurrentWorkspace` is dispatched immediately after `closeAllCollectionTabs`, but before `removeCollectionFromWorkspace` (line 2361). At this point the collection being removed is still in the workspace's `collections` list, so `getWorkspaceCollectionUids` will still include it. This means `ensureActiveTabInCurrentWorkspace` could pick a tab from the collection that's about to be removed (if any survived the close).

In practice this is likely fine because `closeAllCollectionTabs` already removed all tabs for this collection, so no surviving tabs would match. But the ordering is fragile — consider dispatching `ensureActiveTabInCurrentWorkspace` after `removeCollectionFromWorkspace` instead for correctness.

packages/bruno-app/src/providers/ReduxStore/slices/workspaces/getTabToFocusForCurrentWorkspace.js
```diff
@@ -0,0 +1,69 @@
...
+  if (!activeWorkspace) {
+    return null;
+  }
+  const workspaceCollectionUids = getWorkspaceCollectionUids(state, activeWorkspace);
+  if (workspaceCollectionUids.has(activeTab.collectionUid)) {
+    return null;
+  }
+  const inWorkspaceTabs = filter(state.tabs.tabs, (t) => workspaceCollectionUids.has(t.collectionUid));
+  if (inWorkspaceTabs.length > 0) {
+    return { uid: last(inWorkspaceTabs).uid };
```
sanish-bruno, line 57: **Suggestion**: `last(inWorkspaceTabs)` picks the last tab in array order (insertion order), which is the rightmost tab in the tab bar. Consider whether the most recently active tab in the workspace would be a better UX choice. The current behavior is reasonable but may surprise users if they have many open tabs and the last one isn't the one they expect.

packages/bruno-app/src/providers/ReduxStore/slices/workspaces/getTabToFocusForCurrentWorkspace.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+  const scratchCollectionUid = activeWorkspace.scratchCollectionUid;
+  if (!scratchCollectionUid) {
+    return null; // No tabs in current workspace and no scratch; cannot focus a valid tab.
+  }
+  const overviewTabUid = `${scratchCollectionUid}-overview`;
```
sanish-bruno, line 63: **Nit**: The overview tab UID convention `${scratchCollectionUid}-overview` is hardcoded here. If this convention is used elsewhere, consider extracting it to a shared helper (e.g., `getOverviewTabUid(collectionUid)`) to avoid drift if the convention changes.

packages/bruno-app/src/providers/ReduxStore/slices/workspaces/getTabToFocusForCurrentWorkspace.spec.js
```diff
@@ -0,0 +1,111 @@
...
+        tabs: [
+          { uid: 'req-a', collectionUid: 'col-a' },
+          { uid: 'req-b', collectionUid: 'col-b' }
+        ]
+      }
+    });
+    expect(getTabToFocusForCurrentWorkspace(state)).toBeNull();
+  });
+
+  it('returns in-workspace tab when active tab is from another workspace', () => {
```
sanish-bruno, line 57: **Suggestion (test gap)**: This test has both `req-b` (col-b) and `scratch-b-overview` (scratch-b) as in-workspace tabs, and expects the last one (`scratch-b-overview`). Consider adding a test where multiple request tabs from the current workspace exist (e.g., `req-b1`, `req-b2`) to verify the "pick last" behavior is intentional and not just an artifact of having only one candidate.

Also missing: a test for when the active tab has no `collectionUid` at all (malformed state).

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -3105,6 +3107,25 @@ export const scanForBrunoFiles = (dir) => (dispatch, getState) => {
   });
 };
 
+/**
+ * If the current active tab belongs to another workspace, focus a tab in the current workspace.
+ */
+export const ensureActiveTabInCurrentWorkspace = () => (dispatch, getState) => {
```
sanish-bruno, line 3114: **Note (missed call site?)**: `switchWorkspace` in `workspaces/actions.js` calls `setActiveWorkspace` but does NOT dispatch `ensureActiveTabInCurrentWorkspace`. If a user switches workspaces while the active tab belongs to the old workspace, the same bug could occur. Is workspace switching handled elsewhere (e.g., the UI re-renders and selects a workspace-specific tab), or should `ensureActiveTabInCurrentWorkspace` also be called after `switchWorkspace`?

sanish-bruno (approved):

PR #7280, opened 2026-02-24
fix: req.getPath() when URL uses {{BASEURL}} or other vars

### LINKED ISSUE #7248: req.getPath() is empty if the address of the request starts with {{VAR}}

### This bug is:

- [x] annoying

### Bruno version

3.1.3

### Operating System

Windows 11

### Describe the bug

I need to get interpolated path of the request in pre-request script.

If the request looks like this:
https://something.com/path/:p1/:p2
Then req.getPath() properly returns '/path/p1value/p2value'.

But if I try to move the BASE URL to the parameter and the request looks like this:
{{BASEURL}}/path/:p1/:p2
Then req.getPath() returns empty value.

Expected behavior: it should return the interpolated path in both cases.

_No response_

n/a

sanjaikumar-bruno:
[BRU-2778](https://usebruno.atlassian.net/browse/BRU-2778)
Fixes: https://github.com/usebruno/bruno/issues/7248

Pre-request scripts run before URL interpolation, so for URLs like `{{BASEURL}}/path/:p1/:p2`, `req.url` is still a template and `new URL(req.url)` throws, causing `req.getPath()` to return `''`

### DIFF
--- a/packages/bruno-js/src/bruno-request.js
+++ b/packages/bruno-js/src/bruno-request.js
@@ -52,31 +52,47 @@ class BrunoRequest {
   }
 
   getPath() {
+    let pathname;
     try {
       const url = new URL(this.req.url);
-      let pathname = url.pathname;
-
-      // If path params exist, interpolate them into the pathname
-      if (this.req.pathParams && Array.isArray(this.req.pathParams)) {
-        pathname = pathname
-          .split('/')
-          .map((segment) => {
-            if (segment.startsWith(':')) {
-              const paramName = segment.slice(1);
-              const pathParam = this.req.pathParams.find((param) => param.name === paramName);
-              if (pathParam && pathParam.value) {
-                return pathParam.value;
-              }
-            }
-            return segment;
-          })
-          .join('/');
-      }
-
-      return pathname;
+      pathname = url.pathname;
     } catch (e) {
-      return '';
+      if (typeof this.req.url !== 'string') {
+        return '';
+      }
+      const withoutQuery = this.req.url.split('?')[0];
+      if (withoutQuery.startsWith('/')) {
+        pathname = withoutQuery;
+      } else {
+        let searchIn = withoutQuery;
+        const protoEnd = searchIn.indexOf('://');
+        if (protoEnd >= 0) {
+          searchIn = searchIn.substring(protoEnd + 3);
+        }
+        const firstSlash = searchIn.indexOf('/');
+        pathname = firstSlash >= 0 ? searchIn.substring(firstSlash) : '';
+      }
     }
+    return this._interpolatePathParams(pathname) || '';
+  }
+
+  _interpolatePathParams(pathname) {
+    if (!pathname || !this.req.pathParams || !Array.isArray(this.req.pathParams)) {
+      return pathname;
+    }
+    return pathname
+      .split('/')
+      .map((segment) => {
+        if (segment.startsWith(':')) {
+          const paramName = segment.slice(1);
+          const pathParam = this.req.pathParams.find((param) => param.name === paramName);
+          if (pathParam != null && pathParam.value != null) {
+            return pathParam.value;
+          }
+        }
+        return segment;
+      })
+      .join('/');
   }
 
   getQueryString() {
--- a/packages/bruno-js/src/bruno-request.js
+++ b/packages/bruno-js/src/bruno-request.js
@@ -62,14 +62,13 @@ class BrunoRequest {
       if (typeof this.req.url !== 'string') {
         return '';
       }
-      const urlWithoutQuery = this.req.url.split('?')[0];
-      let urlWithoutProtocol = urlWithoutQuery;
-      const protoEnd = urlWithoutProtocol.indexOf('://');
+      let urlWithoutQuery = this.req.url.split('?')[0];
+      const protoEnd = urlWithoutQuery.indexOf('://');
       if (protoEnd >= 0) {
-        urlWithoutProtocol = urlWithoutProtocol.substring(protoEnd + 3);
+        urlWithoutQuery = urlWithoutQuery.substring(protoEnd + 3);
       }
-      const firstSlash = urlWithoutProtocol.indexOf('/');
-      pathname = firstSlash >= 0 ? urlWithoutProtocol.substring(firstSlash) : '';
+      const firstSlash = urlWithoutQuery.indexOf('/');
+      pathname = firstSlash >= 0 ? urlWithoutQuery.substring(firstSlash) : '';
     }
     return interpolatePathParams(pathname, this.req.pathParams) || '';
   }
--- a/packages/bruno-js/src/bruno-request.js
+++ b/packages/bruno-js/src/bruno-request.js
@@ -1,3 +1,5 @@
+const { interpolatePathParams } = require('./utils/url');
+
 class BrunoRequest {
   /**
    * The following properties are available as shorthand:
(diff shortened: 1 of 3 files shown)


packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,42 @@ class BrunoRequest {
...
+    } catch (e) {
+      if (typeof this.req.url !== 'string') {
+        return '';
+      }
+      const withoutQuery = this.req.url.split('?')[0];
+      if (withoutQuery.startsWith('/')) {
+        pathname = withoutQuery;
+      } else {
+        const firstSlash = withoutQuery.indexOf('/');
+        pathname = firstSlash >= 0 ? withoutQuery.substring(firstSlash) : '';
```
sanish-bruno: **Bug (Minor)**: `indexOf('/')` matches the `/` in `://` for URLs with an explicit protocol, producing wrong pathnames.

For `https://{{HOST}}/path`, `indexOf('/')` returns 6 (inside `https://`), so `substring(6)` yields `//{{HOST}}/path` instead of `/path`.

Suggested fix — strip the protocol separator before searching for the path `/`:
```js
      } else {
        let searchIn = withoutQuery;
        const protoEnd = searchIn.indexOf('://');
        if (protoEnd >= 0) {
          searchIn = searchIn.substring(protoEnd + 3);
        }
        const firstSlash = searchIn.indexOf('/');
        pathname = firstSlash >= 0 ? searchIn.substring(firstSlash) : '';
      }
```

This correctly handles `https://{{HOST}}/path` → `/path`, `{{PROTOCOL}}://{{HOST}}/path` → `/path`, and `{{BASEURL}}/path` → `/path` (unchanged).

packages/bruno-js/tests/bruno-request.spec.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    it('strips query string when parsing unparseable URL', () => {
+      const req = new BrunoRequest({
+        url: '{{BASEURL}}/path?foo=bar',
+        method: 'GET',
+        headers: {},
+        pathParams: []
+      });
+      expect(req.getPath()).toBe('/path');
```
sanish-bruno, line 46: **Test gap**: Missing test for URLs with an explicit protocol and template host.

This pattern is common (`https://{{HOST}}/api/resource`) and currently triggers the bug noted above. Adding this test would catch the `://` issue:

```js
it('returns path when URL has protocol with template host (e.g. https://{{HOST}}/path)', () => {
  const req = new BrunoRequest({
    url: 'https://{{HOST}}/api/users',
    method: 'GET',
    headers: {},
    pathParams: []
  });
  expect(req.getPath()).toBe('/api/users');
});
```

sanish-bruno (commented):
## PR Review

Found **2** issue(s): 1 minor bug, 1 test gap.

### Summary
- Good fix for the core issue (`{{BASEURL}}/path` pattern). The fallback string parsing and extraction of `_interpolatePathParams` are clean.
- URLs with an explicit protocol + template host (e.g., `https://{{HOST}}/path`) produce incorrect pathnames because `indexOf('/')` matches the `/` in `://` rather than the path separator.
- The `pathParam.value` null-check improvement (`!= null` instead of truthiness) is a nice fix for edge cases like `value: 0`.

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,47 @@ class BrunoRequest {
   }
 
   getPath() {
+    let pathname;
     try {
       const url = new URL(this.req.url);
```
sanish-bruno, line 59: @sanjaikumar-bruno can you check if we can use `import { patternHasher } from '@usebruno/common/utils';` patternHasher function to handle this? we can pre-process the `this.req.url` with `patternHasher`, before passing it to the `new URL()` i guess it may solve the issue, please verify

sanish-bruno (changes requested):

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,47 @@ class BrunoRequest {
...
-            return segment;
-          })
-          .join('/');
+      pathname = url.pathname;
+    } catch (e) {
+      if (typeof this.req.url !== 'string') {
+        return '';
+      }
+      const withoutQuery = this.req.url.split('?')[0];
+      if (withoutQuery.startsWith('/')) {
```
sanish-bruno: will the req.url ever start with a '/'?

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,47 @@ class BrunoRequest {
...
-            }
-            return segment;
-          })
-          .join('/');
+      pathname = url.pathname;
+    } catch (e) {
+      if (typeof this.req.url !== 'string') {
+        return '';
+      }
+      const withoutQuery = this.req.url.split('?')[0];
```
sanish-bruno: rename it to `urlWithoutQuery`

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,47 @@ class BrunoRequest {
...
+      if (withoutQuery.startsWith('/')) {
+        pathname = withoutQuery;
+      } else {
+        let searchIn = withoutQuery;
```
sanish-bruno: can think of a better descriptive name

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,47 @@ class BrunoRequest {
...
+        const protoEnd = searchIn.indexOf('://');
+        if (protoEnd >= 0) {
+          searchIn = searchIn.substring(protoEnd + 3);
```
sanish-bruno: `urlWithoutProtocol`

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,47 @@ class BrunoRequest {
...
+        }
+        const firstSlash = searchIn.indexOf('/');
+        pathname = firstSlash >= 0 ? searchIn.substring(firstSlash) : '';
```
sanish-bruno: Image

sanish-bruno: please verify is the code also follows the same behaviour

sanish-bruno (changes requested):

packages/bruno-js/src/bruno-request.js
sanjaikumar-bruno: Verified!

sanish-bruno (dismissed):

packages/bruno-js/src/bruno-request.js
```diff
@@ -52,31 +52,43 @@ class BrunoRequest {
...
+      if (protoEnd >= 0) {
+        urlWithoutProtocol = urlWithoutProtocol.substring(protoEnd + 3);
       }
+      const firstSlash = urlWithoutProtocol.indexOf('/');
+      pathname = firstSlash >= 0 ? urlWithoutProtocol.substring(firstSlash) : '';
+    }
+    return this._interpolatePathParams(pathname) || '';
+  }
 
+  _interpolatePathParams(pathname) {
```
bijin-bruno: I think we should use the common util `packages/bruno-app/src/utils/url/index.js`. We could modify this as required.
Any reason to add separate logic?

bijin-bruno (changes requested):

Willt123:
@sanish-bruno just wanted to make sure this branch hadn't been forgotten as its currently broken our whole collection. 
Thanks for the great work team.

sanish-bruno:
Hi @Willt123 , thanks for checking in, and sorry for the disruption this has caused.

We're actively tracking this issue. At the moment, we're focused on a few higher-priority items, so we won't be able to address it immediately, and we appreciate your patience.

Please rest assured that this hasn't been forgotten, we do intend to pick it up, but it may take a little more time before we can get to it. Thanks again for your understanding and for taking the time to report it. : )

Issue #7034, opened 2026-02-03
`Interpolate Variables` in Path is not respecting toggle choice.


Abhinav1217:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

3.0.2

### Operating System

MacOs

### Describe the bug

I keep base part of as env variable so that i can switch test/dev/local URLs easily. I would prefer sharing it as same in curl 
Currently checkbox setting for interoplate variable is working in headers/body  but not in URL 

For example look at following screenshot

<img width="765" height="451" alt="Image" src="https://github.com/user-attachments/assets/c78b78ed-b46b-4309-8e34-aaefe005c93e" />

I am exporting it with interpolation off but still on URL, it is replacing it with current value from env.

<img width="969" height="451" alt="Image" src="https://github.com/user-attachments/assets/d815373e-19ca-4f76-b5d6-54582223387f" />

I was expectiong URL to be like this `{{booking-service}}/subscription/plans/tiers?planId={{planId}}`

### .bru file to reproduce the bug

Can't share current curl from screenshot but you can try something like this
```
{
  google: "https://www.google.com",
  bing: "https://www.bing.com",
};

and then 
{{google}}/search?q={{query}} 
{{bing}}/search?q={{query}}
```

### Screenshots/Live demo link

<img width="969" height="451" alt="Image" src="https://github.com/user-attachments/assets/d815373e-19ca-4f76-b5d6-54582223387f" />

pooja-bruno:
@Abhinav1217, Thanks for creating the issue. We were already discussing this internally and will be working on a fix soon.

helloanoop:
Just verified the fix on the latest main branch.

https://github.com/user-attachments/assets/1375d4e2-3bc8-41ec-a7ed-cfb3a3f59da9

This will be available in our upcoming release `v4.1.0` scheduled to go out on 3rd week of August.
If anyone wants to get this / try out earlier - you can refer our [nightly builds](https://github.com/usebruno/bruno-nightly-builds)

Closing this issue.

Issue #3328, opened 2024-10-20
POST request which works fine from CURL does not in Bruno


Rickcau:
### I have checked the following:

- [X] I use the newest version of bruno.
- [X] I've searched existing issues and found nothing related to my issue.

### Describe the bug

Bruno fails with a 404 error with trying to make a POST request to Azure AI Language service.  The came POST request works fine from CURL.  This fine from Postman and CURL just not Bruno (see the screen shot below for example of this working from CURL)

Steps to Reproduce the issue.
1. Open the Azure Portal and setup an AI Language Service
2. Navigate to the service and copy the endpoint URL as well as the API Key
3. You can verify that the service is working through the Language Studio by trying the PII demo and point it to the service you just provisioned.

Next Steps:
1. Open a Command Prompt on Windows and run either of the CURL commands below, but make sure to replace the endpoint and the key with the service you just provisioned.

Example1: 
curl -v -X POST "https://ai-language-service-cw.cognitiveservices.azure.com/language/:analyze-text?api-version=2023-04-01" -H "Content-Type: application/json" -H "Ocp-Apim-Subscription-Key: 123_your_key" --data-ascii "{\"kind\":\"PiiEntityRecognition\",\"analysisInput\":{\"documents\":[{\"id\":\"documentId\",\"text\":\"Hello, my SSN is 111-11-1111\",\"language\":\"en\"}]},\"parameters\":{\"domain\":\"none\"}}"

Example2: 
curl --request POST \
  --url 'https://ai-language-service.cognitiveservices.azure.com/language/?api-version=2023-04-01' \
  --header 'Content-Type: application/json' \
  --header 'Ocp-Apim-Subscription-Key: 1234-your-key' \
  --header 'content-type: application/json' \
  --data '"{\"kind\":\"PiiEntityRecognition\",\"analysisInput\":{\"documents\":[{\"id\":\"documentId\",\"text\":\"Hello, my SSN is 111-11-1111\",\"language\":\"en\"}]},\"parameters\":{\"domain\":\"none\"}}"'

### .bru file to reproduce the bug

{
      "type": "http",
      "name": "analyze-text4",
      "seq": 8,
      "request": {
        "url": "https://ai-language-service-cw.cognitiveservices.azure.com/language/:analyze-text?api-version=2023-04-01",
        "method": "POST",
        "headers": [
          {
            "name": "Content-Type",
            "value": "application/json",
            "enabled": true
          },
          {
[...]

PR #7308, opened 2026-02-26
fix(generate-code): generate code for url interpolation

pooja-bruno:
### Description

The request URL ignored the "Interpolate Variables" toggle: `{{var}}` / `:pathParam` were always resolved in the generated snippet, even with interpolation off. Headers and body respected the toggle correctly.

[JIRA](https://usebruno.atlassian.net/browse/BRU-2784)
Issue: https://github.com/usebruno/bruno/issues/7034

#### Contribution Checklist:

Before
https://github.com/user-attachments/assets/44f0630a-50da-42e4-9229-912c8a46cd21

After
https://github.com/user-attachments/assets/92a1ab3e-737a-4237-8ad1-43d7e948f2fd

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL validation and error messaging in code snippet generation (UI and errors now use a dedicated validation URL).
  * Corrected construction of generated request URLs to rely on request data and preserved raw URL behavior; path-parameter interpolation now behaves consistently.
  * HAR generation uses interpolated URLs while non-HAR outputs retain original request URLs when interpolation is disabled.

* **Tests**
  * Added/updated tests covering interpolated and non-interpolated URL behaviors, path-parameter scenarios, and interactions between raw and interpolated URLs.


packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
```diff
@@ -81,6 +86,9 @@ const generateSnippet = ({ language, item, collection, shouldInterpolate = false
     if (language.target === 'shell' && language.client === 'curl') {
       result = addCurlAuthFlags(result, effectiveAuth);
     }
+    if (!shouldInterpolate) {
+      result = result.replace(harUrl, templateUrl);
```
sanish-bruno: **Bug (Major)**: This `replace` correctly restores the template URL, but the `encodeUrl` replacement logic below (~lines 95–115) will **undo** it.

After this PR, `request.url` is the template URL (e.g., `https://api.example.com/{{endpoint}}`), while `item.rawUrl` (set by `index.js`) is the interpolated URL (e.g., `https://api.example.com/data`).

The `encodeUrl` logic computes:
- `httpSnippetPath` from `parse(request.url)` → `'/{{endpoint}}'`
- `desiredPath` from `stripOrigin(item.rawUrl)` → `'/data'`

Since these differ, `result.replaceAll('/{{endpoint}}', '/data')` fires on line ~113 and replaces the template path back with the interpolated path — completely undoing this restoration.

The tests don't catch this because they construct items directly without setting `item.rawUrl`, but the real app flow (via `index.js`) always sets it.

Suggested fix: wrap the `encodeUrl` replacement in an `else` so it only runs when interpolation is enabled, since template URLs with `{{variables}}` can't be meaningfully encoded anyway:

```js
if (!shouldInterpolate) {
  result = result.replaceAll(harUrl, templateUrl);
} else {
  // existing encodeUrl replacement logic
  const settings = item.draft ? get(item, 'draft.settings') : get(item, 'settings');
  const rawUrl = item.rawUrl || request.url;
  // ... rest of encodeUrl logic ...
}
```

---

**Bug (Minor)**: `replace()` only replaces the **first** occurrence. Some snippet targets (e.g., Python requests, Node http) may reference the URL more than once. Use `replaceAll()` for consistency with line ~113.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
```diff
@@ -1241,3 +1247,116 @@ describe('generateSnippet – encodeUrl setting', () => {
...
+  const baseCollection = {
+    root: { request: { auth: { mode: 'none' }, headers: [] } },
+    globalEnvironmentVariables: {
+      host: 'https://api.example.com'
+    },
+    runtimeVariables: {},
+    processEnvVariables: {}
+  };
+
+  it('should NOT interpolate URL variables when shouldInterpolate is false', () => {
```
sanish-bruno: **Test gap**: None of the tests in this new `describe` block set `item.rawUrl`, which the real app flow (via `index.js`) **always** sets. This means the interaction between the URL template restoration (`replace(harUrl, templateUrl)`) and the `encodeUrl` path-replacement logic (which reads `item.rawUrl`) is untested — and it's where the major bug above manifests.

Add at least one test with `rawUrl` set on the item to verify the template URL survives the full `encodeUrl` replacement pipeline. For example:

```js
it('should preserve template URL when rawUrl is set and shouldInterpolate is false', () => {
  const item = {
    uid: 'url-test-rawurl',
    request: {
      method: 'GET',
      url: '{{host}}/ping',
      headers: [],
      body: { mode: 'none' },
      auth: { mode: 'none' },
      params: []
    },
    rawUrl: 'https://api.example.com/ping'
  };

  const result = generateSnippet({
    language,
    item,
    collection: baseCollection,
    shouldInterpolate: false
  });

  expect(result).toContain('{{host}}/ping');
  expect(result).not.toContain('https://api.example.com/ping');
});
```

sanish-bruno (commented):
## PR Review

Found **3** issue(s): 1 major, 1 minor, 1 test gap.

### Summary
- The core fix is correct: always interpolate the URL for HTTPSnippet (which needs a valid URL), then swap the template URL back when `shouldInterpolate` is false.
- However, the `encodeUrl` path-replacement logic below the new code will **undo** the template URL restoration in the real app flow, because `request.url` is now the template URL while `item.rawUrl` is the interpolated URL — causing a path mismatch that triggers an unwanted `replaceAll`.
- The new tests don't cover this because they never set `item.rawUrl`, but `index.js` always sets it.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
```diff
@@ -1241,3 +1247,116 @@ describe('generateSnippet – encodeUrl setting', () => {
...
+      item,
+      collection: baseCollection,
+      shouldInterpolate: false
+    });
+
+    expect(result).toContain('/users/:userId');
+    expect(result).not.toContain('/users/123');
+  });
+
+  it('should interpolate both URL variables and path params when shouldInterpolate is true', () => {
```
sanish-bruno: @pooja-bruno it is a far-fetched case, but just want to know what happens when we have another variable as the path param value?

`params: { name: {{userId}}, value: {{numValue}}, type: 'path', enabled: true }` ? do we interpolate it?

sanish-bruno (changes requested):

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
pooja-bruno: Thanks for the review! The bug doesn't actually reproduce — url.parse (line 97) auto-encodes `{{` to `%7B%7B`, so the replaceAll never matches the  
literal curlies in the result.                                                                                                                  
                                                                                                                                                  
I tried the else block approach but it breaks 16 existing encodeUrl tests the encodeUrl logic needs to run regardless of the interpolation toggle . So I've kept the original structure and just applied the replace → replaceAll change + added a test with rawUrl set.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
pooja-bruno: Yes,  variables in path param values get interpolated when shouldInterpolate is true, and preserved as-is when false

packages/bruno-common/src/generate-code/har/index.ts
```diff
@@ -584,6 +589,9 @@ const buildPostData = (body: BrunoBody | undefined): any => {
 // Main
 // ---------------------------------------------------------------------------
 
+// Stand-in scheme for URLs whose real scheme arrives via a `{{var}}`. Only ever
+const SYNTHETIC_SCHEME = 'http://';
+
```
shubh-bruno: please move this to common

packages/bruno-common/src/generate-code/har/index.ts
```diff
@@ -584,6 +589,9 @@ const buildPostData = (body: BrunoBody | undefined): any => {
 // Main
 // ---------------------------------------------------------------------------
 
+// Stand-in scheme for URLs whose real scheme arrives via a `{{var}}`. Only ever
```
sachin-thakur-bruno: incomplete comment?

shubh-bruno (changes requested):

packages/bruno-common/src/generate-code/har/index.ts
```diff
@@ -601,14 +609,29 @@ export async function buildHar(input: BuildHarInput): Promise<BuildHarOutput> {
...
+  // one, so a genuinely schemeless `localhost:6000/x` still renders the `http://` the
+  // client will use. Gating on a *leading* placeholder keeps the gate strict for
+  // everything else: `not a url` must stay rejected, not become `http://not a url`.
+  const needsSyntheticScheme = (working.url || '').startsWith('{{') && !hasExplicitScheme(hashedUrl);
+  const syntheticSchemeIsRedundant = needsSyntheticScheme
+    && hasExplicitScheme(interpolate(working.url || '', variables) || '');
+  // The token the synthetic scheme was prepended to, so the strip below can target
+  // that one position instead of every `http://` in the snippet.
+  const leadingToken = needsSyntheticScheme ? hashedUrl.match(/^[A-Za-z0-9._-]+/)?.[0] ?? '' : '';
+  const urlForPipeline = needsSyntheticScheme ? SYNTHETIC_SCHEME + hashedUrl : hashedUrl;
```
sachin-thakur-bruno: There is bug where in python3 if interpolations is off getting conn = http.client.HTTPConnection("{{host}}") instead of conn = http.client.HTTPSConnection("{{host}}")

tests/request/generate-code/url-interpolation.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+    const interpolated = await readGeneratedSnippet(page);
+    expect(interpolated).toContain('http://localhost:8081/api/echo/anything/ping');
+    expect(interpolated).not.toContain('{{host}}');
+
+    await setInterpolateVariables(page, false);
+    const templated = await readGeneratedSnippet(page);
+    expect(templated).toContain('{{host}}/api/echo/anything/ping');
+    expect(templated).not.toContain('http://localhost:8081/api/echo/anything/ping');
+
+    await closeGenerateCodeDialog(page);
```
sachin-thakur-bruno: if test fails then this will not close. can we close this in afterEach? and then remove it from below test also.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.js
```diff
@@ -31,6 +32,19 @@ const addCurlAuthFlags = (curlCommand, auth) => {
...
+const prependDefaultScheme = (url) => {
+  // No URL to work with — let buildHar report it rather than inventing one here.
+  if (!url) return url;
+
+  if (hasExplicitScheme(url)) return url;
+
+  if (url.startsWith('{{')) return url;
+
+  return `${DEFAULT_SCHEME}${url}`;
+};
```
shubh-bruno:
```
const prependDefaultScheme = (url) => {
  if (!url || hasExplicitScheme(url) || url.startsWith('{{')) {
    return url;
  }

  return `${DEFAULT_SCHEME}${url}`;
};
```
Please update the function

packages/bruno-common/src/generate-code/har/index.ts
```diff
@@ -277,9 +277,14 @@ const hashPathParamPositions = (
   pathParams: BrunoKV[] | undefined
 ): { url: string; restore: (input: string, opts: { encode: boolean }) => string } => {
   const noopRestore = (input: string) => input;
-  if (!url || !Array.isArray(pathParams) || pathParams.length === 0) {
+  if (!url) {
```
shubh-bruno, line 280: Please check this condition change once we might have regression for more context refer to the issue [#8771 ](https://github.com/usebruno/bruno/issues/8771) 

And please add test case for it in `packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js`

For more context refer to the issue : 
`https://github.com/usebruno/bruno/issues/7524`
`https://github.com/usebruno/bruno/issues/3328`  

Test URL : `http://localhost:6000/values:colon`

Please add a unit test and

sid-bruno (commented):
Go through all tests and make sure we have mixed cases handled where one variables value might be missing, verify and validate if the code snippet gets rendered properly with the actual variable name being printed into the snippet

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
sid-bruno, line 1: Missing a case for when interpolation is on but the value isn't provided, can we add one for that as well to see that it generates properly with the `{{` instead of erroring out

shubh-bruno (changes requested):

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/url-validation.js
sid-bruno, line 1: Built har already handles this, you just need to test it, don't need another util for it

sid-bruno (approved):

sanish-bruno (approved):

packages/bruno-common/src/generate-code/har/index.ts
shubh-bruno: Please check this issue, I'm still seeing `http://localhost:6000/values:colon` when URL encoding toggle is ON i was expecting to be  `http://localhost:6000/values%3Acolon.`

PR #7315, opened 2026-02-26
fix(`oauth2`): prevent false callback matches on `root path URLs`

lohit-bruno:
[jira](https://usebruno.atlassian.net/browse/BRU-2004)

### Description

Move error check before callback URL matching in onWindowRedirect so OAuth error responses are rejected immediately. Remove redundant error param check from matchesCallbackUrl and require a code param or hash fragment to match, preventing false positives on intermediate pages when the callback URL is a root path like https://hostname/.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected OAuth callback validation to require OAuth indicators (code or hash) and handle absent URLs safely.
  * Prioritized OAuth error detection during redirects, returning descriptive errors and closing the window before further processing.
  * Improved authorization window lifecycle and listener cleanup to avoid extra processing after error or completion.

* **Tests**
  * Added unit tests covering callback URL matching, implicit/authorization flows, and null-URL handling.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
+++ b/packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
@@ -151,7 +151,7 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session, additionalH
         callbackUrlObj = null;
       }
 
-      // Handle OAuth error responses on the callback URL first, so we reject with
+      // Handle OAuth error responses first, so we reject with
       // a descriptive error instead of resolving with a null authorization code
       if (urlObj.searchParams.has('error')) {
         const error = urlObj.searchParams.get('error');


packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
```diff
@@ -145,14 +151,8 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session, additionalH
...
 
-      // Check if redirect is to the callback URL and contains an authorization code
-      if (callbackUrlObj && matchesCallbackUrl(urlObj, callbackUrlObj)) {
-        finalUrl = url;
-        window.close();
-        return;
-      }
-
-      // Handle OAuth error responses
+      // Handle OAuth error responses on the callback URL first, so we reject with
```
sanish-bruno: **Suggestion**: The error param check applies to **all** redirected URLs, not just the callback URL. If an intermediate IdP page happens to redirect with a `?error=` query parameter (e.g. `https://idp.example.com/authorize?error=session_expired` for a login retry), the flow would be prematurely rejected.

This is pre-existing behavior (the old code had the same un-gated check), but since the comment now says "on the callback URL", consider scoping the check to only fire when the URL matches the callback host+path:

```js
if (callbackUrlObj && urlObj.href.startsWith(callbackUrlObj.href) && urlObj.searchParams.has('error')) {
```

This keeps the benefit of rejecting error responses on the actual callback URL while avoiding false positives from intermediate pages.

PR #7331, opened 2026-02-27
fix: allow user to delete default bruno headers in pre-request

shubh-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2460)

`req.deleteHeader()` and `req.deleteHeaders()` now correctly suppress default headers (`user-agent`, `accept`, `accept-encoding`, `connection`, etc.) added by Bruno's runtime when used in pre-request scripts.

#### What Was Broken

Calling `req.deleteHeader('user-agent')` in a pre-request script had no effect — the header still arrived at the backend. Same for `accept-encoding`, `connection`, and others.

#### Root Causes & Fixes

| # | Problem | Fix |
|---|---------|-----|
| 1 | `config.headers.delete()` removed the key entirely, so the axios adapter re-added its default (`axios/1.x.x`, `gzip, compress, deflate, br`) via a "don't overwrite if already set" guard | Use `config.headers.set(name, null)` — keeps the key present (blocks the guard) while `toJSON()` omits `null` values from the wire |
| 2 | `connection: keep-alive` is injected by `http.Agent` at the socket level, not by axios — deleting from headers had no effect | Override the agent with `keepAlive: false` per request; in Electron, apply this **after** `setupProxyAgents()` to avoid being overwritten |
| 3 | `deleteHeaders([...])` (plural) never wrote to `req.__headersToDelete`, so it couldn't suppress default headers | Delegate each item to `deleteHeader()` |
| 4 | Network Console showed deleted headers because the timeline was logged before the deletion ran | Move timeline header logging to after the deletion block; skip `null`/`false` values |

### Supported Headers

| Header | Deletable | Notes |
|--------|-----------|-------|
| `user-agent` | Yes | |
| `accept` | Yes | |
| `accept-encoding` | Yes | |
| `connection` | Yes | Agent override — `Connection` header is fully absent |
| `request-start-time` | Yes | Bruno internal header |
| `host` | No | Required by HTTP; silently ignored |

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Selective header removal for outgoing HTTP requests, with safeguards for connection-related headers and more reliable header handling.
  * Request initialization now uses a clean header baseline to ensure deletions and overrides behave predictably.

* **Tests**
  * Added comprehensive tests covering header deletion, duplicate handling, and edge cases.

### DIFF
--- a/packages/bruno-cli/src/utils/axios-instance.js
+++ b/packages/bruno-cli/src/utils/axios-instance.js
@@ -99,18 +99,9 @@ function makeAxiosInstance({ requestMaxRedirects = 5, disableCookies, followRedi
     */
     const headersToDelete = config.__headersToDelete;
     if (headersToDelete && Array.isArray(headersToDelete)) {
-      const http = require('http');
-      const https = require('https');
       headersToDelete.forEach((headerName) => {
         const lower = headerName.toLowerCase();
-        if (lower === 'host') return;
-        if (lower === 'connection') {
-          // connection is set by Node.js http.Agent at socket level, not by axios.
-          // keepAlive:false means Node.js does not inject Connection:keep-alive.
-          config.httpAgent = new http.Agent({ keepAlive: false });
-          config.httpsAgent = new https.Agent({ keepAlive: false });
-          return;
-        }
+        if (lower === 'host' || lower === 'connection') return;
         config.headers.set(headerName, null);
       });
       delete config.__headersToDelete;
--- a/packages/bruno-electron/src/ipc/network/axios-instance.js
+++ b/packages/bruno-electron/src/ipc/network/axios-instance.js
@@ -220,15 +220,6 @@ function makeAxiosInstance({
       });
     }
 
-    // Override the agent AFTER setupProxyAgents so keepAlive:true doesn't win.
-    // With keepAlive:false Node.js does not inject Connection:keep-alive at all.
-    if (deleteConnection) {
-      const http = require('http');
-      const https = require('https');
-      config.httpAgent = new http.Agent({ keepAlive: false });
-      config.httpsAgent = new https.Agent({ keepAlive: false });
-    }
-
     config.metadata.timeline = timeline;
     return config;
   });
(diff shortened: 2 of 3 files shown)


packages/bruno-cli/src/utils/axios-instance.js
```diff
@@ -78,14 +78,44 @@ function makeAxiosInstance({ requestMaxRedirects = 5, disableCookies, followRedi
...
+      const http = require('http');
+      const https = require('https');
+      headersToDelete.forEach((headerName) => {
+        const lower = headerName.toLowerCase();
+        if (lower === 'host') return;
+        if (lower === 'connection') {
+          // connection is set by Node.js http.Agent at socket level, not by axios.
+          // keepAlive:false means Node.js does not inject Connection:keep-alive.
+          config.httpAgent = new http.Agent({ keepAlive: false });
+          config.httpsAgent = new https.Agent({ keepAlive: false });
```
sid-bruno: If the `connection` header needs to be removed, you remove it, there's no need for the agents to be re-initialised, this will conflict with any existing proxy agents that were added

packages/bruno-electron/src/ipc/network/axios-instance.js
```diff
@@ -195,6 +219,16 @@ function makeAxiosInstance({
...
       });
     }
+
+    // Override the agent AFTER setupProxyAgents so keepAlive:true doesn't win.
+    // With keepAlive:false Node.js does not inject Connection:keep-alive at all.
+    if (deleteConnection) {
+      const http = require('http');
+      const https = require('https');
+      config.httpAgent = new http.Agent({ keepAlive: false });
+      config.httpsAgent = new https.Agent({ keepAlive: false });
```
sid-bruno: Same as https://github.com/usebruno/bruno/pull/7331/changes#r2876590193

packages/bruno-requests/src/network/axios-instance.ts
```diff
@@ -51,10 +52,35 @@ const makeAxiosInstance = (customRequestConfig?: AxiosRequestConfig) => {
...
+    const headersToDelete = modConfig.__headersToDelete;
+    if (headersToDelete && Array.isArray(headersToDelete)) {
+      headersToDelete.forEach((headerName: string) => {
+        const lower = headerName.toLowerCase();
+        if (lower === 'host') return;
+        if (lower === 'connection') {
+          // connection is set by Node.js http.Agent at socket level, not by axios.
+          // keepAlive:false means Node.js does not inject Connection:keep-alive.
+          modConfig.httpAgent = new http.Agent({ keepAlive: false });
+          modConfig.httpsAgent = new https.Agent({ keepAlive: false });
```
sid-bruno: same as https://github.com/usebruno/bruno/pull/7331/changes#r2876590193

PR #7344, opened 2026-03-02
FIX: Sync URL Input when pasting a cURL command

gitKashish:
### Description

After the recent update when I tried pasting in the a cURL command in the URL text input. The cURL was parsed properly and it was updated in the app state as well. But the URL input did not reflect the API URL.

I think it was due to the `event.PreventDefault()` on the paste operation that was causing the issue and causing the single line editor to not reflect changes, since its paste event was interrupted.

So I explicitly set the URL once the cURL command is parsed. I hope this seems like a valid approach. I made the changes for both `handleGraphqlPaste()` and `handleHttpPaste()` functions.

Thanks.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where URL changes from pasting curl or GraphQL code were not immediately reflected in the editor UI, ensuring real-time synchronization.

### DIFF
--- a/packages/bruno-app/src/components/RequestPane/QueryUrl/index.js
+++ b/packages/bruno-app/src/components/RequestPane/QueryUrl/index.js
@@ -112,6 +112,11 @@ const QueryUrl = ({ item, collection, handleRun }) => {
         url: request.url
       }));
 
+      // Update URL in the editor UI
+      if (editorRef.current?.editor) {
+        editorRef.current.editor.setValue(request.url);
+      }
+
       // Update method
       dispatch(updateRequestMethod({
         method: request.method.toUpperCase(), // Convert to uppercase
@@ -193,6 +198,11 @@ const QueryUrl = ({ item, collection, handleRun }) => {
           url: request.url
         })
       );
+
+      // Update URL in the editor UI
+      if (editorRef.current?.editor) {
+        editorRef.current.editor.setValue(request.url);
+      }
 
       // Update method
       if (request.method) {


gitKashish:
Fixes #7345

chirag-bruno:
Hey @gitKashish, thanks for taking the time to work on this. We’ll review your changes over the next day or two and get back to you with an update.

naman-bruno (approved):

gitKashish:
Hi @chirag-bruno, thanks for responding. Is there some update on this?

cyb3rd0g-cell:
Any update on this bug fix?

sid-bruno:
@chirag-bruno can you validate if this needs to be fixed in the transient request changes or if it's a bug with the Editor instance itself. I'd assume that the redux state is getting the `request.url` correctly, can you validate if changes from https://github.com/usebruno/bruno/pull/7098 are responsible for this. 

@gitKashish , thanks for the PR, we'd like to look deeper into it before commiting to the fix from the PR since the fix from the PR would also mean that an additional save handler will need to be dispatched when not in a Temporary request.

Aman-14:
I am also facing the same problem while importing curl, and investigated the core issue. 

The problem is in this block of code from #7098
```js
if (this.props.value !== prevProps.value && this.props.value !== this.cachedValue && this.editor) {
  // TODO: temporary fix for keeping cursor state when auto save and new line insertion collide PR#7098
  const nextValue = String(this.props.value ?? '');
  const currentValue = this.editor.getValue();
  if (this.editor.hasFocus?.() && currentValue !== nextValue && nextValue !== '') {
    this.cachedValue = currentValue;
  } else {
    const cursor = this.editor.getCursor();
    this.cachedValue = nextValue;
    this.editor.setValue(nextValue);
    this.editor.setCursor(cursor);

    // Update newline markers after value change
    if (this.props.showNewlineArrow) {
      this._updateNewlineMarkers();
    }
  }
}
```

This line: `if (this.editor.hasFocus?.() && currentValue !== nextValue && nextValue !== '')` evaluates to true during paste which prevents updating the editor UI.
- `this.editor.hasFocus()` -> true while pasting curl
- `currentValue !== nextValue` -> true (current is '' and next is URL)
- `nextValue !==''` -> true (next is URL)

mfcallahan:
I had to go back to Postman 😢 This is bug is a showstopper for me and my team.

PR #7403, opened 2026-03-09
fix: validate environment variables in unsaved changes dialog

### LINKED ISSUE #7339: The "unsaved changes" feature does not validate config files

### This bug is:

- [x] annoying

### Bruno version

3.14

### Operating System

Windows 11

### Describe the bug

In a (collection) environment config, add a variable name which includes a "space" character
UI validator reports that the name contains invalid characters (OK)
Close Bruno
The "unsaved changes" dialog is shown.
Select Save
Open Bruno
The environment config is not listet (failed to load)

I.e. the "unsaved changes" dialog does not validate config and  thus allows the user to save an invalid config.

vars {
  valid_name: value
  invalid name: value
}

Image

chirag-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2821)

The unsaved changes dialog was bypassing validation when saving environment configs, allowing invalid variable names (e.g., containing spaces) to be saved. This resulted in corrupted config files that failed to load on app restart.

Added hasInvalidVariableNames utility function in regex.js and validation before saving environments from:
- Tab close unsaved changes dialog (RequestTab)
- App close unsaved changes dialog (SaveRequestsModal)

Also added unit tests for hasInvalidVariableNames function.

Fixes #7339

#### Contribution Checklist:

## Release Notes

* **New Features**
  * Environment variable names are now validated before saving configurations. Invalid variable names trigger detailed error notifications and prevent the save operation, ensuring configurations contain only properly formatted variable names.

* **Tests**
  * Added comprehensive test coverage for variable name validation, including handling of valid formats, invalid characters, and edge cases.

### DIFF
--- a/packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
+++ b/packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
@@ -22,7 +22,7 @@ import NewRequest from 'components/Sidebar/NewRequest/index';
 import GradientCloseButton from './GradientCloseButton';
 import { flattenItems } from 'utils/collections/index';
 import { closeWsConnection } from 'utils/network/index';
-import { hasInvalidVariableNames } from 'utils/common/regex';
+import { getInvalidVariableNames } from 'utils/common/variables';
 import ExampleTab from '../ExampleTab';
 import toast from 'react-hot-toast';
 
@@ -357,8 +357,9 @@ const RequestTab = ({ tab, collection, tabIndex, collectionRequestTabs, folderUi
                 window.addEventListener('dotenv-save-failed', onFailed, { once: true });
                 window.dispatchEvent(new Event('dotenv-save'));
               } else if (draft?.environmentUid && draft?.variables) {
-                if (hasInvalidVariableNames(draft.variables)) {
-                  toast.error('Please fix validation errors before saving');
+                const invalidNames = getInvalidVariableNames(draft.variables);
+                if (invalidNames.length > 0) {
+                  toast.error(`Invalid variable name(s): ${invalidNames.join(', ')}`);
                   return;
                 }
                 dispatch(saveEnvironment(draft.variables, draft.environmentUid, collection.uid))
@@ -407,8 +408,9 @@ const RequestTab = ({ tab, collection, tabIndex, collectionRequestTabs, folderUi
                 window.addEventListener('dotenv-save-failed', onFailed, { once: true });
                 window.dispatchEvent(new Event('dotenv-save'));
               } else if (draft?.environmentUid && draft?.variables) {
-                if (hasInvalidVariableNames(draft.variables)) {
-                  toast.error('Please fix validation errors before saving');
+                const invalidNames = getInvalidVariableNames(draft.variables);
+                if (invalidNames.length > 0) {
+                  toast.error(`Invalid variable name(s): ${invalidNames.join(', ')}`);
                   return;
                 }
                 dispatch(saveGlobalEnvironment({ variables: draft.variables, environmentUid: draft.environmentUid }))
--- a/packages/bruno-app/src/providers/App/ConfirmAppClose/SaveRequestsModal.js
+++ b/packages/bruno-app/src/providers/App/ConfirmAppClose/SaveRequestsModal.js
@@ -6,7 +6,7 @@ import { useSelector } from 'react-redux';
 import { useDispatch } from 'react-redux';
 import { findCollectionByUid, flattenItems, isItemARequest, hasRequestChanges, findEnvironmentInCollection } from 'utils/collections';
 import { pluralizeWord } from 'utils/common';
-import { hasInvalidVariableNames } from 'utils/common/regex';
+import { getInvalidVariableNames } from 'utils/common/variables';
 import { completeQuitFlow } from 'providers/ReduxStore/slices/app';
 import { saveMultipleRequests, saveMultipleCollections, saveMultipleFolders, saveEnvironment, closeTabs } from 'providers/ReduxStore/slices/collections/actions';
 import { saveGlobalEnvironment, clearGlobalEnvironmentDraft } from 'providers/ReduxStore/slices/global-environments';
@@ -153,13 +153,6 @@ const SaveRequestsModal = ({ onClose, forceCloseTabs = false, tabUidsToClose = [
       const collectionEnvironmentDrafts = allDrafts.filter((d) => d.type === 'collection-environment');
       const globalEnvironmentDrafts = allDrafts.filter((d) => d.type === 'global-environment');
 
-      const allEnvironmentDrafts = [...collectionEnvironmentDrafts, ...globalEnvironmentDrafts];
-      const hasInvalidEnvDraft = allEnvironmentDrafts.some((draft) => hasInvalidVariableNames(draft.variables));
-      if (hasInvalidEnvDraft) {
-        toast.error('Please fix validation errors before saving');
-        return;
-      }
-
       // Save all collection drafts
       if (collectionDrafts.length > 0) {
         await dispatch(saveMultipleCollections(collectionDrafts));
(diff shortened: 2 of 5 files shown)


sanish-bruno:
@chirag-bruno please rebase with main, check if tests are still failing

packages/bruno-app/src/providers/App/ConfirmAppClose/SaveRequestsModal.js
```diff
@@ -119,6 +121,13 @@ const SaveRequestsModal = ({ onClose }) => {
       const collectionEnvironmentDrafts = allDrafts.filter((d) => d.type === 'collection-environment');
       const globalEnvironmentDrafts = allDrafts.filter((d) => d.type === 'global-environment');
 
+      const allEnvironmentDrafts = [...collectionEnvironmentDrafts, ...globalEnvironmentDrafts];
+      const hasInvalidEnvDraft = allEnvironmentDrafts.some((draft) => hasInvalidVariableNames(draft.variables));
+      if (hasInvalidEnvDraft) {
```
sanish-bruno: @chirag-bruno how can the users know, what actually went wrong? do we show them what validation got failed? the error seems a bit generic

packages/bruno-app/src/providers/App/ConfirmAppClose/SaveRequestsModal.js
```diff
@@ -119,6 +121,13 @@ const SaveRequestsModal = ({ onClose }) => {
...
+        toast.error('Please fix validation errors before saving');
+        return;
```
sanish-bruno: **Design**: Early return blocks ALL unsaved items from being saved, not just the invalid environment.

If a user has unsaved request changes AND an environment with invalid variable names, clicking "Save All" does nothing — they lose the ability to save any of their valid work during app close. The only options are "Don't Save" (loses everything) or "Cancel" (prevents close, but the user may not know which environment has the issue).

Consider either:
1. Saving what can be saved (requests, collections, folders) and only blocking the invalid environment saves, or
2. At minimum, making the error message specific enough that the user knows which environment to fix (e.g., include the environment name).

packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
```diff
@@ -339,6 +340,10 @@ const RequestTab = ({ tab, collection, tabIndex, collectionRequestTabs, folderUi
                 window.addEventListener('dotenv-save-failed', onFailed, { once: true });
                 window.dispatchEvent(new Event('dotenv-save'));
               } else if (draft?.environmentUid && draft?.variables) {
+                if (hasInvalidVariableNames(draft.variables)) {
+                  toast.error('Please fix validation errors before saving');
```
sanish-bruno: **Suggestion**: The error message `'Please fix validation errors before saving'` is generic.

Consider including the invalid variable name(s) in the toast so the user can quickly locate and fix the issue, e.g.:
```js
const invalidNames = draft.variables
  .filter(v => v.name && v.name.trim() !== '' && !variableNameRegex.test(v.name))
  .map(v => v.name);
toast.error(`Invalid variable name(s): ${invalidNames.join(', ')}`);
```
This applies to all three call sites (both in RequestTab and in SaveRequestsModal).

packages/bruno-app/src/utils/common/regex.js
```diff
@@ -6,6 +6,14 @@ const lastCharacter = /[^.\s<>:"/\\|?*\x00-\x1F]$/; // no dot, space and `invali
 
 export const variableNameRegex = /^[\w-.]*$/;
 
+export const hasInvalidVariableNames = (variables) => {
```
sanish-bruno: **Nit**: This function operates on an array of variable domain objects (checking `.name` properties), which is a higher level of abstraction than the regex patterns and string validators in this file. Consider placing it in a more domain-specific utility (e.g., `utils/common/variables.js` or `utils/environments.js`), or at minimum adding a brief doc comment explaining the coupling to `variableNameRegex`.

packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
sanish-bruno: we may need to make it mot more straightforward, like what is actually wrong within it, either we can have a good error message or some info before the table that tells the user what is a valid variable name, may need to involve product

packages/bruno-app/src/providers/App/ConfirmAppClose/SaveRequestsModal.js
```diff
@@ -166,14 +168,27 @@ const SaveRequestsModal = ({ onClose, forceCloseTabs = false, tabUidsToClose = [
...
       }
 
-      // Save all global environment drafts
-      for (const draft of globalEnvironmentDrafts) {
-        await dispatch(saveGlobalEnvironment({ variables: draft.variables, environmentUid: draft.environmentUid }));
+      if (hasSkippedEnvs) {
+        return;
       }
 
       if (forceCloseTabs) {
```
sanish-bruno, line 194: **Bug (Major)**: Early return leaves already-saved drafts in an inconsistent state.

When `hasSkippedEnvs` is true, the function returns without calling `closeTabs`/`completeQuitFlow`/`onClose`. However, requests, collections, folders, and *valid* environments above have already been saved to disk. Their Redux draft state is never cleared because that happens in the close/quit flow.

On the next "Save All" click, `allDrafts` still includes these already-persisted items and they get re-saved. While likely idempotent, it's confusing — the user sees items listed as "unsaved" that were already written to disk.

Consider clearing drafts for items that were successfully saved before the early return, or restructuring so the quit flow completes for everything except the invalid environments.

packages/bruno-app/src/utils/common/regex.spec.js
```diff
@@ -163,3 +164,85 @@ describe('sanitizeName and validateName', () => {
     });
   });
 });
+
```
sanish-bruno, line 167: **Test gap**: Tests cover `hasInvalidVariableNames` (boolean helper) but the production code in `RequestTab` and `SaveRequestsModal` calls `getInvalidVariableNames` (returns the actual name list). The function that matters in prod is untested — e.g., there are no assertions that `getInvalidVariableNames` returns the correct names for display in toast messages.

Also, these tests belong in a `variables.spec.js` file since they import from `./variables`, not `./regex`.

packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
```diff
@@ -356,6 +357,11 @@ const RequestTab = ({ tab, collection, tabIndex, collectionRequestTabs, folderUi
                 window.addEventListener('dotenv-save-failed', onFailed, { once: true });
                 window.dispatchEvent(new Event('dotenv-save'));
               } else if (draft?.environmentUid && draft?.variables) {
+                const invalidNames = getInvalidVariableNames(draft.variables);
+                if (invalidNames.length > 0) {
+                  toast.error(`Invalid variable name(s): ${invalidNames.join(', ')}`);
```
sanish-bruno, line 362: **Suggestion**: This exact 5-line validation block (call `getInvalidVariableNames`, check length, toast error, return) is duplicated at lines ~408-413 for the global environment path. Consider extracting a small helper like `validateAndToast(draft.variables)` to avoid the duplication, especially since both blocks need to stay in sync if the validation logic or error message format changes.

packages/bruno-app/src/providers/App/ConfirmAppClose/SaveRequestsModal.js
```diff
@@ -166,14 +168,27 @@ const SaveRequestsModal = ({ onClose, forceCloseTabs = false, tabUidsToClose = [
         await dispatch(saveMultipleRequests(requestDrafts));
       }
 
-      // Save all collection environment drafts
-      for (const draft of collectionEnvironmentDrafts) {
-        await dispatch(saveEnvironment(draft.variables, draft.environmentUid, draft.collectionUid));
+      // Save environment drafts, skipping any with invalid variable names
+      const allEnvironmentDrafts = [...collectionEnvironmentDrafts, ...globalEnvironmentDrafts];
+      let hasSkippedEnvs = false;
```
sanish-bruno, line 173: **Bug (Minor)**: The combined loop dispatches `saveEnvironment` or `saveGlobalEnvironment` based on `draft.type`, but after the loop, there's no call to `clearEnvironmentsDraft` or `clearGlobalEnvironmentDraft` for the environments that *were* successfully saved. The original code also didn't clear these (relying on the quit flow), but now that valid and invalid envs are mixed in one loop with a potential early return, valid environment drafts can be persisted to disk yet remain as stale drafts in Redux.

sid-bruno (dismissed):

PR #7407, opened 2026-03-09
fix:  app crash error (Rendered fewer hooks than expected)

sanish-bruno:
[Jira ticket](https://usebruno.atlassian.net/browse/BRU-2823)
## Summary
- Fix React "Rendered fewer hooks than expected" crash in `CollectionItem` component
- Two `useCallback` hooks (`handleCopyItem`, `handlePasteItem`) were defined after an early `return null` (search text filtering), violating React's Rules of Hooks
- Moved the hooks before the early return so they are called consistently on every render

## Root Cause
When the sidebar search text didn't match a collection item, the component returned early on [line 490](https://github.com/usebruno/bruno/blob/main/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js#L490), skipping the `useCallback` hooks defined later. React requires hooks to be called in the same order on every render — the skipped hooks caused the "Rendered fewer hooks than expected" runtime error.

## Test plan
- [ ] Open Bruno, create a collection with multiple requests/folders
- [ ] Use the sidebar search to filter items — verify no crash occurs
- [ ] Clear search — verify items render correctly
- [ ] Right-click context menu copy/paste still works as expected

## Summary by CodeRabbit

* **New Features**
  * Added copy and paste functionality for collection items (folders and requests).
  * Toast notifications confirm successful copy and paste operations or display error messages.
  * Keyboard shortcuts integrated for seamless copy and paste workflows.


wwwnuk:
XD

PR #7414, opened 2026-03-09
fix: preserve hashes in dotenv values

### LINKED ISSUE #7375: .env files feature does not support values with hash

### This bug is:

- [x] annoying

### Bruno version

3.1.4

### Operating System

macOS 26.3 (25D125)

### Describe the bug

Steps to reproduce:

- Open a workspace
- Go to Global Environments
- Add a new .env file (or use any existing one)
- Add a property with any name, and a value like "ABC#DEF"
- Click Save

The table refreshes and shows "ABC" as value.

This makes passwords with a "#" inside it unusable.

The # character should be escaped or the values should be surrounded by double quotes upon save.

When editing the .env file outside Bruno and wrapping inside quotes they are shown correctly when Bruno is loaded. Only when you save from within Bruno it is broken again as the whole file is re-written without the surrounding quotes.

_No response_

Not provided

cryst-hq:
## Summary
- quote dotenv values that would otherwise be parsed as comments or trimmed by dotenv
- keep app and Electron dotenv serialization logic aligned
- add regression tests for quoted hash values in dotenv parsing and table editor serialization

Closes #7375

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment variable handling to properly quote and preserve values containing special characters (hash symbols, newlines, leading/trailing whitespace) in .env files.

* **Tests**
  * Added comprehensive test coverage for environment variable quoting and parsing behavior.

### DIFF
--- a/packages/bruno-app/src/components/Environments/DotEnvFileEditor/utils.js
+++ b/packages/bruno-app/src/components/Environments/DotEnvFileEditor/utils.js
@@ -1,11 +1,22 @@
 import { uuid } from 'utils/common';
 
+const needsQuoting = (value) => {
+  return (
+    value.includes('\n')
+    || value.includes('"')
+    || value.includes('\'')
+    || value.includes('\\')
+    || value.includes('#')
+    || /^\s|\s$/.test(value)
+  );
+};
+
 export const variablesToRaw = (variables) => {
   return variables
     .filter((v) => v.name && v.name.trim() !== '')
     .map((v) => {
       const value = v.value || '';
-      if (value.includes('\n') || value.includes('"') || value.includes('\'')) {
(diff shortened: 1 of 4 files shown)


chirag-bruno:
Hi @cryst-hq, thank you again for taking the time to contribute. We really appreciate your continued effort!

In this case as well, the issue had already been addressed in an earlier PR ([https://github.com/usebruno/bruno/pull/7380](https://github.com/usebruno/bruno/pull/7380)), which was raised and merged before your submission.

Going forward, it would be great if you could check in with us before starting work on an issue. That way, we can assign it to you and avoid situations where multiple people end up working on the same thing.

Thanks again for your contribution and understanding! 🙏

PR #7441, opened 2026-03-11
fix: support response query filtering in safe mode

### LINKED ISSUE #7387: Response query filtering does not work

### This bug is:

- [x] annoying

### Bruno version

3.0.0

### Operating System

macOS

### Describe the bug

I tried to use the response query feature and for simple cases it works like a charm, but once I try to use filtering it doesn't work and produces the following error:

```js
{
  name: "Error",
  message: "missing function for resources", stack: " at <eval> (eval.js) "
}
```

This is my query: `res('resources[?].id', r => r.id === 'test')`. Other queries like `res('resources.id')` work.
I also noticed that bruno replaces the `=>` with this:

Image

Therefore, I tried to use a function instead of an arrow function. Still, I get the same error.

_No response_

I believe the example from the docs won't work, so I suggest trying that.

abhishek-bruno:
### Description

- Forward filter/predicate arguments from the QuickJS sandbox `res()` shim to the host query engine, fixing `res('path[?].prop', filterFn)` in safe mode
- Add `toHostQueryArg` helper that marshals QuickJS function handles into native callbacks across the VM boundary, while passing object predicates as-is
- Remove unnecessary `try/finally` argument disposal that was inconsistent with other shims in the codebase

Closes #7387
Built on top of #7417 by @cryst-hq

[JIRA](https://usebruno.atlassian.net/browse/BRU-2839)

## How it works
The `res()` function in the QuickJS sandbox previously only forwarded the expression string to the host, silently dropping any filter/predicate arguments. This caused `[?]` array filtering to fail with "missing function for resources".

The fix wraps each QuickJS function argument into a native callback that marshals data across the VM boundary (host → sandbox → host), while keeping the filter function safely inside the sandbox. Object predicates (e.g. `{ id: 'test' }`) are simply dumped to plain JS objects.

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Extended the response function to accept and process query arguments, enabling callback-based filtering with proper error handling.

* **Tests**
  * Added comprehensive test suite validating query argument handling, callback execution, and object predicate support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,9 +55,6 @@ bruno.iml
 .agent
 skills-lock.json
 
-.agents
-.agent
-skills-lock.json
 # Playwright
 /blob-report/
 
--- a/packages/bruno-js/src/sandbox/quickjs/shims/bruno-response.spec.js
+++ b/packages/bruno-js/src/sandbox/quickjs/shims/bruno-response.spec.js
@@ -14,11 +14,25 @@ describe('bruno response shim', () => {
   });
 
   afterEach(() => {
-    vm = null;
+    if (vm) {
+      try {
+        vm.dispose();
+      } catch (err) {
+        console.error('Error disposing vm', err);
+      }
+      vm = null;
+    }
   });
 
   afterAll(() => {
-    module = null;
+    if (module) {
+      try {
+        module.dispose();
+      } catch (err) {
+        console.error('Error disposing module', err);
+      }
+      module = null;
+    }
   });
 
   it('forwards response query filter callbacks in safe mode', () => {


.gitignore
```diff
@@ -55,6 +55,9 @@ bruno.iml
 .agent
 skills-lock.json
 
+.agents
+.agent
```
sanish-bruno: **Nit**: These three lines (`.agents`, `.agent`, `skills-lock.json`) are duplicates — they already exist on lines 54-56 of the same file. This hunk should be removed.

packages/bruno-js/src/sandbox/quickjs/shims/bruno-response.spec.js
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  beforeEach(() => {
+    vm = module.newContext();
+  });
+
+  afterEach(() => {
+    vm = null;
+  });
+
```
sanish-bruno, line 26: **Nit**: `vm` is created via `module.newContext()` in `beforeEach` but never disposed. This leaks WASM memory across tests. Add `vm.dispose()` before nulling:

Suggested fix:
```js
  afterEach(() => {
    vm.dispose();
    vm = null;
  });
```

Issue #7353, opened 2026-03-03
SSE: previous connections are not closed when re-sending a request


Fimeo:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

v3.1.4

### Operating System

macOS 26.2 (25C56)

### Describe the bug

 When using SSE (text/event-stream), clicking "Send" again while a stream is active does not close the previous connection. Each send opens a new SSE connection, and the old ones keep running in the background.

  Steps to reproduce:
  1. Create a GET request to an SSE endpoint
  2. Click Send — the stream starts
  3. Click Send again — a new stream starts but the previous one is still active server-side
  4. Click Stop — only the latest connection is closed
  5. Events from the orphaned connections are still visible in the response panel

  Expected behavior:
  - Clicking Send while a stream is active should close the previous connection before opening a new one
  - A visual indicator showing active SSE connections would be helpful
  - Stop should close all active connections for that request

### Screenshots/Live demo link

https://github.com/user-attachments/assets/f3df207b-9358-4570-9da1-6376dfd7d79a

chirag-bruno:
Hi @Fimeo, thanks for bringing this issue to our attention, and apologies for the delay in acknowledging it. I’ve opened a PR to fix it, and the change should be available in the upcoming releases.

PR #7474, opened 2026-03-13
fix: close previous SSE connection before sending new request

chirag-bruno:
### Description

When resending an SSE (Server-Sent Events) request using Cmd+Enter, the previous connection was not being closed, causing connection leaks.

Changes:
- Add SSE cancellation logic to sendRequest action - checks for running stream and cancels it before sending new request
- Add return to cancelRequest action to make it properly chainable
- Simplify RequestTabPanel by removing duplicate cancel logic (now handled centrally in sendRequest)
- Add SSE endpoints to test server for e2e testing
- Add Playwright e2e test to verify SSE connection cancellation

Fixes: #7353
JIRA: https://usebruno.atlassian.net/browse/BRU-2860

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added local Server-Sent Events (SSE) test service with stream, connections, and reset endpoints.
  * Updated run control to avoid dispatching duplicate sends for requests already sending or queued.

* **Bug Fixes**
  * Await cancellation of active SSE streams before starting a new run to prevent duplicate connections.

* **Tests**
  * Added end-to-end tests, fixtures, and test preferences validating SSE connection cancellation.

### DIFF
--- a/tests/sse/sse-cancellation.spec.ts
+++ b/tests/sse/sse-cancellation.spec.ts
@@ -19,8 +19,8 @@ test.describe('SSE Connection Cancellation', () => {
     await expect.poll(async () => {
       const response = await page.request.get('http://localhost:8081/api/sse/connections');
       const data = await response.json();
-      return data.activeConnections;
-    }, { timeout: 5000 }).toBe(1);
+      return data.connectionIds;
+    }, { timeout: 5000 }).toStrictEqual([1]);
 
     // Resend the request (this should cancel the old connection and start a new one)
     const resendShortcut = process.platform === 'darwin' ? 'Meta+Enter' : 'Control+Enter';
@@ -30,7 +30,7 @@ test.describe('SSE Connection Cancellation', () => {
     await expect.poll(async () => {
       const response = await page.request.get('http://localhost:8081/api/sse/connections');
       const data = await response.json();
-      return data.activeConnections;
-    }, { timeout: 5000 }).toBe(1);
+      return data.connectionIds;
+    }, { timeout: 5000 }).toStrictEqual([2]);
   });
 });


tests/sse/sse-cancellation.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    await page.waitForTimeout(1000);
+
+    const initialResponse = await page.request.get('http://localhost:8081/api/sse/connections');
+    const initialData = await initialResponse.json();
+    expect(initialData.activeConnections).toBe(1);
+
+    // Press Cmd+Enter to resend the request (this should cancel the old connection and start a new one)
+    await page.keyboard.press('Meta+Enter');
+    await page.waitForTimeout(1500);
+
```
sanish-bruno: **Bug (Minor)**: `Meta+Enter` is macOS-only. This will fail on Linux/Windows CI.

Other tests in this repo use the platform-aware pattern:

```ts
const resendShortcut = process.platform === 'darwin' ? 'Meta+Enter' : 'Control+Enter';
await page.keyboard.press(resendShortcut);
```

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -529,6 +529,14 @@ export const sendRequest = (item, collectionUid) => (dispatch, getState) => {
       return reject(new Error('Collection not found'));
     }
 
+    if (item.response?.stream?.running && item.cancelTokenUid) {
+      try {
+        await dispatch(cancelRequest(item.cancelTokenUid, item, collection));
```
sanish-bruno: **Bug (Minor)**: This `try/catch` is dead code. `cancelRequest` already has a `.catch((err) => console.log(err))` at the end of its promise chain, which swallows any rejection. So `dispatch(cancelRequest(...))` always resolves (never rejects), and this `catch` block will never execute.

Either:
1. Remove the `.catch` from `cancelRequest` so errors propagate to callers (may affect `QueryUrl` and `Overlay` callers), or
2. Remove the `try/catch` here since it's misleading.

Option 2 is safer since other callers don't expect `cancelRequest` to reject:

```js
if (item.response?.stream?.running && item.cancelTokenUid) {
  await dispatch(cancelRequest(item.cancelTokenUid, item, collection));
}
```

tests/sse/sse-cancellation.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    // Verify that only 1 connection is active (old one was closed, new one opened)
```
sanish-bruno: **Nit**: `waitForTimeout` is a known source of flaky tests. Consider using `expect.poll()` to wait for the connection count, which is both faster and more reliable:

```ts
await expect.poll(async () => {
  const resp = await page.request.get('http://localhost:8081/api/sse/connections');
  const data = await resp.json();
  return data.activeConnections;
}, { timeout: 5000 }).toBe(1);
```

This applies to both `waitForTimeout(1000)` (line 19) and `waitForTimeout(1500)` (line 28).

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -530,6 +530,10 @@ export const sendRequest = (item, collectionUid) => (dispatch, getState) => {
       return reject(new Error('Collection not found'));
     }
 
+    if (item.response?.stream?.running && item.cancelTokenUid) {
```
sanish-bruno, line 533: **Bug (Minor)**: `cancelRequest` swallows all errors via `.catch((err) => console.log(err))` on line 650, so the `await` here always resolves successfully — even if the cancellation actually failed. This means if `cancelNetworkRequest` throws (e.g. the cancel token is stale or the IPC call fails), the code silently proceeds to send a new request while the old SSE stream is still running — exactly the connection leak this PR aims to fix.

Consider either:
1. Letting `cancelRequest` propagate the error and handling it here, or
2. At minimum, logging a warning so silent failures are visible:

Suggested fix:
```js
try {
  await dispatch(cancelRequest(item.cancelTokenUid, item, collection));
} catch (err) {
  console.warn('Failed to cancel previous SSE stream, proceeding with new request:', err);
}
```

tests/sse/sse-cancellation.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Resend the request (this should cancel the old connection and start a new one)
+    const resendShortcut = process.platform === 'darwin' ? 'Meta+Enter' : 'Control+Enter';
+    await page.keyboard.press(resendShortcut);
+
+    // Poll until the old connection is closed and a new one is established
+    await expect.poll(async () => {
+      const response = await page.request.get('http://localhost:8081/api/sse/connections');
+      const data = await response.json();
+      return data.activeConnections;
+    }, { timeout: 5000 }).toBe(1);
```
sanish-bruno: **Suggestion**: The final assertion `toBe(1)` verifies there's exactly one connection, but doesn't confirm it's a *new* connection (as opposed to the original one surviving and the new one failing to open). Consider also checking that the `connectionId` changed:

```ts
// Before resend, capture the current connection ID
const beforeResend = await page.request.get('http://localhost:8081/api/sse/connections');
const beforeData = await beforeResend.json();
const oldConnectionId = beforeData.connectionIds[0];

// ... resend ...

await expect.poll(async () => {
  const response = await page.request.get('http://localhost:8081/api/sse/connections');
  const data = await response.json();
  return data.activeConnections === 1 && data.connectionIds[0] !== oldConnectionId;
}, { timeout: 5000 }).toBe(true);
```

This makes the test more resilient against false positives where the cancel failed but the new request also failed.

packages/bruno-tests/src/sse/index.js
```diff
@@ -0,0 +1,58 @@
+const express = require('express');
+const router = express.Router();
+
+// Track active SSE connections
+let activeConnections = new Set();
+let connectionIdCounter = 0;
+
```
sanish-bruno, line 7: **Nit**: Module-level mutable state (`activeConnections`, `connectionIdCounter`) is fine for a test server, but note that these won't reset between test runs if the server process is shared across test files. The `/reset` endpoint mitigates this, but if a test crashes before calling reset, subsequent tests could see stale state. Consider whether the `beforeEach` reset in the test is sufficient for your CI setup.

sid-bruno:
LGTM, waiting on tests

PR #7499, opened 2026-03-16
fix(import): resolve EEXIST error when importing OpenAPI collections with paths folder arrangement

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2872)

### Problem
Importing an OpenAPI spec using the Paths folder arrangement fails with:

Error EEXIST: file already exists, mkdir '.../{customerID}'
The collection directory is partially created on disk but never opens in Bruno. This affects both macOS and Windows.

### Root cause
The import handler calls fs.mkdirSync(folderPath) without { recursive: true }. When a spec contains paths with the same path parameter name in different casing (e.g. {customerId} and {customerID}), the path-based grouper creates two distinct folder items for them. On case-insensitive filesystems, both map to the same directory — the second mkdirSync call throws EEXIST, aborting the import and preventing main:collection-opened from firing.

### Fix
Pass { recursive: true } to mkdirSync in parseCollectionItems inside the renderer:import-collection handler. This silently succeeds if the directory already exists instead of throwing, allowing the import to complete. Both sets of requests are written into the same folder, which is the correct behavior since they represent the same URL pattern.

#### Contribution Checklist:

## Release Notes

* **Bug Fixes**
  * Fixed an issue where creating nested folder structures would fail if parent directories didn't exist. The system now automatically creates all required parent directories when adding new folders.

### DIFF
--- a/packages/bruno-electron/src/ipc/collection.js
+++ b/packages/bruno-electron/src/ipc/collection.js
@@ -1186,7 +1186,7 @@ const registerRendererEventHandlers = (mainWindow, watcher) => {
             if (item.type === 'folder') {
               let sanitizedFolderName = sanitizeName(item?.filename || item?.name);
               const folderPath = path.join(currentPath, sanitizedFolderName);
-              fs.mkdirSync(folderPath);
+              fs.mkdirSync(folderPath, { recursive: true });
 
               if (item?.root?.meta?.name) {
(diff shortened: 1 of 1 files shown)


bijin-bruno:
@abhishek-bruno Please update playwright tests to cover this scenario

PR #7537, opened 2026-03-20
fix: enhance error handling and context retrieval for script errors

sanish-bruno:
- Updated the error formatter to utilize in-memory script content for error context, improving accuracy when users have unsaved changes.
- Introduced a new utility function, `getSourceContextFromContent`, to extract context lines from in-memory scripts.
- Enhanced tests to verify that draft script errors display the correct code context, ensuring users see the most relevant information during debugging.
- Added new Playwright tests to validate error handling in draft states across pre-request, post-response, and test scripts.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Error views now prefer unsaved (draft) script content when available, yielding correct error lines, context, and preserving stack-trace locations for draft-only code.
  * Improved context extraction for in-memory script content so error cards show the draft script lines.

* **Tests**
  * Added end-to-end tests validating draft script error UI for pre-request, post-response, and tests.
  * Expanded unit/integration tests for in-memory vs on-disk context and stack-trace handling; added fixtures.

* **Chores**
  * UI now shows a draft indicator on modified request tabs.

### DIFF
--- a/packages/bruno-js/src/utils/error-formatter.js
+++ b/packages/bruno-js/src/utils/error-formatter.js
@@ -274,6 +274,8 @@ const resolveSegmentError = (parsed, metadata, scriptType, cache) => {
             filePath: segment.filePath,
             displayPath: segment.displayPath,
             scriptContent: segment.scriptContent,
+            // segment.startLine points to the IIFE wrapper line (`await (async () => {`),
+            // so subtracting it yields a 1-based index into the user's script content.
             lineInScript: scriptRelativeLine - segment.startLine
           };
         }
@@ -285,6 +287,8 @@ const resolveSegmentError = (parsed, metadata, scriptType, cache) => {
         filePath: segment.filePath,
         displayPath: segment.displayPath,
         scriptContent: segment.scriptContent || null,
+        // segment.startLine points to the IIFE wrapper line (`await (async () => {`),
+        // so subtracting it yields a 1-based index into the user's script content.
         lineInScript: scriptRelativeLine - segment.startLine
       };
     }
@@ -595,6 +599,8 @@ const resolveErrorContext = ({ adjustedLine, scriptRelativeLine, metadata, segme
     // would produce misleading results — flag it as draft-only.
     const blockStartLine = findBlockStart(filePath, scriptType, cache);
     const draftOnlyBlock = !blockStartLine && isAllowedSourceFile(filePath);
+    // requestStartLine points to the IIFE wrapper line (`await (async () => {`),
+    // so subtracting it yields a 1-based index into the user's script content.
     const lineInScript = scriptRelativeLine - metadata.requestStartLine;
     const context = getSourceContextFromContent(metadata.requestScriptContent, lineInScript, 3);
     if (context) return { context, fromMemory: true, draftOnlyBlock };


packages/bruno-js/src/utils/error-formatter.js
```diff
@@ -353,6 +368,28 @@ const getSourceContext = (filePath, errorLine, contextLines = DEFAULT_CONTEXT_LI
...
+  for (let i = startLine; i <= endLine; i++) {
+    contextLinesArray.push({
+      lineNumber: i,
+      content: lines[i - 1],
+      isError: i === errorLine
+    });
+  }
+
+  return { lines: contextLinesArray, startLine, errorLine };
+};
```
sanish-bruno, line 387: **Suggestion**: `getSourceContextFromContent` is nearly identical to `getSourceContext` — the only difference is the content source (parameter vs `readFile`). Consider consolidating by making `getSourceContext` accept an optional `content` override:

```js
const getSourceContext = (filePath, errorLine, contextLines = DEFAULT_CONTEXT_LINES, cache = null, content = null) => {
  const fileContent = content || readFile(filePath, cache);
  if (!fileContent) return null;
  // ... rest unchanged
};
```

Then callers with in-memory content would pass it directly:
```js
context = getSourceContext(null, lineInScript, 3, null, metadata.requestScriptContent);
```

This eliminates the duplicated function and keeps one code path for context extraction.

packages/bruno-js/src/utils/error-formatter.js
```diff
@@ -559,24 +596,49 @@ const formatErrorWithContextV2 = (error, scriptType, scriptMetadata, collectionP
...
 
     const { filePath } = parsed;
+    const wrapperOffset = parsed.isQuickJS ? QUICKJS_SCRIPT_WRAPPER_OFFSET : NODEVM_SCRIPT_WRAPPER_OFFSET;
+    const scriptRelativeLine = parsed.line - wrapperOffset;
     const adjustedLine = adjustLineNumber(filePath, parsed.line, parsed.isQuickJS, scriptType, cache, metadata);
 
     let sourceFile = filePath;
     let sourceLine = adjustedLine;
+    let context = null;
+    let blockRelativeAlready = false;
```
sanish-bruno: **Nit**: The name `blockRelativeAlready` is not immediately clear without reading the surrounding comments. Consider `contextFromMemory` or `isInMemoryContext` — these directly communicate *why* the block-relative computation can be skipped (because the context came from in-memory content where line numbers are already 1-indexed relative to the script block).

tests/script-errors/draft-script-errors.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+
+for (const mode of ['safe', 'developer'] as const) {
+  test.describe.serial(`Draft Script Error Context [${mode} mode]`, () => {
+    let scriptErrorLocators: ReturnType<typeof buildScriptErrorLocators>;
+    let commonLocators: ReturnType<typeof buildCommonLocators>;
+
+    test.beforeAll(async ({ pageWithUserData: page }) => {
+      scriptErrorLocators = buildScriptErrorLocators(page);
+      commonLocators = buildCommonLocators(page);
```
sanish-bruno, line 20: **Nit**: `commonLocators` is assigned here but never used in any test. Only `scriptErrorLocators` is used from `beforeAll`. Either remove it or use it in the test steps (e.g., `commonLocators.request.sendButton()` instead of re-building locators in the helpers).

packages/bruno-electron/src/utils/collection.js
```diff
@@ -253,6 +253,9 @@ const mergeScripts = (collection, request, requestTreePath, scriptFlow) => {
     displayPath: config.collectionFile
   };
 
+  const withContent = (source, script) =>
+    script?.trim() ? { ...source, scriptContent: script } : source;
```
sanish-bruno, line 257: **Bug (Minor)**: `withContent` uses `script?.trim()` as a truthiness check, which means a script containing only whitespace (e.g., `"  \n  "`) won't get `scriptContent` attached. This is technically consistent with the existing guard (`if (preReqScript && preReqScript.trim() !== '')`) that prevents such scripts from being pushed to `combinedPreReqScript`. However, for the **collection-level** usage at line 303 (`withContent(collectionSource, collectionPreReqScript)`), there's no such pre-guard — `collectionPreReqScript` could be a non-empty whitespace-only string. In that case, `collectionPreReqSource` would lack `scriptContent` but `collectionPreReqScript` would still be passed to `wrapScriptInClosure` (which returns `''` for whitespace-only scripts, so the segment is effectively skipped). This is safe in practice but the asymmetry is worth being aware of.

sanish-bruno: it is "safe in practice" since wrapScriptInClosure returns '' for whitespace-only scripts, so the segment gets effectively dropped and never produces an error to display context for.

packages/bruno-js/src/utils/error-formatter.js
```diff
@@ -559,35 +581,95 @@ const formatErrorWithContextV2 = (error, scriptType, scriptMetadata, collectionP
...
+
     const resolvedDisplayPath = posixifyPath(
       collectionPath ? path.relative(collectionPath, sourceFile) : sourceFile
     );
 
-    const context = getSourceContext(sourceFile, sourceLine, 3, cache);
+    // Fall back to disk read if in-memory content not available
+    if (!context && sourceLine !== null) {
+      context = getSourceContext(sourceFile, sourceLine, 3, cache);
+    }
```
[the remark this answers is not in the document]
sanish-bruno: The concern is about the disk fallback at line 611 running when draftOnlyBlock would
   be true. Let's trace the two scenarios where draftOnlyBlock could be true:         
                                                                                      
  Case 1: Request-level draft-only block (lines 591-601)                              
  - draftOnlyBlock = true when findBlockStart returns null for a .bru/.yml file       
  - If getSourceContextFromContent returns a context → we return early at line 600
  (never reach disk fallback)
  - If getSourceContextFromContent returns null → we fall through. But in this case,
  sourceLine equals adjustedLine which equals scriptRelativeLine (a meaningless
  offset, not a real file line). The disk read at line 612 would either read the wrong
   line or return null. But it cannot return stale script content — it reads the .bru
  file which contains the entire file (meta block, headers, etc.), not just the
  script. The formatErrorWithContextV2 caller would then fail at the block-relative
  filtering step (lines 680-707 in the main function) because blockEndLine would be
  null for a missing block, causing it to return null at line 696.

  Case 2: Segment draft-only block (lines 604-608)
  - segmentResult.line is null → sourceLine is null (set at line 644)
  - The disk fallback guard if (sourceLine !== null) at line 611 already prevents the
  disk read

  So in both cases, the disk fallback either can't produce stale script context (Case
  1 — the downstream filtering catches it) or is already guarded (Case 2). No fix
  needed.

PR #7549, opened 2026-03-23
fix: allow __Host- prefixed cookies to be stored via script API

### LINKED ISSUE #7452: __Host- Cookies are not settable from a script

### This bug is:

- [ ] annoying

### Bruno version

3.1.4

### Operating System

macOS 15.6.1

### Describe the bug

Bruno ignores cookie set manually when it is prefixed by ["__Host"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#cookie_prefixes).

See the example below

meta {
  name: Fetch bruno website and save __Host cookie
  type: http
  seq: 10
}

get {
  url: https://usebruno.com
  body: none
  auth: inherit
}

script:post-response {
  // This doesn't work
  bru.cookies.jar().setCookie('https://usebruno.com', {
    key: '__Host-SESSION',
    value: '12345',
    path: '/',
    secure: true,
    httpOnly: true
  })
  
  // This does
  bru.cookies.jar().setCookie('https://usebruno.com', {
    key: 'Host-SESSION',
    value: '12345',
    path: '/',
    secure: true,
    httpOnly: true
  })
}

settings {
  encodeUrl: true
  timeout: 0
}

Image

Image

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2910)

- Fixes #7452 
- Bruno always injected `domain: new URL(url).hostname` on cookie objects before passing them to tough-cookie. This caused `hostOnly` to be `false`, which fails tough-cookie's `__Host-` prefix validation (`isHostPrefixConditionMet` requires `hostOnly === true`). The cookie was silently discarded because `ignoreError: true` is used.
- Fix: omit the explicit `domain` for `__Host-` prefixed cookies so tough-cookie derives it from the URL, correctly setting `hostOnly = true`

## Changes
- `packages/bruno-requests/src/cookies/index.ts` — skip setting `domain` when cookie key starts with `__Host-` (3 locations: `setCookie` name/value, `setCookie` object, `setCookies` array)
- `packages/bruno-requests/src/cookies/index.test.ts` — new test suite covering `__Host-` cookie storage, regular cookie regression, and spec-correct rejection cases

## Test plan
- [x] `__Host-SESSION` via object API → stored successfully
- [x] `__Host-SESSION` via array API (`setCookies`) → stored successfully
- [x] `__Host-` via name/value API without `secure`/`path` → correctly dropped per spec
- [x] `__Host-` with user-supplied `domain` in object → correctly dropped per spec
- [x] Regular cookie → still works (regression check)

#### Contribution Checklist:

## Summary by CodeRabbit

* **Tests**
  * Added tests validating secure, host-prefixed cookie handling across different set APIs and regression checks for standard cookies.

* **Bug Fixes**
  * Tightened cookie handling so host-prefixed cookies require proper secure/path attributes and are rejected when domain or required attributes are invalid, improving cookie storage security.

### DIFF
--- a/packages/bruno-requests/src/cookies/index.test.ts
+++ b/packages/bruno-requests/src/cookies/index.test.ts
@@ -1,6 +1,6 @@
 import cookiesModule from './index';
 
-const URL = 'https://example.com';
+const TEST_URL = 'https://example.com';
 
 describe('cookieJarWrapper.setCookie — __Host- prefix handling', () => {
   let jar: ReturnType<typeof cookiesModule.jar>;
@@ -17,14 +17,14 @@ describe('cookieJarWrapper.setCookie — __Host- prefix handling', () => {
     // correctly rejects it per __Host- spec requirements. This verifies the
     // cookie is dropped for the right reason (missing secure/path), not because
     // of the old domain/hostOnly mismatch bug.
-    jar.setCookie(URL, '__Host-TOKEN', 'abc123');
-    const cookies = await jar.getCookies(URL);
+    await jar.setCookie(TEST_URL, '__Host-TOKEN', 'abc123');
+    const cookies = await jar.getCookies(TEST_URL);
     expect(cookies!.find((c: any) => c.key === '__Host-TOKEN')).toBeUndefined();
   });
 
   // T2: object with __Host- prefix (the exact scenario from the bug report)
   it('stores __Host- cookie via cookie object', async () => {
-    jar.setCookie(URL, {
+    await jar.setCookie(TEST_URL, {
       key: '__Host-SESSION',
       value: '12345',
       path: '/',
@@ -32,7 +32,7 @@ describe('cookieJarWrapper.setCookie — __Host- prefix handling', () => {
       httpOnly: true
     });
 
-    const cookies = await jar.getCookies(URL);
+    const cookies = await jar.getCookies(TEST_URL);
     const found = cookies!.find((c: any) => c.key === '__Host-SESSION');
     expect(found).toBeDefined();
     expect(found!.value).toBe('12345');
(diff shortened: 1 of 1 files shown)


packages/bruno-requests/src/cookies/index.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    jar = cookiesModule.jar();
+  });
+
+  // T1: name/value pair with __Host- prefix — dropped because secure/path not set
+  it('drops __Host- cookie via name/value pair (missing secure and path)', async () => {
+    // The name/value API doesn't set secure:true or path:"/", so tough-cookie
+    // correctly rejects it per __Host- spec requirements. This verifies the
+    // cookie is dropped for the right reason (missing secure/path), not because
+    // of the old domain/hostOnly mismatch bug.
+    jar.setCookie(URL, '__Host-TOKEN', 'abc123');
```
sanish-bruno: **Bug (Minor)**: `setCookie` (without a callback) returns a `Promise`, but it's not `await`ed here or in any of the other test cases (T2 line 33, T3 line 50, T4 line 66, T5 line 73).

This works today because `executeSetCookie()` calls `setCookieSync` under the hood (synchronous), so the cookie is already in the jar before the `await getCookies()` flushes. But it's fragile:
1. Any unhandled rejection (e.g. if `new URL(url)` throws) is silently swallowed
2. If the implementation ever becomes truly async, these tests would pass vacuously

Suggested fix:
```ts
await jar.setCookie(URL, '__Host-TOKEN', 'abc123');
```
Apply the same `await` to all `setCookie` / `setCookies` calls in the other tests.

packages/bruno-requests/src/cookies/index.test.ts
```diff
@@ -0,0 +1,86 @@
+import cookiesModule from './index';
+
+const URL = 'https://example.com';
```
sanish-bruno: **Nit**: `const URL` shadows the global `URL` constructor. This doesn't cause a bug right now since the source code receives `url` as a parameter, but it can be confusing and would break any future test code in this file that needs `new URL(...)`.

Suggested fix:
```ts
const TEST_URL = 'https://example.com';
```

PR #7575, opened 2026-03-25
feat: add gRPC proxy support

lohit-bruno:
[jira](https://usebruno.atlassian.net/browse/BRU-2665)

#### Contribution Checklist:

## Release Notes

* **New Features**
  * Added proxy support for gRPC requests with HTTP/SOCKS proxy configuration
  * Added ability to cancel gRPC streaming connections with dedicated UI controls
  * Enhanced response timeline to display active proxy configuration details

* **Tests**
  * Added comprehensive test coverage for proxy configuration and gRPC cancellation flows

### DIFF
--- a/packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
+++ b/packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
@@ -481,3 +481,4 @@ if (app && typeof app.on === 'function') {
 }
 
 module.exports = registerGrpcEventHandlers;
+module.exports.resolveGrpcProxyConfig = resolveGrpcProxyConfig;
--- a/packages/bruno-requests/src/grpc/grpc-client.js
+++ b/packages/bruno-requests/src/grpc/grpc-client.js
@@ -385,6 +385,12 @@ class GrpcClient {
       return { targetHost: originalHost, proxyChannelOptions: baseOptions };
     }
 
+    // We replicate what @grpc/grpc-js's internal mapProxyName() does:
+    // redirect the channel target to the proxy and set http_connect_target/creds
+    // so that getProxiedConnection() performs the HTTP CONNECT handshake.
+    // These channel options are marked "internal" in channel-options.ts, but
+    // there is no public programmatic proxy API — the only alternative is
+    // setting process-wide env vars (http_proxy), which is racy.
     const proxyUrl = new URL(proxyConfig.proxyUrl);
     const proxyChannelOptions = {
       ...baseOptions,
@@ -588,8 +594,6 @@ class GrpcClient {
     includeDirs = [],
     proxyConfig
   }) {
-    this.#cancelAndCloseConnection(request.uid);
-
     const credentials = this.#getChannelCredentials({
       url: request.url,
       rootCertificate,
--- a/packages/bruno-electron/src/ipc/network/grpc-event-handlers.spec.js
+++ b/packages/bruno-electron/src/ipc/network/grpc-event-handlers.spec.js
@@ -1,0 +1,146 @@
+jest.mock('./interpolate-string', () => ({
+  interpolateString: (str) => str
+}));
+
+const { resolveGrpcProxyConfig } = require('./grpc-event-handlers');
+
+const emptyInterpolationOptions = {};
+
+describe('resolveGrpcProxyConfig', () => {
+  describe('proxyMode "off"', () => {
+    it('should return null proxyUrl', () => {
+      expect(resolveGrpcProxyConfig('off', {}, 'grpc://localhost:50051', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+  });
+
+  describe('proxyMode "on"', () => {
+    it('should return proxy URL without auth', () => {
+      const proxyConfig = {
+        protocol: 'http',
+        hostname: 'proxy.example.com',
+        port: '8080',
+        auth: { disabled: true }
+      };
+      expect(resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: 'http://proxy.example.com:8080' });
+    });
+
+    it('should return proxy URL with auth when auth is enabled', () => {
+      const proxyConfig = {
+        protocol: 'http',
+        hostname: 'proxy.example.com',
+        port: '8080',
+        auth: { disabled: false, username: 'user', password: 'pass' }
+      };
+      expect(resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: 'http://user:pass@proxy.example.com:8080' });
+    });
+
+    it('should URL-encode special characters in credentials', () => {
+      const proxyConfig = {
+        protocol: 'http',
+        hostname: 'proxy.example.com',
+        port: '8080',
+        auth: { disabled: false, username: 'user@domain', password: 'p@ss:word' }
+      };
+      const result = resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions);
+      expect(result.proxyUrl).toBe('http://user%40domain:p%40ss%3Aword@proxy.example.com:8080');
+    });
+
+    it('should reject SOCKS proxy protocols', () => {
+      const proxyConfig = {
+        protocol: 'socks5',
+        hostname: 'proxy.example.com',
+        port: '1080'
+      };
+      expect(resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should reject HTTPS proxy protocol', () => {
+      const proxyConfig = {
+        protocol: 'https',
+        hostname: 'proxy.example.com',
+        port: '8080'
+      };
+      expect(resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should return null when request URL is in bypassProxy list', () => {
+      const proxyConfig = {
+        protocol: 'http',
+        hostname: 'proxy.example.com',
+        port: '8080',
+        bypassProxy: 'localhost,api.example.com'
+      };
+      expect(resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should omit port when not provided', () => {
+      const proxyConfig = {
+        protocol: 'http',
+        hostname: 'proxy.example.com',
+        auth: { disabled: true }
+      };
+      expect(resolveGrpcProxyConfig('on', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: 'http://proxy.example.com' });
+    });
+  });
+
+  describe('proxyMode "system"', () => {
+    it('should use https_proxy when available', () => {
+      const proxyConfig = {
+        https_proxy: 'http://system-proxy.example.com:3128',
+        http_proxy: 'http://fallback-proxy.example.com:3128'
+      };
+      expect(resolveGrpcProxyConfig('system', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: 'http://system-proxy.example.com:3128' });
+    });
+
+    it('should fall back to http_proxy when https_proxy is not set', () => {
+      const proxyConfig = {
+        http_proxy: 'http://fallback-proxy.example.com:3128'
+      };
+      expect(resolveGrpcProxyConfig('system', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: 'http://fallback-proxy.example.com:3128' });
+    });
+
+    it('should return null when no system proxy is configured', () => {
+      expect(resolveGrpcProxyConfig('system', {}, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should reject non-HTTP system proxy protocols', () => {
+      const proxyConfig = {
+        https_proxy: 'socks5://system-proxy.example.com:1080'
+      };
+      expect(resolveGrpcProxyConfig('system', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should return null when request URL matches no_proxy', () => {
+      const proxyConfig = {
+        https_proxy: 'http://system-proxy.example.com:3128',
+        no_proxy: 'api.example.com'
+      };
+      expect(resolveGrpcProxyConfig('system', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should return null for invalid system proxy URL', () => {
+      const proxyConfig = {
+        https_proxy: 'not-a-valid-url'
+      };
+      expect(resolveGrpcProxyConfig('system', proxyConfig, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+
+    it('should return null when proxyConfig is null', () => {
+      expect(resolveGrpcProxyConfig('system', null, 'grpc://api.example.com:443', emptyInterpolationOptions))
+        .toEqual({ proxyUrl: null });
+    });
+  });
+});


packages/bruno-app/src/components/ResponsePane/Timeline/GrpcTimelineItem/index.js
```diff
@@ -74,6 +75,28 @@ const GrpcTimelineItem = ({ timestamp, request, response, eventType, eventData,
...
+            {effectiveRequest.proxy && effectiveRequest.proxy.mode !== 'off' && (
+              <div>
+                <div className="content-request-label mb-1">
+                  <IconArrowsRightLeft size={14} strokeWidth={1.5} className="inline-block mr-1" />
+                  Proxy
+                </div>
+                <div className="content-box">
+                  {effectiveRequest.proxy.url ? (
+                    <div>
+                      Using {effectiveRequest.proxy.mode === 'system' ? 'system ' : ''}proxy: {effectiveRequest.proxy.url}
```
sanish-bruno, line 91: **Security**: Proxy URL may contain credentials in the timeline display.

When `proxyMode` is `'on'` and auth is enabled, `resolveGrpcProxyConfig` constructs URLs like `http://user:password@proxy:8080`. This raw URL is passed through `requestSent.proxy.url` → stored in Redux → rendered here.

Users will see proxy credentials in the gRPC timeline panel. This is especially concerning since the timeline is a UI element that may be visible in screenshots or screen shares.

Suggested fix:
```jsx
const sanitizeProxyUrl = (url) => {
  try {
    const parsed = new URL(url);
    if (parsed.username) {
      parsed.username = '***';
      parsed.password = '***';
    }
    return parsed.toString();
  } catch {
    return url;
  }
};

// Then in JSX:
Using {effectiveRequest.proxy.mode === 'system' ? 'system ' : ''}proxy: {sanitizeProxyUrl(effectiveRequest.proxy.url)}
```

packages/bruno-requests/src/grpc/grpc-client.js
```diff
@@ -504,8 +585,11 @@ class GrpcClient {
     pfx,
     verifyOptions,
     channelOptions = {},
-    includeDirs = []
+    includeDirs = [],
+    proxyConfig
   }) {
+    this.#cancelAndCloseConnection(request.uid);
```
sanish-bruno: **Suggestion**: This `#cancelAndCloseConnection` call is redundant with the one inside `#addConnection`.

`startConnection` calls `#cancelAndCloseConnection(request.uid)` here at the top. Later, when `#handleConnection` → `#handle*Response` → `#addConnection` is called, `#addConnection` also calls `#cancelAndCloseConnection(requestId)` as its first action. The second call is always a no-op since the entry was already removed by the first.

Consider picking one canonical location. Since `#addConnection` already guards against stale connections, this early cancel could be removed — or conversely, remove the one from `#addConnection` and document that callers must clean up first.

packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
```diff
@@ -10,10 +10,84 @@ const path = require('node:path');
...
+      );
+      return { proxyUrl: `http://${username}:${password}@${hostname}${portStr}` };
+    }
+    return { proxyUrl: `http://${hostname}${portStr}` };
+  }
+
+  if (proxyMode === 'system') {
+    const { http_proxy, https_proxy, grpc_proxy, no_proxy, no_grpc_proxy } = proxyConfig || {};
+    // Prefer gRPC-specific no_grpc_proxy over general no_proxy
+    const noProxyList = no_grpc_proxy || no_proxy || '';
```
sanish-bruno: **Suggestion**: `no_grpc_proxy` completely overrides `no_proxy` — consider documenting this behavior.

With `no_grpc_proxy || no_proxy || ''`, if a user has `no_proxy=localhost,127.0.0.1` but `no_grpc_proxy=*.internal.com`, then `localhost` and `127.0.0.1` **will** be proxied for gRPC requests. This is a reasonable design choice matching `@grpc/grpc-js` precedence, but it may surprise users who expect `no_proxy` to always apply.

A brief comment explaining the intentional override semantics would help future maintainers:
```js
// no_grpc_proxy fully overrides no_proxy when set (not merged).
// This matches @grpc/grpc-js behavior where gRPC-specific env vars take precedence.
const noProxyList = no_grpc_proxy || no_proxy || '';
```

packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
```diff
@@ -10,10 +10,84 @@ const path = require('node:path');
...
+ * Always returns an object so that @grpc/grpc-js's built-in env var proxy
+ * (grpc_proxy/http_proxy/https_proxy) is disabled — Bruno controls all proxy behavior.
+ *
+ * @param {string} proxyMode - 'on', 'system', or 'off'
+ * @param {Object} proxyConfig - Raw proxy config from getCertsAndProxyConfig
+ * @param {string} requestUrl - The gRPC request URL
+ * @param {Object} interpolationOptions - Variable interpolation options
+ * @returns {{ proxyUrl: string | null }}
+ */
+const resolveGrpcProxyConfig = (proxyMode, proxyConfig, requestUrl, interpolationOptions) => {
```
sanish-bruno, line 32: **Test gap**: `resolveGrpcProxyConfig` has no direct unit tests.

The proxy resolution logic in this function — SOCKS/HTTPS protocol rejection, system proxy URL validation, `no_grpc_proxy` vs `no_proxy` precedence, bypass-proxy logic, and the auth credential encoding — is only tested indirectly through the `GrpcClient` tests in `grpc-client.spec.js`, which test `#resolveProxyTarget` (channel option mapping) but not the proxy *resolution* logic here.

Consider either:
1. Extracting this to a separate module (e.g., `resolve-grpc-proxy.js`) with its own unit tests
2. Or adding tests to the existing `grpc-event-handlers` test suite (if one exists)

Key paths to test:
- `proxyMode='on'` with SOCKS protocol → returns `{ proxyUrl: null }` with warning
- `proxyMode='system'` with `no_grpc_proxy` set → `no_proxy` is ignored
- `proxyMode='system'` with non-HTTP system proxy URL → returns null
- `proxyMode='on'` with auth enabled → URL contains encoded credentials

packages/bruno-requests/src/grpc/grpc-client.js
```diff
@@ -756,18 +874,14 @@ class GrpcClient {
...
 
-    this.activeConnections.forEach((connection) => {
-      if (typeof connection.cancel === 'function') {
-        connection.cancel();
-      }
-    });
-
-    this.activeConnections.clear();
+    for (const id of connectionIds) {
+      this.#cancelAndCloseConnection(id);
```
sanish-bruno, line 882: **Suggestion**: `clearAllConnections` emits N+1 `grpc:connections-changed` events.

Each `#cancelAndCloseConnection(id)` calls `#removeConnection`, which emits `grpc:connections-changed`. Then the `if (connectionIds.length > 0)` block emits one more. For N active connections, this results in N+1 events to the renderer, each triggering a Redux dispatch + re-render.

This is unlikely to cause visible issues in practice (N is typically small), but if you want to optimize, consider a silent-mode approach:
```js
for (const id of connectionIds) {
  // Close without emitting per-connection events
  const entry = this.activeConnections.get(id);
  if (entry?.rpc?.cancel) entry.rpc.cancel();
  if (entry?.client?.close) entry.client.close();
  this.activeConnections.delete(id);
}
// Single event for all
if (connectionIds.length > 0) {
  this.eventCallback('grpc:connections-changed', { activeConnectionIds: [] });
}
```

packages/bruno-requests/src/network/system-proxy/index.ts
```diff
@@ -94,7 +98,9 @@ export async function getSystemProxy(): Promise<ProxyConfiguration> {
     return {
       http_proxy: proxyEnvironmentVariables?.http_proxy || systemProxyEnvironmentVariables?.http_proxy,
       https_proxy: proxyEnvironmentVariables?.https_proxy || systemProxyEnvironmentVariables?.https_proxy,
+      grpc_proxy: proxyEnvironmentVariables?.grpc_proxy || null,
```
sanish-bruno: **Design**: `grpc_proxy` is sourced only from environment variables, never from OS-level system proxy.

Unlike `http_proxy`/`https_proxy` which merge env vars with OS system proxy (macOS `networksetup`, Windows registry), `grpc_proxy` falls back to `null` when the env var isn't set. This is likely intentional since OS-level proxy settings don't have a gRPC proxy concept, but it creates an asymmetry that might confuse maintainers.

Consider adding a brief comment:
```ts
// grpc_proxy is env-var only — OS-level system proxy settings
// (macOS/Windows) don't support gRPC-specific proxy configuration.
grpc_proxy: proxyEnvironmentVariables?.grpc_proxy || null,
```

playwright.config.ts
```diff
@@ -22,7 +22,7 @@ export default defineConfig({
       name: 'default',
       testDir: './tests',
       testIgnore: [
-        'ssl/**' // custom CA certificate tests require separate server setup and certificate generation
+        'ssl/**', // custom CA certificate tests require separate server setup and certificate generation
```
sanish-bruno: **Nit**: This trailing comma addition is unrelated to gRPC proxy support. Consider splitting it into its own commit or omitting it from this PR to keep the changeset focused.

packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
```diff
@@ -99,19 +173,26 @@ const registerGrpcEventHandlers = (window) => {
...
         url: preparedRequest.url,
         method: preparedRequest.method,
         methodType: preparedRequest.methodType,
         headers: preparedRequest.headers,
         body: preparedRequest.body,
-        timestamp: Date.now()
+        timestamp: Date.now(),
+        proxy: {
+          mode: proxyMode,
+          url: grpcProxyConfig.proxyUrl || null
```
sanish-bruno, line 185: **Security**: The proxy URL stored here (and sent to the renderer via `sendEvent`) may contain authentication credentials.

When `proxyMode === 'on'` and proxy auth is enabled, `grpcProxyConfig.proxyUrl` will be `http://user:password@host:port`. This URL is:
1. Stored in Redux state (via the `grpc:request` event listener)
2. Displayed in the GrpcTimelineItem component
3. Potentially included in exported HAR/debug data

Consider stripping credentials before sending to the renderer:
```js
const sanitizeUrl = (url) => {
  if (!url) return null;
  try {
    const parsed = new URL(url);
    parsed.username = '';
    parsed.password = '';
    return parsed.toString();
  } catch { return url; }
};

proxy: {
  mode: proxyMode,
  url: sanitizeUrl(grpcProxyConfig.proxyUrl)
}
```

The full URL with credentials is still used by the gRPC client (passed via `proxyConfig: grpcProxyConfig`) — this change would only affect the display/logging path.

packages/bruno-requests/src/grpc/grpc-client.js
```diff
@@ -563,12 +648,16 @@ class GrpcClient {
     );
     const userAgentValue = userAgentKey ? request.headers[userAgentKey] : null;
 
-    const mergedChannelOptions = userAgentValue
-      ? { 'grpc.primary_user_agent': userAgentValue, ...channelOptions }
-      : channelOptions;
+    // Resolve proxy target and channel options
+    const { targetHost, proxyChannelOptions } = this.#resolveProxyTarget(host, proxyConfig);
+
+    const mergedChannelOptions = { ...channelOptions, ...proxyChannelOptions };
```
sanish-bruno, line 658: may need to change , the user agent tests seems to be failing currently, order or spreading seems to influence the behaviour, may also need to check if this is affecting proxy options

```js
    const userAgentChannelOption = userAgentValue ? { 'grpc.primary_user_agent': userAgentValue } : {};
    const mergedChannelOptions = { ...userAgentChannelOption, ...channelOptions, ...proxyChannelOptions };
```

packages/bruno-requests/src/grpc/grpc-client.js
```diff
@@ -662,7 +754,14 @@ class GrpcClient {
       (key) => key.toLowerCase() === 'user-agent'
     );
     const userAgentValue = userAgentKey ? request.headers[userAgentKey] : null;
-    const mergedChannelOptions = userAgentValue ? { 'grpc.primary_user_agent': userAgentValue, ...channelOptions } : channelOptions;
+
+    // Resolve proxy target and channel options
+    const { targetHost, proxyChannelOptions } = this.#resolveProxyTarget(host, proxyConfig);
+
+    const mergedChannelOptions = { ...channelOptions, ...proxyChannelOptions };
```
sanish-bruno, line 765: may need to change , the user agent tests seems to be failing currently, order or spreading seems to influence the behaviour, may also need to check if this is affecting proxy options same as above

```js
    const userAgentChannelOption = userAgentValue ? { 'grpc.primary_user_agent': userAgentValue } : {};
    const mergedChannelOptions = { ...userAgentChannelOption, ...channelOptions, ...proxyChannelOptions };
```

packages/bruno-app/src/components/ResponsePane/Timeline/GrpcTimelineItem/index.js
```diff
@@ -74,6 +75,28 @@ const GrpcTimelineItem = ({ timestamp, request, response, eventType, eventData,
...
+                    </div>
+                  ) : (
```
sanish-bruno, line 93: I guess when there is no proxy, we do not even have to show anything, what do you think? But i do feel like `Proxy enabled but not applicable for this request`  do have some merit. may need to take opinion on this from product

packages/bruno-app/src/components/Preferences/ProxySettings/SystemProxy/index.js
```diff
@@ -79,12 +79,24 @@ const SystemProxy = () => {
             </label>
             <div className="system-proxy-value">{https_proxy || '-'}</div>
           </div>
+          <div className="mb-1 flex items-center">
+            <label className="settings-label">
```
sanish-bruno: I guess we do not have `http` , `https` like distinction for `grpc`, cause it will always happen over a secured connection?

packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
```diff
@@ -10,10 +10,84 @@ const path = require('node:path');
...
+    const shouldProxy = shouldUseProxy(requestUrl, noProxyList);
+    if (!shouldProxy) return { proxyUrl: null };
+
+    // Prefer grpc_proxy (gRPC-specific) over general http/https proxy
+    // This matches @grpc/grpc-js precedence: grpc_proxy > https_proxy > http_proxy
+    const systemProxy = grpc_proxy || https_proxy || http_proxy;
+    if (!systemProxy) return { proxyUrl: null };
+
+    try {
+      const parsed = new URL(systemProxy);
```
sanish-bruno, line 69: Is it possible that the protocol can be `grpc:`?

packages/bruno-requests/src/network/system-proxy/index.ts
```diff
@@ -66,16 +66,20 @@ export class SystemProxyResolver {
   }
 
   getEnvironmentVariables(): ProxyConfiguration {
-    const { http_proxy, HTTP_PROXY, https_proxy, HTTPS_PROXY, no_proxy, NO_PROXY, all_proxy, ALL_PROXY } = process.env;
+    const { http_proxy, HTTP_PROXY, https_proxy, HTTPS_PROXY, no_proxy, NO_PROXY, all_proxy, ALL_PROXY, grpc_proxy, GRPC_PROXY, no_grpc_proxy, NO_GRPC_PROXY } = process.env;
```
sanish-bruno: Does other clients have `grpc proxy` usage or they just get `http proxy`?  i believe within system settings it will always have http, https

packages/bruno-electron/src/ipc/network/grpc-event-handlers.js
```diff
@@ -99,19 +169,26 @@ const registerGrpcEventHandlers = (window) => {
...
         method: preparedRequest.method,
         methodType: preparedRequest.methodType,
         headers: preparedRequest.headers,
         body: preparedRequest.body,
-        timestamp: Date.now()
+        timestamp: Date.now(),
+        proxy: {
+          mode: proxyMode,
+          url: grpcProxyConfig.proxyUrl || null
+        }
```
sanish-bruno, line 186: **Security**: Proxy credentials are displayed in the gRPC timeline UI.

When `proxyMode === 'on'` with auth enabled, `resolveGrpcProxyConfig` returns a URL like `http://user:pass@host:port`. This is sent verbatim to the renderer via `requestSent.proxy.url` and rendered in the timeline as `Using proxy: http://user:pass@host:port`.

The existing HTTP proxy code in `proxy-util.js` intentionally strips credentials from timeline messages:
```javascript
// proxy-util.js line 155
message: `Using proxy: ${proxyProtocol}://${proxyHostname}${uriPort}`
```

Suggested fix: Strip credentials from the display URL before sending to the UI, while keeping the full URL for the actual proxy connection:
```javascript
const displayProxyUrl = grpcProxyConfig.proxyUrl
  ? (() => { try { const u = new URL(grpcProxyConfig.proxyUrl); return `${u.protocol}//${u.host}`; } catch { return grpcProxyConfig.proxyUrl; } })()
  : null;

const requestSent = {
  // ...
  proxy: {
    mode: proxyMode,
    url: displayProxyUrl
  }
};
```

packages/bruno-requests/src/grpc/grpc-client.js
```diff
@@ -756,18 +878,14 @@ class GrpcClient {
...
+      this.#cancelAndCloseConnection(id);
+    }
 
-    // Emit an event with empty active connection IDs
+    // Emit once if there were connections (individual calls already emitted,
+    // but ensure clean state)
     if (connectionIds.length > 0) {
       this.eventCallback('grpc:connections-changed', {
-        type: 'cleared',
         activeConnectionIds: []
```
sanish-bruno, line 889: **Nit**: The final `connections-changed` event here is redundant — each `#cancelAndCloseConnection` call inside the loop already triggers `#removeConnection`, which emits its own `connections-changed` event. So for N connections, N+1 events are emitted.

Not a bug since the final event guarantees an empty array, but it produces unnecessary re-renders. You could either skip individual events during batch clear (e.g. a flag), or just remove this final emit since the last `#removeConnection` will already produce `{ activeConnectionIds: [] }`.

Issue #5704, opened 2025-10-04
Authorization header passed after redirection to different domain


jkolczasty:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [x] slowing me down but I'm able to continue working
- [x] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.12

### Operating System

Linux

### Describe the bug

Bruno passes Authorization header after rediraction (307 Temporary Redirect) to different domain.

Bruno should not add authorzation headers after redir to different domain.

It's even possible credentials leakage issue.

### Screenshots/Live demo link

In my test I'm using in API redirection to S3. Bruno redirects with Authorization header and call to S3 server fails with error: "multiple authentication types".

Bruno should not add authorzation headers after redir to different domain.

PR #7578, opened 2026-03-26
feat: enhance axios instance to strip sensitive auth headers on cross-origin redirects

### LINKED ISSUE #7570: Stop sending the Authorization header during a 302 redirect if the host is not the same

### This bug is:

- [x] annoying

### Bruno version

3.2.0

### Operating System

Windows 11

### Describe the bug

When you GET an endpoint with a configured Authorization header and this endpoint returns a 302, when the "follow redirect" is enabled then Bruno sends the same request to the new URL (fine) BUT it should NOT send the Authorization header IF this is a different host (to avoid credentials leaking)

This is how cURL does it since 7.58.0 :  https://curl.se/docs/CVE-2018-1000007.html

That is also done by the regular C# HTTP client 

Unfortunately, I am not able to share my .bru file since it contains calls to API endpoints with sensitive information

However, I can tell you my exact current problem

* I GET an endpoint with a Authorization header
* it returns 302 to another host with a signature in the query param
* Bruno follows the redirect
* the other host returns a 400 with an error "can not use Authorization header and signature query param at the same time"

If I use "generate code" to obtain the curl command and use it, I correctly get a 200 return code

Upon review of the curl -v output I can see that the Authorization header is not sent in the second request

Here is the truncated and obfuscated output of the curl command that is given by bruno on "generate code" where I added -L to follow the redirect and -v to see everything that is done by curl

The important part is that in the first request, you can see the authorization header whereas in the second request there is no such header

```
PS D:\curl\bin> .\curl.exe --request GET --url https://privatehost/private path  --header 'authorization: private_bearer' -L -v
...
> GET privatepath HTTP/2
> Host: privatehost
> User-Agent: curl/8.17.0
> Accept: */*
> authorization: privatebearer
>
* Request completely sent off
< HTTP/2 302
< date: Wed, 25 Mar 2026 07:47:35 GMT
< content-length: 0
< location: https://anotherprivatehost/panotherprivatepath?signature=xxx
* Ignoring the response-body
...
* Issue another request to this URL: 'https://anotherprivatehost/panotherprivatepath?signature=xxx'
...
* using HTTP/1.x
> GET /anotherprivatepath?signature=xxx HTTP/1.1
> Host: anotherprivatehost
> User-Agent: curl/8.17.0
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
```

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2941)

- Strip `Authorization` and `Proxy-Authorization` headers when a redirect crosses origins (different host, port, or scheme)
- Matches curl 7.58+ and browser behavior (CVE-2018-1000007)
- Applies to both Electron and CLI axios interceptors

Fixes #7570 and #5704

## Test plan
- [ ] `npx bruno run collection/redirects/Test\ Cross\ Origin\ Redirect\ Strips\ Auth.bru` — auth stripped
- [ ] `npx bruno run collection/redirects/Test\ Same\ Origin\ Redirect\ Preserves\ Auth.bru` — auth preserved
- [ ] Existing redirect tests still pass

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Enhanced HTTP redirect security: Authorization headers are removed when a request follows a redirect to a different origin, and preserved for same-origin redirects.

* **Tests**
  * Added end-to-end tests verifying auth header behavior for cross-origin and same-origin redirects.
  * Test environment extended with a cross-origin redirect endpoint to validate stripping behavior.

### DIFF
--- a/packages/bruno-cli/src/utils/axios-instance.js
+++ b/packages/bruno-cli/src/utils/axios-instance.js
@@ -27,6 +27,21 @@ const createRedirectConfig = (error, redirectUrl) => {
     headers: { ...error.config.headers }
   };
 
+  // Strip sensitive auth headers when redirecting to a different origin
+  // (matches curl 7.58+ and browser behavior, see CVE-2018-1000007)
+  try {
+    const originalOrigin = new URL(error.config.url).origin;
+    const redirectOrigin = new URL(redirectUrl).origin;
+    if (originalOrigin !== redirectOrigin) {
+      delete requestConfig.headers['authorization'];
+      delete requestConfig.headers['Authorization'];
+      delete requestConfig.headers['proxy-authorization'];
+      delete requestConfig.headers['Proxy-Authorization'];
+    }
+  } catch (_) {
+    // If URL parsing fails, skip origin check — headers pass through unchanged
+  }
+
   const statusCode = error.response.status;
   const originalMethod = (error.config.method || 'get').toLowerCase();
 
--- a/packages/bruno-electron/src/ipc/network/axios-instance.js
+++ b/packages/bruno-electron/src/ipc/network/axios-instance.js
@@ -352,6 +352,28 @@ function makeAxiosInstance({
             }
           };
 
+          // Strip sensitive auth headers when redirecting to a different origin
+          // (matches curl 7.58+ and browser behavior, see CVE-2018-1000007)
+          try {
+            // URL.URL is the WHATWG URL class — line 1's `require('url')` shadows the global `URL`
+            const originalOrigin = new URL.URL(error.config.url).origin;
+            const redirectOrigin = new URL.URL(redirectUrl).origin;
+            if (originalOrigin !== redirectOrigin) {
+              delete requestConfig.headers['authorization'];
+              delete requestConfig.headers['Authorization'];
+              delete requestConfig.headers['proxy-authorization'];
+              delete requestConfig.headers['Proxy-Authorization'];
+
+              timeline.push({
+                timestamp: new Date(),
+                type: 'info',
+                message: `Stripped Authorization headers: redirect crosses origin (${originalOrigin} → ${redirectOrigin})`
+              });
+            }
+          } catch (_) {
+            // If URL parsing fails, skip origin check — headers pass through unchanged
+          }
+
           // Apply proper HTTP redirect behavior based on status code
           const statusCode = error.response.status;
           const originalMethod = (error.config.method || 'get').toLowerCase();
--- a/packages/bruno-tests/src/redirect/index.js
+++ b/packages/bruno-tests/src/redirect/index.js
@@ -103,6 +103,11 @@ router.get('/anything', function (req, res) {
   });
 });
 
+router.get('/cross-origin', function (req, res) {
+  // Redirect to a different port = different origin
+  res.status(302).set('Location', 'http://localhost:8082/headers').send();
+});
+
 router.get('/:count', function (req, res) {
   const count = parseInt(req.params.count, 10);
 
--- a/packages/bruno-tests/collection/redirects/Test Cross Origin Redirect Strips Auth.bru
+++ b/packages/bruno-tests/collection/redirects/Test Cross Origin Redirect Strips Auth.bru
@@ -1,0 +1,31 @@
+meta {
+  name: Test Cross Origin Redirect Strips Auth
+  type: http
+  seq: 6
+}
+
+get {
+  url: {{localhost}}/api/redirect/cross-origin
+  body: none
+  auth: none
+}
+
+headers {
+  Authorization: Bearer test-secret
+}
+
+assert {
+  res.status: 200
+}
+
+tests {
+  test("should strip authorization header on cross-origin redirect", function() {
+    const headers = res.getBody();
+    expect(headers).to.not.have.property('authorization');
+  });
+}
+
+settings {
+  followRedirects: true
+  maxRedirects: 5
+}
--- a/packages/bruno-tests/collection/redirects/Test Same Origin Redirect Preserves Auth.bru
+++ b/packages/bruno-tests/collection/redirects/Test Same Origin Redirect Preserves Auth.bru
@@ -1,0 +1,32 @@
+meta {
+  name: Test Same Origin Redirect Preserves Auth
+  type: http
+  seq: 7
+}
+
+get {
+  url: {{localhost}}/redirect-to-headers
+  body: none
+  auth: none
+}
+
+headers {
+  Authorization: Bearer test-secret
+}
+
+assert {
+  res.status: 200
+}
+
+tests {
+  test("should preserve authorization header on same-origin redirect", function() {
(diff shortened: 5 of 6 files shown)


packages/bruno-electron/src/ipc/network/axios-instance.js
```diff
@@ -352,6 +352,28 @@ function makeAxiosInstance({
...
+
+              timeline.push({
+                timestamp: new Date(),
+                type: 'info',
+                message: `Stripped Authorization headers: redirect crosses origin (${originalOrigin} → ${redirectOrigin})`
+              });
+            }
+          } catch (_) {
+            // If URL parsing fails, skip origin check — headers pass through unchanged
+          }
```
sanish-bruno, line 375: **Security (Suggestion)**: The `catch` block silently lets auth headers pass through when URL parsing fails. For a security-oriented feature, a **fail-closed** approach would be safer — strip auth headers when you can't verify same-origin rather than preserving them.

Suggested fix:
```js
} catch (_) {
  // If URL parsing fails, err on the side of caution — strip auth headers
  delete requestConfig.headers['authorization'];
  delete requestConfig.headers['Authorization'];
  delete requestConfig.headers['proxy-authorization'];
  delete requestConfig.headers['Proxy-Authorization'];
}
```

In practice, if the URL is too malformed to parse, axios will also fail the redirect request — so the fail-open behavior is unlikely to cause real harm. But the fail-closed pattern is a better default for security code.

packages/bruno-cli/src/utils/axios-instance.js
```diff
@@ -26,6 +26,21 @@ const createRedirectConfig = (error, redirectUrl) => {
...
+    const redirectOrigin = new URL(redirectUrl).origin;
+    if (originalOrigin !== redirectOrigin) {
+      delete requestConfig.headers['authorization'];
+      delete requestConfig.headers['Authorization'];
+      delete requestConfig.headers['proxy-authorization'];
+      delete requestConfig.headers['Proxy-Authorization'];
+    }
+  } catch (_) {
+    // If URL parsing fails, skip origin check — headers pass through unchanged
+  }
```
sanish-bruno, line 43: **Security (Suggestion)**: Same fail-open concern as the Electron version — consider stripping headers in the `catch` block rather than preserving them.

packages/bruno-tests/src/redirect/index.js
```diff
@@ -103,6 +103,11 @@ router.get('/anything', function (req, res) {
   });
 });
 
+router.get('/cross-origin', function (req, res) {
+  // Redirect to a different port = different origin
+  res.status(302).set('Location', 'http://localhost:8082/headers').send();
```
sanish-bruno, line 108: **Bug (Minor)**: Port `8082` is hardcoded here, but `crossOriginPort` in `src/index.js` is configurable via `process.env.CROSS_ORIGIN_PORT`. If someone changes the env var, the redirect would still point to `:8082` while the server listens elsewhere.

Suggested fix: either export `crossOriginPort` and import it here, or extract a shared constant.
```js
const crossOriginPort = process.env.CROSS_ORIGIN_PORT || 8082;
// ...
res.status(302).set('Location', `http://localhost:${crossOriginPort}/headers`).send();
```

packages/bruno-cli/src/utils/axios-instance.js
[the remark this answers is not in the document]
abhishek-bruno: The bruno-requests makeAxiosInstance relies on axios's built-in redirect handling (via follow-redirects), which already protects against CVE-2018-1000007 since v1.15.4. Bruno pins ^1.15.11. The CLI and electron packages duplicate this logic because they set maxRedirects: 0 and handle redirects manually — but the bruno-requests instance doesn't need that since the underlying library already handles it.

packages/bruno-electron/src/ipc/network/axios-instance.js
abhishek-bruno: You're right that fail-closed is the better default for security code as a principle. That said, as you noted, if the URL is too malformed for new URL() to parse, axios/follow-redirects will also fail to make the redirect request — so the auth headers never actually hit the wire. The practical risk here is zero.

packages/bruno-cli/src/utils/axios-instance.js
abhishek-bruno: You're right that fail-closed is the better default for security code as a principle. That said, as you noted, if the URL is too malformed for new URL() to parse, axios/follow-redirects will also fail to make the redirect request — so the auth headers never actually hit the wire. The practical risk here is zero.

packages/bruno-tests/src/redirect/index.js
abhishek-bruno: The hardcoded 8082 in the redirect test route would only break if someone explicitly sets CROSS_ORIGIN_PORT to something other than 8082, which is an unlikely scenario for a test server.

packages/bruno-cli/src/utils/axios-instance.js
```diff
@@ -27,6 +27,21 @@ const createRedirectConfig = (error, redirectUrl) => {
...
+  // Strip sensitive auth headers when redirecting to a different origin
+  // (matches curl 7.58+ and browser behavior, see CVE-2018-1000007)
+  try {
+    const originalOrigin = new URL(error.config.url).origin;
+    const redirectOrigin = new URL(redirectUrl).origin;
+    if (originalOrigin !== redirectOrigin) {
+      delete requestConfig.headers['authorization'];
+      delete requestConfig.headers['Authorization'];
+      delete requestConfig.headers['proxy-authorization'];
+      delete requestConfig.headers['Proxy-Authorization'];
```
bijin-bruno, line 39: Header names are case-insensitive, use lowercase comparison to ensure consistent matching.

packages/bruno-electron/src/ipc/network/axios-instance.js
```diff
@@ -352,6 +352,28 @@ function makeAxiosInstance({
...
+            if (originalOrigin !== redirectOrigin) {
+              delete requestConfig.headers['authorization'];
+              delete requestConfig.headers['Authorization'];
+              delete requestConfig.headers['proxy-authorization'];
+              delete requestConfig.headers['Proxy-Authorization'];
+
+              timeline.push({
+                timestamp: new Date(),
+                type: 'info',
+                message: `Stripped Authorization headers: redirect crosses origin (${originalOrigin} → ${redirectOrigin})`
```
bijin-bruno, line 370: Is this a valid timeline event?

bijin-bruno:
This requires further product discussion — should we preserve the current behavior and introduce an opt-in override at the request and/or collection settings level?
cc - @sreelakshmi-bruno @anusree-bruno @naman-bruno

sreelakshmi-bruno:
Yes, we should do this after we introduce a "Follow Authorization header" setting for users who need the old behavior.

PR #7600, opened 2026-03-27
fix: resolve process.env variables in global environment level

### LINKED ISSUE #7572: Cannot use process.env at global environment level

### This bug is:

- [x] annoying

### Bruno version

3.2.0

### Operating System

Max OS Sequoia 15.7.3

### Describe the bug

I cannot use process.env on the global environment level. There is no auto-completion, and if I set it manually the variable is still empty.

It works as expected at the collection environment level. However, although the UI is similar between the global environment, it is not behaving the same way.

**Expected situation**
- Ideally, make it possible for the global environment variables to be able to pick the values from the .env file at the workspace level (same behavior than the collections).
- If for some reasons that is not supported, then adapt the UI to hide the .env files on the bottom left.

_No response_

Image

--

Image

abhishek-bruno:
### Description

- Fix process.env variable resolution in the global environment editor UI — tooltip/preview now correctly shows resolved values from workspace .env files
- Fix process.env variable resolution at runtime for workspace-level (scratch pad) requests by mapping scratch collections to their workspace path

### Root Cause
Two missing data connections:

- The global environment editor passes collection={null} to EnvironmentVariablesTable, so getAllVariables() had no access to workspace process.env values
- Scratch collections were never mapped to their workspace via setCollectionWorkspace, so getProcessEnvVars() couldn't find workspace .env values at runtime

### Changes
EnvironmentVariablesTable — when collection is null, read processEnvVariables directly from the active workspace in Redux
mountScratchCollection — call renderer:set-collection-workspace to map scratch collection to workspace path

Fixes #7572
[JIRA](https://usebruno.atlassian.net/browse/BRU-2978)

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed resolution of {{process.env.X}} so workspace/global .env values are applied correctly in requests.
  * Ensured newly mounted workspace scratch collections are associated with their workspace so .env values resolve as expected.

* **Tests**
  * Added end-to-end test and fixtures validating global .env / process.env resolution.
  * Added test selectors across environment UI and raw .env editor to support targeted testing.

### DIFF
--- a/tests/environments/global-env-process-env-resolution/global-env-process-env-resolution.spec.ts
+++ b/tests/environments/global-env-process-env-resolution/global-env-process-env-resolution.spec.ts
@@ -4,17 +4,10 @@ import {
   openEnvironmentSelector,
   openRequest,
   sendRequest,
-  expectResponseContains,
-  closeAllCollections
+  expectResponseContains
 } from '../../utils/page';
 
 test.describe('Global Environment process.env Resolution', () => {
-  test.setTimeout(60000);
-
-  test.afterEach(async ({ pageWithUserData: page }) => {
-    await closeAllCollections(page);
-  });
-
   test('should resolve process.env variables referenced in global environment', async ({
     pageWithUserData: page
   }) => {


packages/bruno-app/src/components/EnvironmentVariablesTable/index.js
```diff
@@ -45,6 +45,10 @@ const EnvironmentVariablesTable = ({
 }) => {
   const { storedTheme } = useTheme();
   const { globalEnvironments, activeGlobalEnvironmentUid } = useSelector((state) => state.globalEnvironments);
+  const activeWorkspaceUid = useSelector((state) => state.workspaces?.activeWorkspaceUid);
+  const activeWorkspace = useSelector((state) =>
+    state.workspaces?.workspaces?.find((w) => w.uid === activeWorkspaceUid)
```
sanish-bruno: **Suggestion**: These two selectors can be combined into one to avoid an intermediate variable and a subtle stale-closure edge case.

When Redux state changes, the second selector re-runs with `activeWorkspaceUid` captured from the *previous* render. If the active workspace UID changes in the same Redux dispatch, the second selector momentarily uses the old UID. React-Redux handles this (the first selector triggers a re-render which fixes it), but a single selector avoids the issue entirely and is simpler:

```js
const activeWorkspace = useSelector((state) => {
  const uid = state.workspaces?.activeWorkspaceUid;
  return state.workspaces?.workspaces?.find((w) => w.uid === uid);
});
```

Since `activeWorkspaceUid` is not used elsewhere in this component, this is a clean simplification.

packages/bruno-app/src/providers/ReduxStore/slices/workspaces/actions.js
```diff
@@ -1001,6 +1001,11 @@ export const mountScratchCollection = (workspaceUid) => {
         brunoConfig
       });
 
+      // Map scratch collection to workspace so getProcessEnvVars can resolve workspace .env values
+      if (workspace.pathname) {
```
sanish-bruno, line 1005: **Nit**: If `renderer:set-collection-workspace` throws, it will prevent `openScratchCollectionEvent` from executing (since it's the next `await`), but the collection watcher was already added on line 998. This could leave a watcher running for a collection that was never opened in Redux.

This is unlikely to matter in practice, but if you want to be defensive, consider placing the `set-collection-workspace` call *after* `openScratchCollectionEvent` (since the workspace mapping is only needed later when a request is actually executed, not at mount time). Or simply catch and log the error:

```js
if (workspace.pathname) {
  await ipcRenderer.invoke('renderer:set-collection-workspace', scratchCollectionUid, workspace.pathname)
    .catch((err) => console.error('Failed to set collection workspace:', err));
}
```

sanish-bruno:
@abhishek-bruno is it possible to have playwright test cases for this?

PR #7649, opened 2026-04-01
fix: add no-case-declarations, react-hooks ESLint rules and fix violations

sanish-bruno:
[Jira](https://usebruno.atlassian.net/browse/BRU-3075)

### Description

Adds ESLint rules globally and fixes all existing violations across the codebase.

#### 1. `no-case-declarations` (global)

Prevents lexical declarations (`const`, `let`) in `case` blocks without braces, which can lead to unexpected scoping issues.

**Fix:** Wrapped all case bodies containing declarations in `{ }` blocks across 13 files:
- `bruno-app`: `interpolation.js`, `collections/index.js` (Redux slice), `auth.js`, `collections/index.js` (utils), `openapi-spec.js`
- `bruno-converters`: `postman-to-bruno.js`
- `bruno-electron`: `prepare-request.js`, `ui-state-snapshot.js`
- `bruno-filestore`: `body.ts`
- `bruno-js`: `assert-runtime.js`
- `tests`: `platform.js`

#### 2. `react-hooks/rules-of-hooks` (bruno-app)

Enforces React's Rules of Hooks — hooks must be called at the top level and in the same order on every render.

**Fixes:**
- **`Oauth2TokenViewer`** — `TokenSection` and `ExpiryTimer` had early returns before hooks. Moved early returns after hooks.
- **`RemoveCollectionsModal`** — `useMemo` was called after an early `return null`. Moved `useMemo` before the early return.
- **`Table`** — `useRef()` was called inside `.map()`. Replaced with a `useRef([])` holding an array of `createRef()` objects.

#### 3. `react-hooks/purity` (bruno-app)

Prevents calling impure functions during render, which can produce unstable results.

**Fix:**
- **`StopWatch`** — `useState(Date.now())` called an impure function during render. Changed to lazy initializer `useState(() => Date.now())`.

#### 4. `react-hooks/set-state-in-render` (bruno-app)

Prevents calling setState during render. No existing violations — rule added for prevention.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed React Hooks-related rendering issues in OAuth2 token UI.
  * Made table column sizing/stability more reliable across renders.
  * Ensured stopwatch/timer initializes without premature timestamps.

* **Chores**
  * Added and enforced React Hooks linting rules (exempting stories).
  * Scoped many switch-case blocks for safer variable handling.
  * Normalized package/config formatting and review instructions.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -78,7 +78,8 @@ module.exports = runESMImports().then(() => defineConfig([
       '@stylistic/max-len': ['off'],
       '@stylistic/jsx-one-expression-per-line': ['off'],
       '@stylistic/max-statements-per-line': ['off'],
-      '@stylistic/no-mixed-operators': ['off']
+      '@stylistic/no-mixed-operators': ['off'],
+      'no-case-declarations': 'error'
     }
   },
   {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -107,7 +107,9 @@ module.exports = runESMImports().then(() => defineConfig([
     },
     rules: {
       'no-undef': 'error',
-      'react-hooks/rules-of-hooks': 'error'
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/purity': 'error',
+      'react-hooks/set-state-in-render': 'error'
     }
   },
   {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -3,6 +3,7 @@ const { defineConfig } = require('eslint/config');
 const globals = require('globals');
 const { fixupPluginRules } = require('@eslint/compat');
 const eslintPluginDiff = require('eslint-plugin-diff');
+const reactHooks = require('eslint-plugin-react-hooks');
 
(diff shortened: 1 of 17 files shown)


eslint.config.js
```diff
@@ -101,7 +106,10 @@ module.exports = runESMImports().then(() => defineConfig([
       }
     },
     rules: {
-      'no-undef': 'error'
+      'no-undef': 'error',
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/purity': 'error',
+      'react-hooks/set-state-in-render': 'error'
```
sid-bruno, line 112: can we make this one `warn`/`1` since there can be a real situation where this is necessary because of how the component is designed but doesn't have to stop the PR for it since it might need a larger change to the component itself.

sid-bruno (dismissed):

bijin-bruno (dismissed):

PR #7673, opened 2026-04-03
feat: add PropertyList API for req.headers and res.headers

sanish-bruno:
[Jira](https://usebruno.atlassian.net/browse/BRU-2893)

## Summary

Adds a `headerList` property to both `req` and `res` in scripts, exposing a structured PropertyList API for header management. Method names are aligned with the [MDN Web API Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) interface. The existing `req.headers` and `res.headers` raw objects remain untouched for full backward compatibility.

### New API

**`req.headerList`** — writable `HeaderList` (extends `PropertyList`, dynamic mode):
```js
// Read
req.headerList.get('Content-Type')              // 'application/json'
req.headerList.has('Authorization')             // true
req.headerList.one('Content-Type')              // { key: 'Content-Type', value: 'application/json' }
req.headerList.all()                            // [{ key, value }, ...]
req.headerList.count()                          // 3

// Search
req.headerList.find(h => h.key.startsWith('X-'))
req.headerList.filter(h => h.value.includes('json'))
req.headerList.indexOf('Content-Type')

// Iteration
req.headerList.forEach((h, i) => { ... })
req.headerList.map(h => h.key)
req.headerList.reduce((acc, h) => ..., {})

// Transform
req.headerList.toObject()                       // { 'Content-Type': 'application/json', ... }
req.headerList.toString()                       // 'Content-Type: application/json\n...'

// Write
req.headerList.append({ key: 'X-Custom', value: 'val' })
req.headerList.append('X-Custom: val')          // string format
req.headerList.append('X-Custom', 'val')        // two-arg form
req.headerList.set({ key: 'Content-Type', value: 'text/plain' })
req.headerList.set('Content-Type', 'text/plain') // two-arg form
req.headerList.delete('X-Custom')
req.headerList.delete(h => h.key.startsWith('X-'))
req.headerList.clear()
req.headerList.populate([{ key: 'A', value: '1' }, ...])
req.headerList.repopulate([{ key: 'A', value: '1' }])
req.headerList.assimilate(otherList, true)
```

**`res.headerList`** — read-only `HeaderList` (static mode, write methods throw):
```js
res.headerList.get('content-type')              // 'application/json'
res.headerList.has('x-request-id')              // true
res.headerList.all()                            // [{ key, value }, ...]
res.headerList.filter(h => h.key.startsWith('x-'))
res.headerList.toObject()                       // plain object map
```

### Key design decisions

- **Separate property**: `req.headerList` / `res.headerList` is distinct from `req.headers` / `res.headers` — no Proxy layer, no risk of breaking existing scripts
- **MDN-aligned naming**: Write methods use `append`, `set`, `delete`, `forEach` (matching [MDN Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers))
- **Case-insensitive**: All key lookups are case-insensitive (HTTP headers are case-insensitive)
- **Disabled headers**: `req.headerList` surfaces disabled headers with `{ disabled: true }` — users can query/filter them
- **Dynamic reads**: `req.headerList` reads from the live `req.headers` object on every access, so mutations via `req.setHeader()` are always reflected
- **QuickJS support**: Full PropertyList bridge wired into the QuickJS sandbox with iterator support and synchronous write hooks

### `append` vs MDN's `Headers.append()`

MDN's `append()` creates duplicate headers with the same key. Bruno does not support multiple headers with the same name, so `append()` delegates to `set()` — it overwrites any existing header with the same key.

### `populate` and `repopulate`

- **`populate(items)`** — Adds headers from the array/string, but **skips any key that already exists** (case-insensitive). Postman's `populate` adds duplicate keys because Postman supports multiple headers with the same name. Bruno does not, so we skip existing keys to preserve the current value.
- **`repopulate(items)`** — Clears all existing headers first, then adds the new items. Matches Postman's `repopulate` behavior.

---

## Complete API Reference

All key lookups are **case-insensitive** per the HTTP spec.

#### Read Methods

| Method | Returns | Description |
|---|---|---|
| `get(name)` | `string \| undefined` | Value of the first header matching the key |
| `one(name)` | `object \| undefined` | Full header object `{ key, value }` for the matching key |
| `all()` | `object[]` | Cloned array of all header objects `{ key, value, disabled? }` |
| `count()` | `number` | Number of headers (including disabled) |

#### Search Methods

| Method | Returns | Description |
|---|---|---|
| `has(name)` | `boolean` | `true` if a header with that key exists |
| `has(name, value)` | `boolean` | `true` if key exists **and** value matches exactly |
| `has(object)` | `boolean` | `true` if a header with `object.key` exists |
| `find(fn, context?)` | `object \| undefined` | First header where the predicate returns truthy |
| `filter(fn, context?)` | `object[]` | All headers where the predicate returns truthy |
| `indexOf(item)` | `number` | Index of a header by string key or `{ key, value }` object; `-1` if not found |

#### Iteration Methods

| Method | Returns | Description |
|---|---|---|
| `forEach(fn, context?)` | `void` | Calls `fn(header, index)` for every header |
| `map(fn, context?)` | `any[]` | Returns a new array of mapped values |
| `reduce(fn, initial?, context?)` | `any` | Reduces headers to a single value |

#### Transform Methods

| Method | Returns | Description |
|---|---|---|
| `toObject(excludeDisabled?, caseSensitive?, multiValue?, sanitizeKeys?)` | `object` | Plain `{ key: value }` map |
| `toString()` | `string` | HTTP wire format `Key: Value\n...` (skips disabled headers) |
| `toJSON()` | `object[]` | Same as `all()` — suitable for `JSON.stringify()` |

#### Write Methods

| Method | Returns | Description |
|---|---|---|
| `append(headerObj \| name, value?)` | `void` | Sets a header; accepts `{ key, value }`, `"Key: Value"` string, or `(name, value)` two-arg form. Delegates to `set()` (overwrites existing). |
| `set(headerObj \| name, value?)` | `boolean \| null` | Sets or replaces a header. Returns `true` if new, `false` if updated, `null` if input was nil. |
| `delete(predicate, context?)` | `void` | Deletes header(s) by key string, `{ key }` object, or predicate function |
| `clear()` | `void` | Removes **all** headers (enabled and disabled) |
| `populate(items \| string)` | `void` | Adds items from array or multi-line string, **skipping keys that already exist** |
| `repopulate(items \| string)` | `void` | Clears all headers, then populates with new items |
| `assimilate(source, prune?)` | `void` | Merges headers from a PropertyList or array; if `prune` is true, removes headers not present in source |

---

### `res.headerList` — Response Headers (read-only)

All write methods (`append`, `set`, `delete`, `clear`, `populate`, `repopulate`, `assimilate`) throw a `"HeaderList is read-only"` error when called on response headers.

#### Read Methods

| Method | Returns | Description |
|---|---|---|
| `get(name)` | `string \| undefined` | Value of the first header matching the key |
| `one(name)` | `object \| undefined` | Full header object `{ key, value }` for the matching key |
| `all()` | `object[]` | Cloned array of all header objects `{ key, value }` |
| `count()` | `number` | Number of headers |

#### Search Methods

| Method | Returns | Description |
|---|---|---|
| `has(name)` | `boolean` | `true` if a header with that key exists |
| `has(name, value)` | `boolean` | `true` if key exists **and** value matches exactly |
| `has(object)` | `boolean` | `true` if a header with `object.key` exists |
| `find(fn, context?)` | `object \| undefined` | First header where the predicate returns truthy |
| `filter(fn, context?)` | `object[]` | All headers where the predicate returns truthy |
| `indexOf(item)` | `number` | Index of a header by string key or `{ key, value }` object; `-1` if not found |

#### Iteration Methods

| Method | Returns | Description |
|---|---|---|
| `forEach(fn, context?)` | `void` | Calls `fn(header, index)` for every header |
| `map(fn, context?)` | `any[]` | Returns a new array of mapped values |
| `reduce(fn, initial?, context?)` | `any` | Reduces headers to a single value |

#### Transform Methods

| Method | Returns | Description |
|---|---|---|
| `toObject(excludeDisabled?, caseSensitive?, multiValue?, sanitizeKeys?)` | `object` | Plain `{ key: value }` map |
| `toString()` | `string` | HTTP wire format `Key: Value\n...` |
| `toJSON()` | `object[]` | Same as `all()` — suitable for `JSON.stringify()` |

---

#### Contribution Checklist:


packages/bruno-js/src/sandbox/quickjs/shims/bruno-request.js
```diff
@@ -23,13 +22,24 @@ const addBrunoRequestShimToContext = (vm, req) => {
...
   pathParams.dispose();
   tags.dispose();
 
+  // Wire req.headers as a PropertyList bridge
+  const headersObj = vm.newObject();
+  const { evalCode: headersEvalCode } = createPropertyListBridge(vm, req.headers, headersObj, {
+    globalPath: 'globalThis.req.headers',
+    syncReadMethods: ['get', 'has', 'count', 'indexOf', 'toObject', 'toString'],
+    syncReadObjectMethods: ['one', 'all', 'idx', 'toJSON'],
+    syncWriteMethods: ['add', 'upsert', 'remove', 'clear'],
```
sanish-bruno: **Bug (Minor)**: `remove` is listed in `syncWriteMethods`, but `HeaderList.remove()` accepts a predicate function as its first argument (`remove(h => h.key === 'X-Foo')`). The sync write bridge does `vm.dump(a)` on each argument, which cannot serialize a QuickJS function handle into a callable JS function. As a result, `typeof predicate === 'function'` will be `false` on the native side, and the predicate path silently falls through.

Only the string and object forms of `remove()` will work in QuickJS scripts. The function-predicate form is broken.

The iterator methods (`filter`, `find`, etc.) already handle this correctly by pulling data into the VM via `all()` and running the callback inside the VM. `remove` with a function predicate likely needs similar treatment — either as an iterator-style evalCode wrapper, or by documenting that only `remove('Header-Name')` and `remove({key, value})` are supported in sandboxed scripts.

packages/bruno-js/src/headers-proxy.js
```diff
@@ -0,0 +1,109 @@
...
+
+  return new Proxy(propertyList, {
+    get(target, prop, receiver) {
+      // Symbol properties and prototype chain — always delegate to the target
+      if (typeof prop === 'symbol') {
+        return Reflect.get(target, prop, receiver);
+      }
+
+      // PropertyList methods and internal properties take precedence
+      if (prop in target) {
```
sanish-bruno: **Design**: The `prop in target` check gives PropertyList methods priority over raw header values for both `get` and `set` traps. This means:

- **Read**: `req.headers['get']` returns the PropertyList `.get()` method, not a header value named `get`
- **Write**: `req.headers['get'] = 'some-value'` overwrites the `.get()` method on the instance (via `target[prop] = value` in the set trap), corrupting the PropertyList

In the old code (`this.headers = req.headers` as a plain object), `headers['get']` would correctly return the header value. With this change, any header whose name collides with a PropertyList method (`get`, `has`, `all`, `one`, `count`, `idx`, `find`, `filter`, `each`, `map`, `reduce`, `toObject`, `toString`, `toJSON`, `add`, `upsert`, `remove`, `clear`, `populate`, etc.) cannot be accessed via bracket notation.

In practice HTTP headers with these names are extremely unlikely, but it's a subtle backward-compat break worth documenting or defending against (e.g., by checking `Object.prototype.hasOwnProperty.call(getRawHeaders(), prop)` first in the get trap when the prop also exists in rawHeaders).

packages/bruno-js/src/sandbox/quickjs/shims/bruno-request.js
sanish-bruno: **Suggestion**: Only `add`, `upsert`, `remove`, and `clear` are bridged to QuickJS via `syncWriteMethods`. The remaining write methods — `append`, `prepend`, `insert`, `insertAfter`, `populate`, `repopulate`, `assimilate` — won't be available in sandboxed scripts.

Since `append`/`prepend`/`insert`/`insertAfter` are all aliases for `add`, they're low-priority. But `populate` and `assimilate` offer distinct functionality (bulk replace / merge) that users might expect to work in scripts. Consider adding them to the bridge if the API surface should be consistent between Node and QuickJS contexts.

sanish-bruno (commented):
## PR Review

Found **3** issue(s): 1 minor bug, 1 design concern, 1 suggestion.

### Summary
- `remove()` with function predicates won't work in QuickJS sandbox because `vm.dump()` can't serialize function handles
- Proxy `get`/`set` traps give PropertyList methods priority over raw header keys, so headers with names like `get`, `has`, `all`, `count` etc. can't be read/written via bracket access (subtle backward-compat edge case)
- Several write method aliases (`append`, `prepend`, `populate`, `repopulate`, `assimilate`) are not bridged to the QuickJS sandbox

sanish-bruno:
@coderabbitai review

packages/bruno-js/tests/header-list.spec.js
```diff
@@ -0,0 +1,1051 @@
...
+      list.set({ key: 'Content-Type', value: 'text/html' });
+      expect(rawReq.headers['Content-Type']).toBe('text/html');
+      expect(list.get('Content-Type')).toBe('text/html');
+    });
+
+    test('replaces existing header with two-arg form', () => {
+      const { list, rawReq } = createReqHeaders();
+      list.set('Content-Type', 'text/html');
+      expect(rawReq.headers['Content-Type']).toBe('text/html');
+    });
```
sid-bruno, line 360: There might be a bug here but would be worth a test 

```js
 test('repopulate() does not leave re-added headers in __headersToDelete', () => {                                                                                                                          
    const { list, rawReq } = createReqHeaders({ 'Content-Type': 'application/json' });                                                                                                                       
    list.repopulate([{ key: 'Content-Type', value: 'text/plain' }]);                                                                                                                                         
    expect(rawReq.__headersToDelete).not.toContain('Content-Type');                                                                                                                                          
    expect(rawReq.headers['Content-Type']).toBe('text/plain');                                                                                                                                               
  });                                                                                             
```

sid-bruno: Another one to keep the current behaviour in check 

```js
test('set() with missing value sets header to undefined', () =>{
    const { list, rawReq } = createReqHeaders({});                                                                                                                                                           
    list.set({ key: 'X-Foo' });                                                                                                                                                                              
    expect(rawReq.headers['X-Foo']).toBeUndefined();
    expect(list.count()).toBe(1);
  });       
```

bijin-bruno (approved):

PR #7713, opened 2026-04-08
make post default gql method

ganesh-bruno:
- Make POST as a default method for gql req which user can change but while creating the default will be POST

https://github.com/user-attachments/assets/9af0b9af-a30f-4107-9749-1be4cbbd06da

## Summary by CodeRabbit

* **Bug Fixes**
  * GraphQL requests now default to POST instead of GET for new requests.
  * Switching a request to GraphQL applies the POST method by default to avoid method/type mismatch.
  * Changing the request type to HTTP or gRPC now forces the HTTP method back to GET to ensure compatibility.
  * Selecting a WebSocket request type sets the request method to WS automatically.

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
+++ b/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
@@ -261,6 +261,11 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
 
   const onSubmit = () => formik.handleSubmit();
 
+  const handleRequestTypeChange = (e) => {
+    formik.handleChange(e);
+    formik.setFieldValue('requestMethod', e.target.value === 'graphql-request' ? 'POST' : 'GET');
+  };
+
   const handlePaste = useCallback(
     (event) => {
       const clipboardData = event.clipboardData || window.clipboardData;
@@ -343,7 +348,7 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
                       name="requestType"
                       value="http-request"
                       checked={formik.values.requestType === 'http-request'}
-                      onChange={formik.handleChange}
+                      onChange={handleRequestTypeChange}
                       data-testid="http-request"
                     />
                     <label htmlFor="http-request" className="ml-1 cursor-pointer select-none">
@@ -357,10 +362,7 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
                       name="requestType"
                       value="graphql-request"
                       checked={formik.values.requestType === 'graphql-request'}
-                      onChange={(e) => {
-                        formik.handleChange(e);
-                        formik.setFieldValue('requestMethod', 'POST');
-                      }}
+                      onChange={handleRequestTypeChange}
                       data-testid="graphql-request"
                     />
                     <label htmlFor="graphql-request" className="ml-1 cursor-pointer select-none">
--- a/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
+++ b/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
@@ -395,7 +395,10 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
                       name="requestType"
                       value="ws-request"
                       checked={formik.values.requestType === 'ws-request'}
-                      onChange={formik.handleChange}
+                      onChange={(e) => {
+                        formik.handleChange(e);
+                        formik.setFieldValue('requestMethod', 'ws');
+                      }}
                       data-testid="ws-request"
                     />
                     <label htmlFor="ws-request" className="ml-1 cursor-pointer select-none">
--- a/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
+++ b/packages/bruno-app/src/components/Sidebar/NewRequest/index.js
@@ -261,11 +261,6 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
 
   const onSubmit = () => formik.handleSubmit();
 
-  const handleRequestTypeChange = (e) => {
-    formik.handleChange(e);
-    formik.setFieldValue('requestMethod', e.target.value === 'graphql-request' ? 'POST' : 'GET');
-  };
-
   const handlePaste = useCallback(
     (event) => {
       const clipboardData = event.clipboardData || window.clipboardData;
@@ -348,7 +343,10 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
                       name="requestType"
                       value="http-request"
                       checked={formik.values.requestType === 'http-request'}
-                      onChange={handleRequestTypeChange}
+                      onChange={(e) => {
+                        formik.handleChange(e);
+                        formik.setFieldValue('requestMethod', 'GET');
+                      }}
                       data-testid="http-request"
                     />
                     <label htmlFor="http-request" className="ml-1 cursor-pointer select-none">
@@ -362,7 +360,7 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
                       name="requestType"
                       value="graphql-request"
                       checked={formik.values.requestType === 'graphql-request'}
-                      onChange={handleRequestTypeChange}
+                      onChange={formik.handleChange}
                       data-testid="graphql-request"
                     />
                     <label htmlFor="graphql-request" className="ml-1 cursor-pointer select-none">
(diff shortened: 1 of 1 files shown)


packages/bruno-app/src/components/Sidebar/NewRequest/index.js
```diff
@@ -374,7 +377,10 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
                       name="requestType"
                       value="grpc-request"
                       checked={formik.values.requestType === 'grpc-request'}
-                      onChange={formik.handleChange}
+                      onChange={(e) => {
```
sanish-bruno, line 380: **Bug (Major)**: The `graphql-request` radio button (new line ~362) was not updated with the same pattern applied to HTTP/gRPC/WS. It still reads:

```js
onChange={formik.handleChange}
```

This means when a user starts with the default `http-request` type (method `GET`) and then clicks the GraphQL radio, `requestMethod` is not switched to `POST`. The resulting request is created as GraphQL-type with method `GET`, which is exactly the method/type mismatch this PR set out to prevent and contradicts the PR description ("Switching a request to GraphQL applies the POST method by default").

Note: the actual issue is on line 362 (outside any diff hunk) — attached to this added block for visibility.

Suggested fix (apply to the graphql radio on line ~362):

```js
onChange={(e) => {
  formik.handleChange(e);
  formik.setFieldValue('requestMethod', 'POST');
}}
```

packages/bruno-app/src/components/Sidebar/NewRequest/index.js
```diff
@@ -111,7 +111,7 @@ const NewRequest = ({ collectionUid, item, isEphemeral, onClose }) => {
       filename: '',
       requestType: getRequestType(collectionPresets),
       requestUrl: collectionPresets.requestUrl || '',
-      requestMethod: 'GET',
+      requestMethod: getRequestType(collectionPresets) === 'graphql-request' ? 'POST' : 'GET',
```
sanish-bruno, line 114: **Bug (Minor)**: This ternary only handles the `graphql-request` preset. If a collection has `presets.requestType === 'ws'`, `getRequestType` returns `'ws-request'`, but `requestMethod` is initialized to `'GET'`. That value is then passed through to `newWsRequest` in `onSubmit` (see line 176), producing a WebSocket request whose method is `GET` instead of `ws` — the same mismatch the new radio `onChange` handler for `ws-request` is trying to prevent when the user switches types manually.

Suggested fix:

```js
const defaultRequestType = getRequestType(collectionPresets);
const defaultRequestMethod =
  defaultRequestType === 'graphql-request' ? 'POST'
  : defaultRequestType === 'ws-request' ? 'ws'
  : 'GET';

// ...
initialValues: {
  // ...
  requestType: defaultRequestType,
  requestMethod: defaultRequestMethod,
  // ...
}
```

(`grpc-request` does not pass `requestMethod` to `newGrpcRequest`, so it does not need special handling, but hoisting `defaultRequestType` also removes the double call flagged in the nit below.)

sanish-bruno, line 114: **Nit**: `getRequestType(collectionPresets)` is now invoked twice in `initialValues` (once for `requestType`, once inside this ternary). Hoisting it to a local `const` just above the `useFormik` call keeps the two derived values in sync and avoids re-running the switch in `getRequestType` for no reason.

sanish-bruno (commented):
## PR Review

Found **3** issue(s): 1 major, 1 minor, 1 nit.

### Summary
- **Major**: The `graphql-request` radio button is missing the matching `onChange` handler, so switching from HTTP to GraphQL does not set the method to `POST` — this contradicts the PR description ("Switching a request to GraphQL applies the POST method by default").
- **Minor**: The initial `requestMethod` only accounts for the `graphql-request` preset; `ws-request` presets will incorrectly start with `GET` and that value is forwarded to `newWsRequest`.
- **Nit**: `getRequestType(collectionPresets)` is now called twice in `initialValues`.

sanish-bruno:
## Follow-up: `requestMethod` is only consumed by `http-request` and `graphql-request`

After tracing how `request.method` is actually used for each type, I want to refine my earlier review. Two of the three `onChange` handlers added in this PR (`grpc-request` → `'GET'`, `ws-request` → `'ws'`) are **dead writes** — nothing downstream reads the value for those types. The only handler that matters is the one for `http-request`, and the only omission that matters is the missing one for `graphql-request`.

### Evidence

**gRPC** — `request.method` is not part of the model:
- `newGrpcRequest` in `packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js:1759-1760` does not destructure `requestMethod`, and the constructed item's `request` object has no `method` field.
- All other callers (`CreateTransientRequest/index.js:162`, `CreateUntitledRequest/index.js:92`, `utils/collections/emptyStateRequest.js:40`) omit it.
- Display: `Sidebar/.../RequestMethod/index.js:11-14` hardcodes `'grpc'` from `item.type`.

**WebSocket** — `request.method` is written on creation, then ignored everywhere:
- `newWsRequest` (`actions.js:1911`) does set `request.method = requestMethod` on the in-memory item.
- But no serializer reads it:
  - `packages/bruno-filestore/src/formats/yml/items/stringifyWebsocketRequest.ts` never references `method`.
  - `parseWebsocketRequest.ts` never sets `method` on the returned Bruno item, so a ws request loaded from disk has `request.method === undefined`.
  - `packages/bruno-lang/v2/src/bruToJson.js:580-584` — the `ws { ... }` block grammar has no `method` field. Only `get`/`post`/etc. produce a method.
- No runtime reads it:
  - `packages/bruno-requests/src/ws/ws-client.js` — no `method` usage.
  - `packages/bruno-electron/src/ipc/network/ws-event-handlers.js` — no `method` usage.
  - The WS client dispatches purely off the URL scheme (`ws://` / `wss://`).
- Display: `RequestMethod/index.js:13` hardcodes `'ws'` from `item.type`.

**HTTP / GraphQL** — method *is* consumed:
- Serialization: `stringifyHttpRequest.ts:41`, `stringifyGraphQLRequest.ts:40` both write `method` (defaulting to `'GET'` / `'POST'`).
- Parsing: `parseHttpRequest.ts:21`, `parseGraphQLRequest.ts:20` both read `method`.
- Network: the HTTP runtime uses `request.method` to issue the outbound verb.

### Impact on the PR

| Radio | `onChange` sets method | Method is consumed for this type | Handler needed? |
|-------|------------------------|----------------------------------|-----------------|
| `http-request` | `'GET'` | yes (serialized + sent) | yes |
| `graphql-request` | *(missing)* | yes (serialized as `method: 'POST'`) | **yes — still the real bug** |
| `grpc-request` | `'GET'` | no (not in model, not serialized, not read) | no — dead write |
| `ws-request` | `'ws'` | no (stored on in-memory item only, never serialized, never read at runtime) | no — dead write |

So the grpc/ws handlers aren't wrong in behavior, but they imply those types need a method, which they don't. That uniform "every radio sets the method" pattern is probably part of why the GraphQL handler got missed during review — at a glance, all four radios look structurally similar.

### Retracting one piece of my earlier review

My inline comment on line 114 ("ws-request preset will incorrectly start with GET and that value is forwarded to `newWsRequest`") is technically correct that the value is forwarded, but the follow-on claim ("producing a WebSocket request whose method is `GET` instead of `ws`") is not actually observable — the stored method is never persisted or read for ws. So the user-facing behavior is unchanged either way. Please disregard that part of the comment. The graphql bug is still real.

### Suggested minimum-correctness patch

```js
const DEFAULT_METHOD_BY_TYPE = {
  'http-request': 'GET',
  'graphql-request': 'POST',
  // grpc-request and ws-request intentionally omitted:
  // request.method is not part of their model / not serialized / not read at runtime.
};

// initialValues
requestMethod: DEFAULT_METHOD_BY_TYPE[getRequestType(collectionPresets)] || 'GET',

// single handler, reused:
const handleTypeChange = (type) => (e) => {
  formik.handleChange(e);
  const nextMethod = DEFAULT_METHOD_BY_TYPE[type];
  if (nextMethod) formik.setFieldValue('requestMethod', nextMethod);
};

// usage:
onChange={handleTypeChange('http-request')}
onChange={handleTypeChange('graphql-request')}   // this is the currently-missing case
onChange={handleTypeChange('grpc-request')}      // no-op for method, keeps symmetry
onChange={handleTypeChange('ws-request')}        // no-op for method, keeps symmetry
```

That encodes the real invariant (only HTTP and GraphQL carry a method) in one place and makes the graphql omission a compile-time-obvious gap rather than a copy-paste oversight.

PR #7714, opened 2026-04-08
fix(converters): ensure insomnia variables are strings before matching

### LINKED ISSUE #7687: Insomnia converter does not serialize values cleanly

### This bug is:

- [ ] annoying

### Bruno version

3.2.0

### Operating System

Arch Linux 6.19.11

### Describe the bug

File packages/bruno-converters/src/insomnia/insomnia-to-bruno.js contains a following normalizing function:
```
const normalizeVariables = (value) => {
  value = value || '';
  const variables = value.match(regexVariable) || [];
  each(variables, (variable) => {
    value = value.replace(variable, variable.replace('_.', '').replaceAll(' ', ''));
  });
  return value;
};

```
**Issue**: Insomnia allows for unquoted values such as dates to exist in exported yaml, maybe some other datatypes are susceptible as well. || only catches null-ish values, while parsing everything else as JS datatypes it can infer. This results in error "t.match is not a function" when importing collection with such values due to calling match() on a non-string.
**Suggestion**: Convert defensively. I'm no JS man, but `value = value != null ? String(value) : '';` should be fine maybe?.. I was able to import successfully after manually quoting all such values in yaml.

_No response_

Image

jayesh70599:
### Description
Fixes #7687

#### The Problem
When importing Insomnia collections in YAML format, unquoted values (such as dates, numbers, or booleans) are parsed as non-string JavaScript types. Because the `normalizeVariables` function immediately calls `.match()` on these values, it triggers a `TypeError: value.match is not a function`, causing the entire import process to crash.

#### The Fix
I have updated the `normalizeVariables` function in `packages/bruno-converters/src/insomnia/insomnia-to-bruno.js` to explicitly cast incoming values to a `String`. 

- Used a nullish check (`value != null`) to ensure that `null` or `undefined` default to an empty string.
- Guaranteed that values like `0` (number) or `false` (boolean) are correctly converted to `"0"` and `"false"` rather than being treated as falsy and erased.

#### Verification
- Created a reproduction Insomnia YAML export containing unquoted integers and dates.
- Verified that the import fails with a `TypeError` in the console before the fix.
- Verified that the import succeeds and correctly populates headers and variables after the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed variable normalization during Insomnia to Bruno conversion to properly preserve falsy values such as `0` and `false`, instead of converting them to empty strings.

### DIFF
--- a/packages/bruno-converters/src/insomnia/insomnia-to-bruno.js
+++ b/packages/bruno-converters/src/insomnia/insomnia-to-bruno.js
@@ -34,7 +34,7 @@ const addSuffixToDuplicateName = (item, index, allItems) => {
 const regexVariable = new RegExp('{{.*?}}', 'g');
 
 const normalizeVariables = (value) => {
-  value = value || '';
+  value = value != null ? String(value) : '';
   const variables = value.match(regexVariable) || [];
   each(variables, (variable) => {
     value = value.replace(variable, variable.replace('_.', '').replaceAll(' ', ''));


prateek-bruno:
Hi @jayesh70599, Thanks for the PR. I checked and found out that this PR doesn't handle the values correctly.

For example:
If I have a value of "2024-01-01" for date in insomnia and import in Bruno after this PR, it shows up like:
`Mon Jan 01 2024 05:30:00 GMT+0530 (India Standard Time)`

Instead of `2024-01-01`

Screenshot for reference:

Screenshot 2026-04-24 at 3 57 00 PM

PR #7743, opened 2026-04-12
performance: for large sse, the ui becomes unresponsive

SuperALKALINEdroiD:
### Description
I was recently testing SSEs on Bruno, and the app crashed so badly that I had to force restart my laptop. This PR attempts to fix the issue by moving SSEs to a separate slice partition, preventing other slices from getting bloated.

#### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/WsResponsePane/WSMessagesList/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/WsResponsePane/WSMessagesList/index.js
@@ -1,4 +1,5 @@
-import React, { useState, useRef, useEffect, useCallback, memo } from 'react';
+import React, { useState, useRef, useEffect, useCallback, useMemo, memo } from 'react';
+import { hexy as hexdump } from 'hexy';
 import classnames from 'classnames';
 import StyledWrapper from './StyledWrapper';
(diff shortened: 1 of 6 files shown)


SuperALKALINEdroiD:
@lohit-bruno @bijin-bruno
@sid-bruno
@helloanoop
Can you guys review this

sid-bruno:
Haven't done a full review but preliminary checks , I'd say I like the changes around the components and the state partition. 

I'm not sure we want the flushing logic since that will add a 100ms delay to each received chunk and on a larger stream, any offset in receiving the data chunk might change the order, if we want to do the flush, I'd like a better flushing logic without the loss order and preferably additional tests for the order checks.

SuperALKALINEdroiD:
@sid-bruno 
I dug into this a bit more and I don't think ordering should be affected here.
The listener execution, buffer mutation and flush callback are all running on the same event loop, so push() ordering for a given stream key remains fixed. The timer only affects batch boundaries, not insertion order within the stream. So,  a chunk arriving slightly after the flush window would move into the next batch, but it shouldn't overtake previously received chunks or reorder dispatches for the same key.

Also the intent behind the 100 ms delay was to add some breathing room between the component re-renders. I would be open to ideas if you have some

SuperALKALINEdroiD:
@lohit-bruno
@bijin-bruno
@sid-bruno
@helloanoop
Can i get some actionable feedback  to get this merged ASAP

SuperALKALINEdroiD:
@sid-bruno hello!!

PR #7792, opened 2026-04-17
fix: don't re-encode already encoded characters

### LINKED ISSUE #6048: Content Type in Headers for File Upload to AWS through Pre Signed URL Wrong

### This bug is:

- [x] annoying

### Bruno version

1.38.1

### Operating System

Windows 11

### Describe the bug

I use a **pre-signed-url** which looks like this:

 `aws_url/bucket/file_path.csv?X-Amz-Algorithm=XXXX-Amz-Credential=XXXXX&X-Amz-Date=XXXX&X-Amz-Expires=3600&X-Amz-SignedHeaders=XXX&X-Amz-Signature=XXXX`

**Bug 1**: If I pass the URL with an env var, then my params do not get populated automatically. So I just pasted the URL without using env var "{{pre-signed-url}} and then I see my params. 

In my Body I am adding file/binary with `content-type text/csv `. 

**Bug 2**:  After sending the request I see this error message: 
`<Error>
    <Code>AuthorizationQueryParametersError</Code>
    <Message>Error parsing the X-Amz-Credential parameter; the Credential is mal-formed; expecting "&lt;YOUR-AKID&gt;/YYYYMMDD/REGION/SERVICE/aws4_request".</Message>`

I checked my settings for AWS and after some experiments I tested the same request with postman & curl. Both works. So I assume it must be Bruno or how I use Bruno. 

meta {
  name: upload-to-s3
  type: http
  seq: 2
}

put {
  url: {{upload_presigned_url}}
  body: file
  auth: none
}

body:file {
  file: @file(C:\path.csv) @contentType(text/csv; charset=utf-8)
}

settings {
  encodeUrl: true
  timeout: 0
}

On Postman sending the same request with same body type (binary + file in csv format) works.

Image

So I **compared the headers with postman headers**:

Image

And in the headers of Bruno I found **application/xml** which seems to be the only difference.

prateek-bruno:
### Description

We were re-encoding already encoded characters in the url, causing them to fail. For example aws pre-signed url's are already encoded.

For example:
```
https://example.com/api?filter=status%257Cactive&sort=name%257Casc
```

Now we safely ignore such pattern while encoding again, so that they don't cause an issue.

Fixes #6048

#### Contribution Checklist:

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/GenerateCodeItem/utils/snippet-generator.spec.js
@@ -1113,17 +1113,17 @@ describe('generateSnippet – encodeUrl setting', () => {
     expect(result).not.toContain('%253D');
   });
 
-  it('should double-encode pre-encoded %20 when encodeUrl is true', () => {
+  it('should preserve pre-encoded sequences when encodeUrl is true', () => {
     const preEncodedUrl = 'https://example.com/api?token=abc%20123%3D%3D&type=test';
     const item = {
       ...makeItem(preEncodedUrl, { encodeUrl: true }),
       rawUrl: preEncodedUrl
     };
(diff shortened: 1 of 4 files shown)


prateek-bruno:
Closing this for now.

PR #7824, opened 2026-04-22
fix(environments): persist selected env per-tab and resolve variables from viewed env

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2372)

Fixes a few related UX issues with environment settings:

1. Selected environment is lost on tab navigation. The environment being edited in the settings view was stored in local React state, so navigating away and back reset the selection to activeEnvironmentUid. Any unsaved draft on a different environment would effectively disappear from view. Now the selected environment is persisted in tab.tabState.envUid.
2. Hover tooltips in the Variables tab showed the wrong environment's values. When viewing "Local" while "Prod" was active in the dropdown, hovering {{baseUrl}} would show Prod's value. EnvironmentVariablesTable now sets _collection.activeEnvironmentUid = environment.uid on the synthetic collection so variable resolution matches the viewed environment.
3. Draft bleed-over when switching environments. setEnvironmentsDraft unconditionally wrote { environmentUid, variables }. If a draft already existed for a different environment, its environmentUid would get overwritten while stale fields could linger. Now it guards: if the existing draft is for the same environment, it only updates variables in place; otherwise it resets.

### Context (relation to EE Secrets Manager)
These fixes were surfaced while building the External Secrets tab in the Enterprise edition, where each environment has a second tab alongside Variables for configuring cloud-secret-manager accounts (Vault, AWS, Azure). That work added:

- An External Secrets tab next to Variables, with its own draft state (environmentsDraft.secrets)
- Reference copying like {{$secrets.<name>.<key>}} that users paste into requests
- Hover tooltips in secret path fields that need to resolve variables from the viewed env

While wiring this up, each of the three bugs above showed up as user-facing regressions:

- Losing the selected environment on tab navigation became very visible because users actively edit secret paths in one env while referencing another
- The wrong-env tooltip bug was noticed in the secret path SingleLineEditor but already existed in the variables table
- The draft bleed-over made it possible for a secrets draft on env A to silently drop state when a variables draft on env B was created

Since each of these is a pre-existing issue in OSS (independent of the secrets feature) and the fixes don't reference secrets in any way, they're being upstreamed here. The secrets-aware variants (preserving sibling secrets drafts, tab-specific "unsaved changes" messaging, etc.) stay in EE.

Test plan

1.  Select a non-active environment in collection settings, navigate to a request tab and back → selection preserved
2.  Same flow for global environments
3.  Hover {{var}} in the Variables value field of env A while env B is active → shows env A's resolved value
4.  Create a draft in env A, switch to env B → draft is not overwritten with env B's variables

#### Contribution Checklist:


.gitignore
```diff
@@ -53,6 +53,7 @@ bruno.iml
 .codex
 .agents
 .agent
+.gstack
```
sanish-bruno: this is intentional right?

packages/bruno-app/src/ui/MenuDropdown/SubMenuItem/index.js
```diff
@@ -3,6 +3,8 @@ import { IconChevronRight, IconChevronLeft } from '@tabler/icons';
 
 const SubMenuItem = ({
   item,
+  selectedItemId,
+  showTickMark,
```
sanish-bruno, line 7: please fallback to sane defaults, where ever we need.

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -820,7 +820,12 @@ export const collectionsSlice = createSlice({
       const { collectionUid, environmentUid, variables } = action.payload;
       const collection = findCollectionByUid(state.collections, collectionUid);
       if (collection) {
-        collection.environmentsDraft = { environmentUid, variables };
+        // Reset the draft when switching environments; otherwise update variables in place.
+        if (!collection.environmentsDraft || collection.environmentsDraft.environmentUid !== environmentUid) {
+          collection.environmentsDraft = { environmentUid, variables };
+        } else {
+          collection.environmentsDraft.variables = variables;
```
sanish-bruno, line 833: will we be adding `collection.environmentsDraft.secrets = secrets` fro EE, now that we are combining, just want to check if secrets draft, now scoped to collection will be handled? just question, haven't thought about whether we will be needing that, just a headup!

sanish-bruno (changes requested):

PR #7866, opened 2026-04-27
feat(api-spec): drag-to-resize split pane with persisted width

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2889)

Adds a draggable divider between the spec source (CodeMirror) and the rendered preview (SwaggerUI) in both the API Designer (`ApiSpecPanel`) and the read-only OpenAPI tab (`OpenAPISpecTab`). Width persists across opens — per-spec for the designer, per-tab for the OpenAPI tab. Double-click the divider to reset to the 50/50 default.

https://github.com/user-attachments/assets/1facabed-cd61-4401-aafc-c589e20f3adb

## What changed

- New `useDragResize` hook (`hooks/useDragResize`) drives the dragbar interaction, clamping, and `ResizeObserver`-based re-clamp on container size changes.
- Width persistence via two new actions: `updateApiSpecPanelLeftPaneWidth` (per-spec, in the `apiSpec` slice) and `updateApiSpecTabLeftPaneWidth` (per-tab, in the `tabs` slice). They're separate because the storage lifetimes differ —
  per-spec width survives close/reopen of the designer; per-tab width is bound to the tab.
- `SpecViewer` is now controlled by `leftPaneWidth` + `onLeftPaneWidthChange` props; the parent component owns where the width is stored.

## Bonus fixes pulled into this PR

1. **`OpenAPISpecTab` no longer re-fetches on unrelated re-renders.** Previously `loadSpec`'s `useCallback` deps included `collection.environments`, `collection.runtimeVariables`, and `collection.globalEnvironmentVariables` — all of which are rebuilt as fresh references on every `RequestTabPanel` render via `produce(...)`. The result: `setIsLoading(true)` would fire on every resize commit (and any other parent re-render), unmounting `SpecViewer` and causing both panes to flash. Env context now flows through a ref; `loadSpec` only changes when `pathname`, `uid`, or `sourceUrl` changes.
2. **Loading state for the Swagger preview pane.** Spinner overlays the right pane on first mount and after spec changes; unmounts after SwaggerUI fires `onComplete` (plus a 2-RAF tail so the operation list has time to paint).
3. **`Swagger` wrapped in `React.memo`** so the preview doesn't re-render on every parent state change during drag.

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Resizable split-pane for the API spec editor with drag handle, double-click reset, and per-tab/spec persisted left-pane width.
  * Preview generation is explicitly controlled and shows a “Generating preview…” overlay until the Swagger preview reports ready.

* **Style**
  * Improved dragbar styling, cursor, and selection behavior for smoother resize UX.

* **Tests**
  * Added comprehensive tests for drag-resize behavior.

* **Chores**
  * .gitignore updated to include generated package dist directories.

### DIFF
--- a/packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
+++ b/packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
@@ -1,26 +1,31 @@
-import React, { useState, useEffect, Suspense } from 'react';
+import React, { useState, useEffect, useRef, useCallback } from 'react';
 import get from 'lodash/get';
 import { useTheme } from 'providers/Theme';
 import { useSelector } from 'react-redux';
-import { IconDeviceFloppy } from '@tabler/icons';
+import { IconDeviceFloppy, IconLoader2 } from '@tabler/icons';
 import CodeEditor from './FileEditor/CodeEditor/index';
 import Swagger from './Renderers/Swagger';
+import { useDragResize } from 'hooks/useDragResize';
+
+const MIN_LEFT_PANE_WIDTH = 300;
+const MIN_RIGHT_PANE_WIDTH = 450;
 
 /**
  * Shared split-pane spec viewer: CodeEditor (left) + Swagger preview (right).
  *
  * Props:
- *  - content    (string)   The spec content (YAML/JSON string)
- *  - readOnly   (boolean)  If true, editor is not editable and save icon is hidden
- *  - onSave     (function) Called with current editor content on save (editable mode only)
+ *  - content               (string)  The spec content (YAML/JSON string)
+ *  - readOnly              (boolean) If true, editor is not editable and save icon is hidden
+ *  - onSave                (fn)      Called with current editor content on save (editable mode only)
+ *  - leftPaneWidth         (number|null) Persisted left pane width in px; null = use 50/50 default
+ *  - onLeftPaneWidthChange (fn)      Persist the new width (called on mouseup / double-click / resize-clamp)
  */
-const SpecViewer = ({ content, readOnly, onSave }) => {
+const SpecViewer = ({ content, readOnly, onSave, leftPaneWidth, onLeftPaneWidthChange }) => {
   const { displayedTheme, theme } = useTheme();
   const preferences = useSelector((state) => state.app.preferences);
 
   const [editorContent, setEditorContent] = useState(content);
 
-  // Sync editor when saved content changes from outside (e.g. after save completes)
   useEffect(() => {
     setEditorContent(content);
   }, [content]);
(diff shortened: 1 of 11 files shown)


packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
```diff
@@ -31,38 +36,81 @@ const SpecViewer = ({ content, readOnly, onSave }) => {
...
+  const leftPaneStyle = effectiveWidth != null
+    ? { width: `${effectiveWidth}px`, flexShrink: 0 }
+    : { flex: '1 1 50%', minWidth: 0 };
+
+  const [swaggerReady, setSwaggerReady] = useState(false);
+  useEffect(() => {
+    setSwaggerReady(false);
+  }, [content]);
+  const handleSwaggerComplete = useCallback(() => {
+    requestAnimationFrame(() => {
```
sanish-bruno, line 62: why are we calling `requestAnimationFrame` two times within the callback?

packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
```diff
@@ -31,38 +36,81 @@ const SpecViewer = ({ content, readOnly, onSave }) => {
...
+    minRight: MIN_RIGHT_PANE_WIDTH
+  });
+
+  const effectiveWidth = dragging ? dragWidth : leftPaneWidth;
+  const leftPaneStyle = effectiveWidth != null
+    ? { width: `${effectiveWidth}px`, flexShrink: 0 }
+    : { flex: '1 1 50%', minWidth: 0 };
+
+  const [swaggerReady, setSwaggerReady] = useState(false);
+  useEffect(() => {
```
sanish-bruno, line 55: can we keep some spacing between these

packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
```diff
@@ -31,38 +36,81 @@ const SpecViewer = ({ content, readOnly, onSave }) => {
...
+        )}
+      </div>
+      <div className="dragbar-wrapper" {...dragbarProps}>
+        <div className="dragbar-handle" />
+      </div>
+      <div
+        className="api-spec-right-pane relative"
+        style={{ flex: '1 1 50%', minWidth: 0 }}
+      >
+        <div style={{ visibility: swaggerReady ? 'visible' : 'hidden', height: '100%' }}>
```
sanish-bruno, line 104: Swagger was previously wrapped by a suspense, any reason we remove that. I think we can use the L103 - L113 section as the fallback and keep the suspense may be?

packages/bruno-app/src/components/OpenAPISpecTab/index.js
```diff
@@ -26,6 +39,16 @@ const OpenAPISpecTab = ({ collection }) => {
   const openApiSyncConfig = collection?.brunoConfig?.openapi?.[0];
   const sourceUrl = openApiSyncConfig?.sourceUrl;
 
+  const envContextRef = useRef({});
+  useEffect(() => {
```
sanish-bruno: keep spacing

packages/bruno-app/src/components/OpenAPISpecTab/index.js
```diff
@@ -26,6 +39,16 @@ const OpenAPISpecTab = ({ collection }) => {
...
+    envContextRef.current = {
```
sanish-bruno: the useEffect runs on every render as id doesn't have any deps array. Is it intensional. I guess we may not need the useEffect in that case.

packages/bruno-app/src/hooks/useDragResize/index.js
```diff
@@ -0,0 +1,105 @@
...
+  const onDoubleClick = useCallback(
+    (e) => {
+      e.preventDefault();
+      if (onWidthChange) onWidthChange(null);
+    },
+    [onWidthChange]
+  );
+
+  useEffect(() => {
+    document.addEventListener('mousemove', handleMouseMove);
```
sanish-bruno, line 80: I think we only need to attach this during dragging cause we are registering the listeners.

```js
  useEffect(() => {
    if (!dragging) return;
    document.addEventListener('mousemove', handleMouseMove);
    document.addEventListener('mouseup', handleMouseUp);
    return () => { ... };
  }, [dragging, handleMouseMove, handleMouseUp]);
```

packages/bruno-app/src/components/OpenAPISpecTab/index.js
```diff
@@ -26,6 +39,16 @@ const OpenAPISpecTab = ({ collection }) => {
   const openApiSyncConfig = collection?.brunoConfig?.openapi?.[0];
   const sourceUrl = openApiSyncConfig?.sourceUrl;
 
+  const envContextRef = useRef({});
```
sanish-bruno, line 44: We can just keep may not need a useEffect

```js
  envContextRef.current = {
    activeEnvironmentUid: collection?.activeEnvironmentUid,
    // ...
  };
```

packages/bruno-app/src/hooks/useDragResize/index.js
```diff
@@ -0,0 +1,105 @@
...
+      return Math.max(minLeft, Math.min(w, rect.width - minRight));
+    },
+    [containerRef, minLeft, minRight]
+  );
+
+  const handleMouseMove = useCallback(
+    (e) => {
+      if (!draggingRef.current || !containerRef.current) return;
+      e.preventDefault();
+      const rect = containerRef.current.getBoundingClientRect();
```
sanish-bruno: should we use the `clamp` function here instead of inlining clamping logic, as we already have it?

packages/bruno-app/src/hooks/useDragResize/index.js
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * Owns the transient drag state (cursor tracking, draft width, dragging flag)
+ * and clamps to [minLeft, container.width - minRight]. The persisted width and
+ * its setter are owned by the caller — pass them in as `width` / `onWidthChange`
+ * and the hook treats it as a controlled value.
+ *
+ * Render pattern in the consumer:
+ *   const effectiveWidth = dragging ? dragWidth : width;
+ */
+export function useDragResize({ containerRef, width, onWidthChange, minLeft, minRight }) {
```
sanish-bruno, line 14: Please add unit test cases to verify the behaviour of this hook

packages/bruno-app/src/hooks/useDragResize/index.js
```diff
@@ -0,0 +1,105 @@
...
+      document.removeEventListener('mouseup', handleMouseUp);
+    };
+  }, [handleMouseMove, handleMouseUp]);
+
+  // Re-clamp the persisted width when the container resizes (e.g. window or
+  // parent pane shrinks). Only dispatches if the clamped value differs.
+  useEffect(() => {
+    if (width == null || !containerRef.current) return;
+    const el = containerRef.current;
+    const ro = new ResizeObserver(() => {
```
sanish-bruno, line 97: observer removes an recreates the `resizeObserver` on every change in width, may be it is a bit wasteful, can we consider using ref for width?

```js
const widthRef = useRef(width);
widthRef.current = width;
 const hasWidth = width != null;                                                                                                                                                         
   
  useEffect(() => {                                                                                                                                                                       
    if (!hasWidth || !containerRef.current) return;                                                                                                                                     
    const ro = new ResizeObserver(() => {
      const clamped = clamp(widthRef.current);
      if (clamped !== widthRef.current && onWidthChange) {                                                                                                                                
        onWidthChange(clamped);
      }                                                                                                                                                                                   
    });                                                                                                                                                                                 
    ro.observe(containerRef.current);
    return () => ro.disconnect();
  }, [hasWidth, clamp, onWidthChange, containerRef]);
```

sanish-bruno (changes requested):

packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
abhishek-bruno: Without it, there's a visible flicker when switching from the "Generating preview…" loader to the Swagger view — the loader disappears a split second before Swagger actually shows up, so you see a blank flash in between. The double rAF waits until Swagger is actually painted on screen before hiding the loader, which makes the transition look clean.

packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
abhishek-bruno: done.

packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
abhishek-bruno: The <Suspense> was actually a no-op — it only triggers for lazy components or thrown Promises, and Swagger is neither. The fallback was never being shown. We need onComplete here because that's the only real signal for when Swagger finishes rendering.

packages/bruno-app/src/components/OpenAPISpecTab/index.js
abhishek-bruno: Fixed.

packages/bruno-app/src/hooks/useDragResize/index.js
abhishek-bruno: Good catch — switched to attaching listeners only while dragging. Also dropped draggingRef since the early-return guards aren't needed anymore.

packages/bruno-app/src/components/OpenAPISpecTab/index.js
abhishek-bruno: Addressed in the earlier comment

packages/bruno-app/src/hooks/useDragResize/index.js
abhishek-bruno: Good call — replaced the inline math with clamp(e.clientX - rect.left).

packages/bruno-app/src/hooks/useDragResize/index.js
abhishek-bruno: Added [useDragResize/index.spec.js](vscode-webview://1vh0csqg6mr5l1r47nmj9snjeot7js8gsupvm5i02urmbmrdc52e/packages/bruno-app/src/hooks/useDragResize/index.spec.js) with 8 tests covering initial state, onMouseDown seeding, mousemove clamping, mouseup commit, onDoubleClick reset, the no-op-when-not-dragging case, and ResizeObserver-driven re-clamp.

packages/bruno-app/src/hooks/useDragResize/index.js
abhishek-bruno: Switched to a widthRef. Observer now stays alive for the lifetime of "we have a width" instead of being rebuilt on every drag commit / width update.

bijin-bruno (approved):

PR #7867, opened 2026-04-27
fix: use stable index in requests tab in report

### LINKED ISSUE #7851: HTML reporter output shows wrong data when "Only Failed" filter is applied

### This bug is:

- [x] annoying

### Bruno version

3.3.0

### Operating System

Ubuntu 22.04

### Describe the bug

We have a fairly large test suite in Bruno (2000+ requests) and the HTML report output shows wrong data when using the "Only Failed" filter.
It shows the first N requests, regardless if they were successful or not as the failed requests, where N is the number of failed requests.

Attached are screenshots of the test result HTML content, with added console logging for the "Only Failed" filter to show that the filter does in fact filter in the correct data.
Some searches regarding the issue indicate a problem with how the data is rendered through the HTML, such as "Vue using the first N DOM nodes instead of remapping them to the correct data items".

_No response_

Main landing:
Image

Requests tab:
Image

Only Failed Requests:
Image

prateek-bruno:
### Description

[Jira](https://usebruno.atlassian.net/browse/BRU-3225)

Fixes #7851 

Previously we were using non-stable id's for elements, causing wrong title or data for the requests when "only failed" was toggled.

After this PR, each request will have a stable identifier, so item at index 0 will always have same index even after "Only failed" is toggled.

After:

https://github.com/user-attachments/assets/5a146a32-7430-437f-afab-ecabfb6e77d6

Before:

https://github.com/user-attachments/assets/60498498-d39d-4499-925e-535ba69f385b

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a centralized filter for request results used by HTML reports.

* **Bug Fixes**
  * When showing only failed results, original request indexes are preserved in the UI.

* **Refactor**
  * Report rendering now uses the centralized filter, simplifying result handling and component data flow.

* **Tests**
  * New unit test verifies filtered results preserve original indexes for failed entries.

### DIFF
--- a/packages/bruno-common/src/runner/reports/html/template.ts
+++ b/packages/bruno-common/src/runner/reports/html/template.ts
@@ -1,10 +1,6 @@
-type ReportRequestResult = {
-  status?: string;
-  testResults?: { status?: string }[];
-  assertionResults?: { status?: string }[];
-};
+import type { T_RunnerRequestExecutionResult } from '../../types';
 
-export const getFilteredRequestResults = (results: ReportRequestResult[] = [], onlyFailed = false) => {
+export const getFilteredRequestResults = (results: T_RunnerRequestExecutionResult[] = [], onlyFailed = false) => {
   const indexedResults = (Array.isArray(results) ? results : []).map((value, index) => ({ value, index }));
 
   if (!onlyFailed) {
--- a/packages/bruno-common/src/runner/reports/html/template.ts
+++ b/packages/bruno-common/src/runner/reports/html/template.ts
@@ -1,14 +1,10 @@
-type T_ReportCheckResult = {
+type ReportRequestResult = {
   status?: string;
+  testResults?: { status?: string }[];
+  assertionResults?: { status?: string }[];
 };
 
-type T_ReportRequestResult = {
-  status?: string;
-  testResults?: T_ReportCheckResult[];
-  assertionResults?: T_ReportCheckResult[];
-};
-
-export const getFilteredRequestResults = (results: T_ReportRequestResult[] = [], onlyFailed = false) => {
+export const getFilteredRequestResults = (results: ReportRequestResult[] = [], onlyFailed = false) => {
   const indexedResults = (Array.isArray(results) ? results : []).map((value, index) => ({ value, index }));
 
   if (!onlyFailed) {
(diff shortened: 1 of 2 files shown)


packages/bruno-common/src/runner/reports/html/template.ts
```diff
@@ -423,6 +440,8 @@ export const htmlTemplateString = (resutsJsonString: string) => `<!DOCTYPE html>
     <script>
       const { createApp, ref, computed, onMounted } = Vue;
 
+      const getFilteredRequestResults = ${getFilteredRequestResults.toString()};
```
sanish-bruno, line 443: @prateek-bruno is it possible to keep the function within the template string itself, toString() serialisation may become fragile if somebody changes the function to include imports or some closures.  

I get the point that we need to test it, but the test itself is testing just the function output, ideally we should be testing the view itself, where we will have to confirm whether the `only failed` toggle is rendering results accurately

sanish-bruno (changes requested):

packages/bruno-common/src/runner/reports/html/template.ts
sanish-bruno: Another option is to test the serialised function itself, thoughts on this @lohit-bruno @sid-bruno 

```js
import { getFilteredRequestResults } from './template';
  import vm from 'vm';

  it('serialized function works in a plain JS context', () => {
    const ctx = vm.createContext({});
    vm.runInContext(`var fn = ${getFilteredRequestResults.toString()}`, ctx);

    const result = vm.runInContext(`fn([
      { status: 'pass', testResults: [{ status: 'pass' }], assertionResults: [] },
      { status: 'pass', testResults: [{ status: 'fail' }], assertionResults: [] }
    ], true)`, ctx);

    expect(result).toEqual([{ value: expect.objectContaining({ status: 'pass' }), index: 1 }]);
  });
  ```

prateek-bruno: @sanish-bruno I've incorporated the suggested code.

sanish-bruno (approved):

sanish-bruno:
LGTM, may need to look in to why playwright test cases are failing @prateek-bruno

sid-bruno (approved):

PR #7876, opened 2026-04-28
Centralize OpenAPI sync state in Redux

abhishek-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2889)

https://github.com/user-attachments/assets/d066a641-8e59-4114-9016-8b0dc8c1abb1

## Problem

The OpenAPI Sync tab freezes for 1–3 seconds every time the user switches between a request tab and the OpenAPI Sync tab. Reproduces on any collection with OpenAPI sync configured.

## Root cause

`<RequestTabPanel key={activeTabUid} />` is keyed on the active tab uid, so React fully unmounts and remounts whatever the active tab renders on every switch. `useOpenAPISync` kept its drift data in local React state,
which was lost on unmount. On remount, a `useEffect` auto-fired `checkForUpdates()` — which calls two heavy IPC handlers (`renderer:compare-openapi-specs` + `renderer:get-collection-drift`) that block the Electron main thread.

Net effect: every tab switch repeated minutes of work the user just paid for. The "freeze" was the main thread blocking on the recompute.

## Fix

Lift drift state into the existing `openapiSync` redux slice so it survives unmount/remount. The auto-fire effect now skips when cached drift exists; manual refresh (Check button), settings change, item-count change, and tab close still trigger fresh fetches as before.

### What changed

1. **New `drift[uid]` redux slot** holds `{ specDrift, collectionDrift, remoteDrift, fetching, lastChecked, itemCountAtLastFetch }`. Survives unmount.
2. **Auto-fire gated** on `!drift?.specDrift && !drift?.fetching`. The `fetching` flag survives unmount too, blocking the rapid-tab-switch race where a second mount mid-fetch would otherwise duplicate the IPC.
3. **`storedSpec` split out from drift** into its own slot. Heavy spec (parsed OpenAPI object) is evicted on tab close; the lightweight `storedSpecMeta` (`title`, `version`, `endpointCount`) survives so the toolbar update badge keeps working when the OAPI Sync tab is closed.
4. **Tab-close eviction** wired into the existing `closeTabs` thunk in `collections/actions.js`. On close, drops `drift` + `storedSpec` + `tabUiState` for that collection. `collectionUpdates` and `storedSpecMeta` are preserved (background polling continues feeding the toolbar).
5. **Manual refresh keeps stale data visible** during reload. Sets `fetching: true` instead of clearing drift to `null` — no flash of empty state, success path overwrites.
6. **Stale-drift detection across closed-tab item changes.** Stamps `itemCountAtLastFetch` on each successful fetch; on reopen, if current `httpItemCount` differs from the cached count, drift refreshes automatically. Catches the case where the user adds/deletes endpoints via the sidebar while the OAPI tab was closed.
7. **Drops the dead `diff` field** from `setCollectionUpdate`. No consumers anywhere — `collectionUpdates[uid].diff` was duplicating what `drift[uid].specDrift` now holds. Saves redundant memory.
8. **Drops a hot tabs subscription.** `useSelector(state.tabs.tabs)` was re-rendering the hook on every tab open/close/focus/edit anywhere in the app, even though `tabs` was only used in callbacks. Swapped to `useStore()` (same pattern as `useIpcEvents.js`).

#### Contribution Checklist:

## Summary by CodeRabbit

* **Refactor**
  * OpenAPI sync now uses centralized per-collection state for specs and drift, improving consistency.
  * Tab interactions and sync checks are driven from central state, reducing duplicated local state and race conditions.
  * Closing sync-related tabs now triggers cleanup so stale sync data is removed.

* **Bug Fixes**
  * Drift detection and automatic reloads are more accurate when collection item counts change.

### DIFF
--- a/packages/bruno-app/src/components/OpenAPISyncTab/hooks/useOpenAPISync.js
+++ b/packages/bruno-app/src/components/OpenAPISyncTab/hooks/useOpenAPISync.js
@@ -1,9 +1,16 @@
 import { useState, useEffect, useMemo, useRef } from 'react';
-import { useDispatch, useSelector } from 'react-redux';
+import { useDispatch, useSelector, useStore } from 'react-redux';
 import toast from 'react-hot-toast';
-import { addTab, focusTab, closeTabs } from 'providers/ReduxStore/slices/tabs';
+import { addTab, focusTab } from 'providers/ReduxStore/slices/tabs';
+import { closeTabs } from 'providers/ReduxStore/slices/collections/actions';
 import { getDefaultRequestPaneTab } from 'utils/collections';
-import { clearCollectionState, setCollectionUpdate, setStoredSpecMeta } from 'providers/ReduxStore/slices/openapi-sync';
+import {
+  clearCollectionState,
+  setCollectionUpdate,
+  setStoredSpec,
+  setStoredSpecMeta,
+  setDrift
+} from 'providers/ReduxStore/slices/openapi-sync';
 import { fetchAndValidateApiSpecFromUrl } from 'utils/importers/common';
 import { isHttpUrl } from 'utils/url/index';
 import { flattenItems } from 'utils/collections/index';
@@ -19,19 +26,22 @@ const useOpenAPISync = (collection) => {
   const [isLoading, setIsLoading] = useState(false);
   const [error, setError] = useState(null);
   const [fileNotFound, setFileNotFound] = useState(false);
-  const [specDrift, setSpecDrift] = useState(null);
-  // Collection drift state
-  const [collectionDrift, setCollectionDrift] = useState(null);
-  const [remoteDrift, setRemoteDrift] = useState(null);
   const [isDriftLoading, setIsDriftLoading] = useState(false);
-  const [storedSpec, setStoredSpec] = useState(null);
 
-  const tabs = useSelector((state) => state.tabs.tabs);
+  const drift = useSelector((state) => state.openapiSync?.drift?.[collection.uid] || null);
+  const specDrift = drift?.specDrift || null;
+  const collectionDrift = drift?.collectionDrift || null;
+  const remoteDrift = drift?.remoteDrift || null;
+  const storedSpec = useSelector((state) => state.openapiSync?.storedSpec?.[collection.uid] || null);
+
+  const updateDrift = (patch) => dispatch(setDrift({ collectionUid: collection.uid, patch }));
+
+  const store = useStore();
 
   const isConfigured = !!openApiSyncConfig?.sourceUrl;
 
   const updateStoredSpec = (spec) => {
-    setStoredSpec(spec);
+    dispatch(setStoredSpec({ collectionUid: collection.uid, spec }));
     dispatch(setStoredSpecMeta({
       collectionUid: collection.uid,
       title: spec?.info?.title || null,
@@ -72,6 +82,7 @@ const useOpenAPISync = (collection) => {
   const openEndpointInTab = (endpointId) => {
     const itemUid = endpointUidMap[endpointId];
     if (!itemUid) return;
+    const tabs = store.getState().tabs?.tabs || [];
     const existingTab = tabs.find((t) => t.uid === itemUid);
     if (existingTab) {
       dispatch(focusTab({ uid: itemUid }));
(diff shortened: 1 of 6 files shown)


packages/bruno-app/src/components/OpenAPISyncTab/hooks/useOpenAPISync.js
```diff
@@ -19,19 +26,22 @@ const useOpenAPISync = (collection) => {
...
-  const tabs = useSelector((state) => state.tabs.tabs);
+  const drift = useSelector((state) => state.openapiSync?.drift?.[collection.uid] || null);
+  const specDrift = drift?.specDrift || null;
+  const collectionDrift = drift?.collectionDrift || null;
+  const remoteDrift = drift?.remoteDrift || null;
+  const storedSpec = useSelector((state) => state.openapiSync?.storedSpec?.[collection.uid] || null);
+
+  const updateDrift = (patch) => dispatch(setDrift({ collectionUid: collection.uid, patch }));
+
+  const store = useStore();
```
sanish-bruno, line 40: @abhishek-bruno is there any specific reason why we have used `useStore` instead of using `useSelector` ? we may need add a comment specifying the reason, so that future changes won't affect the performance regression

packages/bruno-app/src/providers/ReduxStore/slices/openapi-sync.js
```diff
@@ -36,6 +41,30 @@ export const openapiSyncSlice = createSlice({
       delete state.collectionUpdates[collectionUid];
       delete state.tabUiState[collectionUid];
       delete state.storedSpecMeta[collectionUid];
+      delete state.storedSpec[collectionUid];
+      delete state.drift[collectionUid];
+    },
+    setDrift: (state, action) => {
```
sanish-bruno, line 47: we do not have the idea on the functional signature of patch, may need to specify expected values, we may also need to prevent unintended keys getting into drift through patch

packages/bruno-app/src/components/OpenAPISyncTab/hooks/useOpenAPISync.js
sanish-bruno, line 126: Question: in loadCollectionDrift we are using `itemCountAtLastFetch` to understand whether or not we have more endpoints to review, shouldn't the check for updates should also be relying on that right to check if any new endpoints got added

packages/bruno-app/src/providers/ReduxStore/slices/openapi-sync.js
```diff
@@ -2,27 +2,32 @@ import { createSlice } from '@reduxjs/toolkit';
...
   // Last poll timestamp
   lastPollTime: null,
   // Map of collectionUid -> { activeTab, expandedSections, expandedRows }
   tabUiState: {},
-  // Map of collectionUid -> { title, version, endpointCount } (persists across tab navigations)
-  storedSpecMeta: {}
+  // Map of collectionUid -> { title, version, endpointCount }
+  storedSpecMeta: {},
+  // Map of collectionUid -> full parsed OpenAPI spec object
+  storedSpec: {},
```
sanish-bruno, line 18: We did not store specs in redux store before? Now that we are storing the spec, if we have multiple large spec, how will it perform?

sanish-bruno (changes requested):

packages/bruno-app/src/components/OpenAPISyncTab/hooks/useOpenAPISync.js
abhishek-bruno: Yes — useStore is intentional. tabs is only read inside imperative handlers (openEndpointInTab, handleDisconnect), never in render output. If we used useSelector, the hook would re-render every time any tab is opened/closed/focused/edited anywhere in the app, re-running the heavy flattenItems + endpointUidMap memos. useStore + store.getState() reads the current value at click-time without subscribing. Will add a code comment to make this explicit.

packages/bruno-app/src/providers/ReduxStore/slices/openapi-sync.js
abhishek-bruno: Memory is the same as before — the spec was already in useState inside the tab component, now it's in Redux with the same lifetime (cleared via clearOpenApiSyncTabState on tab close). No persistence layer involved, store is purely in-memory. useSelector returns stable references via Immer, so no extra re-renders.

packages/bruno-app/src/providers/ReduxStore/slices/openapi-sync.js
abhishek-bruno: Documented the expected fields inline on drift: in initialState. Runtime validation isn't needed here — only one hook (useOpenAPISync) calls setDrift, all callsites live in this PR, and a typo would surface immediately since the selector would read undefined and the UI would visibly break.

packages/bruno-app/src/components/OpenAPISyncTab/hooks/useOpenAPISync.js
abhishek-bruno: Currently loadCollectionDrift is sufficient for the staleness check:

Added endpoints are local-only by definition. They wouldn't appear in specDrift (no collection involved) or as anything actionable in remoteDrift beyond localOnly — info the user already has since they just created it.
Deleted endpoints are picked up in collectionDrift.missing, which loadCollectionDrift refreshes.

bijin-bruno (approved):

Issue #7568, opened 2026-03-25
App starts (dock icon, processes visible) but no window on macOS after 3.2.0


Nagato-Yuzuru:
**Describe the bug**

After updating to 3.2.0, Bruno starts — dock icon bounces, process is running — but no window ever appears. Restarting doesn't help.

**Environment**

- Bruno: 3.2.0
- macOS: Darwin 24.6.0 (Sequoia)
- Intel Silicon

**Steps to reproduce**

1. Update to 3.2.0
2. Launch Bruno
3. Dock icon appears, process visible in Activity Monitor — no window

**What I found digging into it**

Ran with `--remote-debugging-port=9222`. The debug endpoint (`localhost:9222/json/list`) returns `[]` no matter how long you wait — meaning no BrowserWindow ever gets created.

Process list: main process runs, GPU helper and network helper both spawn, but there's no Renderer process. Something is blocking the ready handler before it gets to `new BrowserWindow(...)`.

The issue is in `src/index.js`:

```js
app.on('ready', async () => {
  await initializeShellEnv(); // hangs here indefinitely
  ...
  mainWindow = new BrowserWindow({...}); // never reached
```

`initializeShellEnv()` (in `@usebruno/requests`) does a dynamic ESM import:

```js
const { shellEnv: e } = await import("shell-env");
return await e();
```

Inside the Electron asar context, that `import()` never resolves and never rejects. The catch block never fires. The whole ready handler just stalls. This was added in 3.2.0 (PR #7223/#7228).

**Workaround**

Patching `fetchShellEnv` with a timeout fixes it:

```js
const fetchShellEnv = async () => {
  if (process.platform === 'win32') return {};
  try {
    const result = await Promise.race([
      import("shell-env").then(({ shellEnv }) => shellEnv()),
      new Promise((_, reject) => setTimeout(() => reject(new Error('timeout')), 5000))
    ]);
    return result;
  } catch {
    return {};
  }
};
```

Nagato-Yuzuru:
Downgrading to 3.14 should resolve this issue.

pkfox:
@Nagato-Yuzuru Downgrading is hardly a fix is it

Nagato-Yuzuru:
> [@Nagato-Yuzuru](https://github.com/Nagato-Yuzuru) Downgrading is hardly a fix is it

In my case, downgrading to 3.14 can work.

anusree-bruno:
@Nagato-Yuzuru @pkfox 

[...]

PR #8159, opened 2026-06-01
fix(system proxy): app needs restart to fetch system proxy when set from terminal

### LINKED ISSUE #7559: System proxy changes require app restart to take effect

### This bug is:

- [x] annoying

### Bruno version

3.2.0

### Operating System

Mac OS

### Describe the bug

When using system proxy variables, Bruno does not pick up changes dynamically. Every time proxy variables are set or modified, the application needs to be restarted for the changes to take effect.

This behavior is inconvenient, especially for users who frequently switch or update proxy configurations during development.

### Steps to reproduce

1. Set system proxy variables 

```bash
export HTTP_PROXY="http://127.0.0.1:8080"
export HTTPS_PROXY="http://127.0.0.1:8080"
```
2. Open Bruno and configure system proxy. 

3.  It won't update the values automatically

```bash
export HTTP_PROXY="http://127.0.0.1:9090"
export HTTPS_PROXY="http://127.0.0.1:9090"
```

4. Click the refresh button inside Bruno and it won't work.

5. Restart the app, and Bruno will update the values.

### Expected Behavior

- Bruno should automatically detect and apply updated system proxy variables

- The refresh action should reload and reflect the latest proxy configuration

_No response_

Image

abhishekp-bruno:
JIRA Link - https://usebruno.atlassian.net/browse/BRU-3202

Fixes: https://github.com/usebruno/bruno/issues/7559

### Description

When users configured proxy settings via shell environment variables (e.g. `http_proxy`, `https_proxy` in `.zshrc`/`.bashrc`), Bruno would only pick them up at startup. Changing or removing those exports in the terminal had no effect until the app was fully restarted — the cached `process.env` proxy values went
stale, and the "refresh" path never re-read the shell.

### Fix

Added `refreshShellEnvProxyVars()` in `bruno-requests`, which re-syncs proxy-related variables from the user's login shell on demand:

- **Clears stale proxy vars first** (`delete process.env[key]`) before spawning the login shell, so a removed export in `.zshrc` doesn't persie subprocess.
- Re-fetches the shell environment and re-applies any proxy values that are still set.
- Covers the full set of proxy keys: `http_proxy`/`HTTP_PROXY`, `https_proxy`/`HTTPS_PROXY`, `no_proxy`/`NO_PROXY`, `all_proxy`/`ALL_PROXY`.

Wired this into the Electron system-proxy store: `fetchSystemProxy({ refresh: true })` now passes `refreshShellEnv: true` down to `loadSystemP re-reads the terminal environment rather than serving cached values.

### Changes

- `packages/bruno-requests/src/utils/shell-env.ts` — new `refreshShellEnvProxyVars()` + exported `PROXY_ENV_KEYS`.
- `packages/bruno-requests/src/index.ts` — export the new function.
- `packages/bruno-electron/src/store/system-proxy.js` — refresh shell env on proxy refresh.

### Testing

Added unit tests covering the refresh behavior (re-syncing updated vars and clearing removed vars).

#### Contribution Checklist:

## Summary by CodeRabbit

* **Improvements**
  * System proxy reloads can now refresh shell environment proxy variables first, ensuring proxy settings stay in sync across the app.
  * Reload behavior now aligns with the existing “refresh” toggle, updating cached proxy values consistently.

* **Public API**
  * Exposed a capability to refresh shell-environment proxy variables programmatically.
  * Added a shared list of supported proxy environment variable keys.

* **Tests**
  * Added Jest coverage to verify proxy env var updates, removal when absent, and cleanup of stale lowercase/uppercase proxy variables.


package-lock.json
```diff
@@ -4949,7 +4949,7 @@
       "version": "7.26.3",
       "resolved": "https://registry.npmjs.org/@babel/helper-create-regexp-features-plugin/-/helper-create-regexp-features-plugin-7.26.3.tgz",
       "integrity": "sha512-G7ZRb40uUgdKOQqPLjfD12ZmGA54PzqDFUv2BKImnC9QIfGhIHKvVML0oN8IUiDq4iRqpq74ABpvOaerfWdong==",
-      "devOptional": true,
```
sanish-bruno: rm these changes

packages/bruno-requests/src/utils/shell-env.ts
```diff
@@ -35,3 +45,30 @@ export const initializeShellEnv = async (): Promise<Record<string, string>> => {
...
+  }
+
+  const shellEnvVars = await fetchShellEnv();
+
+  for (const key of PROXY_ENV_KEYS) {
+    const value = shellEnvVars[key];
+    if (value) {
+      process.env[key] = value;
+    } else {
+      delete process.env[key];
```
sanish-bruno: the delete seems to be redundant, we are already deleting them above,if block will be enough!

packages/bruno-requests/src/utils/shell-env.ts
```diff
@@ -35,3 +45,30 @@ export const initializeShellEnv = async (): Promise<Record<string, string>> => {
...
   return shellEnvVars;
 };
+
+/**
+ * Re-syncs proxy-related process.env values from the user's shell configuration.
+ * Used when refreshing system proxy settings without restarting the app.
+ *
+ * @returns The fetched shell environment variables
+ */
+export const refreshShellEnvProxyVars = async (): Promise<Record<string, string>> => {
```
sanish-bruno, line 86: should we have early return for windows platforms

```js
if (process.platform === 'win32') {
    return {};
  }
```

sanish-bruno (changes requested):

packages/bruno-requests/src/utils/shell-env.ts
abhishekp-bruno: REMOVED

sanish-bruno (approved):

packages/bruno-electron/src/store/system-proxy.js
```diff
@@ -1,10 +1,13 @@
...
+const { getSystemProxy, refreshShellEnvProxyVars } = require('@usebruno/requests');
 
 let cachedSystemProxy;
 let systemProxyPromise;
 
-const loadSystemProxy = async () => {
+const loadSystemProxy = async ({ refreshShellEnv = false } = {}) => {
   try {
+    if (refreshShellEnv) {
+      await refreshShellEnvProxyVars();
```
bijin-bruno, line 9: Have we made sure this is not blocking for the systems that doesn't have access to shellEnv? 
Ref: 
1. code -> `packages/bruno-electron/src/store/shell-env-state.js`
2. prev issue -> https://github.com/usebruno/bruno/issues/7568

bijin-bruno (changes requested):

packages/bruno-electron/src/store/system-proxy.js
abhishekp-bruno: Have added a timeout to refresh to handle this issue

lohit-bruno (dismissed):

packages/bruno-requests/src/utils/shell-env.spec.ts
```diff
@@ -92,3 +92,54 @@ describe('initializeShellEnv', () => {
...
+      http_proxy: 'http://new-proxy:8080',
+      https_proxy: 'http://new-proxy:8443'
+    };
+
+    await refreshShellEnvProxyVars();
+
+    expect(process.env.http_proxy).toBe('http://new-proxy:8080');
+    expect(process.env.https_proxy).toBe('http://new-proxy:8443');
+  });
+
```
helloanoop: The tests appear to be achieveing the same thing
* `should remove proxy env vars missing from shell config`
* `should not keep stale process.env proxy vars after shell config is cleared`

both tests set `mockShellEnvResult = {};` and the check the process env vars to be undefined, seems to be a bit redundant

packages/bruno-requests/src/utils/shell-env.ts
```diff
@@ -6,8 +6,18 @@
...
+  'https_proxy',
+  'HTTPS_PROXY',
+  'no_proxy',
+  'NO_PROXY',
+  'all_proxy',
+  'ALL_PROXY'
+] as const;
+
 const fetchShellEnv = async (): Promise<Record<string, string>> => {
-  // Windows handles environment variables differently - skip
```
helloanoop, line 10: why remove this useful comment ? its the very least saying that the reason we are returning empty object is because `Windows handles environment variables differently` although I wish more context was commented

abhishekp-bruno: My bad I actually tried to explore handling proxy for windows and was working on it but then reverted the changes since it was taking too much time and also it was out of scope for this ticket. 
will add the comment back

packages/bruno-requests/src/utils/shell-env.spec.ts
```diff
@@ -1,5 +1,5 @@
-import path from 'path';
 import { initializeShellEnv } from './shell-env';
+import path from 'path';
```
sanish-bruno, line 1: rm this, i think don't need to reorder, keep the such imports always at the top, like library imports, followed by imports from local files

packages/bruno-requests/src/utils/shell-env-proxy-refresh.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+describe('refreshShellEnvProxyVars — shell inheritance', () => {
+  beforeEach(() => {
+    proxyVarsSeenByShell = {};
+    mockShellEnvResult = {};
+    for (const key of PROXY_ENV_KEYS) {
+      delete process.env[key];
+    }
+  });
+
+  afterEach(() => {
```
sanish-bruno: do we need to do the delete in both afterEach and beforeEach? , i think we can drop afteEach

packages/bruno-requests/src/network/system-proxy/get-system-proxy.spec.ts
```diff
@@ -0,0 +1,407 @@
...
+ * To avoid running platform-restricted commands natively (which would crash tests run
+ * on mismatched host platforms or CI/CD pipelines), this file uses a dynamic sandbox runtime via
+ * `loadForPlatform(platform)`.
+ * * Instead of rigid invocation-sequence queues, platform CLI utilities are mocked using argument-aware
+ * routing logic (`execFile.mockImplementation`). This guarantees that structural changes to the internal
+ * execution sequence inside the proxy resolver will not cause brittle test breaks, keeping tests
+ * strictly tied to structural string contract parsing and logical compliance.
+ */
+
+const PROXY_ENV_KEYS = [
```
sanish-bruno: @abhishekp-bruno we have the option to split this up, may be even separate PR that handles cross-platform integration tests, better to add them to `network/system-proxy/index.spec.js`, can consider doing it.

and trim `refresh.spec.ts` to just the two cases that actually exercise `efreshShellEnvProxyVars` end-to-end (macOS env-wins-over-system, and env-cleared-after-refresh).

Also should we add a no-op test for `refreshShellEnvProxyVars` to assert the windows behaviour?

sanish-bruno (changes requested):

packages/bruno-requests/src/utils/shell-env-proxy-refresh.spec.ts
abhishekp-bruno: removed afterEach

packages/bruno-requests/src/utils/shell-env.spec.ts
abhishekp-bruno: This happened while resolving conflicts have updated this

packages/bruno-requests/src/utils/shell-env.ts
sanish-bruno: Add the early return cause we unconditioanlly deletes all the process.envs without considering whether they are from the shell or not, the refreshShellEnv can also be no-op in windows, just like fetchShellEnv. to avoid any possible regressions due to deletion @abhishekp-bruno

packages/bruno-requests/src/utils/shell-env.ts
```diff
@@ -39,3 +50,36 @@ export const initializeShellEnv = async (): Promise<Record<string, string>> => {
...
+  // would strip process.env proxy vars (registry-propagated, launcher-set, or inherited
+  // from the parent shell) without any means to restore them. Preserve the additive-only
+  // invariant that initializeShellEnv maintains on Windows.
+  if (process.platform === 'win32') {
+    return {};
+  }
+
+  // Clear stale proxy vars first so shell-env does not inherit them into the
+  // login shell subprocess (removed .zshrc exports would otherwise persist).
+  for (const key of PROXY_ENV_KEYS) {
```
sanish-bruno, line 97: Since we are actually deleting all the PROXY_ENVs here, then calling the `fetchShellEnv`, if any error the happens in `fetchShellEnv`, as it returns `{}` within the catch, refresh will remove all the existing values. Edge case, but something to be mindful about.  @abhishekp-bruno cc: @vijayh-bruno 

something like this can work, may also need to update the failing test accordingly

```js
  // New helper that surfaces failure. Keep fetchShellEnv as-is for initializeShellEnv's additive path.
  const tryFetchShellEnv = async (): Promise<Record<string, string> | null> => {
    if (process.platform === 'win32') return {};
    try {
      const { shellEnv } = await import('shell-env');
      return await shellEnv();
    } catch {
      return null;
    }
  };

  export const refreshShellEnvProxyVars = async (): Promise<Record<string, string>> => {
    if (process.platform === 'win32') return {};

    const snapshot: Record<string, string | undefined> = {};
    for (const key of PROXY_ENV_KEYS) {
      snapshot[key] = process.env[key];
      delete process.env[key];
    }

    const shellEnvVars = await tryFetchShellEnv();

    if (shellEnvVars === null) {
      // Subprocess failed — restore prior values rather than leave the user unproxied.
      for (const key of PROXY_ENV_KEYS) {
        if (snapshot[key] !== undefined) process.env[key] = snapshot[key];
      }
      return {};
    }

    for (const key of PROXY_ENV_KEYS) {
      const value = shellEnvVars[key];
      if (value) process.env[key] = value;
    }
    return shellEnvVars;
  };

```

sanish-bruno (changes requested):

sanish-bruno (approved):

vijayh-bruno (dismissed):

packages/bruno-requests/src/utils/shell-env.ts
sid-bruno: @abhishekp-bruno this is still valid in a situation where the race function causes this one to just be left in the timeout, we'll end up with a partial snapshot which isn't something we want. 

Needs to be addressed.

sid-bruno (changes requested):

packages/bruno-requests/src/utils/shell-env.ts
abhishekp-bruno: We are skipping windows here because everything related to windows proxy settings is handled by the system proxy resolver i.e getSystemProxy()

packages/bruno-requests/src/utils/shell-env.ts
abhishekp-bruno: I have updated the code to handle the race condition inside the function itself

sid-bruno: ~~Timeout shouldn't be in bruno-requests, it shouldn't be holding business and rate limiting in here but should be in the app / cli layer only. you can just move the delete logic to be run after fetching has finished instead.~~

The rest looks good, let me discuss with @bijin-bruno and get back

bijin-bruno:
Closed as merged via https://github.com/usebruno/bruno/pull/8535

PR #8168, opened 2026-06-02
fix(headers): case-insensitive header lookup

abhishekp-bruno:
JIRA Link - https://usebruno.atlassian.net/browse/BRU-2763

### Description

HTTP header names are case-insensitive per spec, but Bruno was treating them as case-sensitive internally. As a result, scripts using `req.getHeader('Content-Type')` could return `undefined` when the header was stored as `content-type` (or any other casing), and headers set with different casings could end up duplicated (e.g. both `Authorization` and `authorization`).

This PR makes header handling case-insensitive across the request lifecycle:

**Lowercase normalization at the source**
All header keys are now normalized to lowercase as they are built — when preparing requests, interpolating variables, merging collection/folder headers, and applying auth (Bearer, Basic, WSSE, API key), cookie, and OAuth2 headers. This keeps a single consistent internal convention and prevents duplicate headers across casings.

**Case-insensitive script APIs**
- `req.getHeader()` / `res.getHeader()` now resolve a header regardless of its casing, and return `null` (instead of `undefined`) when the header is missing.
- `req.setHeader()` reuses an existing header's key if one already exists in a different casing, otherwise stores it lowercased — avoiding accidental duplicates.
- `req.deleteHeader()` removes the matching header regardless of casing.
- `req.setHeaders()` normalizes all keys to lowercase.
- `getAuthMode()` and `hasJSONContentType()` now use case-insensitive lookups.
- Added null/type guards so these APIs no longer throw on missing headers or non-string names.

### Testing

Expanded the `getHeader` scripting collection tests to cover case-insensitive lookups (including `Content-Type` and auth-token variants) and to confirm a `null` result for missing headers.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request/response header retrieval for case-insensitive lookups, safer handling when headers are missing, and consistent `null` results when a header isn’t present.
  * Standardized HTTP header casing to lowercase across request building, including merged headers, authentication (Bearer/WSSE), and cookie handling for consistent behavior.

* **Tests**
  * Expanded coverage to confirm case-insensitive `getHeader()` behavior (including `Content-Type` and auth-token variants) and correct `null` results for missing headers.


packages/bruno-js/src/bruno-request.js
```diff
@@ -140,7 +140,12 @@ class BrunoRequest {
   }
 
   getHeader(name) {
-    return this.req.headers[name];
+    if (!this.req?.headers || typeof name !== 'string') {
+      return null;
+    }
+    const headers = this.req.headers;
```
sanish-bruno: check if we can have cased names for headers within requests, as they are getting added by the user, unlike the case with response headers.

packages/bruno-tests/collection/scripting/api/req/getHeader.bru
```diff
@@ -12,6 +12,7 @@ get {
 
 headers {
   bruno: is-awesome
+  Content-Type: application/json
```
sanish-bruno, line 15: we may need to normalize, addition/retrieval of headers to use lowercase values, it may show diff for users who who already have cased values in their bru files, or for those who add the headers manually, through the UI, I guess this may need to be discussed.

sanish-bruno (changes requested):

packages/bruno-electron/src/utils/collection.js
```diff
@@ -37,26 +33,18 @@ const mergeHeaders = (collection, request, requestTreePath, options = {}) => {
       let _headers = get(folderRoot, 'request.headers', []);
       _headers.forEach((header) => {
         if (header.enabled) {
-          if (header.name.toLowerCase() === 'content-type') {
-            headers.set('content-type', header.value);
-          } else {
-            headers.set(header.name, header.value);
-          }
+          headers.set(header.name.toLowerCase(), header.value);
```
sanish-bruno, line 36: check if name is always a string, so that it doesn't throw any error while accessing .toLowerCase(

sanish-bruno (changes requested):

sanish-bruno:
@abhishekp-bruno check why are the CI tests are failing, please modify tests if needed

packages/bruno-cli/src/runner/interpolate-vars.js
```diff
@@ -65,7 +65,8 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
 
   forOwn(request.headers, (value, key) => {
     delete request.headers[key];
-    request.headers[_interpolate(key)] = _interpolate(value);
+    // keep header keys lowercased internally so script/getHeader lookups stay case-insensitive
+    request.headers[_interpolate(key)?.toLowerCase()] = _interpolate(value);
```
helloanoop, line 69: The ask is just to fix `res.getHeader()`
Normalising request headers so that we can do a case insensitive search seems like a bad idea. Have we though about the effects of this messing up the request itself ? And existing workflow's breaking ?

PR #8169, opened 2026-06-02
fix(cli): use path name for classname in JUnit reports instead of request URL

sharan-bruno:
**Jira link: https://usebruno.atlassian.net/browse/BRU-3123**

### Description

The CLI's JUnit reporter used the request URL as the classname on each <testcase>. Since URLs change per environment/region, the same test produced duplicate entries in CI tools (Xray, Azure DevOps) that key tests by classname + name.

This changes classname to use the request name (matching Postman), so the same test maps to a single test case across all environments.

packages/bruno-cli/src/reporters/junit.js — classname now uses result.name instead of result.request.url
Added test coverage asserting classname is the request name
#### Contribution Checklist:

After fix:
image

## Summary by CodeRabbit

* **Bug Fixes**
  * JUnit test reports now derive each testcase's classname from the test result path (file/path-derived name) instead of the request URL, improving report clarity.

* **Tests**
  * Added a unit test to assert classname derivation and updated JUnit test snapshots across platforms to reflect the new classname values.

### DIFF
--- a/packages/bruno-cli/src/reporters/junit.js
+++ b/packages/bruno-cli/src/reporters/junit.js
@@ -15,7 +15,7 @@ const makeJUnitOutput = async (results, outputPath) => {
     const testCount = result.testResults ? result.testResults.length : 0;
     const postResponseTestCount = result.postResponseTestResults ? result.postResponseTestResults.length : 0;
     const totalTests = assertionTestCount + preRequestTestCount + testCount + postResponseTestCount;
-    const classname = result.name;
+    const classname = result.path;
 
     const suite = {
       '@name': result.name,
--- a/packages/bruno-cli/src/reporters/junit.js
+++ b/packages/bruno-cli/src/reporters/junit.js
@@ -15,6 +15,7 @@ const makeJUnitOutput = async (results, outputPath) => {
     const testCount = result.testResults ? result.testResults.length : 0;
     const postResponseTestCount = result.postResponseTestResults ? result.postResponseTestResults.length : 0;
     const totalTests = assertionTestCount + preRequestTestCount + testCount + postResponseTestCount;
+    const classname = result.name;
 
     const suite = {
       '@name': result.name,
@@ -34,7 +35,7 @@ const makeJUnitOutput = async (results, outputPath) => {
       const testcase = {
         '@name': `${assertion.lhsExpr} ${assertion.rhsExpr}`,
         '@status': assertion.status,
-        '@classname': result.request.url,
+        '@classname': classname,
         '@time': (result.runDuration / totalTests).toFixed(3)
       };
 
@@ -52,7 +53,7 @@ const makeJUnitOutput = async (results, outputPath) => {
       const testcase = {
         '@name': test.description,
         '@status': test.status,
-        '@classname': result.request.url,
+        '@classname': classname,
         '@time': (result.runDuration / totalTests).toFixed(3)
       };
 
@@ -70,7 +71,7 @@ const makeJUnitOutput = async (results, outputPath) => {
       const testcase = {
         '@name': test.description,
         '@status': test.status,
-        '@classname': result.request.url,
+        '@classname': classname,
         '@time': (result.runDuration / totalTests).toFixed(3)
       };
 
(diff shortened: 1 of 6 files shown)


packages/bruno-cli/src/reporters/junit.js
```diff
@@ -15,6 +15,7 @@ const makeJUnitOutput = async (results, outputPath) => {
     const testCount = result.testResults ? result.testResults.length : 0;
     const postResponseTestCount = result.postResponseTestResults ? result.postResponseTestResults.length : 0;
     const totalTests = assertionTestCount + preRequestTestCount + testCount + postResponseTestCount;
+    const classname = result.name;
```
sanish-bruno: have we decided on what should be the classname? path or the name?

sanish-bruno (changes requested):

packages/bruno-cli/src/reporters/junit.js
```diff
@@ -34,7 +35,7 @@ const makeJUnitOutput = async (results, outputPath) => {
       const testcase = {
         '@name': `${assertion.lhsExpr} ${assertion.rhsExpr}`,
         '@status': assertion.status,
-        '@classname': result.request.url,
+        '@classname': classname,
```
sanish-bruno, line 39: we need to update the snapshots fro junit playwright tests we have

sanish-bruno (changes requested):

packages/bruno-cli/src/reporters/junit.js
sharan-bruno: updated it to path

packages/bruno-cli/src/reporters/junit.js
sharan-bruno: updated it

tests/runner/collection-run-report/collection-run-report.spec.ts
```diff
@@ -13,8 +13,9 @@ function normalizeJunitReport(xmlContent: string): string {
     .replace(/time="[^"]*"/g, 'time="0.100"')
     // Replace file paths with normalized path
     .replace(/file="[^"]*[\\/][^"]*"/g, 'file="/mock/path/to/file.bru"')
-    // Replace test paths with normalized path
-    .replace(/classname="[^"]*[\\/][^"]*"/g, 'classname="/test/path/collection"');
+    // classname is the request's relative path within the collection (stable across
+    // environments); only normalize the path separator so snapshots match on all platforms
+    .replace(/classname="([^"]*)"/g, (_match, value) => `classname="${value.replace(/\\/g, '/')}"`);
```
sanish-bruno: check if this regex works fine! it seems like it is different from the prev version of it `/classname="[^"]*[\\/][^"]*"/g`

sanish-bruno (changes requested):

tests/runner/collection-run-report/collection-run-report.spec.ts
sharan-bruno: Verified — works correctly, the change is intentional. classname is stripExtension(result.path), the stable relative request path. The old regex only matched paths containing a separator and swapped in a mock; the new one keeps the real path and just normalizes \→/ so Windows matches the snapshot. Confirmed all three platform snapshots store classname="api/v1/users" with forward slashes, so it lines up. Quote-bounded [^"]* + g flag handle the multiple/edge cases fine.

bijin-bruno: In this case, i think we should try to keep the realistic value for file as well for consistency. 
Can we avoid this completely? Windows report will be validated only in a win environment I believe. Should not create any issues with path normalization.

bijin-bruno (changes requested):

tests/runner/collection-run-report/collection-run-report.spec.ts
sharan-bruno: update classname in JUnit report to use backslashes for Windows compatibility

bijin-bruno (approved):

sanish-bruno (approved):

bijin-bruno (approved):

PR #8172, opened 2026-06-03
fix(env):override global environment variables

abhishekp-bruno:
JIRA Link - https://usebruno.atlassian.net/browse/BRU-3253

## Summary

This PR adds support for updating **global environment variables** when running collections with `bru run`.

Previously, only **local environment variables** could be updated from the CLI. With this change, you can now update  global environment variables similar to local environment variables 

---

## Usage

```bash
# Update a local environment variable
bru run --env <local-env-name> --env-var <key>=<value>
# Example
bru run --env dev --env-var token=abc123

# Update a global environment variable
bru run --global-env <global-env-name> --global-env-var <key>=<value>
# Example
bru run --global-env global --global-env-var apiKey=xyz789

# Update both local and global environment variables
bru run \
  --env <local-env-name> \
  --env-var <local-key>=<local-value> \
  --global-env <global-env-name> \
  --global-env-var <global-key>=<global-value>

# Example
bru run \
  --env dev \
  --env-var token=abc123 \
  --global-env global \
  --global-env-var apiKey=xyz789
```
---

## Changes

- Added `--global-env-var` to update variables in the specified global environment.
- Allows updating both local and global environment variables in the same `bru run` invocation.

#### Contribution Checklist:

## Release Notes

* **New Features**
  * Added `--global-env-var` CLI option (repeatable) to override individual variables in the global environment when `--global-env` is set. Validates that `--global-env` is present and parses `name=value` pairs correctly.


packages/bruno-cli/src/commands/run.js
```diff
@@ -498,6 +499,9 @@ const handler = async function (argv) {
             process.exit(constants.EXIT_STATUS.ERROR_INCORRECT_ENV_OVERRIDE);
           }
           envVars[match[1]] = match[2];
+          if (globalEnv) {
```
sanish-bruno: do not think we should be doing this, we shouldn't be overriding, i do not feel this is the right solution, we shouldn't be adding to globalEnvVars object, same goes for envVars, this seems like a mistake, including adding them to envVars.

sanish-bruno (changes requested):

PR #8212, opened 2026-06-09
fix(SSE-text/event-stream):  sse response body is empty in res.getBody() for app and cli

sachin-thakur-bruno:
### Description

Currently users are not able to get SSE response body in scripts. This is needed by some users so fixing this by collecting the responses and adding them to response body so that it is available in post scripts.

[JIRA](https://usebruno.atlassian.net/browse/BRU-3465)

## Bug

The desktop app always requests responses as a stream
(`responseType: 'stream'`). For non-streaming responses it drains the stream
into a full buffer via `promisifyStream` before running post-response logic.

For SSE responses, the single-request path deliberately **skips** that drain so
chunks can be forwarded to the renderer live, and hard-codes the body to empty:

The streamed chunks were forwarded to the renderer for the live view and then
discarded — nothing ever reconstituted them into response.data. So when
post-response scripts, assertions, and tests ran on the stream's close event,
they read an empty body.

## Fix
Accumulate the streamed chunks (in the existing renderer-forwarding data
listener — the single stream consumer) and, on stream close, rebuild
response.data / response.dataBuffer from them before running
post-response scripts. This mirrors what the Collection Runner already does,
without adding a second stream listener (which would change flowing-mode timing
and risk dropping early chunks from the live UI).

All changes are in packages/bruno-electron/src/ipc/network/index.js:

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Streaming responses now buffer and rebuild streamed data on close, ensuring complete response content and reliable post-processing for server-sent events.
* **New Features**
  * Streaming response payloads expose buffered SSE chunks for inspection by consumers.
* **Tests**
  * Added automated test and server route validating a finite SSE stream's headers, message order, and content.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -73,6 +73,12 @@ const hasStreamHeaders = (headers) => {
   return headerSplit.indexOf('text/event-stream') > -1;
 };
 
+const buildResponseBodyFromStreamChunks = (sseChunks, headers, disableParsingResponseJson) => {
+  const dataBuffer = Buffer.concat(sseChunks);
+  const { data } = parseDataFromResponse({ data: dataBuffer, headers }, disableParsingResponseJson);
+  return { data, dataBuffer };
+};
+
 const promisifyStream = async (stream, abortController, closeOnFirst) => {
   const chunks = [];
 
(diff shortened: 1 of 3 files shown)


packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -1254,6 +1272,7 @@ const registerNetworkIpc = (mainWindow) => {
         headers: response.headers,
         data: response.data,
         stream: isResponseStream ? axiosDataStream : null,
+        sseChunks: isResponseStream ? sseChunks : null,
```
sid-bruno, line 1279: Don't think we need to send this anywhere at least for now, without this the above and below should work

sid-bruno: My bad, needed for the round trip of the response

sid-bruno (approved):

sid-bruno (approved):

PR #8219, opened 2026-06-10
fix(perf):quickjs-memory-leak

abhishekp-bruno:
JIRA - https://usebruno.atlassian.net/browse/BRU-3509

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QuickJS sandbox execution to more reliably manage the lifetime of VM contexts and temporary values, including tracked heap handles and pending runtime jobs. This reduces the risk of resource leaks and improves stability for both synchronous and asynchronous script runs.

* **Tests**
  * Added Jest coverage for QuickJS VM lifecycle, handle tracking/disposal behavior, retained-result evaluation, and correct cleanup for successful and failing sync/async executions.


packages/bruno-js/src/sandbox/quickjs/index.js
```diff
@@ -56,9 +56,10 @@ const executeQuickJsVm = ({ script: externalScript, context: externalContext, sc
 
     externalScript = removeQuotes(externalScript);
   }
-
+  let managedQuickJsContext;
   try {
-    const vm = QuickJSModule.newContext();
+    managedQuickJsContext = createManagedQuickJsContext(QuickJSModule);
+    const vm = managedQuickJsContext.vm;
```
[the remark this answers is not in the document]
abhishekp-bruno: let QuickJSModule;
const loader = memoizePromiseFactory(() => newQuickJSWASMModule());
loader().then((mod) => (QuickJSModule = mod));

I think this  code line 18-20 handles that can you check and confirm ?

packages/bruno-js/src/sandbox/quickjs/utils/index.js
```diff
@@ -1,3 +1,84 @@
...
+  if (!vm?.alive) {
+    return;
+  }
+
+  if (typeof disposeTracked === 'function') {
+    disposeTracked();
+  }
+
+  if (vm.runtime?.executePendingJobs) {
+    for (let i = 0; i < 10; i++) {
```
chirag-bruno: What is the significance of looping until i === 9?
Is that an arbitrary number?

scripts/memory-profile.js
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env node
```
chirag-bruno: We can probably remove the profiling from the actual merge. You can probably run the benchmark locally and post it in the PR conversation thread. This would be useful if we could run this on every future PR, which is not the case here.

packages/bruno-js/src/sandbox/quickjs/utils/index.js
```diff
@@ -1,3 +1,84 @@
...
+  }
+
+  return () => {
+    for (let i = handles.length - 1; i >= 0; i--) {
+      if (handles[i]?.alive) {
+        handles[i].dispose();
+      }
+    }
+  };
+};
```
chirag-bruno, line 98: nit (readability): the `vm[method] = (...args) => {}` reassignment is hard to parse at a glance. Pulling out a small `track` helper and naming the wrap (plus `forEach` instead of the index loops) makes the intent self-describing — behavior is unchanged and existing shims still get tracked for free:

```suggestion
const trackQuickJsContext = (vm) => {
  const handles = [];

  const track = (handle) => (handles.push(handle), handle);

  // Replace an allocator with a wrapper that records every handle it returns,
  // so teardown can dispose them all. Behaviour is otherwise identical.
  const trackAllocations = (method) => {
    const original = vm[method]?.bind(vm);
    if (!original) {
      return;
    }

    vm[method] = (...args) => track(original(...args));
  };

  ['newObject', 'newFunction', 'newArray'].forEach(trackAllocations);

  // Dispose newest-first: later handles may reference earlier ones.
  return () => handles.reverse().forEach((handle) => handle?.alive && handle.dispose());
};
```

scripts/memory-profile.js
abhishekp-bruno: will remove this, it was for testing while development

packages/bruno-js/src/sandbox/quickjs/utils/index.js
abhishekp-bruno: Executing a job can schedule more jobs (chained `.then()`s) 
have updated this to handle it properly

packages/bruno-js/src/sandbox/quickjs/utils/index.js
```diff
@@ -1,3 +1,87 @@
...
+  const trackAllocations = (method) => {
+    const original = vm[method]?.bind(vm);
+    if (!original) {
+      return;
+    }
+
+    vm[method] = (...args) => track(original(...args));
+  };
+
+  ['newObject', 'newFunction', 'newArray'].forEach(trackAllocations);
```
sid-bruno, line 88: any reason we're leaving out numbers and strings or other primitives?

packages/bruno-js/src/sandbox/quickjs/index.js
```diff
@@ -126,17 +131,20 @@ const executeQuickJsVmAsync = async ({ script: externalScript, context: external
 
     const script = wrapScriptInClosure(externalScript, SANDBOX.QUICKJS);
 
-    const result = vm.evalCode(script, scriptPath);
+    const result = vm.evalCodeRetained(script, scriptPath);
     const promiseHandle = vm.unwrapResult(result);
     const resolvedResult = await vm.resolvePromise(promiseHandle);
     promiseHandle.dispose();
```
sid-bruno, line 137: If this throws, the `promiseHandle` is never disposed, is that okay?

sid-bruno (changes requested):

packages/bruno-js/src/sandbox/quickjs/utils/index.js
abhishekp-bruno: We intentionally only track the allocators that produce owned handles which the shims retain for the lifetime of the context — newObject, newArray, newFunction. Those build the persistent sandbox surface (bru, req, res, etc.), stay alive until teardown, and would otherwise leak.

- undefined / null / true / false aren't disposable at all. They're StaticLifetime constants whose dispose() is a literal no-op ([lifetime.ts](https://github.com/justjake/quickjs-emscripten/blob/main/packages/quickjs-emscripten-core/src/lifetime.ts)), so there's nothing to track.
- newNumber / newString / newBigInt do return owned handles that must be disposed, but they're consumed inline — created, set as a property, and disposed at the call site (or via a Scope). They never become part of the long-lived shim surface, so tracking them would just bloat the array with thousands of already-dead handles and risk masking a genuine local-disposal bug rather than fixing one.

**So the inclusion criterion is "owned + retained by the shim layer," not "is it a handle." **

packages/bruno-js/src/sandbox/quickjs/index.js
abhishekp-bruno: The finally block should handle it if the promise throws any error

sid-bruno: my bad, we're throwing the error and `finally` handles the `dispose` for the entire context

sid-bruno (approved):

sid-bruno (dismissed):

bijin-bruno (approved):

sid-bruno (approved):

lohit-bruno (approved):

PR #8222, opened 2026-06-10
feat: faster collection mount via file cache (Beta)

chirag-bruno:
## Summary

[Logic JIRA](https://usebruno.atlassian.net/browse/BRU-3258)
[UI JIRA](https://usebruno.atlassian.net/browse/BRU-2407)

Adds an opt-in **File cache** for collection mounting. When enabled, opening a collection reads its parsed tree from a local SQLite snapshot instead of re-walking and re-parsing every file on disk. Live edits during the session continue to flow through the existing `collection-watcher`, which now also writes parsed files back into the snapshot so the next launch loads instantly.

This is gated behind a toggle in **Preferences → Cache → File cache** (shown with a `Beta` badge). It's off by default; existing behavior is unchanged.

## How to try it

1. Open **Preferences → Cache**.
2. Toggle **File cache** on.
3. Open / reopen a collection.

The first open populates the cache (same speed as today). Subsequent opens of that collection — including after restarting the app — load from cache.

A **Clear cache** button next to the toggle wipes the snapshot DB (`VACUUM`s after delete so the on-disk file actually shrinks).

## How it works

The mount path splits into two phases:

```
trigger mount  →  diff vs git-lite cache  →  parse changes via workers
               →  build tree              →  hand off to collection-watcher
               →  push tree + ui state to the renderer
```

- **`services/mount/manager.js`** owns the cold-start lifecycle.
- **`services/mount/gitlite.js`** is the SQLite-backed snapshot store. Status diffs walk the disk and compare mtime → hash → row, returning `{ added, updated, removed }`.
- **`@usebruno/pool`** (new package, foundational commits 1–2) provides the worker pool used to parse changed files in parallel.
- **`@usebruno/storage`** (new package, foundational commits 3–4) wraps `node:sqlite` with migrations + transactions.
- **`services/mount/tree-builder.js`** assembles the renderer-shape tree, hydrating per-item uids deterministically (`sha256(absPath#section#index)`).
- After cold start, the existing `collection-watcher.js` takes over for live changes. A small hook (`stageToCache` / `unstageFromCache`) stages parsed files back into the snapshot when a collection is registered. Unregistered collections are byte-for-byte unaffected.

The renderer:
- Routes mount/unmount through new v2 IPC channels (`renderer:mount-collection-v2`, `renderer:unmount-collection-v2`).
- Consumes parallel outgoing channels (`main:collection-tree-loaded`, `main:bruno-config-update-v2`, `main:collection-loading-state-updated-v2`, `main:hydrate-app-with-ui-state-snapshot-v2`).
- Adds a `collectionLoadedFromTree` reducer that merges the tree blob by uid — overwrites file-derived fields, preserves runtime fields like `response`, `requestSent`, `cancelTokenUid`, and conditionally clears `draft` only when it matches what's on disk.

Strict isolation: when the flag is off, none of the new code path runs.

## File-by-file overview

**Foundational packages**
- `packages/bruno-pool` — `worker_threads` job pool with a `ParseFile` job type
- `packages/bruno-storage` — `node:sqlite` wrapper with migrations + transactions

**Backend (`bruno-electron`)**
- `services/mount/manager.js` — `MountManager` lifecycle (mount, unmount, shutdown, cache size, clear)
- `services/mount/gitlite.js` — snapshot store (status, stage, entries, clear, transaction)
- `services/mount/tree-builder.js` — renderer-shape tree + deterministic uid hydration
- `utils/mount.js` — shared classify + uid helpers
- `ipc/mount.js` — v2 IPC channels
- `app/collection-watcher.js` — optional `snapshotIndex` hook (no-op when absent)
- `store/preferences.js` — `cache.file.enabled` + `isFileCacheEnabled()` helper
- `index.js` — register the IPC; shut it down in the deferred-quit IIFE

**Frontend (`bruno-app`)**
- `components/Preferences/Cache/` — File cache + SSL session cache redesigned as card-style rows with toggle + Clear cache action
- `providers/App/useIpcEvents.js` — listeners for the four v2 channels
- `providers/ReduxStore/slices/collections/actions.js` — mountCollection routes to v2 when the flag is on; removeCollection fires v2 unmount alongside the existing remove
- `providers/ReduxStore/slices/collections/index.js` — `collectionLoadedFromTree` reducer (uid-keyed merge, `isTransient` annotation under `tempDirectories[uid]`)
- `providers/ReduxStore/slices/workspaces/actions.js` — scratch collection picks v2 channel when the flag is on
- `providers/ReduxStore/middlewares/tasks/middleware.js` — `OPEN_REQUEST` tasks also fire on the v2 tree event

## Notes for reviewers

- Old code is preserved exactly: `collection-watcher.js` changes are additive (a small write-through hook + the `ignoreInitial`/`snapshotIndex` option in `addWatcher`). With the flag off, behavior is unchanged.
- Stable uids: per-child uids in the tree are now `sha256(absPath#section#index).slice(0,21)` so anything that pins to them survives a restart. Matches Bruno's existing `uidSchema` (21 alphanumeric chars).
- Cache size shown in the UI is `fs.statSync(dbPath).size`. After `Clear cache`, the DB runs `VACUUM` so the displayed size actually drops to ~0.
- Tests were intentionally left out of this PR to keep it focused on the feature. A follow-up will add unit + integration coverage.

## Test plan

- [ ] Toggle file cache on → open a real collection → verify items load and edits propagate
- [ ] Restart the app → reopen the same collection → confirm cache is reused (faster cold start)
- [ ] Modify a file outside the app → confirm the live watcher picks it up and the tree updates
- [ ] Save a request that has a visible response → response should remain after the save
- [ ] Create a new transient request → it appears in the tree and opens in a tab
- [ ] Rename a file → uid migrates so any open tab/draft survives
- [ ] Clear cache → cache size drops to ~0; next open repopulates
- [ ] Toggle file cache off → fall back to the existing chokidar flow with no surprises
- [ ] Scratch workspace collection → routes through v2 when the flag is on, v1 when off

## Release Notes

* **New Features**
  * Cache preferences now provide independent toggles for **File cache (Beta)** and **SSL session cache**.
  * **File cache (Beta)** displays the current cache size and adds a **Clear cache** action.
  * **SSL session cache** adds a separate **Clear cache** action.

* **Enhancements**
  * Improved cached collection mounting and collection-tree loading to reduce delays and better handle queued open requests.
  * Refined cache-related preferences visuals (including the File cache beta badge).

* **Tests**
  * Updated Playwright coverage for the revised cache toggle controls.


packages/bruno-app/src/components/Preferences/Cache/index.js
```diff
@@ -1,120 +1,152 @@
...
-    enabled: Yup.boolean()
-  })
-});
+const humanSize = (bytes) => {
+  if (bytes === null || bytes === undefined) return '—';
+  if (bytes < 1024) return `${bytes} B`;
+  if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(2)} KB`;
+  if (bytes < 1024 * 1024 * 1024) return `${(bytes / (1024 * 1024)).toFixed(2)} MB`;
+  return `${(bytes / (1024 * 1024 * 1024)).toFixed(2)} GB`;
+};
```
sid-bruno: can be moved to app utils

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -2389,6 +2389,7 @@ export const removeCollection = (collectionUid) => (dispatch, getState) => {
       }
     }
 
+    ipcRenderer.invoke('renderer:unmount-collection-v2', { collectionUid }).catch(() => {});
```
sid-bruno:
```suggestion

```

packages/bruno-electron/src/services/mount/gitlite.js
```diff
@@ -0,0 +1,179 @@
...
+    const stored = this.#loadStored(root);
+    const denylist = resolveDenylist(options.denylist);
+    const added = [];
+    const updated = [];
+    const removed = [];
+    const seen = new Set();
+
+    for (const { relativePath, absolutePath } of walk(root, denylist)) {
+      seen.add(relativePath);
+      const stat = fs.statSync(absolutePath, { bigint: true });
```
sid-bruno: see if this and the callers of `status` are always async, if yes, change this to `fs.promises.stat`

packages/bruno-electron/src/services/mount/gitlite.js
```diff
@@ -0,0 +1,179 @@
...
+
+      if (!prior) {
+        const hash = hashFile(absolutePath);
+        added.push({ relativePath, absolutePath, mtime, hash });
+        continue;
+      }
+
+      if (prior.mtime === mtime) continue;
+
+      const hash = hashFile(absolutePath);
```
sid-bruno: same, if the callers of `status` are async, then I'd prefer an async hasher instead

packages/bruno-electron/src/services/mount/gitlite.js
```diff
@@ -0,0 +1,179 @@
...
+        hash TEXT NOT NULL,
+        data TEXT NOT NULL,
+        PRIMARY KEY (collection_path, relative_path)
+      ) WITHOUT ROWID;
+      CREATE INDEX IF NOT EXISTS idx_collection_path ON snapshot_entries(collection_path);
+    `
+  }
+];
+
+class GitLite {
```
sid-bruno:
```suggestion
class Cache {
```

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -0,0 +1,179 @@
...
+  resolveDenylist,
+  isDenied,
+  walk
+} = require('../../utils/mount');
+
+const MIGRATIONS = [
+  {
+    version: 1,
+    up: `
+      CREATE TABLE IF NOT EXISTS snapshot_entries (
```
sid-bruno:
```suggestion
      CREATE TABLE IF NOT EXISTS cache_entries (
```

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -0,0 +1,179 @@
...
+        collection_path TEXT NOT NULL,
+        relative_path TEXT NOT NULL,
+        id TEXT NOT NULL,
+        mtime INTEGER NOT NULL,
+        hash TEXT NOT NULL,
+        data TEXT NOT NULL,
+        PRIMARY KEY (collection_path, relative_path)
+      ) WITHOUT ROWID;
+      CREATE INDEX IF NOT EXISTS idx_collection_path ON snapshot_entries(collection_path);
```
sid-bruno:
```suggestion
      CREATE INDEX IF NOT EXISTS idx_collection_path ON cache_entries(collection_path);
```

packages/bruno-electron/src/services/mount/file-index.js
sid-bruno, line 1: Can the file be changed to `cache.js` instead?

packages/bruno-electron/src/services/mount/manager.js
```diff
@@ -0,0 +1,243 @@
...
+
+    if (this.#mounts.has(collectionUid)) {
+      // renderer reload — pull fresh state from cache and re-emit
+      const existing = this.#mounts.get(collectionUid);
+      existing.win = win;
+      existing.emit = emit;
+      existing.brunoConfig = brunoConfig || existing.brunoConfig;
+      existing.state = this.#getIndex().entries(existing.collectionPath);
+      await this.#emitTree(collectionUid, existing);
+      existing.emit.uiState(computeUiStateSnapshot(existing.collectionPath));
```
sid-bruno: would remove this from here, will collide with existing snapshot

packages/bruno-electron/src/services/mount/manager.js
```diff
@@ -0,0 +1,243 @@
...
+        pathname: collectionPath,
+        environmentPath: snapshotState?.environment?.collection || '',
+        selectedEnvironment: snapshotState?.selectedEnvironment || ''
+      };
+    }
+  } catch (err) {
+    console.error('[mount] snapshot lookup failed', err);
+  }
+  return null;
+};
```
sid-bruno: would remove this from here, will collide with existing snapshot

packages/bruno-electron/src/utils/mount.js
```diff
@@ -0,0 +1,90 @@
...
+const path = require('node:path');
+const crypto = require('node:crypto');
+
+const DENY_DIRS = new Set(['node_modules', '.git', '.svn', '.hg', '.bruno']);
+const DEFAULT_DENYLIST = ['**/.DS_Store', '**/Thumbs.db'];
+
+const sha256 = (input) => crypto.createHash('sha256').update(input).digest('hex');
+const hashFile = (absPath) => sha256(fs.readFileSync(absPath));
+const normalize = (p) => path.resolve(p);
+const toPosix = (p) => (path.sep === '/' ? p : p.split(path.sep).join('/'));
```
sid-bruno: we have a posixfy util already, can be reused here

sid-bruno (changes requested):
Would move the pool and storage into the electron implementation, not sure we want another set of packages to manage this when it's mostly going to be tied to the electron implementation

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -0,0 +1,188 @@
...
+    const root = normalize(collectionPath);
+    const stored = this.#loadStored(root);
+    const denylist = resolveDenylist(options.denylist);
+    const added = [];
+    const updated = [];
+    const removed = [];
+    const seen = new Set();
+
+    const files = walk(root, denylist);
+    const results = await Promise.all(files.map(async ({ relativePath, absolutePath }) => {
```
sachin-thakur-bruno, line 56: Just a thought.
Promise.all() here launches work for every file concurrently which may results in excessive memory usage. 
Is it worth processing files with a concurrency limit instead?

chirag-bruno: So the idea behind `Promise.all()` here is to let the thread context switch when it is waiting on the stat syscall, and the DB call. Since both of them are IO bound, blocking the thread while we wait can become expensive. Having said that, we can evaluate using a worker pool for this job. Can be considered.

sid-bruno (approved):

PR #8253, opened 2026-06-12
fix(electron): Install packages dialog reports npm not on PATH when Node.js and npm are installed

utkarsh-bruno:
### Description

https://usebruno.atlassian.net/browse/BRU-3531

Postman package install showed “npm was not found on your PATH” even when Node/npm worked in the terminal..
Bruno’s main process couldn’t reliably find the user’s Node/npm at install time.

Apps launched from Finder/Dock get a minimal PATH, without NVM/FNM/Homebrew paths that the terminal has.

Wrong spawn target — Spawning npm / npm.cmd directly failed on Windows after CVE-2024-27980 (EINVAL for .cmd without shell).

The fix resolves the user’s system Node, runs npm-cli.js via node (no shell), prepends shell PATH from login config, and waits for shell-env before installing.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Installer waits for the shell environment before installing and returns clearer, specific error codes for install failures.

* **Improvements**
  * More reliable, cross-platform npm invocation via Node.
  * Shell PATH is now merged/prepended into the process environment.
  * Client-side package name validation before install.

* **Tests**
  * Expanded tests for resolution, spawn options, env merging, timeouts, output truncation, and error mappings.


packages/bruno-electron/src/utils/install-packages.js
```diff
@@ -16,13 +24,81 @@ const appendCapped = (buffer, chunk, cap) => {
...
+
+  return null;
+};
+
+const resolveNpmCli = (nodePath) => {
+  const nodeDir = path.dirname(nodePath);
+  const candidates = [
+    path.join(nodeDir, 'node_modules', 'npm', 'bin', 'npm-cli.js'),
+    path.join(nodeDir, '..', 'lib', 'node_modules', 'npm', 'bin', 'npm-cli.js')
+  ];
```
sid-bruno, line 51: For *nix systems I can see it right beside the `node` binary itself, so should be faster to just look it closer to node binary first assuming nvm would have the same setup for both linux and windows, though worth checking if it needs to `npm.cmd` (you can self resolve otherwise)

```suggestion
  const candidates = [
    path.join(nodeDir, 'npm'),
    path.join(nodeDir, 'node_modules', 'npm', 'bin', 'npm-cli.js'),
    path.join(nodeDir, '..', 'lib', 'node_modules', 'npm', 'bin', 'npm-cli.js')
  ];
```

sid-bruno (dismissed):

packages/bruno-electron/src/utils/install-packages.js
utkarsh-bruno: You're right. I was looking at the npm installation paths, great catch. Thanks. 
I checked the commands on windows as well, they seem to be working fine.

sid-bruno (approved):

engineering-bruno (approved):

PR #8274, opened 2026-06-15
security(ipc): confine file-write handlers to known collections and api specs

chirag-bruno:
## Problem

Many `renderer:*` IPC handlers wrote, moved, renamed or deleted pathnames supplied by the renderer without checking that the path resolved inside a collection Bruno has open. `preload.js` forwards any channel with no allowlist, so the safety contract has to live in the handler. Where it was missing, any renderer-side script-execution primitive became a path-traversal file write — or, via `delete-item`, a recursive delete — running with the user's full filesystem privileges.

Ref: BRU-971

## Fix

Confined to a watched collection via the existing `validatePathIsInsideCollection`:

- **Writes** — `save-request`, `save-multiple-requests`, `save-folder-root`, `save-collection-root`, `update-variable-in-file`
- **Renames / structure** — `rename-item-name`, `rename-item-filename`, `new-folder`, `clone-folder`, `resequence-items` (per item), `migrate-collection-to-yml`
- **Moves** — `move-file-item`, `move-item`, `move-item-cross-format`, `move-folder-item`, validating both ends so cross-collection moves between two open collections still work
- **Deletes** — `delete-item`, which was a delete-anywhere primitive (`fs.rmSync` with `recursive: true`)
- **Dotenv** — `save-dotenv-variables`, `save-dotenv-raw`, `create-dotenv-file`, `delete-dotenv-file`. The filename was already constrained by `isValidDotEnvFilename`; the directory it was joined onto was not.

Three cases where the collection path itself was legitimate but an interpolated name escaped it:

- **Environment name** — built as ``path.join(envDirPath, `${environmentName}.${format}`)`` with no sanitization, so a name containing path separators escaped the collection. Reachable on `delete-environment` (`unlinkSync`) and `rename-environment` (`renameSync`). All environment handlers now build paths through `resolveEnvironmentFilePath`, which requires the result to be a direct child of the collection's `environments/`.
- **API spec filename** — `create-api-spec` joined `apiSpecName` onto the location with only an `existsSync` check on the result, accepting any extension. It must now be a bare filename passing `validateApiSpec`, the same extension allowlist the open path already uses, with an existing directory as the location. The location stays renderer-supplied because the user picks it through a native browse dialog.
- **Environment export type** — `export-environment` now requires an absolute, existing export directory, but `environmentType` was interpolated into `bruno-${environmentType}-environments` and passed to `generateUniqueName`, which does not sanitize — escaping the directory that had just been validated. It is now restricted to the two values the renderer sends.

## Deliberately not covered

- `save-api-spec` and `rename-collection`: anchoring these to watcher state breaks real flows. Collections mount lazily and the sidebar Rename entry does not mount, and the API-spec registry is keyed by the raw path passed to `addWatcher`, which `workspace.yml` persists posix-style while chokidar reports native separators back. Both need a registry of *opened* collections and specs rather than watcher state.
- `update-bruno-config`, `ignore-folder` and `save-openapi-spec` remain unconfined.
- `targetFilename` in `save-transient-request` / `save-scratch-request`, and nested `item.filename` in `clone-folder`, are not reduced to basenames.
- The check is lexical, so a symlink inside a collection still resolves outward.
- `create-collection` and `import-collection*` take a user-picked location by design.

Legitimate UI flows already operate on paths that pass these checks, so no UX change is expected.

### DIFF
--- a/packages/bruno-electron/src/app/apiSpecs.js
+++ b/packages/bruno-electron/src/app/apiSpecs.js
@@ -116,5 +116,6 @@ const openApiSpec = async (win, watcher, apiSpecPath, options = {}) => {
 module.exports = {
   openApiSpec,
   openApiSpecDialog,
+  validateApiSpec,
   INVALID_EXTENSION_MESSAGE
 };
--- a/packages/bruno-electron/src/ipc/apiSpec.js
+++ b/packages/bruno-electron/src/ipc/apiSpec.js
@@ -1,6 +1,6 @@
 const { ipcMain } = require('electron');
-const { openApiSpecDialog, openApiSpec } = require('../app/apiSpecs');
-const { writeFile } = require('../utils/filesystem');
+const { openApiSpecDialog, openApiSpec, validateApiSpec } = require('../app/apiSpecs');
+const { writeFile, isDirectory } = require('../utils/filesystem');
 const { removeApiSpecUid } = require('../cache/apiSpecUids');
 const { removeApiSpecFromWorkspace } = require('../utils/workspace-config');
 const { getCertsAndProxyConfig } = require('./network/cert-utils');
@@ -25,7 +25,6 @@ const registerRendererEventHandlers = (mainWindow, watcher, lastOpenedApiSpecs) 
   ipcMain.handle('renderer:save-api-spec', async (event, pathname, content) => {
     try {
       await writeFile(pathname, content);
-      Promise.resolve();
     } catch (error) {
       return Promise.reject(error);
     }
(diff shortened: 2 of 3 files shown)


sid-bruno:
This should cover the sibling renderer-controlled file mutations too: `renderer:update-variable-in-file`, `renderer:create-environment`, `renderer:save-environment`, `renderer:rename-environment`, `renderer:delete-environment`, `renderer:update-environment-color`, `renderer:rename-item-name`, `renderer:rename-item-filename`, `renderer:new-folder`, and `renderer:resequence-items`.

Probably things we've added after the original document was created, worth checking if necessary

sid-bruno (approved):

PR #8327, opened 2026-06-22
feat(collections): add option to delete collection folder from disk o…

daniilsimakhin:
### Description

  Adds an opt-in **"Also delete folder from disk"** checkbox to the Remove Collection / Close Collections modals.

  When the box is checked, in addition to detaching the collection from the workspace, the collection folder is moved to the **OS trash**
   via Electron's `shell.trashItem`. The default behaviour is unchanged — the checkbox is off by default, so existing flows still only
  detach the collection.
  
  **Motivation.** Today the only way to fully get rid of a collection is to remove it from Bruno and then delete the folder manually in a
   file manager. This is friction for users who created a collection in the wrong place, are cleaning up experiments, or want to retire a
   workspace. Using `shell.trashItem` (not `fs.rm`) keeps the operation reversible — the folder lands in the OS trash and can be
  restored.
  
  **Changes.**                                                                          

  - `bruno-electron/src/ipc/collection.js` — `renderer:remove-collection` now accepts an `options` object. When `options.deleteFolder ===
   true`, after the in-app cleanup the handler calls `shell.trashItem(collectionPath)`. Errors are surfaced back to the renderer so the
  toast can show the real message.
  - `bruno-app/.../RemoveCollection/index.js`, `ConfirmCollectionCloseDrafts.js`, `RemoveCollectionsModal/index.js` — new `deleteFolder`
  checkbox with a short hint that the folder will be moved to the OS trash. Toast text and the primary-button label switch based on the
  choice.                                                                               
  - `bruno-app/.../slices/collections/actions.js` — `removeCollection(uid, options)` forwards `{ deleteFolder }` through IPC.
  - Error toasts now surface `error.message` instead of the generic "An error occurred…" string.

  **UX summary.**

  | Checkbox | Result | Toast |
  |---|---|---|
  | off (default) | Collection detached from workspace, folder untouched | `Collection removed from workspace` |
  | on | Collection detached, folder moved to OS trash | `Collection removed and folder moved to trash` |
  | on, error | Collection stays in workspace, folder untouched | real `error.message` |

  **Test plan.**

  - Remove a collection without checking the box → folder remains on disk, workspace updated.
  - Remove a collection with the box checked → folder appears in the OS trash (Recycle Bin on Windows, Trash on macOS, Trash on
  Linux/GIO).
  - Remove a collection with unsaved drafts → both "Save and Delete" and "Discard and Delete" paths honour the checkbox.
  - Remove multiple collections at once → all selected folders go to trash.
  - Read-only or locked folder → error toast shows the real OS error, collection stays in the workspace.

  **Notes.**

  - Intentionally restricted to `shell.trashItem`; hard delete (`fs.rm`) is not exposed.
  - Could be hardened by validating `collectionPath` against the list of registered collections before calling `shell.trashItem` (same
  class of concern raised in #8274 for `renderer:delete-item`). Happy to add that here if maintainers prefer.

  #### Contribution Checklist:
  
  Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please
  consider submitting them as separate pull requests.

## Summary by CodeRabbit

* **New Features**
  * Added an option to also delete collection folders from disk when removing collections.
  * Updated confirmation dialogs and button labels to reflect the chosen deletion behavior.
* **Bug Fixes**
  * Improved removal error handling with more informative messages when operations fail.
* **Other**
  * Enhanced “remove all” behavior to better handle partial failures, with success messaging that reflects how many removals completed.

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/RemoveCollection/ConfirmCollectionCloseDrafts.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/RemoveCollection/ConfirmCollectionCloseDrafts.js
@@ -200,6 +200,7 @@ const ConfirmCollectionCloseDrafts = ({ onClose, collection, collectionUid }) =>
         <label className="mt-6 flex items-start gap-2 cursor-pointer text-sm select-none">
           <input
             type="checkbox"
+            data-testid="confirm-close-drafts-delete-folder-checkbox"
             checked={deleteFolder}
             onChange={(e) => setDeleteFolder(e.target.checked)}
             className="mt-1"
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/RemoveCollection/ConfirmCollectionCloseDrafts.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/RemoveCollection/ConfirmCollectionCloseDrafts.js
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 import filter from 'lodash/filter';
 import { useDispatch, useSelector } from 'react-redux';
 import { flattenItems, isItemARequest, hasRequestChanges, findCollectionByUid } from 'utils/collections';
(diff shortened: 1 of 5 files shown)


daniilsimakhin:
Thanks for the review. Addressed all three findings in e2c81433a:

  - **`collection.js`** — `shell.trashItem` now runs first; on failure the
    handler throws before any workspace/watcher/spec cleanup, so renderer
    state and disk stay consistent.
  - **`RemoveCollectionsModal`** — switched to `Promise.allSettled`. Each
    failure emits its own toast; on partial success a follow-up toast
    reports `succeeded / total`.
  - **Checkboxes** — added `data-testid` to all three modals
    (`remove-collection-delete-folder-checkbox`,
    `confirm-close-drafts-delete-folder-checkbox`,
    `remove-collections-delete-folder-checkbox`,
    `remove-collections-drafts-delete-folder-checkbox`).

PR #8374, opened 2026-06-25
fix(app): make multi-file environment import resilient and idempotent

### LINKED ISSUE #8373: Importing multiple environment files is fragile: one bad file aborts the batch, partial failures show a generic error, and re-import stacks duplicates
### Summary

Importing **multiple** environment files at once (the "Import Environment" modal, `ImportEnvironmentModal`) is fragile in three ways:

1. **One bad/mismatched file aborts the entire batch.** The format is detected from only the first file (`detectEnvironmentFormat(parsedFiles[0].content)`) and a single importer is then run over *all* files. A malformed file, or a mixed Bruno + Postman selection, throws inside the importer and nothing gets imported.
2. **A mid-batch failure leaves a partial import behind a generic error.** `processEnvironments` loops with `await dispatch(...)` inside one `try`. If environment #3 of 5 fails, #1–2 are already written to disk, the loop throws, and the user only sees `An error occurred while importing the environment(s)` with no idea what partially succeeded.
3. **Re-importing silently stacks duplicates.** The client never checks existing names; the Electron handler `renderer:create-environment` calls `generateUniqueName()`, so re-importing the same file produces `X copy`, `X copy 2`, `X copy 3`, …

### Steps to reproduce

- Open a collection → Environments → Import, and select several `.json` environment files where one is malformed (or mix Bruno and Postman exports). → The whole import fails; nothing is imported.
- Or: import a valid set, then import the same set again. → Duplicates named `… copy`, `… copy 2` accumulate.

### Expected

- A single bad file should be reported but not block the others.
- Mixed Bruno/Postman selections should each import with their detected format.
- Re-importing an environment that already exists should be a no-op (or clearly skipped), not a duplicate.
- Clear feedback on how many were imported / skipped / failed.

### Environment

- Affects current `main`. Component: `packages/bruno-app/src/components/Environments/Common/ImportEnvironmentModal/index.js`.

I have a fix ready and will link a PR.

NicolasCV:
Fixes #8373

### Description

Importing several environment files at once is fragile. One bad file makes the whole batch fail, a failure partway through leaves some environments written to disk behind a generic error, and re-importing the same files quietly creates `… copy` duplicates. This PR reads and parses each file on its own so the batch survives bad input, skips environments that already exist, and replaces the generic toast with a short summary of what happened.

### Root cause

Three separate problems in `handleImportEnvironment` and `processEnvironments`:

1. Every file was read up front in a single `readMultipleFiles(Array.from(files))` call and the format was detected from the first file only (`detectEnvironmentFormat(parsedFiles[0].content)`), then one importer was run over every file. A malformed file makes the read throw and aborts the whole import; a selection that mixes Bruno and Postman exports throws inside the importer and aborts the rest. Either way nothing gets imported.

2. `processEnvironments` looped with `await dispatch(action)` inside a single `try`. If the third of five environments failed, the first two were already written to disk, the loop threw, and the user saw `An error occurred while importing the environment(s)` with no idea which ones made it.

3. The client never checks existing names. The only guard is the Electron handler `renderer:create-environment`, which calls `generateUniqueName()`, so re-importing the same file produces `X copy`, then `X copy 2`, and so on.

### The fix

Each file is now **read and parsed on its own** — the loop calls `readMultipleFiles([file])` per file and detects its format individually, so a single malformed JSON only fails that file and is collected as a failure instead of aborting the batch. A mixed Bruno/Postman selection imports each file with its own detected format.

Each environment is dispatched in its own `try/catch` and failures are collected, so one failure doesn't take down the rest of the batch.

Before importing, the code builds a set of existing names (`collection.environments` for collection scope, the `globalEnvironments` selector for global scope) keyed by `sanitizeName`, the same function the Electron handler uses, and skips any name that already exists. The incoming batch is also deduped against itself. The name is normalized **once** with `sanitizeName` and that value is reused for the dedupe check and for the `importEnvironment` / `addGlobalEnvironment` payloads, so the collection and global branches behave identically and never re-import raw names. Re-importing the same file is now a no-op instead of stacking copies.

Feedback is one summary: an `Imported N environments.` toast, plus a note of how many were skipped, unnamed, or failed. The modal stays open only when nothing was imported and nothing was skipped so the user can retry; otherwise it closes and fires `onEnvironmentCreated` only if something was actually imported.

The change is contained to the modal component. Importers, actions, and the Electron handler are untouched, and it reuses the existing `readMultipleFiles`, `importPostmanEnvironment`, `importBrunoEnvironment`, `sanitizeName`, `importEnvironment`, and `addGlobalEnvironment`.

### Before / After

| Scenario | Before | After |
| --- | --- | --- |
| One malformed file among several | Whole import aborts, nothing imported | Bad file reported, the rest imported |
| Mixed Bruno + Postman selection | Wrong importer throws, import aborts | Each file imported with its own detected format |
| One env fails mid-batch | Partial write plus a generic error | The rest import, the failure is listed |
| Re-import the same file | Stacks `X copy`, `X copy 2` | Names that already exist are skipped |
| Result feedback | `N environments imported successfully` or a generic error | `Imported N environments.` plus a skipped/unnamed/failed note |

### Verification

Tested manually in the app, in both collection and global scope:

- A folder of valid Bruno envs imports cleanly with one success toast.
- A corrupt JSON in the set is reported while the rest import.
- A mixed Bruno and Postman selection imports both formats.
- Re-importing the same set imports nothing and skips everything already present, with no `copy` duplicates on disk.
- An env with no name is counted under "had no name".

ESLint passes on the changed files. The change is contained to the modal component, with no API or behavior changes to importers, actions, or the Electron handler. Added a Jest spec for the modal (`ImportEnvironmentModal/index.spec.js`, 8 tests, all passing) covering mixed Bruno/Postman selections, a malformed file alongside valid ones, duplicate-name skips, unnamed environments, partial dispatch failures, the "nothing valid found" path, and name normalization across the collection and global branches.

---

- [ ] I've used AI significantly to create this pull request
- [x] The pull request only addresses one issue or adds one feature.
- [x] The pull request does not introduce any breaking changes
- [ ] I have added screenshots or gifs to help explain the change if applicable.
- [x] I have read the contribution guidelines.
- [x] Create an issue and link to the pull request.

## Summary by CodeRabbit

* **New Features**
  * Multi-file environment import can handle mixed Postman/Bruno files in one batch.
  * Import now presents a detailed batch summary (imported, skipped duplicates, unnamed) and shows specific failure notes when issues occur.

* **Bug Fixes**
  * Duplicate detection and import routing are improved by normalizing environment names.
  * Import failures for individual environments no longer prevent other valid environments from importing.
  * The import dialog remains open when no environments were effectively imported.

* **Tests**
  * Added coverage for mixed imports, partial failures, duplicates, unnamed handling, and summary/toast behavior (including name normalization).

### DIFF
--- a/packages/bruno-app/src/components/Environments/Common/ImportEnvironmentModal/index.js
+++ b/packages/bruno-app/src/components/Environments/Common/ImportEnvironmentModal/index.js
@@ -8,7 +8,6 @@ import importBrunoEnvironment from 'utils/importers/bruno-environment';
 import { readMultipleFiles } from 'utils/importers/file-reader';
 import { importEnvironment } from 'providers/ReduxStore/slices/collections/actions';
 import { addGlobalEnvironment } from 'providers/ReduxStore/slices/global-environments';
-import { toastError } from 'utils/common/error';
 import { sanitizeName } from 'utils/common/regex';
 import { IconFileImport } from '@tabler/icons';
 
@@ -44,13 +43,13 @@ const ImportEnvironmentModal = ({ type = 'collection', collection, onClose, onEn
     const toImport = [];
     let skipped = 0;
     for (const env of named) {
-      const key = sanitizeName(env.name);
-      if (seen.has(key)) {
+      const name = sanitizeName(env.name);
+      if (seen.has(name)) {
         skipped++;
-      } else {
-        seen.add(key);
-        toImport.push(env);
+        continue;
       }
+      seen.add(name);
+      toImport.push({ ...env, name });
     }
 
     let imported = 0;
@@ -107,17 +106,18 @@ const ImportEnvironmentModal = ({ type = 'collection', collection, onClose, onEn
   };
 
   const handleImportEnvironment = async (files) => {
-    let parsedFiles;
-    try {
-      parsedFiles = await readMultipleFiles(Array.from(files));
-    } catch (err) {
-      toastError(err, 'Import environment failed');
-      return;
-    }
-
     const environments = [];
     const parseFailures = [];
-    for (const parsedFile of parsedFiles) {
+    for (const file of Array.from(files)) {
+      let parsedFile;
+      try {
+        [parsedFile] = await readMultipleFiles([file]);
+      } catch (err) {
+        console.error(`Failed to read ${file.name}:`, err);
+        parseFailures.push({ name: file.name, message: err?.message || String(err) });
+        continue;
+      }
+
       try {
         const format = detectEnvironmentFormat(parsedFile.content);
         const result
--- a/packages/bruno-app/src/components/Environments/Common/ImportEnvironmentModal/index.js
+++ b/packages/bruno-app/src/components/Environments/Common/ImportEnvironmentModal/index.js
@@ -2,17 +2,19 @@ import React, { useState } from 'react';
 import Portal from 'components/Portal';
 import Modal from 'components/Modal';
 import toast from 'react-hot-toast';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import importPostmanEnvironment from 'utils/importers/postman-environment';
 import importBrunoEnvironment from 'utils/importers/bruno-environment';
 import { readMultipleFiles } from 'utils/importers/file-reader';
 import { importEnvironment } from 'providers/ReduxStore/slices/collections/actions';
 import { addGlobalEnvironment } from 'providers/ReduxStore/slices/global-environments';
 import { toastError } from 'utils/common/error';
+import { sanitizeName } from 'utils/common/regex';
 import { IconFileImport } from '@tabler/icons';
 
 const ImportEnvironmentModal = ({ type = 'collection', collection, onClose, onEnvironmentCreated }) => {
   const dispatch = useDispatch();
+  const globalEnvironments = useSelector((state) => state.globalEnvironments.globalEnvironments) || [];
   const [isDragOver, setIsDragOver] = useState(false);
 
   const isGlobal = type === 'global';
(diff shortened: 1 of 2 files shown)


NicolasCV:
Went through CodeRabbit's review. Here's where each point landed:

- Read each file on its own so one bad file doesn't abort the batch (Major). Done in ef9bccb. Each file goes through its own `readMultipleFiles([file])` call, so a malformed JSON only fails that file and the valid ones still import. CodeRabbit already marked this one resolved.
- Normalize the name once before dedupe and import (Major). The name is sanitized once in the named/unnamed split and reused for the dedupe key and for both the collection and global payloads, so the two branches can't diverge. I also added the type guard CodeRabbit suggested (3d81bff): a non-string name no longer reaches `sanitizeName` and throws, and a name that sanitizes to empty is now counted as unnamed instead of importing a blank-named environment.
- Test coverage for the batch outcomes. Added a jest spec next to the component (`ImportEnvironmentModal/index.spec.js`, 9 tests) covering mixed Bruno/Postman, a malformed file alongside valid ones, duplicate-name skips, unnamed environments, non-string and empty names, a mid-batch dispatch failure, the "nothing valid" path, and the sanitized name in both import payloads. I kept it as a unit spec rather than extending the `tests/environments/import-environment` Playwright suite, because it asserts the toast summary and the imported/skipped/unnamed/failures counts directly.
- Unnamed-environment test fidelity. The Bruno importer defaults a missing name to "Imported Environment", so I moved the unnamed case to the Postman path, which is where a genuinely nameless environment actually comes from.

Attachments:

- `bruno-env-import-demo.zip`: the six environment files I tested with (valid Bruno and Postman exports, a duplicate name, a malformed JSON, and a valid-but-not-an-environment file). Dropping all six at once imports 3, skips the duplicate, and reports the 2 bad files instead of failing the whole batch.

https://github.com/user-attachments/assets/2704448c-1542-4bfc-891c-709c86f621a3

[bruno-env-import-demo.zip](https://github.com/user-attachments/files/29392361/bruno-env-import-demo.zip)

- A short screen recording showing the before/after.

NicolasCV:
Pushed 79ce252 for the two follow-up review comments.

The within-batch duplicate path now has its own test. Two environments in the same drop whose names sanitize to the same value ('Prod/Env' and 'Prod-Env' both become 'Prod-Env'): the first one imports, the second is skipped, and neither gets a 'copy' suffix. It asserts the toast summary and that dispatch only fires once, so the `seen.add(key)` branch is actually exercised rather than only the collision-against-existing case.

I also made the Bruno importer mock faithful to the real one. It now normalizes a missing name to 'Imported Environment' before returning, the same as `validateBrunoEnvironment`, instead of handing the raw env straight to `processEnvironments`. With that fixed I added a test that an unnamed Bruno environment imports under that default name rather than being counted as "had no name". The genuinely nameless case still goes through the Postman path, since that is where an environment with no name actually shows up.

sachin-thakur-bruno:
Thank you @NicolasCV for raising the PR but we have solved this internally with muti environment imports effort and will be part of 4.2.0 release. Closing this PR.

PR #8398, opened 2026-06-27
feat(app): add "Ignore Folder" sidebar context menu item (closes #1890)

### LINKED ISSUE #1890: Add menu item to ignore folder
[JIRA](https://usebruno.atlassian.net/browse/BRU-3462)

It would be nice if folders could be ignored without adding them to the config file manually:

Screenshot

justadityaraj:
### Description

Adds an **Ignore Folder** item to the folder context menu in the sidebar, closing #1890.

Today the only way to add a folder to a collection's `ignore` list is to hand-edit `bruno.json`. This adds a right-click action that does it from the UI:

- Right-click a folder → **Ignore Folder**
- The folder's collection-relative path is appended to `brunoConfig.ignore` and persisted to `bruno.json`
- The folder is removed from the sidebar immediately; `bruno.json` keeps it ignored on subsequent loads

The path is stored with forward slashes so the value stays portable when `bruno.json` is committed and shared across operating systems.

### How it works

- `getCollectionRelativePath(collectionPathname, itemPathname)` derives the portable relative path
- `addPathToIgnoreList(brunoConfig, relativePath)` returns a new config with the path added (immutable, dedupes)
- The menu handler dispatches the existing `updateBrunoConfig` thunk to persist, then `deleteItem` to drop the folder from the tree

### Tests

Added `utils/collections/ignore.spec.js` covering relative-path derivation (nested paths, Windows separators, trailing slashes, root/empty inputs) and ignore-list merging (append, create, dedupe, empty paths, immutability). 11 assertions, all passing.

### Notes

- Only shown for folders.
- Reuses existing persistence (`updateBrunoConfig`) and tree-removal (`deleteItem`) machinery; no watcher changes.

## Summary by CodeRabbit

* **New Features**
  * Added an **Ignore Folder** action to the folder menu in the sidebar.
  * Ignoring a folder updates the app’s ignore list and removes the folder from view right away.

* **Bug Fixes**
  * Improved ignore-path computation, including nested paths, trailing slashes, Windows path separators, and boundary/case handling.
  * Avoids duplicate ignore entries and skips invalid inputs.

* **Tests**
  * Added Jest coverage for ignore-path handling and ignore-list update behavior.

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
@@ -19,12 +19,14 @@ import {
   IconSettings,
   IconInfoCircle,
   IconTerminal2,
-  IconAppWindow
+  IconAppWindow,
+  IconEyeOff
 } from '@tabler/icons';
 import { useSelector, useDispatch } from 'react-redux';
 import { addTab, focusTab, makeTabPermanent } from 'providers/ReduxStore/slices/tabs';
-import { handleCollectionItemDrop, sendRequest, showInFolder, pasteItem, saveRequest } from 'providers/ReduxStore/slices/collections/actions';
-import { toggleCollectionItem, addResponseExample } from 'providers/ReduxStore/slices/collections';
+import { handleCollectionItemDrop, sendRequest, showInFolder, pasteItem, saveRequest, updateBrunoConfig } from 'providers/ReduxStore/slices/collections/actions';
+import { toggleCollectionItem, addResponseExample, deleteItem } from 'providers/ReduxStore/slices/collections';
+import { getCollectionRelativePath, addPathToIgnoreList } from 'utils/collections/ignore';
 import { insertTaskIntoQueue } from 'providers/ReduxStore/slices/app';
 import { uuid } from 'utils/common';
 import { copyRequest, setFocusedSidebarPath } from 'providers/ReduxStore/slices/app';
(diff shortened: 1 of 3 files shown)


justadityaraj:
Thanks for the review. Addressed below:

**1. `deleteItem` deleting the folder from disk** — this one's a false positive. The `deleteItem` used in `handleIgnoreFolder` is the slice reducer (`slices/collections`, imported at the top of `CollectionItem`), not the `deleteItem` *thunk* in `slices/collections/actions.js`. The reducer only removes the node from the in-memory tree via `deleteItemInCollection` (a plain array filter) — no IPC, no filesystem call — and the `{ itemUid, collectionUid }` payload matches the reducer's `action.payload`. So ignoring a folder never touches disk; the files stay put and `bruno.json`'s `ignore` list keeps it hidden on reload.

**3. `getCollectionRelativePath` fallback** — good catch, fixed. It now returns `''` for anything not under the collection root (matching its documented contract) instead of the absolute path.

**4. Missing edge-case tests** — added regression tests for non-descendant paths, false prefix boundaries (`coll` vs `collection`), and case-only mismatches.

**2. Re-reading `brunoConfig` before the previous update lands** — fair point. Since this is a manual, one-folder-at-a-time context-menu action and the only downside is a lost ignore-list entry (no data loss), I've left it as-is to keep the change small, but happy to harden it if you'd prefer.

justadityaraj:
Bumping.

The CodeRabbit points are answered in the thread above. The main one was a false positive: the `deleteItem` used here is the slice reducer, so it just drops the node from the in-memory tree, not the thunk that touches disk. I also added the relative-path fallback it flagged, with tests.

Let me know if need to rebase if it has gone stale in the meantime. Thanks!

justadityaraj:
Closing this — the feature landed in #8725, which added the "Ignore Folder" context menu item along with the electron-side watcher and IPC changes. `CollectionItem/index.js` now carries that implementation, so this branch conflicts with it and there's nothing left here worth rebasing.

Thanks all the same, and thanks to @ravindra-bruno for picking it up.

PR #8405, opened 2026-06-29
fix(oauth2): prevent code injection in OAuth2 callback handling

abhishekp-bruno:
JIRA - https://usebruno.atlassian.net/browse/BRU-3546

### Description
Bruno didn't validate the state returned on the OAuth2 callback, and sent none at all when the user left it blank — leaving auth flows open to CSRF / code injection.

Changes:

Always issue a state — random when unset, or a random nonce appended to the user's value so it can't be predicted/replayed.
Validate the returned state against the issued one and abort on mismatch, in both the embedded-window and system-browser.
Covers authorization code + implicit grants (query params and hash fragments).

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Enhanced OAuth2 security by validating callback `state` for both authorization-code and implicit flows, with nonce-protected state handling.
* **Bug Fixes**
  * OAuth2 fetch/refresh errors now reliably appear in the response pane (in addition to notifications), with clearer error details.
* **Tests**
  * Added unit and E2E coverage for `state` match/mismatch scenarios (query + hash) and introduced new OAuth2 state-related test fixtures.


packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
```diff
@@ -70,7 +93,9 @@ const Oauth2ActionButtons = ({ item, request, collection, url: accessTokenUrl, c
       if (error?.message && error.message.includes('cancelled by user')) {
         return;
       }
-      toast.error(error?.message || 'An error occurred while fetching token!');
+      const errorMessage = formatIpcError(error) || 'An error occurred while fetching token!';
+      toast.error(errorMessage);
+      showOauth2Error(errorMessage);
```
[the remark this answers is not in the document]
abhishekp-bruno: This is the pattern that we are following in our codebase should I handle this case flagged by CodeRabbit 
cc: @sanish-bruno @vijayh-bruno

packages/bruno-electron/src/utils/oauth2-protocol-handler.js
```diff
@@ -80,6 +81,23 @@ const handleOauth2ProtocolUrl = (url) => {
...
     }
 
+    // Validate the state parameter to protect against CSRF / authorization code
+    // injection. The returned state must match the cryptographically random state
+    // issued when the flow was initiated.
+    const expectedState = oauth2AuthorizationRequest?.expectedState;
+    if (expectedState) {
+      const returnedState
+        = urlObj.searchParams.get('state')
+          || (urlObj.hash ? new URLSearchParams(urlObj.hash.substring(1)).get('state') : null);
```
sanish-bruno: @abhishekp-bruno nitpick: can we create a small util to get the state? i guess we have the same logic in couple of places

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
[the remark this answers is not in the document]
sanish-bruno: we will keep the way we were doing so far

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
```diff
@@ -97,14 +122,17 @@ const Oauth2ActionButtons = ({ item, request, collection, url: accessTokenUrl, c
...
         return;
       }
 
       toast.success('Token refreshed successfully!');
     } catch (error) {
       console.error(error);
       toggleRefreshingToken(false);
-      toast.error(error?.message || 'An error occurred while refreshing token!');
+      const errorMessage = formatIpcError(error) || 'An error occurred while refreshing token!';
+      toast.error(errorMessage);
```
sanish-bruno, line 131: can consider removing the toast error, i feel like  showing toast for error happening during the request is probably not needed like you pointed out @abhishekp-bruno , as long as we are showing the error within response pane. can consider showing the error properly inside a ErrorCard like we do for pre-request error and all, but  that can be done in a separate PR at later stage, not needed for now.

packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
```diff
@@ -202,6 +202,22 @@ const authorizeUserInWindow = ({ authorizeUrl, callbackUrl, session, additionalH
...
+          // random state issued when the flow was initiated.
+          if (expectedState) {
+            const finalUrlObj = new URL(finalUrl);
+            const returnedState
+              = finalUrlObj.searchParams.get('state')
+                || (finalUrlObj.hash ? new URLSearchParams(finalUrlObj.hash.substring(1)).get('state') : null);
+
+            if (returnedState !== expectedState) {
+              return reject(
+                new Error('OAuth2 state mismatch: the returned state does not match the issued state. Aborting to prevent authorization code injection.')
```
sanish-bruno: I don't think we need to provide the reason for it, just the mismatch message will do IMO, can consider unit tests if needed once extracted

packages/bruno-electron/src/utils/oauth2-protocol-handler.js
sanish-bruno: `urlObj.hash` how does this work? what the the hash do? can you please help me understand `urlObj.hash.substring(1)).get('state')`

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -707,6 +712,16 @@ const generateCodeVerifier = () => {
   return crypto.randomBytes(22).toString('hex');
 };
 
+// Generate a cryptographically random state value used to protect OAuth2
+// authorization flows against CSRF / authorization code injection.
+const generateState = ({ userState }) => {
+  let cryptographicallyRandomString = crypto.randomBytes(16).toString('hex');
+  if (userState && userState.length) {
+    return userState + cryptographicallyRandomString;
```
sanish-bruno: I guess we have verified this right that the provider won't be trying to match the state, it just go through the round-trip. check whether the state has any character limit

tests/auth/oauth2/init-user-data/preferences.json
```diff
@@ -0,0 +1,15 @@
+{
+  "maximized": false,
+  "lastOpenedCollections": ["{{collectionPath}}"],
+  "preferences": {
+    "request": {
+      "oauth2": {
+        "useSystemBrowser": true
+      }
+    },
+    "onboarding": {
```
sanish-bruno, line 5: can remove the onboarding section

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      );
+    });
+  });
+
+  describe('implicit flow (state in hash fragment)', () => {
+    it('should resolve with tokens when the returned state matches', () => {
+      registerOauth2AuthorizationRequest(resolve, reject, null, 'expected-state');
+
+      handleOauth2ProtocolUrl(
+        'bruno://oauth2/callback#access_token=token-abc&token_type=bearer&state=expected-state'
```
sanish-bruno: just verify if this  is a actual use case when it comes to hash fragments

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
```diff
@@ -0,0 +1,122 @@
...
+      );
+
+      expect(resolve).not.toHaveBeenCalled();
+      expect(reject).toHaveBeenCalledWith(
+        expect.objectContaining({ message: expect.stringContaining('state mismatch') })
+      );
+    });
+  });
+
+  describe('when no expected state was registered (backward compatibility)', () => {
```
sanish-bruno: I guess we can mock cryptographic state, and verify against, let me know if it not possible. not sure how is this test succeeding when not state is received in the callback

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
```diff
@@ -0,0 +1,122 @@
...
+
+  describe('error responses are handled before state validation', () => {
+    it('should reject with the provider error even if state is absent', () => {
+      registerOauth2AuthorizationRequest(resolve, reject, null, 'expected-state');
+
+      handleOauth2ProtocolUrl('bruno://oauth2/callback?error=access_denied');
+
+      expect(resolve).not.toHaveBeenCalled();
+      expect(reject).toHaveBeenCalledWith(
+        expect.objectContaining({ message: expect.stringContaining('Authorization Failed') })
```
sanish-bruno, line 131: add hash fragment error `bruno://oauth2/callback#error=some_error`

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+  return {
+    state: url.searchParams.get('state'),
+    code: url.searchParams.get('code'),
+    access_token: null
+  };
+};
+
+test.describe('OAuth2 callback state validation', () => {
+  test('authorization code: rejects callback when returned state does not match issued state', async ({ restartApp }) => {
+    const app = await restartApp();
```
sanish-bruno: I guess we can user launchEletronApp utility rather than restartApp

sanish-bruno: why aren't we using `pageWithUserdata`?, check if we have access to app when we are using `pageWithUserData`? else we can use launchEletronApp instead i guess

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+    const page = await waitForReadyPage(app);
+    await stubOpenExternal(app);
```
sanish-bruno: why. can't this steps be in a beforeEach block?

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+      page.getByTestId('response-pane').getByText(STATE_MISMATCH_ERROR)
+    ).toBeVisible({ timeout: 15_000 });
+  });
+
+  test('implicit grant: rejects callback when returned state does not match issued state', async ({ restartApp }) => {
+    const app = await restartApp();
+    const page = await waitForReadyPage(app);
+    await stubOpenExternal(app);
+    await installCallbackCapture(app);
+
```
sanish-bruno: rm new line

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+    const { state: returnedState, access_token } = getCallbackParams(receivedCallbackUrl as string, 'hash');
+    expect(returnedState).toBe(WRONG_STATE);
+    expect(returnedState).not.toBe(issuedState);
+    expect(access_token).toBe(PROVIDER_ACCESS_TOKEN);
+    await expect(
+      page.getByRole('status').filter({ hasText: STATE_MISMATCH_ERROR })
+    ).toBeVisible({ timeout: 15_000 });
+    await expect(
+      page.getByTestId('response-pane').getByText(STATE_MISMATCH_ERROR)
```
sanish-bruno: use commonlocators where-ever possible

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+    test.setTimeout(60_000);
+    const app = await restartApp();
+    const page = await waitForReadyPage(app);
+    await stubOpenExternal(app);
+    await installCallbackCapture(app);
+
+    await clickGetAccessToken(page, 'Implicit');
+    await waitForAuthorizationStarted(app);
+
+    const issuedState = await getIssuedState(app);
```
sanish-bruno: add steps to logical steps for better legibility of the test flow, use `test.step`

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+    expect(receivedCallbackUrl).toBe(callbackUrl);
+
+    const { state: returnedState, access_token } = getCallbackParams(receivedCallbackUrl as string, 'hash');
+    expect(returnedState).toBe(issuedState);
+    expect(access_token).toBe(PROVIDER_ACCESS_TOKEN);
+
+    await expect(page.getByText('Token fetched successfully!')).toBeVisible({ timeout: 15_000 });
+  });
+
+  test('authorization code (user-supplied state): issues userState + nonce and accepts matching callback', async ({ restartApp }) => {
```
sanish-bruno: Check if we need test for implicit,  user provided state

sanish-bruno (changes requested):

packages/bruno-electron/src/utils/oauth2-protocol-handler.js
abhishekp-bruno: added a common function  to handle this

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
abhishekp-bruno: Removed toast

packages/bruno-electron/src/ipc/network/authorize-user-in-window.js
abhishekp-bruno: UPDATED the error message to "OAuth2 state mismatch"

packages/bruno-electron/src/utils/oauth2-protocol-handler.js
abhishekp-bruno: const urlObj = new URL('https://app.example.com/callback#access_token=abc123&state=xyz');

urlObj.hash;                    // "#access_token=abc123&state=xyz"
urlObj.hash.substring(1);       // "access_token=abc123&state=xyz"  (drops the "#")

const params = new URLSearchParams('access_token=abc123&state=xyz');
params.get('access_token');     // "abc123"
params.get('state');            // "xyz"
params.get('code');             // null  (not present)

example of how it works

packages/bruno-electron/src/utils/oauth2.js
abhishekp-bruno: Yes 
The provider treats state as opaque and simply echoes it back — unchanged. It does not parse it, validate it, or derive anything from it. The provider's only job is to return the exact same value on the redirect.

tests/auth/oauth2/init-user-data/preferences.json
abhishekp-bruno: I have removed it, are we handling it at config level ?

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
abhishekp-bruno: For implicit this used to be the norm but it is discouraged now. It can still be used by some providers

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
abhishekp-bruno: so actually the test should fail if no state is passed, but here if the user did not pass it is created by us and passed in the params, hence it should not fail.

abhishekp-bruno: I will try to think of a better test case for this situation since this would have passed before this fix as well

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
abhishekp-bruno: added one test to check hash fragment error

tests/auth/oauth2/oauth2-state-validation.spec.ts
abhishekp-bruno: restartApp is intentional here. It bundles three things these tests need per-test: (1) copies init-user-data (enables useSystemBrowser: true), (2) resolves {{collectionPath}} from the collection fixture so oauth2-state auto-loads, and (3) closePrevious: true for a clean instance per test. launchElectronApp is the low-level primitive and would make us re-implement all three inline in every test. Happy to switch if we see a concrete benefit, but for now restartApp keeps these tests isolated and readable.

pageWithUserData doesn't work for these tests: it only exposes a Page, but we need the ElectronApplication handle for app.evaluate(...) (that's how stubOpenExternal, installCallbackCapture, and fireCallback operate). It also reuses a single cached app across the whole file (no closePrevious), so global stubs and in-flight OAuth flows would leak between tests. restartApp gives us the app plus a fresh instance per test, which is exactly what this suite needs.

tests/auth/oauth2/oauth2-state-validation.spec.ts
abhishekp-bruno: can only add these two to beforeEach block 
    const app = await restartApp();
    const page = await waitForReadyPage(app);
 
 imo just adding these two to for each reduces readability, right now it looks simple and straightforward for a test file

tests/auth/oauth2/oauth2-state-validation.spec.ts
abhishekp-bruno: removed

tests/auth/oauth2/oauth2-state-validation.spec.ts
abhishekp-bruno: replaced with locators

tests/auth/oauth2/oauth2-state-validation.spec.ts
abhishekp-bruno: have added a test for implicit user supplied state

tests/auth/oauth2/oauth2-state-validation.spec.ts
abhishekp-bruno: ADDED steps for legibility

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
```diff
@@ -59,6 +81,7 @@ const Oauth2ActionButtons = ({ item, request, collection, url: accessTokenUrl, c
         const errorMessage = result?.error || 'No access token received from authorization server';
         console.error(errorMessage);
         toast.error(errorMessage);
+        showOauth2Error(errorMessage);
```
sanish-bruno, line 82: i think we can keep toast.error in both, i guess. only issue was the toast error will be gone, and user might not know the issue, i guess that is covered with error inside the response pane. also add a screen shot of the error, and recording of this in action

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -707,6 +712,16 @@ const generateCodeVerifier = () => {
   return crypto.randomBytes(22).toString('hex');
 };
 
+// Generate a cryptographically random state value used to protect OAuth2
+// authorization flows against CSRF / authorization code injection.
+const generateState = ({ userState }) => {
```
sanish-bruno, line 718: add additional info to the comment, mentioning what will happen to the user provided state

sanish-bruno (changes requested):

sanish-bruno:
@lohit-bruno need your eyes here!

packages/bruno-electron/src/utils/oauth2.js
abhishekp-bruno: ADDED comment for user supplied state condition

packages/bruno-electron/src/utils/oauth2-protocol-handler.js
```diff
@@ -80,6 +90,21 @@ const handleOauth2ProtocolUrl = (url) => {
...
     }
 
+    // Validate the state parameter to protect against CSRF / authorization code
+    // injection. The returned state must match the cryptographically random state
+    // issued when the flow was initiated.
+    const expectedState = oauth2AuthorizationRequest?.expectedState;
+    if (expectedState) {
+      const returnedState = getParamFromUrl(urlObj, 'state');
+
+      if (returnedState !== expectedState) {
```
sanish-bruno: should we use `crypto.timingSafeEqual` instead of `!==`?

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
```diff
@@ -41,6 +44,25 @@ const Oauth2ActionButtons = ({ item, request, collection, url: accessTokenUrl, c
   const credentialsData = find(collection?.oauth2Credentials, (creds) => creds?.url == interpolatedAccessTokenUrl && creds?.collectionUid == collectionUid && creds?.credentialsId == credentialsId);
   const creds = credentialsData?.credentials || {};
 
+  const showOauth2Error = (errorMessage) => {
```
sanish-bruno, line 47: @abhishekp-bruno please verify the response structure is same for other IPC error scenarios.

sanish-bruno (changes requested):

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
```diff
@@ -0,0 +1,133 @@
...
+      );
+
+      expect(resolve).not.toHaveBeenCalled();
+      expect(reject).toHaveBeenCalledWith(
+        expect.objectContaining({ message: expect.stringContaining('state mismatch') })
+      );
+    });
+  });
+
+  describe('when no expected state was registered ', () => { // it stops someone from later making state mandatory and breaking flows when state is not passed
```
sanish-bruno: what is the need for this test? , is this expected behaviour? validation should be called either way right, if no state comes back from the callback, we don't validate at all? That might cause problems right?

sanish-bruno (changes requested):

packages/bruno-electron/src/utils/oauth2-protocol-handler.js
abhishekp-bruno: that's not needed, as per my understanding that is needed when someone can make multiple hits trying to guess some key/value which might give the attacker the access but here if the callback gets wrong state even once it fails there itself. So not needed here.

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
abhishekp-bruno: added this again as per our discussion earlier

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/Oauth2ActionButtons/index.js
abhishekp-bruno: yes this matches the structure with other error scenarios

packages/bruno-electron/tests/utils/oauth2-protocol-handler.spec.js
abhishekp-bruno: This was  actually a vulnerability, have added the fix to make sure that the state is always present

sanish-bruno (approved):

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -324,8 +328,8 @@ const getOAuth2AuthorizationCode = (request, codeChallenge, collectionUid) => {
       authorizationUrlWithQueryParams.searchParams.append('code_challenge', codeChallenge);
       authorizationUrlWithQueryParams.searchParams.append('code_challenge_method', 'S256');
     }
-    if (state) {
-      authorizationUrlWithQueryParams.searchParams.append('state', state);
+    if (effectiveState) {
```
vijayh-bruno, line 331: What is the point of this condition? Will `effectiveState` ever be false?

abhishekp-bruno: Yes, that is a vulnerability we are fixing here, we need to make sure that at any given time if state does not exist we terminate the process.  
An attacker can inject auth code like this when we are waiting for callback from the auth provider. 
`open bruno://app/oauth2/callback?code=ATTACKER_CODE` 
So we need to make sure the state exists. From our side the state will always be there.

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -844,8 +863,8 @@ const getOAuth2TokenUsingImplicitGrant = async ({ request, collectionUid, forceF
   if (scope) {
     authorizationUrlWithQueryParams.searchParams.append('scope', scope);
   }
-  if (state) {
-    authorizationUrlWithQueryParams.searchParams.append('state', state);
+  if (effectiveState) {
+    authorizationUrlWithQueryParams.searchParams.append('state', effectiveState);
   }
```
sid-bruno:
```suggestion
  authorizationUrlWithQueryParams.searchParams.append('state', effectiveState);
```

sid-bruno (changes requested):

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -707,6 +712,17 @@ const generateCodeVerifier = () => {
   return crypto.randomBytes(22).toString('hex');
 };
 
+// Build an OAuth2 state string to help prevent CSRF and forged auth codes.
+// If the user passes a state, it goes first; we append random bytes after it.
+// The user keeps their custom data, and the random suffix keeps the flow secure.
+const generateState = ({ userState }) => {
+  let cryptographicallyRandomString = crypto.randomBytes(16).toString('hex');
+  if (userState && userState.length) {
+    return userState + cryptographicallyRandomString;
```
sid-bruno: This might lead to a breaking change. if the user has been using a local testing setup for a while that validates the tokens based on hard coded state strings, the string will now come with an added crypto string which isn't providing much value since making it a secure string doesn't change anything for the provider. Nonce's are generated by the users to just validate a secure transfer and retrieval.

I'd recommend just having `userState` left as is and only use the secure string if nothing is passed by the user.

sid-bruno (changes requested):

packages/bruno-electron/src/utils/oauth2.js
abhishekp-bruno: `RFC 6749 §10.12 (CSRF) requires the client to bind the redirect callback to the user-agent's authenticated state, and recommends carrying that binding value in state. The whole point of CSRF protection is that an attacker can't forge a callback the victim's client will accept. If state is constant or predictable, an attacker who knows/guesses it can stitch their own authorization response onto the victim's session (the code-injection class of attack), which is exactly what the parameter is supposed to prevent.`

This is the reason I thought appending the random value to user's input would be the correct approach for to handle this vulnerability. 

Should we still pass only user's input ?

abhishekp-bruno: as per our discussion I have removed appending random string to the user's state value 
we return generatedRandomValue if user does not enter any value or return user's state as is

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/index.js
```diff
@@ -182,13 +182,23 @@ const OAuth2AuthorizationCode = ({ save, item = {}, request, handleRun, updateAu
...
 
         return (
           <div className="flex items-center gap-4 w-full" key={`input-${key}`}>
-            <label className="block min-w-[140px]">{label}</label>
+            <label className="min-w-[140px] flex items-center gap-4">
+              {label}
+              {tooltip && (
+                <div className="relative group cursor-pointer inline-flex items-center">
+                  <IconHelp size={16} className="text-gray-500" />
+                  <span className="group-hover:opacity-100 pointer-events-none opacity-0 max-w-60 absolute left-0 top-full mt-1 w-max p-2 bg-gray-700 text-white text-xs rounded-md transition-opacity duration-200 z-10">
```
sid-bruno: `bg-gray-700` and `text-white` colors to come from the themes and not tailwind

sid-bruno (changes requested):

packages/bruno-app/src/components/RequestPane/Auth/OAuth2/AuthorizationCode/index.js
abhishekp-bruno: @sid-bruno just checked both of these properties are not getting applied from the themes.

abhishekp-bruno: Updated the tooltip I added and the existing tooltips in the component to use styles from the useTheme hook.

sid-bruno (approved):

PR #8461, opened 2026-07-02
fix(bruno-response): normalize response header name lookup to be case-insensitive

### LINKED ISSUE #1412: accept any casing in scripts for getHeader
As headers are – per spec – case insensitive, bruno could be changed such that `getHeader` would accept any casing. This would avoid some confusion when writing scripts. From what I gather Bruno uses Axios, it seems Axios already lowercases all headers (see https://github.com/axios/axios/issues/413), so bruno might do that too.

abhishekp-bruno:
BRU-2763

Fixes: #1412

### Description

res.getHeader(name) is now case-insensitive, ensuring header lookups work regardless of casing. This aligns Bruno's behavior with Postman's pm.response.headers.get().

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Header lookups are now case-insensitive by normalizing header names before retrieving values, returning consistent results across different casing.
  * Non-string header name inputs now return `null` instead of producing an unexpected value.

* **Tests**
  * Added coverage for mixed-case header queries, array-valued headers (for example, `set-cookie`), missing headers, and non-string inputs to prevent regressions.

### DIFF
--- a/packages/bruno-js/tests/bruno-response-get-header.spec.js
+++ b/packages/bruno-js/tests/bruno-response-get-header.spec.js
@@ -1,0 +1,36 @@
+const BrunoResponse = require('../src/bruno-response');
+
+describe('BrunoResponse getHeader (case-insensitive lookup)', () => {
+  const createRes = () => ({
+    status: 200,
+    statusText: 'OK',
+    headers: {
+      'content-type': 'application/json',
+      'cache-control': 'no-cache',
+      'x-request-id': 'abc-123',
+      'set-cookie': ['a=1', 'b=2'],
+      'content-length': '512'
+    },
+    data: {}
+  });
+
+  it('looks up headers case-insensitively', () => {
+    const res = new BrunoResponse(createRes());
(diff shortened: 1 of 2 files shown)


morsh-bruno (approved):

packages/bruno-js/tests/bruno-response-get-header.spec.js
```diff
@@ -0,0 +1,43 @@
+const BrunoResponse = require('../src/bruno-response');
+
+describe('BrunoResponse getHeader (case-insensitive lookup)', () => {
+  const createRes = () => ({
```
helloanoop: I would recommend just having this as a variable instead of `createRes()`
```javascript
const mockResponse = {
    status: 200,
    statusText: 'OK',
    ....
}
```

abhishekp-bruno: switched to a plain mockResponse variable

vijayh-bruno (approved):

PR #8531, opened 2026-07-08
fix(mount): persist raw file content in file-index for file mode

chirag-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-3762)

In file mode, request nodes were not carrying their raw file content, leaving the `raw` field empty. This change persists the original file content through the mount pipeline so it is available on request nodes.

### Changes

- **parse-file job**: return the raw file `content` as `raw` alongside the parsed data.
- **file-index**: add a `raw TEXT` column (schema migration v2) and read/write it in `entries()`, `stage()`, and the on-disk read path. The migration forces a re-index of existing entries (`mtime = 0, hash = ''`) so the column gets populated.
- **mount manager**: propagate `raw` when updating in-memory state and staging index entries.
- **tree-builder**: include `raw` on request nodes (defaulting to `null`).
- **collections slice**: add `raw` to `FILE_DERIVED_REQUEST_FIELDS` so it is treated as a file-derived field.

### How to test

1. Open a collection in file mode.
2. Verify request nodes now carry their raw file content.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end preservation of original raw request/file text, alongside parsed content, across mounting, indexing, and request tree building.
  * File index now stores and returns raw content so views can access it for accurate editing and review.

* **Bug Fixes**
  * Improved reload/merge behavior to ensure raw request content is retained instead of being dropped when items are updated or reloaded.

### DIFF
--- a/packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
+++ b/packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
@@ -35,6 +35,7 @@ const FILE_DERIVED_REQUEST_FIELDS = [
   'request',
   'settings',
   'examples',
+  'raw',
   'filename',
   'pathname',
   'partial',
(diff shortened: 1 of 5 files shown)


sid-bruno (dismissed):

sid-bruno:
Tests

sid-bruno (approved):

PR #8533, opened 2026-07-08
feat(db): sqlite as a re-usable SDK for secondary queries

chirag-bruno:
### Description

BRU-3411

Adds `@usebruno/sqlite`, a reusable SQLite SDK for secondary/derived data (network history, timelines, caches, and other non-collection state) that keeps SQL out of the renderer and gives the app a typed, cached data layer.

**What it provides**

- **A new `@usebruno/sqlite` package** with two entry points:
  - `@usebruno/sqlite` (`/node`) — main process. Owns the DB (built-in `node:sqlite`), runs migrations, and executes statements.
  - `@usebruno/sqlite/web` — renderer. `SQLiteProvider` + `useSqliteQuery` / `useSqliteMutation` React Query hooks that call statements over IPC. SQL never ships to the renderer.
- **Migrations** authored as sequence-prefixed `.ts` files exporting `up`/`down`. They apply automatically in order on open, support downgrade rollback via DB-stored `down` statements, and are content-hashed into a `_migrations` table so an already-applied migration can't be silently modified.
- **Typed, named statements** authored with sqlc-style annotations (`-- name: X :one|:many|:exec`), compiled at build time into a typed registry. Params are object-only (named `@param` binding).
- **Automatic cache invalidation** — write statements broadcast `{ name, tables }` over a single IPC channel; the renderer invalidates any read query whose tables intersect the changed set, so main- and renderer-initiated writes both refresh live queries (across windows).
- **Migration verification** — `npm run migration:verify` replays every migration against a throwaway `VACUUM INTO` copy of a real database (set via `DB_PATH`), exercising them over real data without touching the source DB.

**Wiring**

- `bruno-electron`: registers SQLite IPC on startup and closes the DB gracefully on quit.
- `bruno-app`: wrapped in `<SQLiteProvider>` over the existing `ipcRenderer` bridge.

**Tests**: migration runner (apply, idempotent, drift detection, rollback), statement execution, VACUUM statement construction, and React Query invalidation behavior.

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added a new `@usebruno/sqlite` package and wired it into the desktop app with main-process IPC support and renderer integration.
  * Introduced migration management and typed, named SQL statements.
  * Added a React provider and React Query hooks for SQLite queries and mutations, including automatic cache refresh after mutations.
* **Bug Fixes**
  * Improved shutdown behavior to stop the SQLite IPC subsystem cleanly.
* **Documentation**
  * Added full documentation for setup, migrations, statement authoring, and hook usage.
* **Tests**
  * Added coverage for migrations, statement execution, and React Query invalidation behavior.


packages/bruno-electron/src/ipc/sqlite.js
```diff
@@ -0,0 +1,48 @@
+const path = require('path');
+const { app, ipcMain } = require('electron');
+const { createDatabase, registerSQLiteIpc, SQLITE_MUTATION_CHANNEL } = require('@usebruno/sqlite');
+
+let ipc = null;
+
+class SqliteIPC {
```
sid-bruno:
```suggestion
class SqliteEventModel {
```

packages/bruno-sqlite/scripts/lib/sources.ts
```diff
@@ -0,0 +1,125 @@
...
+const commentBody = (line: string): string => {
+  const trimmed = line.trim()
+  if (!trimmed.startsWith("--")) return ""
+  return trimmed.slice(2).trim().toUpperCase()
+}
+
+const parseMigration = (fileName: string, content: string): Migration => {
+  const lines = content.split("\n")
+  const upIdx = lines.findIndex(line => commentBody(line).startsWith("UP"))
+  const downIdx = lines.findIndex(line => commentBody(line).startsWith("DOWN"))
```
sid-bruno: Easy to mess up, just do `.up.sql` and `.down.sql` files for the same migration folder, the folder name becomes the migration

packages/bruno-sqlite/scripts/lib/sources.ts
sid-bruno, line 1: Would prefer an API / Arch like the following instead of messing with sql files and parsing them 
https://www.npmjs.com/package/sqlite-up

packages/bruno-sqlite/scripts/verify-migrations.ts
sid-bruno, line 1: Disconnect with a real life scenario with real data where existing rows might cause a crash. Do we really want to use this? 

Example: 
I make an existing column required and don't provide a seed value in the migration this script would pass because there's no data but the actual migration would fail since now there's empty col values which were marked as required.

packages/bruno-electron/src/ipc/sqlite.js
chirag-bruno: Will change this

packages/bruno-sqlite/scripts/lib/sources.ts
chirag-bruno: The reason I did this is to not have too many files in our migrations. I understand the mess up potential and hence the generate script. Is that fine or do we want strict separate files?

packages/bruno-sqlite/scripts/lib/sources.ts
chirag-bruno: We could do this. I wanted to keep the implementation as dependency free as possible. The inspiration was SQLC where sql is the source of truth and we generate domain specific implementations.

packages/bruno-sqlite/scripts/verify-migrations.ts
chirag-bruno: I mean, yeah. This is a valid point. The idea behind this script was to help users verify that their SQL statements for migrations are right (with respect to SQLITE). Is there any better approach?

packages/bruno-sqlite/scripts/lib/sources.ts
chirag-bruno: And the only parsing we are doing in an SQL file is for the comments.

packages/bruno-sqlite/scripts/verify-migrations.ts
```diff
@@ -0,0 +1,49 @@
...
+import { vacuumIntoStatement } from "./lib/sql";
+import { DatabaseSync } from 'node:sqlite';
+import os from 'node:os';
+import fs from 'node:fs';
+import path from 'node:path';
+
+
+const main = () => {
+  const dbPath = process.env.DB_PATH;
+  if (dbPath === undefined || dbPath.trim().length === 0) throw new Error('undefined environment variable DB_PATH. create a .env file.');
```
sid-bruno:
```suggestion
  if (dbPath === undefined || dbPath.trim().length === 0) throw new Error('undefined environment variable DB_PATH.');
```

packages/bruno-app/package.json
```diff
@@ -21,6 +21,7 @@
     "@usebruno/common": "0.1.0",
     "@usebruno/graphql-docs": "0.1.0",
     "@usebruno/schema": "0.7.0",
+    "@usebruno/sqlite": "0.1.0",
```
sid-bruno, line 25: Add @tanstack/react-query as an app runtime dependency; @usebruno/sqlite/web imports it in production.

packages/bruno-sqlite/package.json
```diff
@@ -0,0 +1,75 @@
...
+  ],
+  "scripts": {
+    "clean": "rimraf dist",
+    "generate": "tsx scripts/generate-artifacts.ts",
+    "prepare": "npm run generate",
+    "prebuild": "npm run clean && npm run generate",
+    "build": "rollup -c",
+    "watch": "rollup -c -w",
+    "pretest": "npm run generate",
+    "test": "jest",
```
sid-bruno: Run sqlite tests with Electron bundled Node or enable --experimental-sqlite so plain npm test works consistently. 

Since the `.nvmrc` and electron version that's used has different node versions

packages/bruno-electron/src/ipc/sqlite.js
```diff
@@ -0,0 +1,50 @@
...
+  constructor(window) {
+    this._window = window;
+    const { db, statements } = createDatabase(path.join(app.getPath('userData'), 'bruno.db'), {
+      onMutation: (event) => {
+        this._window?.webContents?.send(SQLITE_MUTATION_CHANNEL, event);
+      }
+    });
+    this._db = db;
+    this._statements = statements;
+    registerSQLiteIpc(ipcMain, statements);
```
sid-bruno, line 20: Make SQLite IPC registration idempotent or remove the handler on shutdown to avoid duplicate or stale handlers. I can't run the same register call twice, will crash if done accidentally in something that happens in runtime, let's add a check or avoid it altogether

packages/bruno-sqlite/scripts/lib/sources.ts
```diff
@@ -0,0 +1,129 @@
...
+const { Parser } = require("node-sql-parser")
+const sqlParser = new Parser()
+
+const extractTables = (sql: string): string[] => {
+  try {
+    let list: string[] = sqlParser.tableList(sql, { database: "Sqlite" })
+    // This is in the form of statement::db::table, hence the second index access
+    return Array.from(list.map(entry => entry.split("::")[2]))
+  } catch {
+    return []
```
sid-bruno: Do not silently return no tables on parse failure; fail or warn so cache invalidation cannot miss affected reads.

packages/bruno-app/package.json
chirag-bruno: `@tanstack/react-query` is a dependency of the sqlite package. Sqlite package wraps few things around `react-query`. Should it become a direct dependency of `package/bruno-app`? The build step for `bruno-sqlite` already bundles `react-query`, isn't it?

chirag-bruno: Added it to package.json. Can ignore my response.

packages/bruno-electron/src/ipc/sqlite.js
chirag-bruno: Done. added a `if (ipc) return;` block in the `registerSQLiteIpc`.

packages/bruno-sqlite/scripts/lib/sources.ts
chirag-bruno: Added a warning.

packages/bruno-sqlite/package.json
chirag-bruno: Done

sid-bruno (dismissed):

sid-bruno (approved):

sachin-thakur-bruno (approved):

PR #8585, opened 2026-07-13
feat(code-mirror-search): add replace and replace-all with search position, debouncing and tests

sachin-thakur-bruno:
### Description

[JIRA](https://usebruno.atlassian.net/browse/BRU-2623)

Curently users can only search a text in codemirror. There is no option to replace the selected text.

Adding replace and replace all feature in code mirror. Users can select a text for search and can replace the selected text by replace one-by-one or replace all. Adding the keyboard shortcuts for opening the serach and replace.

https://github.com/user-attachments/assets/41d44a65-8acc-4b56-97f3-4a664fb55739

### CodeMirrorSearch/index.js

**Replace functionality**
- Added `replaceText` state and `replaceVisible` toggle
- `handleReplace` — replaces the current match via `editor.replaceRange`, invalidates the search cache, and re-runs search at the same index
- `handleReplaceAll` — replaces all matches in reverse order (to preserve character positions) inside `editor.operation()` for a single undo step

**Layout**
- Chevron (`IconChevronRight`) extracted as a full-height left column — spans both rows when expanded, matching VS Code's widget
- Rows restructured into `.toggle-replace-btn` + `.search-replace-rows` (column of `.search-row` / `.replace-row`)
- Replace action buttons use `IconReplace` and `IconArrowsExchange2` with `ToolHint` tooltips

**Keyboard & focus**
- Native `keydown` listener attached in the **capture phase** (`{ capture: true }`) on the container — ensures `Escape` is intercepted before CodeMirror or the browser can act on it
- `type="button"` added to all buttons for correct focusability

**Imperative handle** — exposes four methods via `useImperativeHandle`:

| Method | Purpose |
|--------|---------|
| `focus()` | Focus the search input |
| `setSearch(text)` | Set search text, reset match index, focus and select input |
| `openReplace()` | Expand replace row and focus replace input |
| `close()` | Clean up marks, collapse replace, call `onClose` |

---

### CodeMirrorSearch/StyledWrapper.js

- Container changed to `flex-direction: row` — chevron column on the left, rows column on the right
- `.toggle-replace-btn` uses `align-self: stretch` for full-height
- Each input has its own `border` box, fixed `180px` width — both rows visually identical
- `.search-row` gets a `border-bottom` only when replace is expanded
- Removed `.searchbar-text-btn` styles (replaced by icon buttons)

---

### CodeEditor/index.js

**`Cmd/Ctrl+H`** — replaced `'replace'` string command with a handler that opens the search bar and calls `searchBarRef.current?.openReplace()`

**`Cmd/Ctrl+F`** — updated to check `cm.getSelection()`:
- Selection present → calls `setSearch(selectedText)` to prefill and re-run search (works whether bar is open or closed)
- No selection → calls `focus()` to move cursor to the input without clearing existing search text

**`Esc` extraKey** — updated from bare `setState` to `searchBarRef.current?.close()` so marks are properly cleaned up when Esc is pressed while the editor has focus

### 1. Only highlight what's visible

The search still finds **all matches** in the document, but it only creates highlights (`markText()`) for matches that are currently visible in the editor (plus a small buffer of ±100 lines). The currently selected match is always highlighted, even if it's outside the viewport.

For example, in a 10,000-line file with 500 matches, the editor typically creates only **5–10 highlights instead of 500**, significantly reducing DOM work. As you scroll, old highlights are removed and new ones are drawn for the visible area.

**Files**
- `components/CodeMirrorSearch/markingUtils.js`

---

### 2. Cache search results

Scanning the entire document with CodeMirror's `getSearchCursor()` can be expensive on large files.

Search results are cached and reused until something that affects the results changes:
- Document content
- Search text
- Regex mode
- Case sensitivity
- Whole-word mode

This means actions like **Next**, **Previous**, scrolling, or component re-renders reuse the existing match list instead of rescanning the document every time.

**Files**
- `components/CodeMirrorSearch/searchUtils.js` (`createCacheKey`)
- `components/CodeMirrorSearch/index.js` (`searchCacheKey`, `isCacheHit`)

---

### 3. Debounce search while typing

Instead of performing a full document scan on every keystroke, the search waits until the user stops typing for **250 ms** before running.

Typing stays responsive because intermediate keystrokes only update the input field without triggering an expensive search.

**Files**
- `components/CodeMirrorSearch/index.js` (`useDebounce(searchText, 250)`)

---

### 4. Jump directly to visible matches

When drawing highlights, the code uses a **binary search** to quickly locate the first match inside the current viewport instead of checking every match from the beginning.

This reduces the work from scanning the entire match list to only processing the matches that are actually visible.

**Files**
- `components/CodeMirrorSearch/markingUtils.js`

---

### 5. Limit extremely large searches

To avoid freezing the UI on pathological searches (for example, searching for a single character in a very large minified file), the search stops after **99,999 matches**.

This provides a reasonable safety limit while still supporting very large result sets.

**Files**
- `components/CodeMirrorSearch/searchUtils.js` (`MAX_MATCHES`)

---

### 6. Scroll only when it makes sense

Searching no longer automatically scrolls the editor every time results are updated.

The editor scrolls only when explicitly requested, such as:
- Next / Previous
- Replace
- Opening search with selected text

It does **not** scroll during:
- Debounced searches while typing
- Automatic re-scans after document changes
- Reopening the search bar

This prevents the editor from unexpectedly jumping away from the user's current position.

**Files**
- `components/CodeMirrorSearch/index.js` (`doSearch(..., shouldScroll = false)`)

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved editor Search/Replace UX: opens based on current selection/cursor, adds robust match navigation, and enhances Replace/Replace all interactions (with read-only safeguards).
  * Refined Search/Replace layout and controls, including a dedicated replace toggle and clearer keyboard behavior.
* **Bug Fixes**
  * Stabilized match counting/highlighting through typing and document edits, with reliable close/dismiss behavior (including Escape).
* **Tests**
  * Added unit tests for search/replace utilities and the search component, plus Playwright E2E coverage for large-document behavior, state persistence, and viewport stability.
* **Documentation**
  * Added a targeted test identifier for the documentation editor UI.

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -103,17 +103,37 @@ class CodeEditor extends React.Component {
       theme: this.props.theme === 'dark' ? 'monokai' : 'default',
       extraKeys: {
         'Cmd-F': (cm) => {
+          const selected = cm.getSelection();
           this.setState({ searchBarVisible: true }, () => {
-            this.searchBarRef.current?.focus();
+            if (selected) {
+              this.searchBarRef.current?.setSearch(selected, cm.getCursor('from'));
+            } else {
+              this.searchBarRef.current?.focus();
+            }
           });
         },
         'Ctrl-F': (cm) => {
+          const selected = cm.getSelection();
           this.setState({ searchBarVisible: true }, () => {
-            this.searchBarRef.current?.focus();
+            if (selected) {
+              this.searchBarRef.current?.setSearch(selected, cm.getCursor('from'));
+            } else {
+              this.searchBarRef.current?.focus();
+            }
           });
         },
-        'Cmd-H': this.props.readOnly ? false : 'replace',
-        'Ctrl-H': this.props.readOnly ? false : 'replace',
+        'Cmd-H': this.props.readOnly ? false : (cm) => {
+          this.setState({ searchBarVisible: true }, () => {
+            this.searchBarRef.current?.focus();
+            this.searchBarRef.current?.openReplace();
+          });
+        },
+        'Ctrl-H': this.props.readOnly ? false : (cm) => {
+          this.setState({ searchBarVisible: true }, () => {
+            this.searchBarRef.current?.focus();
+            this.searchBarRef.current?.openReplace();
+          });
+        },
         'Cmd-Enter': runShortcut,
         'Ctrl-Enter': runShortcut,
         'Tab': function (cm) {
@@ -148,7 +168,7 @@ class CodeEditor extends React.Component {
         },
         'Esc': () => {
           if (this.state.searchBarVisible) {
-            this.setState({ searchBarVisible: false });
+            this.searchBarRef.current?.close();
           }
         }
       },
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -21,6 +21,7 @@ import { setupLinkAware } from 'utils/codemirror/linkAware';
 import { setupLintErrorTooltip } from 'utils/codemirror/lint-errors';
 import { setupCodeMirrorResizeRefresh } from 'utils/codemirror/resize';
 import CodeMirrorSearch from 'components/CodeMirrorSearch/index';
+import { buildSearchKeyBindings } from 'components/CodeMirrorSearch/searchKeyBindings';
 import {
   applyEditorState,
   captureEditorState,
@@ -102,52 +103,12 @@ class CodeEditor extends React.Component {
       scrollbarStyle: 'overlay',
       theme: this.props.theme === 'dark' ? 'monokai' : 'default',
       extraKeys: {
-        'Cmd-F': (cm) => {
-          const selected = cm.getSelection();
-          const cursor = cm.getCursor('from');
-          this.setState({ searchBarVisible: true }, () => {
-            if (selected) {
-              this.searchBarRef.current?.setSearch(selected, cursor);
-            } else {
-              this.searchBarRef.current?.focusAtCursor(cursor);
-            }
-          });
-        },
-        'Ctrl-F': (cm) => {
-          const selected = cm.getSelection();
-          const cursor = cm.getCursor('from');
-          this.setState({ searchBarVisible: true }, () => {
-            if (selected) {
-              this.searchBarRef.current?.setSearch(selected, cursor);
-            } else {
-              this.searchBarRef.current?.focusAtCursor(cursor);
-            }
-          });
-        },
-        'Cmd-H': this.props.readOnly ? false : () => {
-          this.setState({ searchBarVisible: true }, () => {
-            this.searchBarRef.current?.focus();
-            this.searchBarRef.current?.openReplace();
-          });
-        },
-        'Ctrl-H': this.props.readOnly ? false : () => {
-          this.setState({ searchBarVisible: true }, () => {
-            this.searchBarRef.current?.focus();
-            this.searchBarRef.current?.openReplace();
-          });
-        },
-        'Cmd-Alt-F': this.props.readOnly ? false : () => {
-          this.setState({ searchBarVisible: true }, () => {
-            this.searchBarRef.current?.focus();
-            this.searchBarRef.current?.openReplace();
-          });
-        },
-        'Ctrl-Alt-F': this.props.readOnly ? false : () => {
-          this.setState({ searchBarVisible: true }, () => {
-            this.searchBarRef.current?.focus();
-            this.searchBarRef.current?.openReplace();
-          });
-        },
+        ...buildSearchKeyBindings({
+          setState: (update, cb) => this.setState(update, cb),
+          searchBarRef: this.searchBarRef,
+          isSearchBarVisible: () => this.state.searchBarVisible,
+          readOnly: this.props.readOnly
+        }),
         'Cmd-Enter': runShortcut,
         'Ctrl-Enter': runShortcut,
         'Tab': function (cm) {
@@ -178,11 +139,6 @@ class CodeEditor extends React.Component {
             this.editor.toggleComment({ lineComment: '//', blockComment: '/*' });
           } else {
             this.editor.toggleComment();
-          }
-        },
-        'Esc': () => {
-          if (this.state.searchBarVisible) {
-            this.searchBarRef.current?.close();
           }
         }
       },
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -104,31 +104,45 @@ class CodeEditor extends React.Component {
       extraKeys: {
         'Cmd-F': (cm) => {
           const selected = cm.getSelection();
+          const cursor = cm.getCursor('from');
           this.setState({ searchBarVisible: true }, () => {
             if (selected) {
-              this.searchBarRef.current?.setSearch(selected, cm.getCursor('from'));
+              this.searchBarRef.current?.setSearch(selected, cursor);
             } else {
-              this.searchBarRef.current?.focus();
+              this.searchBarRef.current?.focusAtCursor(cursor);
             }
           });
         },
         'Ctrl-F': (cm) => {
           const selected = cm.getSelection();
+          const cursor = cm.getCursor('from');
           this.setState({ searchBarVisible: true }, () => {
             if (selected) {
-              this.searchBarRef.current?.setSearch(selected, cm.getCursor('from'));
+              this.searchBarRef.current?.setSearch(selected, cursor);
             } else {
-              this.searchBarRef.current?.focus();
+              this.searchBarRef.current?.focusAtCursor(cursor);
             }
           });
         },
-        'Cmd-H': this.props.readOnly ? false : (cm) => {
+        'Cmd-H': this.props.readOnly ? false : () => {
           this.setState({ searchBarVisible: true }, () => {
             this.searchBarRef.current?.focus();
             this.searchBarRef.current?.openReplace();
           });
         },
-        'Ctrl-H': this.props.readOnly ? false : (cm) => {
+        'Ctrl-H': this.props.readOnly ? false : () => {
+          this.setState({ searchBarVisible: true }, () => {
+            this.searchBarRef.current?.focus();
+            this.searchBarRef.current?.openReplace();
+          });
+        },
+        'Cmd-Alt-F': this.props.readOnly ? false : () => {
+          this.setState({ searchBarVisible: true }, () => {
+            this.searchBarRef.current?.focus();
+            this.searchBarRef.current?.openReplace();
+          });
+        },
+        'Ctrl-Alt-F': this.props.readOnly ? false : () => {
           this.setState({ searchBarVisible: true }, () => {
             this.searchBarRef.current?.focus();
             this.searchBarRef.current?.openReplace();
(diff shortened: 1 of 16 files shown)


tests/code-editor/search-replace.spec.ts
```diff
@@ -0,0 +1,362 @@
...
+}
+
+async function closeSearchBar(page: Page) {
+  await page.keyboard.press('Escape');
+  await page.getByTestId(EDITOR_ID).getByTestId('search-bar').waitFor({ state: 'hidden' });
+}
+
+async function expectMatchCount(page: Page, expected: string) {
+  await expect(page.getByTestId(EDITOR_ID).getByTestId('match-count')).toHaveText(expected, { timeout: 1500 });
+}
```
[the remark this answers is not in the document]
sachin-thakur-bruno: updated the tests to follow the playwright guidelines. 
Removed all 6 inline helper functions
Every locator access goes through buildCodeEditorSearchLocators(page, EDITOR_ID)
Every interaction goes through the page module actions
expectMatchCount stays in the spec as a local helper since it calls expect() — per the rule that page modules must not call expect

packages/bruno-app/src/components/CodeMirrorSearch/index.js
```diff
@@ -71,125 +47,222 @@ const CodeMirrorSearch = forwardRef(({ visible, editor, onClose }, ref) => {
...
+    if (!text) {
       setMatchCount(0);
       setMatchIndex(0);
+      currentMatchIndex.current = 0;
       searchMatches.current = [];
-      searchMarks.current.forEach((mark) => mark.clear());
-      searchMarks.current = [];
+      clearMarks(searchMarks.current);
       return;
     }
```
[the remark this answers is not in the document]
sachin-thakur-bruno: added  searchCacheKey.current = '' to reset the cache key for stale closure bug.

packages/bruno-app/src/components/CodeMirrorSearch/index.js
```diff
@@ -71,125 +47,222 @@ const CodeMirrorSearch = forwardRef(({ visible, editor, onClose }, ref) => {
...
+      if (initialIndexRef.current.forText === debouncedSearchText) {
+        const idx = initialIndexRef.current.idx;
+        initialIndexRef.current = null;
+        doSearch(debouncedSearchText, idx);
+      }
+    } else {
+      const cursor = editor?.getCursor('from');
+      const cursorLine = cursor ? cursor.line : null;
+      doSearch(debouncedSearchText, 0, cursorLine);
+    }
```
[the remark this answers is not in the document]
sachin-thakur-bruno: added a else block where a mismatched initialIndexRef the same as an absent one. clears the stale ref and runs the normal cursor-anchored search.

packages/bruno-app/src/components/CodeMirrorSearch/searchUtils.js
```diff
@@ -0,0 +1,60 @@
+const MAX_MATCHES = 99_999;
```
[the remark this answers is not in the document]
sachin-thakur-bruno: added an optional limit param, pass Infinity from handleReplaceAll.

packages/bruno-app/src/components/CodeMirrorSearch/index.js
```diff
@@ -210,48 +283,130 @@ const CodeMirrorSearch = forwardRef(({ visible, editor, onClose }, ref) => {
...
-    doSearch(next);
+    doSearch(debouncedSearchText, next, null, true);
   };
 
   const handlePrev = () => {
     if (!searchMatches.current || !searchMatches.current.length) return;
     const prev = (matchIndex - 1 + searchMatches.current.length) % searchMatches.current.length;
-    doSearch(prev);
+    doSearch(debouncedSearchText, prev, null, true);
   };
```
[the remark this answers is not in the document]
sachin-thakur-bruno: The debounced window is very small (250ms) for user to type and then press enter. But added a early return in handleNext and handlePrev if debouncing is happening.

packages/bruno-app/src/components/CodeEditor/index.js
```diff
@@ -103,17 +103,51 @@ class CodeEditor extends React.Component {
...
+            this.searchBarRef.current?.openReplace();
+          });
+        },
+        'Cmd-Alt-F': this.props.readOnly ? false : () => {
+          this.setState({ searchBarVisible: true }, () => {
+            this.searchBarRef.current?.focus();
+            this.searchBarRef.current?.openReplace();
+          });
+        },
+        'Ctrl-Alt-F': this.props.readOnly ? false : () => {
```
chirag-bruno: The actions for `(Cmd+F or Ctrl+F), (Cmd+H or Ctrl+H), and (Cmd+Alt+F) or (Ctrl+Alt+F)` have the same function implementation. Can we extract these into methods which takes in the `cm` instance and runs something. Helps us not drift the implementation across two bindings when some logic changes.

packages/bruno-app/src/components/CodeEditor/index.js
```diff
@@ -148,7 +182,7 @@ class CodeEditor extends React.Component {
         },
         'Esc': () => {
           if (this.state.searchBarVisible) {
-            this.setState({ searchBarVisible: false });
+            this.searchBarRef.current?.close();
```
chirag-bruno: Do we need the `null/undefined` check here with `?`? Isn't it always available when the searchBarVisible is true?

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
```diff
@@ -0,0 +1,334 @@
...
+  font: { size: { base: '13px', xs: '11px' } },
+  brand: '#007acc',
+  codemirror: { border: '#555', searchLineHighlightCurrent: 'rgba(255,255,0,0.1)' }
+};
+
+function makeMockEditor(matchResults = []) {
+  const marks = [];
+  return {
+    getSearchCursor: jest.fn(() => {
+      let i = -1;
```
chirag-bruno, line 22: This is a doubt that I have. Maybe I am wrong here. Doesn't invoking `getSearchCursor` again reset the value of `i` to `-1`? Should we have it outside the `getSearchCursor` method? This is just to ensure that the state remains stable. Let me know if I am wrong here and my assumptions are a little off.

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
```diff
@@ -0,0 +1,334 @@
...
+        <ThemeProvider theme={theme}>
+          <CodeMirrorSearch visible={false} editor={makeMockEditor()} onClose={jest.fn()} ref={createRef()} />
+        </ThemeProvider>
+      );
+      expect(container.firstChild).toBeNull();
+    });
+
+    it('renders search input when visible', () => {
+      renderSearch();
+      expect(screen.getByPlaceholderText('Search...')).toBeInTheDocument();
```
chirag-bruno: Placeholders are fine, but is it better to have `data-testid`? Maybe useful for later E2E tests? Same for every other place we are doing a DOM search via placeholders and text.

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
```diff
@@ -0,0 +1,334 @@
...
+      renderSearch();
+      fireEvent.click(screen.getByTitle('Show replace'));
+      expect(screen.getByPlaceholderText('Replace...')).toBeInTheDocument();
+    });
+  });
+
+  describe('result count', () => {
+    it('shows "0 results" when search is empty', () => {
+      renderSearch();
+      expect(screen.getByText('0 results')).toBeInTheDocument();
```
chirag-bruno: Can this test be more precise. For whatever reason, if we have a text saying "0 results" in our editor instance, this will fail. This is not a case in this specific test but can be an issue if the test evolves. Maybe we can target the specific text label which shows the results.

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
```diff
@@ -0,0 +1,334 @@
...
+
+    it('shows "1 / 3" for first match of 3', () => {
+      const matches = [
+        { from: { line: 0, ch: 0 }, to: { line: 0, ch: 4 } },
+        { from: { line: 1, ch: 0 }, to: { line: 1, ch: 4 } },
+        { from: { line: 2, ch: 0 }, to: { line: 2, ch: 4 } }
+      ];
+      renderSearch({ editor: makeMockEditor(matches) });
+      typeSearch('test');
+      expect(screen.getByText('1 / 3')).toBeInTheDocument();
```
chirag-bruno: Same here, maybe we can do a more targeted search than to depend on the screen instance. Let me know your thoughts.

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
```diff
@@ -0,0 +1,334 @@
...
+      setupWithMatches(3);
+      fireEvent.keyDown(screen.getByPlaceholderText('Search...'), { key: 'Enter', shiftKey: true });
+      expect(screen.getByText('3 / 3')).toBeInTheDocument();
+    });
+  });
+
+  describe('imperative handle', () => {
+    it('focus() focuses the search input', () => {
+      const ref = createRef();
+      renderSearch({}, ref);
```
chirag-bruno, line 180: Am I missing something here? `renderSearch` renders the `CodeMirror` instance. But `onFocus`, why is the search bar appearing?  Is this because of some property in makeMockEditor?

packages/bruno-app/src/components/CodeEditor/index.js
sachin-thakur-bruno: Yes it will be there but since we don't have typescript currently here its added for safety. Also ?. costs nothing at runtime and protects against any edge case where the ref hasn't been committed yet

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
sachin-thakur-bruno: actually this is the correct behavior, since each getSearchCursor call should represent an independent cursor scan from the top of the document.

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
sachin-thakur-bruno: When the parent opens the search bar (sets visible={true}), it also needs to move keyboard focus into the search input so the user can start typing immediately. It does that by calling ref.current.focus(). 
So the test is for: calling focus() on the ref moves focus to the search input

packages/bruno-app/src/components/CodeEditor/index.js
sachin-thakur-bruno: Yes sure good idea. created a util and used it from there

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
sachin-thakur-bruno: using the test ids now

packages/bruno-app/src/components/CodeMirrorSearch/CodeMirrorSearch.spec.jsx
sachin-thakur-bruno: using the test ids now for these assertions.

PR #8588, opened 2026-07-13
fix: avoid network probe in PAC myIpAddress

### LINKED ISSUE #8576: System Proxy (PAC) mode is ~2s slower per request due to myIpAddress() calling out to 8.8.8.8:53

### This bug is:

- [ ] annoying

### Bruno version

3.5.1

### Operating System

Windows 11 Enterprise 26100.8655

### Describe the bug

When System Proxy is configured with a PAC URL, every single request takes an extra **~2 seconds** compared to configuring the resolved proxy directly. This happens even though the PAC script itself (compiled) is cached for 5 minutes per `packages/bruno-requests/src/utils/pac-resolver.ts`.

**Root cause**
`FindProxyForURL()` is re-executed on every request (only the compiled script/quickjs context is cached, not the per-request result). Our corporate PAC script calls `myIpAddress()` unconditionally at the top of the script:
```js
function FindProxyForURL(url, host) {
var myip = myIpAddress();
...
}
```

`pac-resolver`'s `myIpAddress()` implementation (`packages/pac-resolver/src/myIpAddress.ts`) determines the local IP by opening a real TCP connection to `8.8.8.8:53` (Google Public DNS) and inspecting the local socket address, falling back to `os.networkInterfaces()` only on error:
```ts
export default async function myIpAddress(): Promise<string> {
return new Promise((resolve, reject) => {
const socket = net.connect({ host: '8.8.8.8', port: 53 });
...
});
}
```

In many corporate networks, outbound TCP to `8.8.8.8:53` is blocked/dropped by the firewall (while ICMP ping to the same host succeeds), so this connection attempt reliably takes ~2 seconds to fail before falling back — and this cost is paid on **every single HTTP request** made through Bruno while System Proxy (PAC) mode is enabled.

### Steps to reproduce

1. Configure OS/Bruno System Proxy to use a PAC URL whose script calls `myIpAddress()` unconditionally.
2. Ensure the network blocks outbound TCP 53 to `8.8.8.8` (common in corporate firewalls; `ping 8.8.8.8` succeeds but `Test-NetConnection 8.8.8.8 -Port 53` fails).
3. Send any request in Bruno with System Proxy (PAC) enabled — observe ~2s added latency.
4. Disable PAC and set the resolved proxy directly (e.g. `PROXY proxy.example.com:8080`) for the same host — latency drops to ~50ms.

# Evidence
Reproduced the exact `myIpAddress()` logic in isolation:
```powershell
node -e "const net=require('net'); const t=Date.now(); const s=net.connect({host:'8.8.8.8',port:53}); s.once('error',(e)=>{console.log('error after', Date.now()-t, 'ms:', e.message); process.exit();}); s.once('connect',()=>{console.log('connected after', Date.now()-t, 'ms'); s.destroy(); process.exit();});"
```

Result:
```
error after 2058 ms: connect ECONNREFUSED 8.8.8.8:53
```
This ~2058ms matches the observed per-request overhead in Bruno almost exactly.

# Suggested fix directions
1. Change `pac-resolver`'s `myIpAddress()` to avoid a real outbound network connection (e.g. use `os.networkInterfaces()` directly, or add a short timeout so the fallback triggers quickly instead of waiting for a slow/blocked connect).
2. And/or cache the PAC evaluation result (`FindProxyForURL` return value) per-host for a short TTL in Bruno, so `myIpAddress()`/other PAC functions aren't re-invoked on every single request.

### Collection to reproduce

N/A (network/proxy configuration issue, not request-specific)

N/A

cyphercodes:
### Description

Fixes #8576.

This overrides `pac-resolver`'s PAC `myIpAddress()` helper with a local network-interface lookup when Bruno compiles PAC scripts. That avoids the upstream helper's outbound TCP probe to `8.8.8.8:53`, which can add ~2 seconds to each PAC evaluation on networks where that connection is blocked or dropped.

Tests updated to assert the PAC resolver receives the `myIpAddress` sandbox override and that it selects a non-internal IPv4 address without using the upstream network probe.

Tests run:
- `npx eslint packages/bruno-requests/src/utils/pac-resolver.ts packages/bruno-requests/src/utils/pac-resolver.spec.ts`
- `npm run test --workspace=packages/bruno-requests -- pac-resolver.spec.ts --runInBand`
- `npm run test --workspace=packages/bruno-requests -- --runInBand`
- `npm run build --workspace=packages/bruno-requests` (completed successfully; Rollup/TypeScript warnings from `@faker-js/faker` declaration syntax were emitted)

#### Contribution Checklist:

- [ ] **I have added screenshots or gifs to help explain the change if applicable.** (N/A)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PAC proxy resolution by providing the local IPv4 address to PAC scripts.
  * Added a fallback to `127.0.0.1` when no suitable local network address is available.
  * Enhanced handling for PAC files loaded from downloads and local file sources.

### DIFF
--- a/packages/bruno-requests/src/utils/pac-resolver.spec.ts
+++ b/packages/bruno-requests/src/utils/pac-resolver.spec.ts
@@ -9,6 +9,12 @@ describe('pac-resolver (shared)', () => {
     jest.clearAllMocks();
   });
 
+  const expectedPacOptions = expect.objectContaining({
+    sandbox: expect.objectContaining({
+      myIpAddress: expect.any(Function)
+    })
+  });
+
   /** Mock pac-resolver (v7: { createPacResolver }) and quickjs-emscripten */
   const setupPacMocks = (resolverFn: (...args: any[]) => Promise<any> = async () => 'PROXY p.example:8080; DIRECT') => {
     jest.doMock('quickjs-emscripten', () => ({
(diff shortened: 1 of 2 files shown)


helloanoop:
Thanks for the PR @cyphercodes !

pooja-bruno:
jira: BRU-3874

sid-bruno (approved):

PR #8608, opened 2026-07-14
feat(migration): enhance folder migration process to preserve display order

utkarsh-bruno:
### Description

- Introduced `stampFolderSeqFromDisplayOrder` to maintain the sequence of folders during migration.
- Updated folder parsing and stringification to conditionally include the sequence number only when set.
- Improved handling of folder data during the migration process, ensuring proper YML file generation and cleanup of .bru files.

BRU-3854

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved folder migration so saved folder order now matches the order shown before migration.
  * Preserved custom folder ordering within each parent directory during migration.
  * Prevented missing/invalid folder sequence values from being replaced with a default, maintaining alphabetical ordering when no custom order is set.
* **Tests**
  * Added Jest coverage to verify sequence stamping aligned to display order, including dragged items and independent ordering per parent folder.

### DIFF
--- a/packages/bruno-filestore/src/formats/yml/parseFolder.ts
+++ b/packages/bruno-filestore/src/formats/yml/parseFolder.ts
@@ -17,7 +17,7 @@ const parseFolder = (ymlString: string): FolderRoot => {
     const folderRoot: FolderRoot = {
       meta: {
         name: ensureString(info?.name, 'Untitled Folder'),
-        seq: info?.seq || 1
+        ...(typeof info?.seq === 'number' ? { seq: info.seq } : {})
       },
       request: {
         headers: [],
--- a/packages/bruno-filestore/src/formats/yml/stringifyFolder.ts
+++ b/packages/bruno-filestore/src/formats/yml/stringifyFolder.ts
@@ -37,12 +37,14 @@ const stringifyFolder = (folderRoot: FolderRoot): string => {
   try {
     const ocFolder: Folder = {};
 
-    // info block
+    // omit seq when unset so alphabetical folder order is preserved
     const info: FolderInfo = {
       name: folderRoot.meta?.name || 'Untitled Folder',
-      type: 'folder',
-      seq: folderRoot.meta?.seq || 1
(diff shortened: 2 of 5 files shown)


packages/bruno-filestore/src/formats/yml/parseFolder.ts
```diff
@@ -17,7 +17,7 @@ const parseFolder = (ymlString: string): FolderRoot => {
     const folderRoot: FolderRoot = {
       meta: {
         name: ensureString(info?.name, 'Untitled Folder'),
-        seq: info?.seq || 1
+        ...(typeof info?.seq === 'number' ? { seq: info.seq } : {})
```
helloanoop, line 20: @utkarsh-bruno Why are we updating this piece of code? I don't think this is what is causing the bug.

I completely agree with the type check to ensure we only use the incoming `seq` if it's a number.

My concern is with removing the default sequence value of `1`.

Keep in mind that this is part of the parsing step, where we parse the file and construct the Bru schema. I think it's beneficial to default `seq` to `1` when it isn't specified. That guarantees that every parsed object has a valid `seq`, allowing the UI layer to safely rely on it without having to handle missing values everywhere.

packages/bruno-filestore/src/formats/yml/stringifyFolder.ts
```diff
@@ -37,12 +37,14 @@ const stringifyFolder = (folderRoot: FolderRoot): string => {
   try {
     const ocFolder: Folder = {};
 
-    // info block
+    // omit seq when unset so alphabetical folder order is preserved
```
helloanoop, line 40:
```javascript
// omit seq when unset so alphabetical folder order is preserved
```

This comment is in the wrong place. It should be above the seq check if block.
As per my earlier comment, I am not sure if changing the piece of code here is the solution.

The migration code should put the preserve the sequence numbers at the time of migration properly.

PS: I have limited context, I could be wrong here.

PR #8618, opened 2026-07-15
fix(environment-table)/autosave removes the already added characters in environment table

sachin-thakur-bruno:
### Description

BRU-3655

With auto save enabled  and user is typing in Environment table, unexpectedly move the cursor. This caused characters to be inserted in the wrong place, making it feel like recent input had been lost.

This PR fixes the issue by distinguishing the app's own autosave updates from real external changes. The form now refreshes only when something has actually changed outside the current editing session, so typing remains uninterrupted.

### Problem
 
`EnvironmentVariablesTable` uses Formik with `enableReinitialize: true` and
`initialValues` derived from `environment.variables` in Redux.
 
Every autosave writes the environment back to Redux with a **new array
reference**. Because `enableReinitialize` was on and the old code produced a new
`initialValues` reference whenever the **content** differed, each autosave reset
the form while the user was still typing. During fast typing the saved snapshot
legitimately *lags* the live form, so the content always looked "different" — and
the reset fired, moving the cursor.
 
The core issue is that `environment.variables` serves three roles that are
indistinguishable downstream:
 
1. persisted/saved state,
2. the runtime-mutation channel (`bru.setEnvVar`), and
3. the reinitialization source for the open editor.
Formik couldn't tell "the app just saved my own edits" from "something external
changed the environment."
 
### Root cause
 
`enableReinitialize` reinitializes whenever the `initialValues` **reference**
changes. The previous gate returned the same reference only when *stripped
content* was equal — which never held during active typing, so the form reset on
every autosave.

### Fix
 
A revision marker recorded at the mutation site, consumed by the component.
 
**1. The reducer marks a genuine external mutation (`slices/collections/index.js`)**
 
- `scriptEnvironmentUpdateEvent` (`bru.setEnvVar`) — bumps `_revision`.
- `saveEnvironment` (the self-save reducer) — does **not** bump `_revision`.
```js
// External mutation — bump _revision so the open editor reinitializes (self-saves don't).
activeEnvironment._revision = (activeEnvironment._revision || 0) + 1;
```
 
**2. The save thunk keeps the marker out of persistence (`actions.js`)**
 
`_revision` is transient in-memory state, so it is stripped before schema
validation and the disk write. This is required, not just tidy: `environmentSchema`
is `.noUnknown(true).strict()`, so a stray `_revision` would make `validate()` throw
on the next save.
 
**3. The component gates reinitialization on `_revision` (`EnvironmentVariablesTable/index.js`)**
 
`enableReinitialize` is **kept**. The `initialValues` memo now returns the
previous reference unless `_revision` changed, so:
 
- self-save echoes → same reference → **no reinit** → cursor stays put;

### Testing

Existing `draft-env-merge.spec.js` and `environment-events.spec.js` still pass
(40/40 across the three suites).
 
**E2E** (`autosave-typing/autosave-env-typing.spec.ts`) enables Auto Save at
500 ms and types into Name, Value, and Description with a deliberate 600 ms
per-keystroke delay, asserting each field retains its text and that the values
are persisted to disk. This test **failed before** the fix and passes after.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment variable editing so autosave no longer interrupts typing or resets fields unexpectedly.
  * Environment changes from scripts or external file updates now refresh the editor only when content has actually changed.
  * Prevented temporary revision data from being included when saving environment files.

* **Tests**
  * Added coverage for autosaving edited environment variable names, values, and descriptions, including verifying persisted file contents.


tests/environments/environment-tabs/autosave-env-typing.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+      await expect(environment.varRowValueLine('apiKey')).toContainText('token12345');
+      await expect(nameInput).toHaveValue('apiKey');
+    });
+
+    await test.step('Type the Description while autosave keeps firing', async () => {
+      await environment.varRowDescriptionEditor('apiKey').click();
+      await page.keyboard.type('my api key', { delay: 600 });
+      await page.waitForTimeout(AUTOSAVE_SETTLE_MS);
+      await expect(environment.varRowDescriptionLine('apiKey')).toContainText('my api key');
+      await expect(nameInput).toHaveValue('apiKey');
```
shubh-bruno, line 57: `await expect(environment.varRowValueLine('apiKey')).toContainText('token12345');` can we add this expect here as well for checking all the values

shubh-bruno (changes requested):

tests/environments/environment-tabs/autosave-env-typing.spec.ts
sachin-thakur-bruno: Not needed here as this test is focused for Description only. This assertion is already added in above tests.

shubh-bruno (approved):
LGTM

PR #8629, opened 2026-07-15
feat: improved masked inputs management

albertasaftei:
### Description

Changed the masked inputs logic so if the user starts by inserting an env variable, it won't show masked by default. While writing it manually, it will normally be shown as masked

This is the same behaviour Postman is using and I find it useful. It removes an additional click to unmask the env var.

https://github.com/user-attachments/assets/6375c53e-3cc0-4864-a8e3-7d300869238b

Added a reusable utility that will be used for masked inputs, approach is simple yet effective.

#### Contribution Checklist:

- [ x ] **The pull request only addresses one issue or adds one feature.**
- [ x ] **The pull request does not introduce any breaking changes**
- [ x ] **I have added screenshots or gifs to help explain the change if applicable.**
- [ x ] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ x ] **Create an issue and link to the pull request.**

## Summary

* **New Features**
  * Added a `shouldMaskValue` utility to determine when secret values should be masked.
  * Updated authentication editors (AWS V4, Basic, Bearer, Digest, OAuth1/OAuth2, WSSE, NTLM, EdgeGrid, and Client Certificate passphrases) to apply masking based on the current field value instead of always masking.

* **Bug Fixes**
  * Secret masking now refreshes immediately when input state changes (removed brief input-read debounce).


helloanoop:
Hey @albertasaftei 

We're aligned with the change, but the PR diff is much larger than it needs to be. It looks like your editor has introduced a number of formatting-only changes. Would you be able to remove those and keep the diff limited to just the changes required for this PR?

albertasaftei:
hello @helloanoop,  i'm removing all the formatting edits.

Wouldn't it be better to add a .pretierrc file just to keep formatting edits the same for all the contributors?

helloanoop:
Thank you. We have [eslint](https://github.com/usebruno/bruno/blob/main/eslint.config.js) to help with lint.

As far as our stance on prettier, I'd let @bijin-bruno / @sid-bruno comment it.

sid-bruno:
**TLDR;**
We'd love to use `prettier` but we're not that opinionated about the code style and 2nd it breaks community PR's and adds to our engg bandwidth for our distribution sync. 

--- 

Prettier for whitespace-only changes sounds reasonable in theory, but as the trail of changes you saw shows, it diverges quite a bit from our existing code style and what it currently allows.

There's also a secondary risk: larger merge conflicts between the OSS and Distribution repo, which adds to maintainer overhead. I want to make contributing easier, but not at the cost of reviewers having to manually re-read every line to confirm which changes were actually relevant, or losing the guarantee that OSS and Distribution changes stay validated together. That's real engineering bandwidth we'd be trading away.

On the conflict side specifically - GitHub and git do support ignoring whitespace changes (e.g. `git merge -Xignore-space-change`), but that caused issues during our last attempt at a distribution sync, so we haven't gone back to it.

[Antfu's post](https://antfu.me/posts/why-not-prettier) also covers the line-wrap noise this introduces, which is worth avoiding.

Practically, it makes more sense to just ask contributors to save files without their formatters running, rather than having a few team members repeatedly spend time untangling merges.

There's also the problem where if we now run `prettier` on the entire codebase, we will have quite a few community PR's that would have to be updated just because someone from bruno decided an arrow function just has to have enclosing braces.

PR #8651, opened 2026-07-16
feat(collections): enhance collection management by adding reset functionality and improving tab restoration logic

utkarsh-bruno:
### Description

- Added `resetCollectionForReopen` action to clear stale items and environments when reopening collections.
- Updated `openCollectionEvent` to ensure proper handling of existing collections and prevent stale data.
- Enhanced tab restoration logic to avoid wiping open tabs when all snapshot tabs are stale.
- Introduced tests for new functionality and improved existing tests for collection and tab management.

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added the ability to cancel an in-progress “migrate to YML” from the migration dialog.

- **Bug Fixes**
  - Reopening collections now refreshes contents, clears stale in-memory data, and avoids restoring outdated tabs.
  - Removing collections (and removing from workspace) now clears relevant cached indices and request UID/config data.
  - Folder `seq` metadata no longer defaults to position 1 and is preserved only when explicitly provided.

- **Improvements**
  - Migration to YML is safer with conflict checks, rollback on failure, cancellation support, and automatic reload on completion.


sid-bruno:
Assume I have a git tracked folder and I revert the migration on the same collection, I no longer see the Migrate To YML button anymore. Is that expected ? Can be moved out of scope for the current version but would like to know if it is considered a valid scenario

utkarsh-bruno:
> Assume I have a git tracked folder and I revert the migration on the same collection, I no longer see the Migrate To YML button anymore. Is that expected ? Can be moved out of scope for the current version but would like to know if it is considered a valid scenario

added a small change to read it the format from the disk and not falling back to the prev. 
The issue currently is that there will be duplicate file (.yml and .bru) that the user will be able to see on the sidebar. 
This is because of another issue with the watcher, for example, if I delete the collection folder from the directory, the collection is still visible on bruno. That fix should take care of the duplicate files as well.

PR #8660, opened 2026-07-18
fix(mount): redact large text blocks so big .bru files parse fast

chirag-bruno:
## Problem

Large `.bru` files (>= 64KB) are fed whole to the ohm parser when mounting a collection, which is slow and memory-heavy. A ~900KB request takes ~860ms to parse and scales badly.

## Approach

Add a redaction pass that splits the unbounded flat text blocks out of the file before parsing, so ohm only ever sees a small structural skeleton:

- **`redactLargeBruTextBlocks(content)`** — line-by-line scan that replaces each top-level text block's content (`body:json/text/xml/sparql/graphql`, `body:graphql:vars`, `script:pre-request`, `script:post-response`, `tests`, `docs`, legacy `body`) with a content-hash sentinel token, returning the skeleton plus the extracted values.
- **`restoreRedactedBlocks(parsed, blocks)`** — re-attaches the extracted values wherever their tokens land in the parsed object (no hard-coded AST paths).
- `parse-file` uses this for **request, collection and folder** `.bru` above the size threshold.

Extracted values are byte-identical to a normal parse — leading blank lines, the CRLF terminator's stray `\r`, and the 2-space outdent are all reproduced to match ohm's grammar boundaries. This is enforced by round-trip tests that deep-equal the redacted path against a normal parse (bodies, scripts, tests, docs, graphql vars, nested braces, leading/trailing blank lines, CRLF, collection, folder).

## Result

A 900KB `.bru` drops from **~860ms → ~0.6ms** to parse, with output identical to the normal path. Examples and `'''`-delimited dictionary values (grpc/ws content, app code) are intentionally out of scope.

Ticket - BRU-3891

## Summary by CodeRabbit

* **New Features**
  * Improved handling of large Bruno files for requests, collections, and folders based on size.
  * Added utilities to redact and restore large text blocks during parsing (covering JSON/XML/GraphQL/script and more).
  * Exposed `outdentString` as a new formatting helper.

* **Tests**
  * Added comprehensive Jest coverage for redaction/restoration round-trips, nested/edge cases, and scenarios with no large text blocks.

### DIFF
--- a/packages/bruno-electron/src/services/pool/jobs/parse-file.js
+++ b/packages/bruno-electron/src/services/pool/jobs/parse-file.js
@@ -5,23 +5,25 @@ const filestore = require('@usebruno/filestore');
 
 const sha256 = (buf) => crypto.createHash('sha256').update(buf).digest('hex');
 
-const REDACT_BODY_THRESHOLD = 512 * 1024;
+const LARGE_FILE_REDACT_THRESHOLD = 64 * 1024;
 
-const parseLargeBruRequest = (content, format) => {
-  const { bruFileStringWithRedactedBody, extractedBodyContent } = filestore.parseRequestAndRedactBody(content, { format });
-  const data = filestore.parseRequest(bruFileStringWithRedactedBody, { format });
-  data.request = data.request || {};
-  data.request.body = data.request.body || {};
-  const body = extractedBodyContent || {};
-  if (body.json) data.request.body.json = body.json;
-  if (body.text) data.request.body.text = body.text;
-  if (body.xml) data.request.body.xml = body.xml;
-  if (body.sparql) data.request.body.sparql = body.sparql;
-  if (body.graphql) {
-    data.request.body.graphql = data.request.body.graphql || {};
-    data.request.body.graphql.query = body.graphql;
+const parseLargeBru = (content, format, type) => {
+  const { skeleton, blocks } = filestore.redactLargeBruTextBlocks(content);
+  let data;
+  switch (type) {
+    case 'request':
+      data = filestore.parseRequest(skeleton, { format });
+      break;
+    case 'collection':
+      data = filestore.parseCollection(skeleton, { format });
+      break;
+    case 'folder':
+      data = filestore.parseFolder(skeleton, { format });
+      break;
+    default:
+      return null;
   }
-  return data;
+  return filestore.restoreRedactedBlocks(data, blocks);
 };
 
 const extractBruMeta = (content) => {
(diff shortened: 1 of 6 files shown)


packages/bruno-filestore/src/formats/bru/utils/redact-large-text-blocks.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    (openTag === null ? skeleton : openContent).push(line);
+  }
+
+  if (openTag !== null) {
+    skeleton.push(openTag, ...openContent);
+  }
+
+  if (blocks.some((block) => source.includes(block.token))) {
+    return { skeleton: source, blocks: [] };
```
sachin-thakur-bruno, line 76: should we add a warning `console.warn('[bruno-filestore] Token collision detected; skipping redaction');` ?

packages/bruno-filestore/src/formats/bru/utils/redact-large-text-blocks.ts
```diff
@@ -0,0 +1,99 @@
...
+  'tests',
+  'docs',
+  'body'
+];
+
+const BLOCK_OPENING = new RegExp(`^(${BLOCK_TAGS.join('|')})[ \\t]*\\{\\r?$`);
+
+const isOpening = (line: string): boolean => BLOCK_OPENING.test(line);
+
+const isClosing = (line: string): boolean => line.startsWith('}');
```
sachin-thakur-bruno, line 34: nit: can we please add a comment as the reasoning is hidden.
//.bru format requires closing braces at column 0,
//so nested braces (indented) are correctly treated as content.

packages/bruno-electron/src/services/pool/jobs/parse-file.js
```diff
@@ -5,23 +5,25 @@ const filestore = require('@usebruno/filestore');
 
 const sha256 = (buf) => crypto.createHash('sha256').update(buf).digest('hex');
 
-const REDACT_BODY_THRESHOLD = 512 * 1024;
+const LARGE_FILE_REDACT_THRESHOLD = 512 * 1024;
```
sachin-thakur-bruno: can we decide on this using some benchmark test? We may also needed it for smaller files.

packages/bruno-filestore/src/formats/bru/tests/redact-large-text-blocks.spec.js
```diff
@@ -0,0 +1,290 @@
...
+
+docs {
+  # Folder docs
+}
+`;
+
+describe('redactLargeBruTextBlocks', () => {
+  describe('matches a normal parse (ohm oracle)', () => {
+    const blankLinesBru = `meta {
+  name: Blanks
```
sachin-thakur-bruno, line 152: nit: Template String Formatting is bit odd in this file. can we format it?

packages/bruno-electron/src/services/pool/jobs/parse-file.js
chirag-bruno: So I had a local script which checks this for multiple file sizes. The result was that the heap usage by ohm was around `900x` the size of the input. I have updated the threshold to be 64KB (heap might go upto 58MB). This is safe because our worker is provisioned a max heap size of 512 MB.

sachin-thakur-bruno (approved):

bijin-bruno (approved):

PR #8681, opened 2026-07-20
fix(converters): default enabled to true when importing Postman environment variables

### LINKED ISSUE #8680: Importing a Postman environment fails when a variable has no "enabled" field
Internal - [BRU-3960]

### This bug is:

- [ ] annoying

### Bruno version

3.5.3

### Operating System

macOS 26.4.1 (25E253)

### Describe the bug

Importing Postman environments as **Global** environments fails with:

```
variables[12].enabled must be defined
An error occurred while importing the environment(s)
```

The import aborts and no environments are created after that error.

The root cause is that Postman omits the `enabled` field on some environment variables (or sets it to `null`). Bruno's converter passes that value straight through, producing `enabled: undefined`, which then fails the environment schema's `enabled: Yup.boolean().defined()` validation on the global-environment save path.

Importing the **same** files at the *collection* level appears to work, because that path doesn't run the same schema validation — so the behavior is inconsistent between global and collection imports.

### Steps to reproduce

1. Create a Postman environment containing a variable with no `enabled` key, e.g.:

   ```json
   {
     "id": "some-id",
     "name": "My Environment",
     "values": [
       { "key": "api_url", "value": "http://localhost:8080", "enabled": true },
       { "key": "gymfinder-backend-url", "value": "https://example.com", "type": "default" }
     ]
   }
   ```

   (the second variable has no `enabled` field — exactly how Postman exports some variables)

2. In Bruno, open **Global Environments → Import**.
3. Select the JSON file.
4. Import fails with `variables[N].enabled must be defined`.

### Collection to reproduce

Minimal JSON above; happy to attach a full Postman export if useful.

Image

_Screenshot of the error toast._

https://github.com/user-attachments/assets/a7925064-f0a0-428c-a971-631dc8e8c2ec

_Video of the error toast._

ricciardelli:
### Description

Closes #8680
Internal - [BRU-3960]

**Cause:** When importing a Postman environment, `postmanToBrunoEnvironment` mapped `enabled: i.enabled` verbatim. Postman omits the `enabled` field on some variables (or sets it to `null`), so the resulting Bruno variable had `enabled: undefined`.

The environment schema declares `enabled: Yup.boolean().defined()`, which is validated on the **global** environment save path (`store/global-environments.js`), so the whole import aborted with `variables[N].enabled must be defined`. Collection-level import appeared to work only because it doesn't run that same validation — an inconsistency between the two import paths.

**Resolution:** Default `enabled` to `true` when Postman doesn't provide it (`enabled: i.enabled ?? true`), matching Postman's own semantics that a variable without the flag is enabled. This handles both the missing and `null` cases and fixes both import paths at the source (the converter). Added a regression test covering the omitted/`null` `enabled` cases.

#### Contribution Checklist:

#### Before

Screenshot 2026-07-20 at 15 52 51

#### After

Screenshot 2026-07-20 at 16 21 30

## Summary by CodeRabbit

* **Bug Fixes**
  * Postman environment variables without an `enabled` value, or with a null value, are now enabled by default during conversion.
  * Other converted variable properties remain unchanged.

### DIFF
--- a/packages/bruno-converters/src/postman/postman-env-to-bruno-env.js
+++ b/packages/bruno-converters/src/postman/postman-env-to-bruno-env.js
@@ -14,7 +14,7 @@ const importPostmanEnvironmentVariables = (brunoEnvironment, values = []) => {
       uid: uuid(),
       name: (i.key ?? '').replace(invalidVariableCharacterRegex, '_'),
       value: i.value ?? '',
-      enabled: i.enabled,
+      enabled: i.enabled ?? true,
       type: 'text',
       secret: isSecret(i.type)
     };
--- a/packages/bruno-converters/tests/postman/postman-env-to-bruno-env.spec.js
+++ b/packages/bruno-converters/tests/postman/postman-env-to-bruno-env.spec.js
@@ -112,6 +112,52 @@ describe('postmanToBrunoEnvironment Function', () => {
     expect(brunoEnvironment).toEqual(expectedEnvironment);
   });
 
+  it('should default enabled to true when Postman omits or nulls it', async () => {
+    const postmanEnvironment = {
+      id: 'some-id',
+      name: 'My Environment',
+      values: [
+        {
+          key: 'noEnabled',
+          value: 'value1',
+          type: 'default'
+        },
+        {
+          key: 'nullEnabled',
+          value: 'value2',
+          enabled: null,
+          type: 'default'
+        }
+      ]
+    };
+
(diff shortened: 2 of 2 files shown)


helloanoop:
Thanks for the PR @ricciardelli 

We'll review and get back.

helloanoop:
Thanks @ricciardelli 

We've merged this to an internal branch to do some polish before merge. This is being actively worked on and is expected to go out in the `v4.2.0` release scheduled for Sep 2026

PR #8714, opened 2026-07-22
feat(filesystem): decouple display names from filesystem names & resolve collisions silently

sachin-thakur-bruno:
### Description

BRU-894

Previously, we used the same value for both an item's **display name** (what users see) and its **filesystem name** (the file/folder on disk). This meant duplicate names caused filesystem conflicts, leading to failed clone/paste operations, `path already exists` errors, and inconsistent error messages exposing filesystem paths and IPC details.

This PR separates those two concepts.

Users can now freely use duplicate display names, while we automatically resolves any filesystem collisions by appending numeric suffixes (`name`, `name1`, `name2`, ...). Filesystem naming becomes an implementation detail instead of something users have to manage.

https://github.com/user-attachments/assets/ede39e23-63d1-4784-bc13-a2dd242d14cd

## Behavior changes

- Duplicate display names are now allowed for collections, folders, and requests.
- Clone and copy/paste always use `<source> copy` as the display name.
  - Requests and folders clone immediately without a modal.
  - Collections still show a clone dialog because a destination location must be chosen.
- Creating or renaming an item preserves exactly what the user entered.
- Moving or dragging items preserves the display name, even when filesystem collisions occur.
- Filesystem collisions are resolved automatically by suffixing only the on-disk name.
- Raw filesystem paths, IPC errors, and inconsistent duplicate-name errors are no longer shown to users.
- The only user-facing error is when the duplicate limit is reached:
  - `Too many items named "<name>" (limit 200). Please use a different name.`

---

### Shared naming utilities (`@usebruno/common`)

Introduced a shared `src/utils/naming.ts` containing:

- `sanitizeName`
- `validateName`
- `validateNameError`

Both the renderer and Electron use these utilities, ensuring validation rules and suffix generation remain consistent.

---

### Collision-safe filesystem primitives (`bruno-electron`)

Added a set of reusable helpers in `packages/bruno-electron/src/utils/filesystem.js`.

#### Atomic creation

`writeFileUnique` and `mkdirUnique` use exclusive filesystem operations (`wx` and non-recursive `mkdir`) so the filesystem itself resolves races.

If two operations attempt the same name simultaneously:

- one succeeds,
- the other retries with the next suffix.

This fixes the rapid double-paste race that previously resulted in `path already exists` errors.

#### Efficient suffix generation

A single `readdir` pass is used as a hint to jump directly to the first available suffix while still filling gaps, avoiding repeated filesystem scans.

#### Collision-safe rename and move

`getUniqueRenamePath` and `getUniqueTargetPath` ensure rename and move operations also avoid collisions.

They additionally detect when the source and destination refer to the same inode, allowing case-only renames (`login → Login`) to occur in place on case-insensitive filesystems instead of incorrectly generating suffixed names.

#### Long filename handling

Names are truncated before writing to disk, reserving space for:

- the file extension
- the largest possible numeric suffix

This guarantees the final filename never exceeds the filesystem's 255-character limit.

#### Duplicate limit

A `MAX_DUPLICATE_NAMES = 200` safeguard prevents infinite retry loops. If no unique name can be found within that limit, a friendly validation message is shown instead of exposing filesystem errors.

---

### IPC and renderer integration

All create, clone, paste, rename, move, transient, and scratch-save flows now use the shared filesystem primitives.

During rename or move operations, item UIDs are also remapped (`moveRequestUid`) so open tabs, drafts, and item identity remain intact.

---

### Bug fix: duplicate folder after case-only rename

On case-insensitive filesystems, renaming a folder such as:

```text
auth → Auth
```

previously produced two sidebar entries even though only a single directory existed on disk.

The issue was caused by folders reconciling watcher events using case-sensitive filenames, while requests already reconciled using UIDs.

The fix has two parts:

1. The Electron rename flow now remaps the folder UID using `moveRequestUid(oldPath, newPath)`, ensuring watcher events continue using the existing identity.
2. The renderer now reconciles `collectionAddDirectoryEvent` by UID instead of filename, updating the existing node rather than creating a duplicate.

This also preserves folder identity during normal renames while leaving new-folder creation and cross-collection moves unchanged.

---

## Tests

End-to-end tests live under `tests/naming-collisions/`.

Each test validates both:

- the display name shown in the sidebar
- the actual filename or directory created on disk

### Coverage highlights

- Requests and folders
  - Clone, create, rename, copy/paste, and drag/drop all handle filesystem collisions correctly.
  - Duplicate display names render side by side.
  - Works for both `.bru` and `.yml` formats.
  - Reserved names remain blocked.
  - Open tabs continue tracking renamed or moved items.
  - Case-only renames work correctly.
  - Existing filesystem names created outside Bruno are preserved or suffixed as needed.

- Collections
  - Clone still uses a modal because a destination must be selected.
  - New collections never reuse an existing directory name.

- App items
  - Clone remains unavailable for app items while continuing to work for folders and requests.

- Race conditions
  - Two immediate paste operations (`Cmd/Ctrl + V`) create two separate files without errors.

The `MAX_DUPLICATE_NAMES` limit is covered by Electron unit tests rather than end-to-end tests, since creating 200 duplicate items through the UI would be impractical.

Two existing drag-and-drop tests were updated to reflect the new silent collision behavior instead of expecting duplicate-name error toasts.

#### Contribution Checklist:

- [x] **I have run the code review locally.**

## Summary by CodeRabbit

* **New Features**
  * Added one-click cloning for eligible requests and folders.
  * Automatically creates unique filesystem names for duplicate items while preserving display names.
  * Improved copy, paste, move, rename, and save workflows across collections.

* **Bug Fixes**
  * Prevented duplicate-name failures and race conditions.
  * Added validation for invalid, reserved, and overly long names.
  * Improved error messages and saved-request navigation.

* **Tests**
  * Expanded coverage for naming collisions, cloning, drag-and-drop, renaming, and repeated actions.


packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -615,6 +619,15 @@ const CollectionItem = ({ item, collectionUid, collectionPathname, searchText })
     toast.success(`${itemType} copied`);
   };
 
+  // One-click clone: display name becomes "<source> copy"; the filesystem name
+  // uniqueness is resolved silently by electron.
+  const handleCloneItem = () => {
+    if (!isCloneable) return;
+    dispatch(cloneItem(`${item.name} copy`, `${sanitizeName(item.name)} copy`, item.uid, collectionUid))
+      .then(() => toast.success(`${isFolder ? 'Folder' : 'Request'} cloned!`))
+      .catch((err) => toast.error(err?.message || `An error occurred while cloning the ${isFolder ? 'folder' : 'request'}`));
```
chirag-bruno: If I have not mistaken, the value for `err?.message` comes in a weird manner. It's format is off and it comes with the IPC name and stuff. Can you verify that once?

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -830,12 +798,32 @@ export const newFolder = (folderName, directoryName, collectionUid, itemUid) =>
...
+            mode: 'inherit'
+          }
+        }
+      };
+
+      ipcRenderer
+        .invoke('renderer:new-folder', { pathname: fullName, folderData, format: collection.format })
+        .then(resolve)
+        .catch((error) => {
+          toast.error('Failed to create a new folder!');
```
chirag-bruno: It's just a broader question. Should we trigger toasts from here?

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
```diff
@@ -830,12 +798,32 @@ export const newFolder = (folderName, directoryName, collectionUid, itemUid) =>
     }
 
     if (!itemUid) {
-      const folderWithSameNameExists = find(
-        collection.items,
-        (i) => i.type === 'folder' && trim(i.filename) === trim(directoryName)
-      );
-      if (!folderWithSameNameExists) {
-        const fullName = path.join(collection.pathname, directoryName);
+      const fullName = path.join(collection.pathname, directoryName);
```
chirag-bruno: Also, do we need two separate branches for `!itemUid`? The code for both branches are almost similar.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
chirag-bruno: `formatIpcError` - `packages/bruno-app/src/utils/common/error.js`

packages/bruno-common/src/utils/naming.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Returns true if `name` is a valid file/dir name.
+ *
+ * Reconciled version: guards against falsy input (the renderer copy guarded
+ * this; the Electron copy did not and would throw on `undefined`).
+ */
+export const validateName = (name: string): boolean => {
+  if (!name) return false;
+  if (name.length > 255) return false; // max name length
+  if (reservedDeviceNames.test(name)) return false; // windows reserved names
```
chirag-bruno, line 41: nit: if the reservation is for windows, do we prevent it in all other OS too?

packages/bruno-common/src/utils/naming.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  return firstCharacter.test(name) && middleCharacters.test(name) && lastCharacter.test(name);
+};
+
+/**
+ * Human-readable validation error for `name`, or '' if valid.
+ */
+export const validateNameError = (name: string): string => {
```
chirag-bruno, line 49: The naming kinda sounds like the function might throw (since usage is in JS and js does not know the return type), but we are just returning a string. Makes sense?

chirag-bruno: Maybe returns a boolean but throws? Just asking. If this is used in a not of places and might need a try/catch wrapping at call site, we can chill.

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -132,6 +135,107 @@ const withFileLock = async (pathname, fn) => {
...
+  const start = await firstFreeSuffix(dirname, safeBase, normalizedExt);
+
+  for (let counter = start; counter < MAX_DUPLICATE_NAMES; counter++) {
+    const candidate = nextSuffixedName(safeBase, normalizedExt, counter);
+    const pathname = path.join(dirname, candidate);
+    try {
+      await fsPromises.writeFile(pathname, content, { flag: 'wx' });
+      return { pathname, filename: path.basename(pathname) };
+    } catch (err) {
+      if (err && err.code === 'EEXIST') continue;
```
chirag-bruno, line 192: Is the `err.code` standard across OSs?

chirag-bruno (commented):
| Severity | Location | Finding |
|---|---|---|
| suggestion | `ipc/collection.js:1649` | `move-item` remaps paths with `p.replace(sourcePathname, targetPathname)`; a target basename with `$` (e.g. `$&`) is read as a replacement pattern → corrupted paths, lost uid mapping. **Verified.** Use split/join or a function replacer. |
| suggestion | `ipc/collection.js:586, :2322` | `save-transient-request` / `save-scratch-request` feed renderer-supplied `targetFilename` to `writeFileUnique` with no `validateName` guard (unlike `:530`); `../../foo` escapes the validated dir, now writes silently. **Verified.** One-line `validateName` fix. |
| suggestion | `ipc/collection.js:195/:254` | `validateName` imported directly but `validateNameError` reached via `utils.` namespace; pick one style. |
| suggestion | `filesystem.js:454` | Self/descendant-copy guard compares case-sensitively; on case-insensitive volumes a case-only-different target escapes and recurses into itself. |
| suggestion | `filesystem.js:486` | `withDirLock` keys the mutex on the raw `dirname`; differently-cased paths to the same dir aren't serialized on Windows/macOS. Normalize the key. |
| suggestion | `filesystem.js:140` | `MAX_FILENAME_LENGTH = 255` truncates by UTF-16 code units, but Linux enforces 255 **bytes**; CJK/emoji names can still throw `ENAMETOOLONG`. |
| suggestion | `filesystem.js:490` | `dirLockChains` entries are never pruned; the map grows unbounded across a long session. Slow leak. |
| suggestion | `actions.js:793` | `newFolder` dereferences `parentItem.items` before the `!collection`/parent-not-found guards → synchronous `TypeError` on stale `itemUid` instead of a clean rejection. |
| suggestion | `actions.js:800–847` | Both `newFolder` branches hold identical `folderData` + invoke blocks differing only in `fullName`; resolve `fullName` first, share one block. |
| suggestion | `CollectionItem/index.js:664` | `CloneCollectionItem` modal is dead code — keybinding (`:117`) and context-menu (`:398`) both call `handleCloneItem()`. Remove state (`:95`), render block, import (`:36`). |
| suggestion | `naming.ts:20, :34` | Comments narrate the migration ("Behavior is intentionally unchanged…"); restate as timeless rationale or drop. |
| suggestion | `tests/utils/page/naming-collisions.ts:237` | `chooseCloneLocation` monkey-patches `dialog.showOpenDialog` on the worker-scoped app and never restores it — leaks into later tests. |
| suggestion | `tests/naming-collisions/{folder,request}/drag-drop.spec.ts:101/75` | Hardcoded `page.waitForTimeout(1000)` before a negative disk assertion; sync on an observable signal instead. |
| nit | `actions.js:108` | Drop the "no longer guesses" situational comment phrasing. |
| nit | `CollectionItem/index.js:624` | `handleCloneItem` inlines the `${name} copy` convention that `actions.js` centralizes in `copyDisplayName`. |
| nit | `tests/utils/page/naming-collisions.ts:243/234` | `setTextBody` + `bodyMode/bodyEditor/modalCardByTitle` unused; `electronApp: any` / `({ dialog }: any)` should use `ElectronApplication`. |
| nit | `cross-collection-drag-drop-*.spec.ts:30/~81` | Inline raw selectors instead of the new `namingCollisions` page module. |

Review by Claude

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
sachin-thakur-bruno: we can discuss this. But currently we have triggers for errors in this.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
sachin-thakur-bruno: Yes, we should use formatIpcError. Updating it in `handleCloneItem`. Same updating it for `handlePasteItem` also.

packages/bruno-app/src/providers/ReduxStore/slices/collections/actions.js
sachin-thakur-bruno: the only diff is the the parent path the folder is created under. Rest of the things are same. Updated it.

packages/bruno-common/src/utils/naming.ts
sachin-thakur-bruno: It's intentional to prevent it in all other OS. If a macOS/Linux user could name a request CON, it'd work fine locally but the collection would become un-clonable / un-openable on Windows for a collaborator

packages/bruno-common/src/utils/naming.ts
sachin-thakur-bruno: This is the existing function moved to common folder. It's currently being used in many places. So kept it like this only. Changing it require changes in existing places which will increase the testing area.

packages/bruno-electron/src/utils/filesystem.js
sachin-thakur-bruno: Yes it is standard across OSs.
On Linux/macOS the C errno is literally EEXIST.
on Windows the native failure is something else (ERROR_FILE_EXISTS / ERROR_ALREADY_EXISTS), but libuv maps it to the same UV_EEXIST, and Node surfaces it as err.code === 'EEXIST'. 
So the string is portable by design

packages/bruno-common/src/utils/naming.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * This is the single definition of the collision suffix scheme. The Electron
+ * `writeFileUnique` / `mkdirUnique` (filesystem authority) use it to generate
+ * candidate names while resolving collisions atomically against the real
+ * filesystem.
+ */
+export const nextSuffixedName = (base: string, ext: string, n: number): string => {
+  const suffix = n === 0 ? '' : String(n);
+  return ext ? `${base}${suffix}.${ext}` : `${base}${suffix}`;
+};
```
sid-bruno: can we move this to electron since I don't see this being used in CLI for the most part, cli can't create only run

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -132,6 +135,107 @@ const withFileLock = async (pathname, fn) => {
...
+    existing = new Set(await fsPromises.readdir(dirname));
+  } catch (err) {
+    return 0;
+  }
+  let counter = 0;
+  while (existing.has(nextSuffixedName(baseName, ext, counter))) {
+    counter++;
+  }
+  return counter;
+};
```
sid-bruno: can we not just do a try and catch for the next suffix and on an EEXIST error try the next value instead of doing a full directory read? 

The entire file directory could have a lot of files

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -329,6 +407,103 @@ const safeToRename = (oldPath, newPath) => {
...
+  }
+
+  for (let counter = 1; counter < MAX_DUPLICATE_NAMES; counter++) {
+    const candidate = nextSuffixedName(base, extNoDot, counter);
+    const candidatePath = path.join(dir, candidate);
+    const isFree = existing ? !existing.has(normalizeName(candidate)) : safeToRename(oldPath, candidatePath);
+    if (isFree) {
+      return candidatePath;
+    }
+  }
```
sid-bruno: Could be simpler?

```suggestion
  for (let counter = 0; counter < MAX_DUPLICATE_NAMES; counter++) {
    const candidate = nextSuffixedName(base, extNoDot, counter);
    const candidatePath = path.join(dir, candidate);
    if (safeToRename(oldPath, candidatePath)) {
      return candidatePath;
    }
  }
```

tests/collection/moving-requests/cross-collection-drag-drop-folder.spec.ts
```diff
@@ -168,30 +168,29 @@ test.describe('Cross-Collection Drag and Drop for folder', () => {
...
+    // No error toast is shown, and no flow fails just because a folder name already exists.
+    await expect(page.getByText(/already exists/i)).toHaveCount(0);
 
-    // source and target collection request should remain unchanged
+    // The folder is moved out of the source collection.
     const sourceCollectionContainer = page
       .locator('.collection-name')
       .filter({ hasText: 'source-collection' })
       .locator('..');
+    await page.locator('#sidebar-collection-name').filter({ hasText: 'source-collection' }).click();
```
sid-bruno: we have an existing locator that can be reused

tests/utils/page/naming-collisions.ts
```diff
@@ -0,0 +1,249 @@
...
+    page
+      .locator('.collection-name')
+      .filter({ hasText: collectionName })
+      .locator('..')
+      .locator('.collection-item-name')
+      .filter({ hasText: itemName })
+      .first(),
+
+  collectionDropTarget: (collectionName: string): Locator =>
+    page.locator('.collection-name').filter({ hasText: collectionName }),
```
sid-bruno:
```suggestion
  itemInCollection: (collectionName: string, itemName: string): Locator => {
    const { sidebar } = buildCommonLocators(page);
    return sidebar.collectionScope(collectionName).locator('.collection-item-name').filter({ hasText: itemName }).first();
  },

  collectionDropTarget: (collectionName: string): Locator =>
    buildCommonLocators(page).sidebar.collection(collectionName),
```

tests/utils/page/naming-collisions.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  // Sidebar items with an exact display name, scoped to one collection's container
+  // (for counting duplicates / asserting presence within a specific collection).
+  itemsByTitleInCollection: (collectionName: string, title: string): Locator =>
+    page
+      .locator('.collection-name')
+      .filter({ hasText: collectionName })
+      .locator('..')
+      .locator('.item-name')
+      .and(page.getByTitle(title, { exact: true })),
```
sid-bruno:
```suggestion
  itemsByTitleInCollection: (collectionName: string, title: string): Locator =>
    buildCommonLocators(page)
      .sidebar.collectionScope(collectionName)
      .locator('.item-name')
      .and(page.getByTitle(title, { exact: true })),
```

tests/utils/page/naming-collisions.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  toast: (text: string | RegExp): Locator => page.getByText(text),
+
+  anyModal: (): Locator => page.locator('.bruno-modal'),
+  modalByTitle: (title: string): Locator => page.locator('.bruno-modal').filter({ hasText: title }),
+  modalCardByTitle: (title: string): Locator => page.locator('.bruno-modal-card').filter({ hasText: title }),
```
sid-bruno:
```suggestion
  modalByTitle: (title: string): Locator => buildCommonLocators(page).modal.byTitle(title),
  modalCardByTitle: (title: string): Locator => buildCommonLocators(page).modal.card().filter({ hasText: title }),
```

tests/utils/page/naming-collisions.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  requestNameInput: (): Locator => page.getByPlaceholder('Request Name'),
```
sid-bruno:
```suggestion
  requestNameInput: (): Locator => buildCommonLocators(page).request.requestNameInput(),
```

tests/utils/page/naming-collisions.ts
```diff
@@ -0,0 +1,249 @@
...
+  collectionLocationInput: (): Locator => page.locator('#collection-location'),
+  browseButton: (): Locator => page.getByText('Browse', { exact: true }),
+
+  // Save Request (transient) modal
+  saveRequestModal: (): Locator => page.locator('.bruno-modal-card').filter({ hasText: 'Save Request' }),
+  saveRequestNameInput: (): Locator => page.locator('#request-name'),
+
+  // Request body editor (used to set/inspect a request body)
+  bodyModeSelector: (): Locator => page.locator('.body-mode-selector'),
+  bodyEditor: (): Locator => page.locator('.request-pane .CodeMirror').first()
```
sid-bruno:
```suggestion
  bodyModeSelector: (): Locator => buildCommonLocators(page).request.bodyModeSelector(),
  bodyEditor: (): Locator => buildCommonLocators(page).request.bodyEditor().locator('.CodeMirror').first()
```

tests/utils/page/naming-collisions.ts
```diff
@@ -0,0 +1,249 @@
...
+  await actions.collectionItemActions(name).first().click();
+};
+
+const openCollectionActionsMenu = async (page: Page, collectionName: string) => {
+  const { sidebar, actions } = buildCommonLocators(page);
+  await sidebar.collection(collectionName).hover();
+  const action = actions.collectionActions(collectionName);
+  await action.waitFor({ state: 'visible' });
+  await action.click();
+};
```
sid-bruno:
```suggestion
export const openItemActionsMenu = async (page: Page, name: string) => {
  const { sidebar } = buildCommonLocators(page);
  const menu = sidebar.rowMenu(name);
  await sidebar.itemRow(name).first().hover();
  await menu.trigger().first().click({ force: true });
};

const openCollectionActionsMenu = async (page: Page, collectionName: string) => {
  const { sidebar } = buildCommonLocators(page);
  const menu = sidebar.rowMenu(collectionName, 'collection');
  await sidebar.collectionRow(collectionName).hover();
  const trigger = menu.trigger();
  await trigger.waitFor({ state: 'visible' });
  await trigger.click();
};
```

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -132,6 +135,107 @@ const withFileLock = async (pathname, fn) => {
...
+ * max-length name can't trigger ENAMETOOLONG.
+ *
+ * @param {string} base   basename without extension
+ * @param {string} ext    extension without a leading dot ('' for directories)
+ */
+const truncateBaseForSuffix = (base, ext) => {
+  const extLength = ext ? ext.length + 1 : 0;
+  const maxBaseLength = Math.max(1, MAX_FILENAME_LENGTH - extLength - SUFFIX_RESERVE);
+  return base.length > maxBaseLength ? base.slice(0, maxBaseLength) : base;
+};
```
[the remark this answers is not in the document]
sachin-thakur-bruno: fixed this

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -1126,11 +1123,18 @@ const registerRendererEventHandlers = (mainWindow, watcher) => {
...
           const newRequestFilePath = requestFile.replace(oldPath, newPath);
           moveRequestUid(requestFile, newRequestFilePath);
         }
 
+        // Remap the folder's own uid so its identity survives the rename. Without
+        // this the watcher's addDir for the new path mints a fresh uid and the
+        // sidebar ends up with a duplicate folder node (notably on case-only
+        // renames on case-insensitive filesystems, where the directory is the
+        // same on disk but the path casing differs).
+        moveRequestUid(oldPath, newPath);
```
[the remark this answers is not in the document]
sachin-thakur-bruno: UIDs are intentionally remapped before the rename so the watcher cache is ready
when the addDir event fires. Moving them after would reintroduce a sidebar duplicate bug.

packages/bruno-electron/src/utils/tests/filesystem/index.spec.js
```diff
@@ -22,46 +27,51 @@ describe('File System Operations', () => {
...
         const sourcePath = path.join(tempDir, 'folder_1', 'folder_1_1', 'folder_2', 'folder_2_1');
         const destDir = path.join(tempDir);
-        await copyPath(sourcePath, destDir);
+        await moveInto(sourcePath, destDir);
         await removePath(sourcePath);
       }
 
       const result = await verifyFilesAndFolders(tempDir, finalCollectionStructure);
       expect(result).toBe(true);
     });
```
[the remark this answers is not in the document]
sachin-thakur-bruno: fixed this tests

packages/bruno-common/src/utils/naming.ts
sachin-thakur-bruno: moved this it electron

packages/bruno-electron/src/utils/filesystem.js
sachin-thakur-bruno: removed the full directory read and just tried the next value.

packages/bruno-electron/src/utils/filesystem.js
sachin-thakur-bruno: simplified this

tests/collection/moving-requests/cross-collection-drag-drop-folder.spec.ts
sachin-thakur-bruno: reused the existing locator

tests/utils/page/naming-collisions.ts
sachin-thakur-bruno: fixed this

tests/utils/page/naming-collisions.ts
sachin-thakur-bruno: fixed this

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -329,6 +409,122 @@ const safeToRename = (oldPath, newPath) => {
...
+
+    if (lst.isSymbolicLink()) {
+      // Dereference the link (copy what it points at) rather than recreating it.
+      const real = await fsPromises.stat(src).catch(() => null); // follows the link
+      if (!real) return; // dangling link(target doesn't exist). skip
+      if (real.isFile()) {
+        await fsPromises.copyFile(src, dest); // copyFile follows the link → real file
+      }
+
+      return;
```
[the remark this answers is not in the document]
sachin-thakur-bruno: This is a theoretical edge case need not to be handled. Its existing scenario

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -329,6 +409,122 @@ const safeToRename = (oldPath, newPath) => {
...
+ * Recursively copies `source` to `targetPath`.
+ */
+const copyPathTo = async (source, targetPath) => {
+  const resolvedSource = canonicalPath(source);
+  const resolvedTarget = path.join(canonicalPath(path.dirname(targetPath)), path.basename(targetPath));
+
+  // Prevent copying a path into itself or one of its descendants.
+  const srcLower = resolvedSource.toLowerCase();
+  const tgtLower = resolvedTarget.toLowerCase();
+  if (tgtLower === srcLower || tgtLower.startsWith(srcLower + path.sep)) {
```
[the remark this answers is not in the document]
sachin-thakur-bruno: Comparing canonical paths directly

tests/naming-collisions/collection/create.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    fs.mkdirSync(path.join(location, 'MyColl'), { recursive: true });
+
+    await createCollection(page, 'MyColl', location, 'bru');
+
+    await test.step('On disk: the empty dir is reused (bruno.json written into it); no "MyColl1" created', async () => {
+      await expect
+        .poll(() => fs.existsSync(path.join(location, 'MyColl', 'bruno.json')), { timeout: 10000 })
+        .toBe(true);
+      expect(fs.existsSync(path.join(location, 'MyColl1'))).toBe(false);
+    });
```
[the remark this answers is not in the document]
sachin-thakur-bruno: added the yml also

tests/naming-collisions/request/copy-paste.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    await copyItem(page, 'login');
+    await pasteIntoFolder(page, 'Auth');
+    await pasteIntoFolder(page, 'Auth');
+
+    await test.step('Two "login copy" in the folder; second file suffixed', async () => {
+      const files = listRequestFiles(path.join(testDir, 'Paste Folder Twice', 'Auth'));
+      expect(files).toContain('login copy.bru');
+      expect(files).toContain('login copy1.bru');
+    });
```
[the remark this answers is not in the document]
sachin-thakur-bruno: its a test to assert the file system name not the user visible name

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -116,12 +116,14 @@ const CollectionItem = ({ item, collectionUid, collectionPathname, searchText })
   const itemIsCollapsed = hasSearchText ? false : item.collapsed;
   const isFolder = isItemAFolder(item);
```
chirag-bruno, line 118: Do we need to memoize this? I understand that this is not a crazy complicated computation but kinda does branch operations on both conditions (isFolder and isItemARequest). Just curious if that would reduce the branching during rerenders.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
```diff
@@ -631,6 +634,15 @@ const CollectionItem = ({ item, collectionUid, collectionPathname, searchText })
     toast.success(`${itemType} copied`);
   };
 
+  // One-click clone: display name becomes "<source> copy"; the filesystem name
+  // uniqueness is resolved silently by electron.
+  const handleCloneItem = () => {
+    if (!isCloneable) return;
+    dispatch(cloneItem(`${item.name} copy`, `${sanitizeName(item.name)} copy`, item.uid, collectionUid))
```
sid-bruno: Sanitize after appending ` copy` so clone matches paste (e.g. `api v2.` → `api v2. copy`, not `api v2 copy`).

```suggestion
    dispatch(cloneItem(`${item.name} copy`, sanitizeName(`${item.name} copy`), item.uid, collectionUid))
```

sid-bruno (commented):
nitpicks

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -1672,61 +1701,78 @@ const registerRendererEventHandlers = (mainWindow, watcher) => {
...
-      return Promise.reject(error);
-    }
+        return { newPathname: targetPathname };
+      } catch (error) {
+        return Promise.reject(error);
+      }
+    });
   });
 
   ipcMain.handle('renderer:move-folder-item', async (event, folderPath, destinationPath) => {
```
sid-bruno, line 1778: I think this is a dead call now, please check if it's being called anywhere and delete / leave accordingly.

packages/bruno-electron/src/utils/filesystem.js
```diff
@@ -132,6 +135,105 @@ const withFileLock = async (pathname, fn) => {
   }
 };
 
+const MAX_DUPLICATE_NAMES = 200;
```
sid-bruno, line 138: Can we have tests for checking the limit

packages/bruno-electron/src/ipc/collection.js
sachin-thakur-bruno: Its a dead path.Also move-file-item also a dead path. I will clear these 2 in a separate PR for clear separation from this feature.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
sachin-thakur-bruno: no we do n't need this. Both are booleans with O(1) work. memoize will not help here.

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/index.js
sachin-thakur-bruno: updated this

packages/bruno-electron/src/utils/filesystem.js
sachin-thakur-bruno: added a unit tests for this

packages/bruno-common/src/utils/naming.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Single source of truth for sanitizing display names into filesystem-safe
+ * names and validating names. Consumed by the Electron main process and the
+ * renderer (and re-exported from each package's local helper so existing import
+ * paths keep working). Collision-suffix resolution (`nextSuffixedName`) lives in
+ * the Electron filesystem utils, since that's the only place collisions are
+ * resolved against the real filesystem.
+ */
+
+// Characters that are illegal in file/dir names across Windows/macOS/Linux.
```
sid-bruno:
```suggestion
```

Redundant

tests/naming-collisions/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+  for (const entry of fs.readdirSync(testDir, { withFileTypes: true })) {
+    if (entry.isDirectory()) {
+      const candidate = path.join(testDir, entry.name);
+      if (markers.some((m) => fs.existsSync(path.join(candidate, m)))) return candidate;
+    }
+  }
+  throw new Error(`No collection (bruno.json / opencollection.yml) found under ${testDir}`);
+};
+
+/** Minimal valid .bru request content, for seeding files directly on disk. */
```
sid-bruno:
```suggestion
```
redundant

tests/naming-collisions/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+      out.push(entry.name);
+    }
+  }
+  return out;
+};
+
+/**
+ * Find the collection root under `testDir`. Bruno marks a collection root with
+ * `bruno.json` (bru format) or `opencollection.yml` (yml format).
+ */
```
sid-bruno: redundant
```suggestion
```

tests/utils/page/naming-collisions.ts
sid-bruno, line 1: shouldn't these be just the sidebar / collection items instead of being their own set of locators?

tests/utils/page/locators.ts
```diff
@@ -34,6 +35,7 @@ export const buildCommonLocators = (page: Page) => ({
   },
   preferences: buildPreferencesLocators(page),
   ai: buildAiPreferencesLocators(page),
+  namingCollisions: buildNamingCollisionLocators(page),
```
sid-bruno: https://github.com/usebruno/bruno/pull/8714/changes#r3850772114

tests/utils/page/toast.ts
```diff
@@ -4,5 +4,6 @@ import { Page } from '../../../playwright';
  * Toast (react-hot-toast) locators, parameterised by message.
  */
 export const buildToastLocators = (page: Page) => ({
-  byMessage: (message: string | RegExp) => page.getByText(message)
+  byMessage: (message: string | RegExp) =>
+    page.locator('.toast-container').getByText(message)
```
sid-bruno: Add a data test id since you are adding this class in the same PR.

tests/utils/page/naming-collisions.ts
sachin-thakur-bruno: moved the locators to sidebar. Now this only contains the functions needed for na,ing collissions

tests/utils/page/toast.ts
sachin-thakur-bruno: We can't add a data test id as we are using the Toaster library and it doesn't support a test id. Now using page.getByRole('status')

sid-bruno (approved):

PR #8730, opened 2026-07-22
fix(mountv2): give duplicate-named env variables distinct uids

chirag-bruno:
### Problem

In the mountv2 mount path, deleting one environment variable also deletes every other variable that shares the same name. Reproduces with an environment holding multiple variables with the same key but different values (e.g. 3 × `TOKEN`). Only occurs on mountv2 with the file cache enabled.

### Root cause

Cold-load environment hydration minted each variable's `uid` from the variable **name** only:

```js
const key = variable.name || `index:${i}`;
variable.uid = uidForSeed(`${env.uid}#var#${key}`);
```

`uidForSeed` is a pure hash of the seed, so variables sharing a name produced an identical `uid`. The renderer keys variable delete/edit (and its React row keys) on `uid`, so removing one duplicate filtered out all of them.

The legacy (non-mountv2) watcher path assigns a fresh `uuid()` per variable, which is why it was unaffected.

### Fix

Fold the array index into the seed so every variable gets a distinct `uid` regardless of duplicate names. Nothing depends on env-variable uid stability across reloads — secret hydration matches by name, and the legacy path already re-mints random uids each reload.

### Test

Added a regression test in `manager.spec.js` that mounts an environment with three same-named variables and asserts three distinct uids.

Ticket: BRU-3981

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed environment variable identification during remounts to ensure variables with identical names receive distinct identifiers.
  * Improved reliability when emitting environments containing duplicate variable names.

### DIFF
--- a/packages/bruno-electron/src/services/mount/manager.js
+++ b/packages/bruno-electron/src/services/mount/manager.js
@@ -24,8 +24,7 @@ const hydrateEnvironments = (collectionPath, environments) => {
   for (const env of environments) {
     if (!Array.isArray(env.variables)) continue;
     env.variables.forEach((variable, i) => {
-      const key = variable.name || `index:${i}`;
-      variable.uid = uidForSeed(`${env.uid}#var#${key}`);
+      variable.uid = uidForSeed(`${env.uid}#var#${i}#${variable.name || ''}`);
     });
     if (!envHasSecrets(env)) continue;
     try {
--- a/packages/bruno-electron/src/services/mount/manager.spec.js
+++ b/packages/bruno-electron/src/services/mount/manager.spec.js
@@ -44,6 +44,7 @@ jest.mock('../../cache/requestUids', () => ({
 
 const { MountManager } = require('./manager');
 const collectionWatcher = require('../../app/collection-watcher');
+const treeBuilder = require('./tree-builder');
 
 const makeEmit = () => ({ loading: jest.fn(), tree: jest.fn(), config: jest.fn() });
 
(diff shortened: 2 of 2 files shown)


packages/bruno-electron/src/services/mount/manager.js
```diff
@@ -24,8 +24,7 @@ const hydrateEnvironments = (collectionPath, environments) => {
   for (const env of environments) {
     if (!Array.isArray(env.variables)) continue;
     env.variables.forEach((variable, i) => {
-      const key = variable.name || `index:${i}`;
-      variable.uid = uidForSeed(`${env.uid}#var#${key}`);
+      variable.uid = uidForSeed(`${env.uid}#var#${i}#${variable.name || ''}`);
```
sid-bruno, line 27: Why not add the index everytime? are we re-ordering them anywhere? if the order changes, we change the cache (or I'd say don't have the env in the cache at all)

packages/bruno-electron/src/services/mount/manager.spec.js
```diff
@@ -101,6 +102,30 @@ describe('MountManager.remount', () => {
     expect(collectionWatcher.addTempDirectoryWatcher).not.toHaveBeenCalled();
   });
 
+  it('gives each environment variable a distinct uid when names collide', async () => {
```
sid-bruno, line 105: Add another one with the same value as well, would want both restored

packages/bruno-electron/src/services/mount/manager.js
chirag-bruno: You mean storing the index in cache? We can do that but there is no real benefit from it. Also, we need the env in cache because that is being tracked by the watcher. If we are not tracking environment in the cache, we have to parse that too, adding to the bottleneck. If I missed something, let me know

sid-bruno: ignore, just realised you've added `${i}`

packages/bruno-electron/src/services/mount/manager.spec.js
sid-bruno: as in TOKEN:value, TOKEN:value

chirag-bruno: Both get restored because we are tracking by UID and that is unique.

sid-bruno: Haan just add another test for later

chirag-bruno: Yeah. Cool.

sid-bruno (approved):

PR #8769, opened 2026-07-24
fix(workspace): heal legacy object entries in lastOpenedWorkspaces

vijayh-bruno:
### Description

**Issue:**

Workspaces disappearing after upgrade when `preferences.json` →`workspaces.lastOpenedWorkspaces` contains legacy object entries mixed with string paths. Startup code assumed every entry was a string and called
`path.join(entry, 'workspace.yml')`. On an object entry, this threw a `TypeError`, which aborted workspace resolution and made the handler return an empty list. So one bad entry wiped out every workspace, leaving only the default "My Workspace".

**Fix:** 

- Normalise entries in the store: keep strings, recover the path from an
  object's `pathname`, drop anything unusable, and de-duplicate. So callers
  only ever see a string-only list.

- `migrate()` runs once on load and rewrites `preferences.json` to the clean string-only array, scrubbing the legacy/experimental build-related objects from disk.

- Guard `isValidWorkspacePathOnDisk` against non-string input as defense.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of previously saved workspaces by removing duplicates and ignoring invalid entries.
  * Added compatibility for older saved workspace formats.
  * Prevented workspace validation errors when invalid or empty paths are provided.

* **Tests**
  * Added coverage for invalid workspace path inputs across supported validation options.

### DIFF
--- a/packages/bruno-electron/src/store/last-opened-workspaces.js
+++ b/packages/bruno-electron/src/store/last-opened-workspaces.js
@@ -1,4 +1,31 @@
+const _ = require('lodash');
 const Store = require('electron-store');
+
+/**
+* Older/experimental builds persisted some entries as a workspace object (`{ pathname, ... }`)
+* instead of a plain path string. Recover the path from `pathname`, drop anything unusable, and
+* de-duplicate so callers only ever see a string-only list.
+*/
+
+const normalizeWorkspaceEntry = (entry) => {
+  if (typeof entry === 'string') {
+    return entry;
+  }
+
+  if (entry && typeof entry === 'object' && typeof entry.pathname === 'string') {
+    return entry.pathname;
+  }
+
+  return null;
+};
+
+const normalizeWorkspaceEntries = (entries) => {
+  if (!Array.isArray(entries)) {
+    return [];
+  }
+
+  return _.uniq(entries.map(normalizeWorkspaceEntry).filter(Boolean));
+};
 
 class LastOpenedWorkspaces {
   constructor() {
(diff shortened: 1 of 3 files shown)


helloanoop (approved):

packages/bruno-electron/src/store/last-opened-workspaces.js
```diff
@@ -1,15 +1,58 @@
...
+* instead of a plain path string. Recover the path from `pathname`, drop anything unusable, and
+* de-duplicate so callers only ever see a string-only list.
+*/
+
+const normalizeWorkspaceEntry = (entry) => {
+  if (typeof entry === 'string') {
+    return entry;
+  }
+
+  if (entry && typeof entry === 'object' && typeof entry.pathname === 'string') {
```
bijin-bruno, line 15: I think we don't need to support this. I believe these were ignored in v3+. We can just clean it up?

packages/bruno-electron/src/store/last-opened-workspaces.js
```diff
@@ -1,15 +1,58 @@
...
+    const normalized = normalizeWorkspaceEntries(rawWorkspaces);
+
+    if (!_.isEqual(rawWorkspaces, normalized)) {
+      this.store.set('workspaces.lastOpenedWorkspaces', normalized);
+    }
   }
 
   getAll() {
-    return this.store.get('workspaces.lastOpenedWorkspaces', []);
+    return normalizeWorkspaceEntries(this.store.get('workspaces.lastOpenedWorkspaces', []));
```
bijin-bruno, line 55: is normalizeWorkspaceEntries required here since constructor should have already taken care of it?

packages/bruno-electron/src/store/last-opened-workspaces.js
```diff
@@ -1,15 +1,58 @@
...
+
+  /**
+  * migrate()
+  * One-shot heal on load: recover the path from any legacy object entries and scrub the objects
+  * from preferences.json so only string paths persist. Runs once at construction, before any read.
+  */
+
+  migrate() {
+    const rawWorkspaces = this.store.get('workspaces.lastOpenedWorkspaces', []);
+    const normalized = normalizeWorkspaceEntries(rawWorkspaces);
```
bijin-bruno, line 47: Would be better if we also check if the path still exist in the filesystem and remove if it doesn't

bijin-bruno (changes requested):

packages/bruno-electron/src/store/last-opened-workspaces.js
vijayh-bruno: are we going with clean up approach? I thought we didn't want to run cleanup. 

I'd added the pathname handling early assuming lots of users were hit, but it was a tiny dev-only group on the experimental build.

packages/bruno-electron/src/store/last-opened-workspaces.js
vijayh-bruno: Firstly, have we decided whether we want to run migrate() on startup?

I added both normalize and migrate() approaches so we can decide which one to pick. If we don't want to clean up, then we can get rid of migrate() altogether.

packages/bruno-electron/src/store/last-opened-workspaces.js
vijayh-bruno: if we plan to not run migrate on constructor then this will be needed

PR #8787, opened 2026-07-27
fix(openapi): sync enabled query params into the URL + enableOptionalParameters option

max-melhuish-depop:
### Description

Fixes to the OpenAPI → Bruno importer (and a matching exporter fix), all about query parameters.

**1. Enabled query params are now written into the request URL (bug fix).**

The importer set a request's URL to the path only and listed query params separately in the `params:query` block. Bruno derives the query string it *sends* from the URL, so an imported query param that was ticked was shown enabled but not actually applied until the user toggled it off and back on (which rewrites the URL from the params table).

Example — `GET /api/v2/embeddings/product/average/` with a ticked `normalise` param imported as:

```
get {
  url: {{baseUrl}}/api/v2/embeddings/product/average/
}
params:query {
  normalise: true
}
```

`normalise` is ticked but never sent. After this change the URL is kept in sync with the enabled query params (matching how Bruno persists a saved request), so it works on first open:

```
get {
  url: {{baseUrl}}/api/v2/embeddings/product/average/?normalise=true
}
```

**2. New `enableOptionalParameters` option (opt-in, non-breaking).**

`getParameterEntries` ticks every required parameter, plus any optional one that carries an example/default value, so whether an *optional* param is ticked depends on how thoroughly the spec documents it — a single spec can end up with all / some / none of its optional params ticked. Passing `enableOptionalParameters: false` (mirroring the option other OpenAPI importers such as Postman expose) disables every non-required parameter, so only required params are ticked. Behaviour is unchanged when the option is unset.

**3. Exporter: exclude the query string from the exported OpenAPI path.**

Change 1 means a saved request URL now canonically carries its enabled query params (e.g. `{{baseUrl}}/users?status=active`). The Bruno → OpenAPI exporter derived the path by stripping `{{baseUrl}}` and using the remainder verbatim, so it produced a path like `/users?status=active` instead of `/users`. Query params are already exported from the request's `params` array, so the exporter now drops the query string when deriving the path. (This also fixes the existing round-trip unit test, which surfaced the exporter bug once the importer started producing canonical URLs.)

Covered by new tests in `openapi-query-parameters.spec.js` plus the existing exporter round-trip test; the full `tests/openapi` suite passes (218 tests).

#### Contribution Checklist:

> Note: this PR bundles a few closely-related query-parameter changes (importer URL sync, an opt-in option, and the matching exporter path fix). Happy to split if you'd prefer.

## Summary by CodeRabbit

* **New Features**
  * OpenAPI imports now provide control over whether optional query parameters are enabled by default.
  * Imported requests now synchronize their URL query string based on which query parameters are enabled (including example/default values).

* **Bug Fixes**
  * Disabled optional query parameters are omitted from the request URL while remaining present as disabled parameters.
  * OpenAPI URL paths derived from base URL templates now exclude any query string portion.

* **Tests**
  * Added Jest coverage for enabled/disabled query parameter URL inclusion rules, including optional parameters with enums.

### DIFF
--- a/packages/bruno-app/src/utils/exporters/openapi-spec.js
+++ b/packages/bruno-app/src/utils/exporters/openapi-spec.js
@@ -119,21 +119,24 @@ export const exportApiSpec = ({ variables, items, name, environments }) => {
       if (baseUrlOverride) {
         const reqVarsMap = {};
         requestVars.filter((v) => v.enabled).forEach((v) => { reqVarsMap[v.name] = v.value; });
-        const path = rawUrl.slice('{{baseUrl}}'.length) || '/';
+        // A saved request keeps its enabled query params in the URL (e.g.
+        // "{{baseUrl}}/users?status=active"); the OpenAPI path must exclude the
+        // query string — query params are exported from the params array below.
+        const path = rawUrl.slice('{{baseUrl}}'.length).split('?')[0] || '/';
         return { url: interpolate(path, reqVarsMap), operationLevelServer: buildServerEntry(baseUrlOverride.value, reqVarsMap) };
       }
     }
 
     // URL uses {{baseUrl}} placeholder — strip it and resolve remaining path
     if (rawUrl.startsWith('{{baseUrl}}')) {
-      const path = rawUrl.slice('{{baseUrl}}'.length) || '/';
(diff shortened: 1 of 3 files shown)


max-melhuish-depop:
@bijin-bruno coderabbit has suggested you may be best placed to review this 🙂

PR #8796, opened 2026-07-28
fix: preserve elapsed time for streaming requests

### LINKED ISSUE #8792: Elapsed-time counter resets after switching away from a running request
Internal - [BRU-4095]

### I have checked the following

### This bug is

### Bruno version

3.5.1

### Operating System

macOS 26.5.2

### Describe the bug

When a long-running or streaming API request is in progress, switching to another request and then returning to the running request resets the elapsed-time display in the response pane.

The request itself continues running, but the displayed duration starts over instead of showing the total elapsed time since the request began. This makes the timing information inaccurate and makes it difficult to tell how long a streaming request has actually been running.

### Steps to reproduce

1. Open a request that returns a long-running or streaming response.
2. Send the request and wait for the elapsed-time counter to increase.
3. While the request is still running, switch to another API request.
4. Return to the original running request.
5. Observe that the elapsed-time counter has reset.

### Expected behavior

The elapsed-time counter should retain the original request start time and continue showing the total elapsed duration after navigating away and back.

### Actual behavior

The counter resets after returning to the running request, even though the request/stream is still active.

### Collection to reproduce

Any endpoint that keeps the response open long enough to switch between requests should reproduce the issue (for example, a streaming/SSE endpoint).

### Screenshots / live demo

The issue is visible in the response status bar: the elapsed-time value resets after switching away from and returning to the active streaming request.

Sudemc:
Internal - [BRU-4095]

### Description

Preserves the elapsed-time counter for active streaming HTTP responses when navigating away from and back to the request.

### Problem

`ResponseStopWatch` kept elapsed time only in component-local state. Switching requests unmounted the component, so returning to a still-running stream initialized the counter from the original response duration and appeared to reset it.

Fixes #8792.

### Fix

- Derive elapsed time from the stable `requestSent.timestamp` when available.
- Keep the existing duration-based ticking as a compatibility fallback.
- Synchronize immediately when the request-sent event arrives after the initial render.
- Clamp future timestamps to zero.
- Add focused regression coverage for navigation/remounting and event ordering.

### Screenshots

Not applicable; this changes timer continuity rather than layout or styling.

#### Contribution Checklist:

### Testing

- `ResponseStopWatch`: 5/5 tests passed
- ResponsePane scope: 48/48 tests passed
- ESLint on all three changed files
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response timing display accuracy when requests are actively running.
  * Preserved elapsed time across component updates and remounts.
  * Prevented future timestamps from showing negative durations.
  * Updated timing when request start information becomes available after rendering.

* **Tests**
  * Added coverage for elapsed-time calculation, fallback timing, timer progression, and future timestamps.

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/ResponseStopWatch/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/ResponseStopWatch/index.js
@@ -1,27 +1,41 @@
 import React, { useState, useEffect } from 'react';
 import StyledWrapper from './StyledWrapper';
 
-const ResponseStopWatch = ({ startMillis }) => {
-  const [milliseconds, setMilliseconds] = useState(startMillis);
+const TICK_INTERVAL = 100;
 
-  const tickInterval = 100;
-  const tick = () => {
-    setMilliseconds((_milliseconds) => _milliseconds + tickInterval);
-  };
+const isValidStartTime = (startTime) => Number.isFinite(startTime);
+
+const getElapsedTime = (startTime, startMillis) => (
+  isValidStartTime(startTime)
+    ? Math.max(0, Date.now() - startTime)
+    : startMillis
+);
+
+const ResponseStopWatch = ({ startTime, startMillis = 0 }) => {
+  const [milliseconds, setMilliseconds] = useState(() => getElapsedTime(startTime, startMillis));
 
   useEffect(() => {
-    let timerID = setInterval(() => {
-      tick();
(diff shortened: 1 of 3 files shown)


sreelakshmi-bruno:
Hi, we're dedicating bandwidth to review this PR and get it merged for release v4.2.0, which is planned to go out in September. Thank you for your contribution!

PR #8835, opened 2026-07-31
fix(e2e): Playwright failures from mock server and lazy mount

sid-bruno:
## Summary
- Update e2e call sites to use `createFolder` / `createRequest` (and edit-mode example testids) so Create no longer collides with the mock-server sidebar button and sidebar-created examples match `openInEditMode`.
- Fix cross-format / unmounted-collection drops: path confinement now includes last-opened collections, watchers reserve paths immediately, and dropping onto a collection mounts it first.
- Scope `createFolder` visibility asserts to the parent collection so duplicate folder names across collections do not trip strict mode.

## Test plan
- [x] Local Playwright (line reporter) for previously failing specs: save, draft-indicator, copy-*, move-tabs, cross-collection folder D&D, cross-format D&D, sidebar-state, save-as-example-multipart
- [ ] Confirm Linux Playwright CI shards go green on this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved drag-and-drop between collections by automatically loading target collections before moving items.
  * Improved collection file-watcher replacement and cleanup for more reliable updates.
  * Prevented folder creation checks from matching same-named folders in unrelated locations.

* **Tests**
  * Updated coverage for moving, copying, saving, examples, sidebar states, and environment selection.
  * Improved test reliability by using shared actions and reducing timing-dependent waits.

### DIFF
--- a/packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
+++ b/packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
@@ -284,9 +284,18 @@ const Collection = ({ collection, searchText }) => {
         setDropType('above');
       }
     },
-    drop: (draggedItem, monitor) => {
+    drop: async (draggedItem, monitor) => {
       const itemType = monitor.getItemType();
       if (isCollectionItem(itemType)) {
+        // Lazy-unmounted workspace collections are droppable in the sidebar but
+        // have no watcher yet — mount first so the move writes through and the UI updates.
+        if (collection.mountStatus !== 'mounted' && collection.mountStatus !== 'mounting') {
+          await dispatch(mountCollection({
+            collectionUid: collection.uid,
+            collectionPathname: collection.pathname,
+            brunoConfig: collection.brunoConfig
+          }));
+        }
         dispatch(handleCollectionItemDrop({ targetItem: collection, draggedItem, dropType: 'inside', collectionUid: collection.uid }));
       } else {
         dispatch(moveCollectionAndPersist({ draggedItem, targetItem: collection }));
--- a/packages/bruno-electron/src/ipc/collection.js
+++ b/packages/bruno-electron/src/ipc/collection.js
@@ -125,6 +125,8 @@ const envHasSecrets = (environment = {}) => {
   return secrets && secrets.length > 0;
 };
 
+const LastOpenedCollections = require('../store/last-opened-collections');
+
 const findCollectionPathByItemPath = (filePath) => {
   const normalizedFilePath = path.normalize(filePath);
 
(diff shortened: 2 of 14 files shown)


packages/bruno-electron/src/ipc/collection.js
```diff
@@ -125,6 +125,8 @@ const envHasSecrets = (environment = {}) => {
   return secrets && secrets.length > 0;
 };
 
+const LastOpenedCollections = require('../store/last-opened-collections');
+
```
sid-bruno: Drops lastOpenedCollections from path checks; mount already reserves the watcher path.

```suggestion
```

packages/bruno-electron/src/ipc/collection.js
```diff
@@ -152,11 +154,17 @@ const findCollectionPathByItemPath = (filePath) => {
...
+  // but not yet watched). Dropping into those must still pass path confinement.
+  const knownCollectionPaths = _.uniq([
+    ...collectionWatcher.getAllWatcherPaths(),
+    ...new LastOpenedCollections().getAll()
+  ]);
 
   // Find the collection path that contains this file
   // Sort by length descending to find the most specific (deepest) match first
-  const sortedPaths = allCollectionPaths.sort((a, b) => b.length - a.length);
+  const sortedPaths = knownCollectionPaths.sort((a, b) => b.length - a.length);
```
sid-bruno: Keeps confinement to mounted watchers once drop mounts first.

```suggestion
  const allCollectionPaths = collectionWatcher.getAllWatcherPaths();

  // Find the collection path that contains this file
  // Sort by length descending to find the most specific (deepest) match first
  const sortedPaths = allCollectionPaths.sort((a, b) => b.length - a.length);
```

sid-bruno:
Ignoring coderabbit since they would add intrusive changes to the watcher that I'd like to avoid. All we've done is made sure the watcher doesn't have to wait for 100ms before getting initialised which I don't see any documentation or reason for

and some spec file fixes for more reliable tests

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
```diff
@@ -284,9 +284,18 @@ const Collection = ({ collection, searchText }) => {
         setDropType('above');
       }
     },
-    drop: (draggedItem, monitor) => {
+    drop: async (draggedItem, monitor) => {
       const itemType = monitor.getItemType();
       if (isCollectionItem(itemType)) {
+        // Lazy-unmounted workspace collections are droppable in the sidebar but
```
vijayh-bruno, line 290: ensureCollectionIsMounted() exists in the same file https://github.com/usebruno/bruno/pull/8835/changes#diff-ebcb2be14e0022e5735a93742050fd7d566d6fa2dc36b02a780090db8b543032L129. 

can we reuse that here?

vijayh-bruno (approved):

packages/bruno-app/src/components/Sidebar/Collections/Collection/index.js
sid-bruno: We could but we don't need to wait for the mount, we want to just fire and let it do it so we can move on to the drop part.

Issue #8813, opened 2026-07-29
GraphQL Load from introspection fails to use inherited oauth2 auth - Unauthorized error


mykhailos635:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] Making Bruno unusable for me.
- [x] Slowing me down but I'm able to continue working.
- [x] Annoying.
- [ ] This feature was working in a previous version but is broken in the current release.

### Bruno version

4.0.0

### 💻 Operating System

Windows

### 🐞 Describe the bug

My GraphQL endpoint is secured with bearer auth.
When I setup oauth2 authentication on a folder level, I can successfully call my endpoints.
However, introspection call fails with a unauthorized error.

A similar issue was recently fixed: #4921
Another similar issue is still open: #4559 (potentially could be closed)

Here is a breakdown of the relation between a request auth values and interpolation calls:
1) Authentication mode Bearer - hardcoded bearer token - success
2) Authentication mode Bearer - interpolated variable (e.g. {{myAccessToken}} - success
3) Authentication mode Bearer - interpolated oauth2 variable (e.g. {{oauth2.myAccessToken.access_token}}) - fails
4) Authentication mode inherit(oauth2) - no explicit value - fails

### 📝 Steps to reproduce

- find a bearer token protected GraphQL API. 
- create a folder in bruno, setup oauth2 authentication for it
- create a request in that folder
- set request auth to "inherit"
- observe that queries work successfully for that request
- attempt to load introspection
- observe "unauthorized" error.

### 📷 Screenshots/Live demo link

<img width="2251" height="328" alt="Image" src="https://github.com/user-attachments/assets/322b0c8c-0c24-44ec-a78c-d9d0a55d4dc7" />

This image features all 3 key points:
- graphql query was successful (200 response)
- auth is inherit (oauth2)
- Unathorized error after pressing "Load from Introspection"

Rohan-Bruno:
Hi @mykhailos635, thank you for reporting this issue. To help us better understand and reproduce the issue, could you please provide a few additional details?

* The OAuth2 grant type you're using (e.g. Authorization Code, Client Credentials, PKCE).
* A minimal collection (with any sensitive information removed) that reproduces the issue.
* If possible, a GraphQL endpoint we can use to reproduce the issue, or let us know if you're able to reproduce it against a public GraphQL API.

This will help our engineering team investigate the issue further.

PR #8842, opened 2026-08-01
fix(graphql): resolve inherited folder auth for introspection requests

marcoarnulfo:
### Description

GraphQL introspection ignores auth inherited from a folder. With auth configured on a folder and the request's auth set to `inherit`, queries succeed but **Load from Introspection** returns Unauthorized.

### Problem

A normal request resolves inheritance in `prepareRequest`, which calls `mergeAuth(collection, request, requestTreePath)` before `setAuthHeaders` — the same thing `prepare-grpc-request` and `ws-event-handlers` already do.

`fetchGqlSchemaHandler` computes the same `requestTreePath` and calls `mergeVars`, but never `mergeAuth`. It then passes the **original** request to `prepareGqlIntrospectionRequest` instead of the resolved clone it just built, so even the merge it does do is thrown away. `setAuthHeaders` only looks at collection-level auth for an `inherit` request, so a folder's auth never reached the introspection call.

### Fix

Three lines in `fetchGqlSchemaHandler`:

- call `mergeAuth` next to the existing `mergeVars`;
- forward `resolvedRequest`, which is what the surrounding code already intended — the comment above it says the clone exists "to avoid mutating the original";
- normalise a missing `auth` block to `{ mode: 'none' }`. It is nullable in `bruno-schema`, and both `mergeAuth` and `setAuthHeaders` read `auth.mode` unguarded, so on this path a request without one failed outright whenever the collection root carried auth. The existing specs in this file build requests with no auth block, so the shape is not hypothetical.

The caller's request object is still left untouched, and there is a test for that.

**Tests.** The new spec stubs only the axios instance, so `prepareGqlIntrospectionRequest` runs for real and the assertions are on the `Authorization` header that would go on the wire. All four fail if any one of the three lines above is reverted.

The six existing variable-precedence tests compared the third argument for deep structural equality against the caller's object. The forwarded clone carries the merged variables, so that comparison no longer holds; they now pin the request's `uid`, which is the only part of that argument those tests are about.

`packages/bruno-electron`: 51 suites, 724 passing.

### Screenshots

Not applicable — the change is an HTTP request header. The observable behaviour is asserted in `fetch-gql-schema-handler-auth.spec.js`:

```js
await fetchGqlSchemaHandler(null, 'https://example.com/graphql', { variables: [] }, request, collection);

expect(sentRequest.headers['Authorization']).toBe('Bearer folder-token');
```

### Notes, deliberately out of scope

Three things found while working here. Happy to send any of them as separate PRs — none belongs in this one.

1. **#8813 also reports case 3** — a bearer token written as `{{$oauth2.<id>.access_token}}` — which still fails on introspection. `oauth2CredentialVariables` is never populated on this path (`prepareRequest` sets it, `fetchGqlSchemaHandler` does not), so the header goes out with the placeholder unresolved. This PR does not fix that, which is why it says "addresses" rather than "fixes".
2. **Folder-level headers have the same gap as auth had.** `prepareGqlIntrospectionRequest` merges only collection-root and request headers, while `prepareRequest` runs `mergeHeaders` over the whole folder chain — so headers defined on a folder are still missing from introspection.
3. **The six existing tests in `fetch-gql-schema-handler.spec.js` make real HTTPS calls to `example.com`.** They mock `prepareGqlIntrospectionRequest` but not `makeAxiosInstance`, so `await axiosInstance(request)` goes on the wire; with an unreachable proxy all six fail with `ECONNREFUSED`. They pass online only because example.com's 405 arrives as `error.response`. The file's top comment says the mock is there "to avoid network calls". The new spec added here shows the pattern that makes it hermetic.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * GraphQL schema introspection now correctly applies authentication inherited from parent folders.
  * Requests without authentication are handled consistently.
  * Request-specific authentication is preserved when fetching schemas.
  * Existing request details remain unchanged while authentication is resolved.
* **Tests**
  * Added coverage for inherited, request-specific, and missing authentication scenarios.
  * Improved validation of request forwarding and variable handling.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -27,7 +27,7 @@ const { uuid, safeStringifyJSON, safeParseJSON, parseDataFromResponse, parseData
 const { chooseFileToSave, writeFile, getCollectionFormat, hasRequestExtension } = require('../../utils/filesystem');
 const { addCookieToJar, getDomainsWithCookies, getCookieStringForUrl } = require('../../utils/cookies');
 const { createFormData } = require('../../utils/form-data');
-const { findItemInCollectionByPathname, sortFolder, getAllRequestsInFolderRecursively, getEnvVars, getTreePathFromCollectionToItem, mergeVars, sortByNameThenSequence } = require('../../utils/collection');
+const { findItemInCollectionByPathname, sortFolder, getAllRequestsInFolderRecursively, getEnvVars, getTreePathFromCollectionToItem, mergeVars, mergeAuth, sortByNameThenSequence } = require('../../utils/collection');
 const { getOAuth2TokenUsingAuthorizationCode, getOAuth2TokenUsingClientCredentials, getOAuth2TokenUsingPasswordCredentials, getOAuth2TokenUsingImplicitGrant, updateCollectionOauth2Credentials, clearOauth2CredentialsByCredentialsId } = require('../../utils/oauth2');
 const { preferencesUtil } = require('../../store/preferences');
 const { getProcessEnvVars } = require('../../store/process-env');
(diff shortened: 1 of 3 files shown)


marcoarnulfo:
Gentle ping on this one. The three workflows (Lint Checks, Windows Tests, Linux Tests) are still sitting in `action_required` — it's my first PR from a fork, so they need a maintainer to approve the runs. Could someone start them off?

The branch is still conflict-free against `main`, and the suite passes locally with `npm run test:ci` in `packages/bruno-electron`. The Snyk check reports an error from the org's quota rather than anything in this change.

Happy to rebase or split this up if that would make it easier to review.

Issue #5755, opened 2025-10-09
Generate a new sample message based on schema is missing fields with gRPC


expe-elenigen:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] making Bruno unusable for me
- [ ] slowing me down but I'm able to continue working
- [ ] annoying
- [ ] this feature was working in a previous version but is broken in the current release.

### Bruno version

2.12.0

### Operating System

macOS 15.6.1

### Describe the bug

When I use `Generate a new sample message based on schema` it would successfully generate a message with the first level of fields having native types (boolean, string, arrays of strings, etc) but for more complex sub-types, it's only generating `{}` instead of going recursively.

### Screenshots/Live demo link

Actual:
<img width="311" height="369" alt="Image" src="https://github.com/user-attachments/assets/478a4514-895f-4546-a210-a0da72d2bd30" />
Here `field4` is left empty

Expected:
<img width="331" height="423" alt="Image" src="https://github.com/user-attachments/assets/7b6bc033-c1db-40a2-8e88-b2f1fbe711ac" />
Here `field4` is filled with all the possible values.

The expected behavior would match for example what Postman is doing and while doing this, it also help us to understand what's the full schema for that query.

helloanoop:
Thanks for reporting this @expe-elenigen . We'll investigate and get back.

alejo-rivera:
Some more details https://github.com/usebruno/bruno/issues/6397

Edit: correct issue

Davidos533:
Any solutions about it? Problem still exists

PR #8847, opened 2026-08-02
fix(grpc): expand nested message fields when generating sample messages

thomas-maurice:
### Description

This PR fixes the fact that Bruno cannot generate example payloads for protobufs messages that are nested and imported from other proto files.

This addresses issue #5755

### Problem

"Generate sample message" only fills top level scalar fields. Any field whose
type is another message comes out as `{}`, in both reflection and proto file
mode.

The generator only recurses when the field descriptor carries an inlined
`messageType`, but proto-loader definitions never have that: message fields
only carry a `typeName` string, and nothing resolved it. So the recursion
branch was effectively dead code.

While testing the fix inside the app I found two more problems sitting on top:

1. `getCircularReplacer` replaces any object it has already seen with the
   string `"[Circular]"`, not just actual cycles. Methods of a service share
   the same `fileDescriptorProtos` array, so after the load-methods IPC
   response goes through `safeStringifyJSON`, every method except the first
   ends up with `"fileDescriptorProtos": "[Circular]"`. That corrupted copy
   gets persisted in the renderer's reflection cache and was passed back as
   `methodMetadata`, which always won over the clean cache in the main process.
2. Buffers that cross IPC or come back from persisted state turn into
   `{ "type": "Buffer", "data": [...] }` objects, which the decoder rejects.

### Fix

- `grpcMessageGenerator.js`: build a type index from the
  `fileDescriptorProtos` that proto-loader already attaches to every method
  definition, and resolve `typeName` against it. Files are decoded one by one
  with `protobufjs/ext/descriptor` (building a full Root from the set throws on
  some valid descriptors, e.g. proto3 optional fields). Handles imported and
  nested messages, enums (first declared value name), maps (rendered as an
  object, detected by the key/value entry shape since the map_entry option is
  not reliable in synthesized descriptors), recursive types (falls back to
  `{}`), and Buffers that were JSON serialized.
- `grpc-client.js`: only prefer `methodMetadata` when its descriptors are
  actually a usable array, otherwise fall back to the method cache. This works
  around the `"[Circular]"` corruption described above. The replacer itself
  probably deserves its own fix, happy to open a separate issue for it.
- `protobufjs` added as a direct dependency of `bruno-requests`. It is already
  in the tree through `@grpc/proto-loader` and the range matches the locked
  version, so nothing changes in resolved dependencies.

### Screenshots

| Before | After |
| --- | --- |
| image | image  |

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Improved gRPC sample message generation using protobuf descriptors.
  * Added support for nested messages, referenced types, map fields, and enum values.
  * Supports descriptors provided in multiple encoded and serialized formats.

* **Bug Fixes**
  * Improved fallback behavior when request method metadata is incomplete or unavailable.
  * Added safeguards against recursive message expansion.
  * More reliably generates usable sample requests for complex protobuf definitions.


thomas-maurice:
@sid-bruno do you think this is mergeable ?

PR #8855, opened 2026-08-03
fix: remove the 50 cap on maxRedirects and normalise it on every read path

prateek-bruno:
## Description

Ref: BRU-3808

Importing a Postman collection that sets `maxRedirects` above 50 no longer fails the entire import: the schema cap is gone, and a value Bruno cannot honour is reported as a warning instead of aborting. Every parser and writer now applies one normalization rule, which also fixes a round-trip bug where a ceiling Bruno itself wrote as `1e+21` was read back as `1`.

### Problem

Postman allows an arbitrarily large `maxRedirects`, but `itemSchema` capped it at 50, so one request over the cap failed validation and took the whole import down with it. Two quieter defects sat behind that. The bru parser read the setting with `parseInt`, which stops at the exponent, so any ceiling large enough to serialize in exponential form silently became `1` the next time the request was opened. The yml and OpenCollection paths used `toNumber`, which returns any number untouched, letting `Infinity`, `NaN` and negatives reach the request pipeline from a hand-edited file.

### Fix

The schema now accepts any whole number of 0 or more (`.integer().min(0)` in place of `.min(0).max(50)`), which is only safe because every parser normalizes before validation runs - the two halves have to ship together.

**Unusable values warn instead of aborting the import**

The Postman importer accepts any number of 0 or more, truncating fractions, and routes anything else (negative, non-numeric, non-finite) to the existing import-issues toast as a per-request warning. The request still imports and falls back to the default, so one malformed value costs a warning rather than the collection.

**One normalization rule on every read and write**

`parseMaxRedirects` in `@usebruno/common` is the single rule, used by both yml parsers, all four OpenCollection converter directions, and the yml writers - normalizing on the way out too, so `parse(stringify(x))` is stable. The bru parser applies the same rule inline, since `bruno-lang` is a leaf package that cannot import `bruno-common`. This changes how existing files are read:

```
on-disk value    before                      after
1e+21            1                           1e+21
3.5              3.5 (yml), 3 (bru)          3
-1               -1 (yml), dropped (bru)     default
.inf / .nan      Infinity / NaN              default
51               rejected on save            51
```

A previously stored out-of-range value is therefore replaced on the next save. Note also that a value ≥ 1e21 written by this version reads back as `1` in older Bruno versions, since their `parseInt` is the bug being fixed here.

**The request pipeline is deliberately untouched**

The electron and CLI runners still honour whatever `req.setMaxRedirects()` or an unsaved draft sets, without re-validating. No upper bound is reintroduced anywhere.

### Screenshots

Before:
Screenshot 2026-08-03 at 4 22 47 PM

After:
Screenshot 2026-08-03 at 4 22 01 PM

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - Supports larger and more flexible maximum redirect limits, including zero and exponent-form values.
  - Preserves valid redirect settings across imports, exports, and saved collection formats.

- **Bug Fixes**
  - Postman imports now validate redirect limits and report warnings for invalid values while continuing to import affected requests.
  - Invalid or unsupported values safely fall back to the default redirect limit.
  - Fractional values are consistently truncated.


packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -553,11 +553,11 @@ const runSingleRequest = async function (
     const forwardAuthorizationHeader = request.settings?.forwardAuthorizationHeader ?? true;
 
     // Get maxRedirects from request settings, fallback to request.maxRedirects, then default to 5
-    let requestMaxRedirects = request.settings?.maxRedirects ?? request.maxRedirects ?? 5;
+    let requestMaxRedirects = request.settings?.maxRedirects ?? request.maxRedirects ?? DEFAULT_MAX_REDIRECTS;
 
     // Ensure it's a valid number
     if (typeof requestMaxRedirects !== 'number' || requestMaxRedirects < 0) {
-      requestMaxRedirects = 5; // Default to 5 redirects
+      requestMaxRedirects = DEFAULT_MAX_REDIRECTS;
```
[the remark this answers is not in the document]
prateek-bruno: Script level maxRequests changes are not covered in this ticket.

packages/bruno-schema/src/collections/itemSchema.spec.js
```diff
@@ -88,4 +88,47 @@ describe('Item Schema Validation', () => {
...
+      });
+    });
+
+    it.each(['100', 'abc'])('item schema must throw an error if maxRedirects is the string %p', async (maxRedirects) => {
+      await expect(itemSchema.validate(itemWithMaxRedirects(maxRedirects))).rejects.toMatchObject({
+        path: 'settings.maxRedirects',
+        type: 'typeError'
+      });
+    });
+  });
```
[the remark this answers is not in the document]
prateek-bruno: This is false positive. Locally running ESLint proves semicolons are not missing.

tests/import/postman/import-postman-max-redirects.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  importCollection,
+  selectRequestPaneTab
+} from '../../utils/page';
+import { buildCommonLocators } from '../../utils/page/locators';
+import { captureClipboardWrites } from '../../utils/clipboard';
+
+const COLLECTION_NAME = 'Max Redirects Collection';
+
+// Requests whose maxRedirects cannot be honoured: each produces one warning and falls back to 5.
+const OFFENDERS = ['Null Limit', 'Negative Limit', 'Quoted Limit', 'Fractional Limit'];
```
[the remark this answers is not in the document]
prateek-bruno: Added!

packages/bruno-lang/v2/src/bruToJson.js
```diff
@@ -561,11 +561,15 @@ const sem = grammar.createSemantics().addAttribute('ast', {
...
-        parsedSettings.maxRedirects = maxRedirects;
+    // Number, not parseInt: parseInt stops at the exponent, so 1e+21 reads back as 1. The truthy
+    // check skips a blank, which Number reads as 0, meaning "no redirects" rather than "unset"; a
+    // real 0 survives as '0'. Mirrors toMaxRedirects in @usebruno/common (unimportable: bruno-lang
+    // is a leaf), but leaves the key unset instead of defaulting, since jsonToBru writes back only
+    // the keys it is given.
+    if (settings.maxRedirects) {
+      const maxRedirects = Number(settings.maxRedirects);
+      if (Number.isFinite(maxRedirects) && maxRedirects >= 0) {
+        parsedSettings.maxRedirects = Math.trunc(maxRedirects);
```
sanish-bruno, line 572: I see yml sets default max-redirect at 5, when we have it missing from imports, are not doing the same for bru, or are we doing it at elsewhere?

packages/bruno-filestore/src/maxRedirects.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    method: 'GET',
+    headers: [],
+    params: [],
+    body: { mode: 'none' },
+    auth: { mode: 'none' },
+    script: {},
+    vars: {},
+    assertions: [],
+    tests: '',
+    settings: { maxRedirects }
```
sanish-bruno: is this intentional, why there are settings at root level as well as within request object

packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -553,11 +553,11 @@ const runSingleRequest = async function (
     const forwardAuthorizationHeader = request.settings?.forwardAuthorizationHeader ?? true;
 
     // Get maxRedirects from request settings, fallback to request.maxRedirects, then default to 5
-    let requestMaxRedirects = request.settings?.maxRedirects ?? request.maxRedirects ?? 5;
+    let requestMaxRedirects = request.settings?.maxRedirects ?? request.maxRedirects ?? DEFAULT_MAX_REDIRECTS;
```
sanish-bruno, line 556: Okay during request execution we are defaulting to 5 if the value is undefined, so undefined values from bru, will eventually get 5 as max redirect value.If yml imports are adding max-redirects as 5, i guess, bru ones should also do the same! Thoughts? @prateek-bruno

packages/bruno-common/src/utils/index.spec.ts
```diff
@@ -33,3 +33,40 @@ describe('resolveTimeoutSetting', () => {
     expect(resolveTimeoutSetting(true)).toBe(0);
   });
 });
+
+describe('toMaxRedirects', () => {
+  // Every fallback assertion below goes through the constant, so pin its value here or a change to
+  // it would pass silently.
+  test('the documented default is 5', () => {
```
sanish-bruno: is this test needed?

packages/bruno-converters/tests/common/postman-collection.js
```diff
@@ -0,0 +1,19 @@
...
+  info: {
+    _postman_id: 'test-id',
+    name: 'Test Collection',
+    schema: 'https://schema.getpostman.com/json/collection/v2.1.0/collection.json'
+  },
+  item: items,
+  ...overrides
+});
+
+export const makeRequest = (name, { method = 'GET', url = 'https://example.com', protocolProfileBehavior } = {}) => ({
```
sanish-bruno, line 40: @prateek-bruno can we have a jsdoc of the structure of protocolProfileBehaviuor, just to make it easier to work with in future?

packages/bruno-converters/tests/postman/postman-to-bruno/max-redirects.spec.js
```diff
@@ -0,0 +1,76 @@
...
+        message: 'Invalid maxRedirects, ignored (must be a whole number of 0 or more)'
+      });
+    }
+  );
+
+  it.each(['1e309', '-1e309', `1${'0'.repeat(400)}`])(
+    'should warn for a maxRedirects literal of %s that overflows to a non-finite number',
+    async (literal) => {
+      const postmanJson = JSON.stringify(makeCollection([makeRequest('Req', { protocolProfileBehavior: {} })]));
+      const withOverflow = postmanJson.replace('"protocolProfileBehavior":{}', `"protocolProfileBehavior":{"maxRedirects":${literal}}`);
```
sanish-bruno: why are we doing the withOverflow step? why are we stringifying json created makeCollection and parse and pass it to the postmanToBruno function? Is the stringification needed?

packages/bruno-schema/src/collections/itemSchema.spec.js
```diff
@@ -88,4 +88,54 @@ describe('Item Schema Validation', () => {
...
+
+    it.each([0, 50, 51, 1000, Number.MAX_SAFE_INTEGER, 1e21, 9.999999999998865e21])(
+      'item schema must accept a maxRedirects of %p',
+      async (maxRedirects) => {
+        const validated = await itemSchema.validate(itemWithMaxRedirects(maxRedirects));
+        expect(validated.settings.maxRedirects).toBe(maxRedirects);
+      }
+    );
+
+    it('item schema must accept a null maxRedirects', async () => {
```
sanish-bruno, line 115: maybe test for `undefined` cause that is most likely to happen, when it comes to bru cases, yml already defaults to 5 while importing if the value is undefined i believe?

tests/import/postman/fixtures/postman-with-max-redirects.json
```diff
@@ -0,0 +1,187 @@
...
+        "url": {
+          "raw": "https://api.acme.test/boolean",
+          "protocol": "https",
+          "host": ["api", "acme", "test"],
+          "path": ["boolean"]
+        }
+      }
+    },
+    {
+      "name": "Reporting",
```
sanish-bruno, line 182: 1e30 fixtures for import?

tests/import/postman/import-postman-max-redirects.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+      const issuesSummary = (await readCopiedText()) ?? '';
+      for (const requestName of OFFENDERS) {
+        expect(issuesSummary).toContain(`[WARNING] ${requestName}`);
+      }
+      expect(issuesSummary.match(/Invalid maxRedirects, ignored \(must be a whole number of 0 or more\)/g)).toHaveLength(
+        OFFENDERS.length
+      );
+    });
+
+    await test.step('warnings carry no request data, so no include-items checkbox appears', async () => {
```
sanish-bruno: I think this step is probably not needed, is is more of component behaviour

packages/bruno-filestore/src/maxRedirects.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  },
+  settings: { maxRedirects }
+});
+
+// A number is serialized by Bruno; a string is spliced in as raw text, for values Bruno would never
+// write (.inf, blank, quoted). Assert on `stored` when a value must survive verbatim, since bru
+// omits a setting it will not honour and `effective` would hide that behind the runtime's `?? 5`.
+const roundTripRequest = (maxRedirects: number | string, format: CollectionFormat) => {
+  const seed = typeof maxRedirects === 'number' ? maxRedirects : DEFAULT_MAX_REDIRECTS;
```
sanish-bruno:
```js
const seed = typeof maxRedirects === 'number' ? maxRedirects : DEFAULT_MAX_REDIRECTS;
```

defaulting to DEFAULT_MAX_REDIRECTS will cause false positives, seed the actual value to the `parseRequest` as that's what we are testing here!

packages/bruno-filestore/src/maxRedirects.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  let text = stringifyRequest(requestWithMaxRedirects(seed) as any, { format });
+  if (typeof maxRedirects === 'string') {
```
sanish-bruno: If we are testing the parser we should also avoid doing this, this will also cause false positives

packages/bruno-filestore/src/maxRedirects.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    text = text.replace(`maxRedirects: ${seed}`, `maxRedirects: ${maxRedirects}`);
+  }
+  const stored = parseRequest(text, { format }).settings.maxRedirects;
+  return { stored, effective: stored ?? DEFAULT_MAX_REDIRECTS };
```
sanish-bruno: This line also will have false positive problem

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -140,11 +140,11 @@ const configureRequest = async (
   const forwardAuthorizationHeader = request.settings?.forwardAuthorizationHeader ?? true;
 
   // Get maxRedirects from request settings, fallback to request.maxRedirects, then default to 5
-  let requestMaxRedirects = request.settings?.maxRedirects ?? request.maxRedirects ?? 5;
+  let requestMaxRedirects = request.settings?.maxRedirects ?? request.maxRedirects ?? DEFAULT_MAX_REDIRECTS;
```
sanish-bruno, line 143: can we have  `toMaxRedirects(request.settings?.maxRedirects ?? request.maxRedirects)`

packages/bruno-cli/src/runner/run-single-request.js
sanish-bruno, line 556: can use `toMaxRedirects` , also here `packages/bruno-electron/src/ipc/network/index.js:140`

sanish-bruno (changes requested):

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: I didn't wanted to tweak the existing behaviour of bru returning undefined instead of 5 that's the only reason I didn't changed that. If we want to do that I guess we can do it.

packages/bruno-electron/src/ipc/network/index.js
prateek-bruno: I don't think we need toMaxRedirects here. Cause Script can still set it to Infinity and we are not changing that.

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: Same reasoning as above: https://github.com/usebruno/bruno/pull/8855#discussion_r3719696753

packages/bruno-filestore/src/maxRedirects.spec.ts
prateek-bruno: Actually here this is fine, cause in the upcoming lines we're replacing the DEFAULT_MAX_REDIRECTS to the string value. This is done cause the file writer might normalise the invalid maxRedirects, and we want it to test what happens if file has the invalid value.

packages/bruno-lang/v2/src/bruToJson.js
prateek-bruno: Replied here: https://github.com/usebruno/bruno/pull/8855#discussion_r3719678355

packages/bruno-converters/tests/postman/postman-to-bruno/max-redirects.spec.js
prateek-bruno: If we just pass maxRequests in makeRequest then it would get normalised. So .inf won't even reach in the postmanJson. So we're constructing invalid postmanJson by first constructing a valid postmanJson then replacing the maxRequest with an invalid one. This is also followed in other spec where we test similar malformed behaviour.

packages/bruno-filestore/src/maxRedirects.spec.ts
prateek-bruno: Related: https://github.com/usebruno/bruno/pull/8855#discussion_r3719791169

packages/bruno-converters/tests/postman/postman-to-bruno/max-redirects.spec.js
prateek-bruno: The alternate is to construct the postman json manually so that we can set any value there. That would look more readable, what do you think?

packages/bruno-schema/src/collections/itemSchema.spec.js
prateek-bruno: There won't be a case for this in bru or yml so I didn't added test for it:

```
settings: { maxRedirects: undefined } 
```

But I can add it.

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: Right now basically yml writes maxRequests 5 for every request if it doesn't exist but bru just omits the maxRequests key in write. That's a different behaviour and if we do it bru writer will start writing maxRequests 5 if undefined for every .bru file.

packages/bruno-filestore/src/maxRedirects.spec.ts
prateek-bruno: Done! (will push soon)

packages/bruno-common/src/utils/index.spec.ts
prateek-bruno: Removed it, it wasn't needed! (will push soon)

packages/bruno-converters/tests/common/postman-collection.js
prateek-bruno: Done! (will push soon)

packages/bruno-filestore/src/maxRedirects.spec.ts
prateek-bruno: Resolved!

packages/bruno-filestore/src/maxRedirects.spec.ts
prateek-bruno: Resolved! It is better now, I have rewritten it to construct the hand-modified maxRequests.

packages/bruno-filestore/src/maxRedirects.spec.ts
prateek-bruno: I have rewritten this for hand-modified maxRequests.

tests/import/postman/fixtures/postman-with-max-redirects.json
prateek-bruno: Added!

tests/import/postman/import-postman-max-redirects.spec.ts
prateek-bruno: Removed!

packages/bruno-converters/tests/postman/postman-to-bruno/max-redirects.spec.js
prateek-bruno: I have rewritten this for hand-modified maxRequests.

packages/bruno-cli/src/runner/run-single-request.js
sanish-bruno: makes sense, let't not have any surprises to existing bru users

packages/bruno-electron/src/ipc/network/index.js
sanish-bruno: understood

packages/bruno-filestore/src/maxRedirects.spec.ts
sanish-bruno: Nice

packages/bruno-schema/src/collections/itemSchema.spec.js
sanish-bruno: let's add it considering more defensive approach

tests/import/postman/import-postman-max-redirects.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    expectIssues: true
+  });
+};
+
+const expectMaxRedirects = async (page: Page, requestName: string, expected: string) => {
+  const locators = buildCommonLocators(page);
+  await locators.sidebar.request(requestName).click();
+  await selectRequestPaneTab(page, 'Settings');
+  await expect(locators.requestSettings.maxRedirectsInput()).toHaveValue(expected);
+};
```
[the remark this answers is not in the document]
prateek-bruno: Fixed!

packages/bruno-schema/src/collections/itemSchema.spec.js
prateek-bruno: Added!

sanish-bruno (approved):

packages/bruno-converters/tests/common/postman-collection.js
```diff
@@ -0,0 +1,48 @@
...
+ * The `item.protocolProfileBehavior` fields the Bruno importer reads. Postman types them, but real
+ * exports are not guaranteed to honour that, so importer tests pass unexpected values here too.
+ * Any other field Postman sends is ignored on import.
+ *
+ * @typedef {Object} ProtocolProfileBehavior
+ * @property {boolean} [followRedirects] - Maps to Bruno's followRedirects setting.
+ * @property {number} [maxRedirects] - Maps to Bruno's maxRedirects setting.
+ * @property {boolean} [followAuthorizationHeader] - Maps to Bruno's forwardAuthorizationHeader setting.
+ * @property {boolean} [disableUrlEncoding] - Inverted into Bruno's encodeUrl setting.
+ */
```
vijayh-bruno, line 28: What is this comment block for?

prateek-bruno: It was requested by @sanish-bruno here:
https://github.com/usebruno/bruno/pull/8855#discussion_r3719344815

vijayh-bruno (approved):

PR #8862, opened 2026-08-03
fix(mount): hydrate websocket message uids in the cached mount path

chirag-bruno:
### Problem

With **File Cache** (mount v2) enabled, WebSocket message bodies could not be expanded after reopening the app.

WS message expand/collapse is keyed by the message `uid`, and the toggle bails out on a falsy one:

```js
const toggleMessage = useCallback((uid) => {
  if (!uid) return;
```

The v1 mount path hydrates `request.body.ws` (`hydrateRequestWithUuid`), but mount v2's `REQUEST_UID_PATHS` omitted it. With the cache on, the tree is built from the cache on cold start, so every WS message arrived with `uid: undefined` — nothing expandable, and the whole list rendering with the same undefined React key.

It only reproduced after a restart because live edits during a session still flow through the collection watcher's v1 hydration.

### Fix

- Add `request.body.ws` to `REQUEST_UID_PATHS` in the mount v2 tree builder. Since v2 seeds uids deterministically from the file path, expanded-message state now also survives a restart instead of resetting.
- Add `tree-builder.spec.js` covering uid presence/uniqueness for every hydrated request list, plus WS-specific uid stability across rebuilds and no cross-request collisions. One test pins the spec's path list against `REQUEST_UID_PATHS` so adding or removing a path there can't silently go uncovered.

### Steps to verify

1. Open a collection with WS requests that have several messages, confirm the bodies expand.
2. Enable **Preferences → Cache → File cache**, close and reopen the app.
3. Expand the same message bodies — they now open as before.

### Notes

gRPC and GraphQL are unaffected: gRPC messages carry no uid in either mount path (`GrpcBody` addresses them by array index), and `body.graphql` is an object rather than a list. `EXAMPLE_UID_PATHS` needs no `ws` entry either — WS requests don't support examples at any layer (the UI gates "Create Example" to `http-request`, the yml parser hardcodes `examples: []`, and the `.bru` example grammar has no `body:ws`).

Ticket: BRU-4089

### DIFF
--- a/packages/bruno-electron/src/services/mount/tree-builder.js
+++ b/packages/bruno-electron/src/services/mount/tree-builder.js
@@ -21,7 +21,8 @@ const REQUEST_UID_PATHS = [
   ['request.assertions', 'assertions'],
   ['request.body.formUrlEncoded', 'body.formUrlEncoded'],
   ['request.body.multipartForm', 'body.multipartForm'],
-  ['request.body.file', 'body.file']
+  ['request.body.file', 'body.file'],
+  ['request.body.ws', 'body.ws']
 ];
 
 const EXAMPLE_UID_PATHS = [
@@ -241,4 +242,4 @@ const buildTree = (collectionPath, parserResults, options = {}) => {
   return tree;
 };
 
-module.exports = { buildTree };
+module.exports = { buildTree, REQUEST_UID_PATHS };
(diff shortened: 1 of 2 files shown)


sachin-thakur-bruno (approved):
LGTM. Failed grpc tests are not due to this PR.

PR #8878, opened 2026-08-05
fix(datatype): Variable type validation flag reference mismatch issue

sharan-bruno:
BRU-3849

### Description

Fixes the false data-type warning shown on variable values that reference other variables. DataTypeSelector now resolves the reference before validating.

### Problem

variable.value was type-checked literally, so {{port}} typed as number was seen as the string "{{port}}". Result: a bogus "Value is not a valid number" flag even when port is a number.

### Fix

Added getReferencedVariableType — extracts the {{...}} name, resolves it via getAllVariables, and returns that variable's real type (supports dotted paths and dotted names). References compare against the referenced type; non-references still go through validateDataTypeValue. Adds a collection prop, already passed by all four call sites.

### Screenshots

| Before | After |
| --- | --- |
| image | image |

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Variable references are now validated according to their inferred data types.
  * Collection environment references, including nested values, are supported.
  * Existing validation and coercion behavior remains unchanged for literal values.

* **Tests**
  * Added coverage for matching and mismatched references across multiple data types.
  * Added collection and environment scenarios to verify consistent type validation.


tests/variable-datatypes/reference-types/fixtures/workspace/workspace.yml
```diff
@@ -0,0 +1,12 @@
+opencollection: 1.0.0
+info:
+  name: "Reference Types Workspace"
+  type: workspace
+
+collections:
+  - name: "reference-types"
+    path: "collections/reference-types"
+
+specs:
```
sanish-bruno: remove spec: , docs: ?

packages/bruno-app/src/components/DataTypeSelector/index.js
```diff
@@ -8,10 +8,23 @@ import {
   IconToggleRight
 } from '@tabler/icons';
 import { Tooltip } from 'react-tooltip';
-import { BRUNO_VARIABLE_DATATYPES, parseValueByDataType, validateDataTypeValue } from '@usebruno/common/utils';
+import { BRUNO_VARIABLE_DATATYPES, parseValueByDataType, validateDataTypeValue, getDataTypeFromValue } from '@usebruno/common/utils';
+import get from 'lodash/get';
```
sanish-bruno: nit: external lib imports always at the top. then @usebruno scoped libraries, then utils from package, then component imports

packages/bruno-app/src/components/DataTypeSelector/index.js
```diff
@@ -26,10 +39,14 @@ const TYPE_ICONS_SIZES = {
   object: 18
 };
 
-const DataTypeSelector = ({ variable, onChange, compact = false }) => {
+const DataTypeSelector = ({ variable, collection, onChange, compact = false }) => {
+  const resolvableVariables = useMemo(() => getAllVariables(collection), [collection]);
```
sanish-bruno: `getAllVariables` re-runs on every mounted row whenever collection, this will cause performance issues, can we hoist this to the parent `EnvironmentVariablesTable`, add this  `const resolvableVariables = useMemo(() => getAllVariables(_collection), [_collection]);` as a sibling memo to the _collection, and pass it down, so that each row doesn't have to call `getAllVariables` which is very computationally complex! Also DataTypeSelector shouldn't know about collection at all, Change the prop to resolvableVariables and drop collection

This should be done at all the places we need to have this. so more files needs changes. Be mindful about where all we need to add this.

packages/bruno-app/src/components/DataTypeSelector/index.js
```diff
@@ -26,10 +39,14 @@ const TYPE_ICONS_SIZES = {
   object: 18
 };
 
-const DataTypeSelector = ({ variable, onChange, compact = false }) => {
+const DataTypeSelector = ({ variable, collection, onChange, compact = false }) => {
```
sanish-bruno: Also i don't see any files that changed to pass down `collection` to this component.

packages/bruno-app/src/components/DataTypeSelector/index.js
```diff
@@ -26,10 +39,14 @@ const TYPE_ICONS_SIZES = {
...
+  const resolvableVariables = useMemo(() => getAllVariables(collection), [collection]);
+
+  const referencedType = getReferencedVariableType(variable.value, resolvableVariables);
   const selectedType = variable.dataType || 'string';
+  const referencedTypeError = referencedType === selectedType ? null : `Value is not a valid ${selectedType}`;
```
sanish-bruno: Better to give the responsibility to `validateDataTypeValue` that having referencedTypeError here , keep the concerns together may be something like this

```js
export const validateTypedVariableValue = (value, dataType, resolvableVariables) => {
    const referencedType = getReferencedVariableType(value, resolvableVariables);
    if (referencedType !== null) {
      return referencedType === dataType ? null : `Value is not a valid ${dataType}`;
    }
    return validateDataTypeValue(parseValueByDataType(value, dataType), dataType);
  };

```

packages/bruno-app/src/components/DataTypeSelector/index.js
sanish-bruno, line 36: Taking a tangent from existing implementation , considering unit testability and responsibility, why can't we think of something like this

In `packages/bruno-common/src/utils/datatype.ts`
```js


  // Resolves `path` on `obj`, matching lodash.get semantics we rely on:
  // a literal key that contains dots wins over walking the same-looking dotted path.
  // (Env vars can be named e.g. "globalEnvObject.port" as a flat key.)
  const getByPath = (obj: Record<string, any>, path: string): any => {
    if (obj == null) return undefined;
    if (Object.prototype.hasOwnProperty.call(obj, path)) return obj[path];
    return path.split('.').reduce<any>((acc, key) => (acc == null ? undefined : acc[key]), obj);
  };

  // Whole-string template reference only: '{{a.b}}' resolves, 'x-{{a}}-y' does not.
  const resolveWholeReference = (value: any, resolvableVariables: Record<string, any>): any => {
    if (typeof value !== 'string') return undefined;
    const match = value.trim().match(/^\{\{([^}]+)\}\}$/);
    if (!match) return undefined;
    return getByPath(resolvableVariables, match[1].trim());
  };

  // string-form → typed JS value, or raw on failure. When `resolvableVariables`
  // is provided and `value` is a whole-string template reference, the resolved
  // JS value is returned as-is (its declared type is authoritative — we don't
  // re-coerce it).
  export const parseValueByDataType = (
    value: any,
    dataType?: BrunoVariableDataType,
    resolvableVariables?: Record<string, any>
  ): any => {
    if (resolvableVariables) {
      const resolved = resolveWholeReference(value, resolvableVariables);
      if (resolved !== undefined) return resolved;
    }
    if (!dataType || dataType === 'string') return value;
    // ... existing number / boolean / object branches unchanged ...
  };
```

in `packages/bruno-app/src/components/DataTypeSelector/index.js`
```js
  const DataTypeSelector = ({ variable, resolvableVariables, onChange, compact = false }) => {
    const selectedType = variable.dataType || 'string';
    const coercedValue = parseValueByDataType(variable.value, selectedType, resolvableVariables);
    const typeError = validateDataTypeValue(coercedValue, selectedType);
    // ... rest unchanged
  };

```

and unit test cases in `packages/bruno-common/src/utils/datatype.spec.ts `

sanish-bruno (changes requested):

packages/bruno-common/src/utils/datatype.ts
```diff
@@ -62,9 +84,10 @@ export const valueToString = (value: unknown, indent?: number): string => {
 
 // Returns an error message when post-coerce value's JS type doesn't match dataType.
 export const validateDataTypeValue = (value: any, dataType?: BrunoVariableDataType): string | null => {
-  if (!dataType || dataType === 'string') return null;
+  if (!dataType) return null;
```
sanish-bruno: We were returning null `previously` when `dataType === 'string'` , please check the consumers of this function, and ensure that we are not introducing any regressions

packages/bruno-common/src/utils/datatype.ts
```diff
@@ -5,9 +5,31 @@ export const BRUNO_VARIABLE_DATATYPES: readonly BrunoVariableDataType[] = ['stri
...
+};
+
 // string-form → typed JS value, or raw on failure.
-export const parseValueByDataType = (value: any, dataType?: BrunoVariableDataType): any => {
+export const parseValueByDataType = (
+  value: any,
+  dataType?: BrunoVariableDataType,
+  resolvableVariables?: Record<string, any>
+): any => {
+  if (resolvableVariables) {
```
sanish-bruno, line 29: we will need to pass down the resolved value to check if they are parsable as a number, boolean, object, than return the value as we were doing before, we should try to parse/coerce the value to a a certain type then return it. Thoughts on this!

sanish-bruno (changes requested):

packages/bruno-common/src/utils/datatype.spec.ts
```diff
@@ -198,6 +204,47 @@ describe('validateDataTypeValue', () => {
   });
 });
 
+describe('parseValueByDataType — {{var}} references', () => {
+  const variables = { count: 7, flag: true, payload: { a: 1 }, label: 'hi', nested: { count: 3 } };
+
+  it('resolves a lone reference to the referenced variable value', () => {
```
sanish-bruno: coercion test cases should also be part of the test suite, especially variableReferences, where

```js
obj : {
 "count": 42
}

{{obj.count}} should get parsed as string as well as number
```

Please also add the basic coercion test cases if they are missing

sanish-bruno (changes requested):

packages/bruno-common/src/utils/datatype.spec.ts
```diff
@@ -231,7 +297,7 @@ describe('valueToString — round-trip with parseValueByDataType', () => {
 
   it('returns empty string for functions and symbols', () => {
     expect(valueToString(() => 42)).toBe('');
-    expect(valueToString(function named() {})).toBe('');
+    expect(valueToString(function named() { })).toBe('');
```
sanish-bruno: @sharan-bruno this is intentional right, Linting?

tests/variable-datatypes/reference-types/init-user-data/preferences.json
```diff
@@ -0,0 +1,8 @@
+{
+  "maximized": false,
+  "preferences": {
+    "general": {
+      "defaultWorkspacePath": "{{workspacePath}}"
+    }
+  }
+}
```
sanish-bruno: add a newline

packages/bruno-common/src/utils/datatype.ts
```diff
@@ -5,9 +5,31 @@ export const BRUNO_VARIABLE_DATATYPES: readonly BrunoVariableDataType[] = ['stri
...
+const resolveWholeReference = (value: any, resolvableVariables: Record<string, any>): any => {
+  if (typeof value !== 'string') return undefined;
+  const match = value.trim().match(/^\{\{([^}]+)\}\}$/);
+  if (!match) return undefined;
+  return getByPath(resolvableVariables, match[1].trim());
+};
+
 // string-form → typed JS value, or raw on failure.
-export const parseValueByDataType = (value: any, dataType?: BrunoVariableDataType): any => {
+export const parseValueByDataType = (
```
sanish-bruno, line 24: may need to update the title as there is an additional change including `resolvableVariables`

sanish-bruno (changes requested):

packages/bruno-common/src/utils/datatype.spec.ts
sharan-bruno: formatting issue, i will remove it

tests/variable-datatypes/reference-types/reference-variable-type.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  test('validates a {{reference}} against the referenced variable type', async ({ pageWithUserData: page }) => {
+    await test.step('Open the global environment', async () => {
+      const locators = buildCommonLocators(page);
+      await locators.sidebar.collection('reference-types').click();
+      await selectEnvironment(page, 'global-environment', 'global');
+      await openEnvironmentConfigTab(page, 'global');
+    });
+
+    await test.step('Valid literal values are not flagged', async () => {
+      await expectNotFlagged(page, 'globalEnvString');
```
sanish-bruno: @sharan-bruno In the current state of the test cases, it is very difficult to track which paths are tested, which are not, i think we may need to tweak the test cases do that it is easier to review, maintain. One should be able to understand, what is being tested, what is missed quickly and should be able to decisively add things so that the testing is robust.

Now there is a lot of mental overhead to read each of the variable names, and keeping all those context in mind and work with these files will be difficult to pull through.

Similar approach can also be adopted for unit test cases

Please check https://github.com/usebruno/bruno/pull/8943 , Please use discretion, while making these changes

sanish-bruno (changes requested):

tests/variable-datatypes/reference-types/reference-variable-type.spec.ts
sharan-bruno: @sanish-bruno Agreed. I’ve refactored the test cases using a matrix-based approach, making the coverage and expected behavior easier to understand and maintain

packages/bruno-common/src/utils/datatype.references.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    it('a literal dotted key wins over walking the dotted path', () => {
+      const variables = { 'server.port': 9090, 'server': { port: 8080 } };
+      expect(parseValueByDataType('{{server.port}}', 'number', variables)).toBe(9090);
+    });
+
+    it('surrounding whitespace inside braces is tolerated', () => {
+      expect(parseValueByDataType('  {{ server.port }}  ', 'number', server)).toBe(8080);
+    });
+
+    // Matches ../interpolate's `[^}]+` — such a name resolves at request time, so validation must agree.
```
sanish-bruno: rm

packages/bruno-common/src/utils/datatype.references.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    it('an identifier containing spaces', () => {
+      expect(parseValueByDataType('{{api key}}', 'string', { 'api key': 'secret' })).toBe('secret');
+    });
+  });
+
+  describe('does not resolve — falls through to normal coercion', () => {
+    it('partial reference (user-{{userId}})', () => {
```
sanish-bruno: partial references are treated as string, so can be omitted for this function

packages/bruno-common/src/utils/datatype.references.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    it('the referenced type is authoritative — declared type is ignored', () => {
+      expect(parseValueByDataType('{{port}}', 'string', { port: 8080 })).toBe(8080);
+      expect(parseValueByDataType('{{port}}', undefined, { port: 8080 })).toBe(8080);
+    });
+  });
+});
+
+describe('parseValueByDataType + validateDataTypeValue — mismatch surfaces', () => {
+  const validateRef = (value: string, dataType: BrunoVariableDataType, variables: Record<string, any>) => {
+    const referenced = resolveVariableReference(value, variables);
```
sanish-bruno: one  of  the issues with this approach is that, as multiple functions are involved, it is more of an integration test,  extract `validateRef` (better name if possible) put it in it dataType.ts file, unit test here, just need to import the function from datatype, tests remains the same. the component import and use the `validateRef`

sanish-bruno (changes requested):

sanish-bruno (dismissed):

tests/utils/page/actions.ts
```diff
@@ -981,8 +986,10 @@ const saveEnvironment = async (page: Page) => {
  */
 const closeEnvironmentPanel = async (page: Page, type: EnvironmentType = 'collection') => {
   await test.step('Close environment tab', async () => {
-    const tabLabel = type === 'collection' ? 'Environments' : 'Global Environments';
-    const envTab = page.locator('.request-tab').filter({ hasText: tabLabel });
+    const locators = buildCommonLocators(page);
+    const envTab = type === 'global'
```
vijayh-bruno, line 1042: Tooltip tests in https://github.com/usebruno/bruno/blob/main/tests/variable-tooltip/variable-tooltip.spec.ts leaned on the old loose match. Three tooltip tests now time out and fail because of this change. Fix them

tests/variable-datatypes/reference-types/reference-variable-type.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+const openEnvironment = async (page: Page, scope: EnvironmentScope) => {
+  await buildCommonLocators(page).sidebar.collection('reference-types').click();
+  await selectEnvironment(page, `${scope}-environment`, scope);
+  await openEnvironmentConfigTab(page, scope);
+};
+
+const mismatchIcon = async (page: Page, name: string) => {
+  const locators = buildCommonLocators(page);
+  const row = locators.environment.varRow(name);
+  await scrollVirtuosoRowIntoView(page, row);
```
vijayh-bruno, line 27: This flaked on CI. The row was missing because the table had not rendered yet. Can we wait for it first?

tests/utils/page/actions.ts
sharan-bruno: fixed

tests/variable-datatypes/reference-types/reference-variable-type.spec.ts
sharan-bruno: added wait before checking this

vijayh-bruno: This is still flaky

packages/bruno-app/src/components/RequestPane/Vars/VarsTable/index.js
```diff
@@ -32,6 +33,8 @@ const VarsTable = ({ item, collection, vars, varType, initialScroll = 0, isDraft
   const onSave = () => dispatch(saveRequest(item.uid, collection.uid));
   const handleRun = () => dispatch(sendRequest(item, collection.uid));
 
+  const resolvableVariables = useMemo(() => getAllVariables(collection), [collection]);
```
vijayh-bruno: Should this pass item too? 

getAllVariables(collection, item)

packages/bruno-app/src/components/FolderSettings/Vars/VarsTable/index.js
```diff
@@ -31,6 +32,8 @@ const VarsTable = ({ folder, collection, vars, varType, initialScroll = 0, isDra
 
   const onSave = () => dispatch(saveFolderRoot(collection.uid, folder.uid));
 
+  const resolvableVariables = useMemo(() => getAllVariables(collection), [collection]);
```
vijayh-bruno: Should this pass folder too?

getAllVariables(collection, folder)

cc: @sanish-bruno

packages/bruno-app/src/components/RequestPane/Vars/VarsTable/index.js
sharan-bruno: added

packages/bruno-app/src/components/FolderSettings/Vars/VarsTable/index.js
sharan-bruno: added folder

vijayh-bruno (approved):

PR #8889, opened 2026-08-06
fix (transient-request): unable to create transient request when preset is websocket

chirag-bruno:
### Problem
When `websocket` is the selected preset in collection settings, creating a transient request by `left-click` on the `+` icon does not work.

### Cause
The preset setting used the key `ws` for `websocket` and the `CreateTransientRequest` component was switching on the type with `case 'websocket': `. This would fail as no branch would satisfy the case and the function returns without dispatching any action.

### Fix
Change the key from `websocket` to `ws` in `CreateTransientRequest`

Ticket - BRU-4064

## Summary by CodeRabbit

- **New Features**
  - Creating a transient request now follows the collection’s saved request-type preset.
  - Supported presets include HTTP, GraphQL, gRPC, and WebSocket requests.

- **Bug Fixes**
  - WebSocket transient requests now open with the correct request type and method label.
  - Existing request creation and interaction behavior remains unchanged.

### DIFF
--- a/packages/bruno-app/src/components/CreateTransientRequest/index.js
+++ b/packages/bruno-app/src/components/CreateTransientRequest/index.js
@@ -15,7 +15,7 @@ const REQUEST_TYPE = {
   HTTP: 'http',
   GRAPHQL: 'graphql',
   GRPC: 'grpc',
-  WEBSOCKET: 'websocket'
+  WEBSOCKET: 'ws'
 };
 
 /**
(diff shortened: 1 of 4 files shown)


packages/bruno-app/src/components/CreateTransientRequest/index.js
sachin-thakur-bruno: can we use PRESET_REQUEST_TYPES from `packages/bruno-app/src/utils/common/constants.js` instead of creating our own here?

chirag-bruno: Done

sachin-thakur-bruno (approved):
LGTM

PR #8906, opened 2026-08-07
fix(network): substitute path params in comma-packed url segments

### LINKED ISSUE #8699: path parameters variable substitution not working when separated by a comma
Internal - [BRU-3982]

### This bug is:

- [x] Annoying.

### Bruno version

3.5.3

### 💻 Operating System

macOS

_No response_

### 🐞 Describe the bug

Collection imported from OpenAPI spec with path parameters separated by comma do not perform variable substitution.
 

### 📝 Steps to reproduce

1. Import OpenAPI spec from https://api.weather.gov/openapi.json
2. Send the "GET point" request

### 📁 Collection to reproduce

https://api.weather.gov/openapi.json

[openapi.json](https://github.com/user-attachments/files/30227647/openapi.json)

Image

YuriNachos:
Ref: BRU-3982

Closes #8699.

## Problem

An OpenAPI path that packs multiple params into one segment, like `/points/{lat},{lon}`, is imported as `/points/:lat,:lon` with two correct path-param rows. At send time, the leading-colon branch in `interpolate-vars.js` treated the whole post-colon string as one param name (`path.slice(1)` → `lat,:lon`), so the lookup matched nothing and the segment was sent literally (`/points/:lat,:lon`) instead of `/points/39.7,-104.9`. Reported in #8699 (BRU-3982).

## Fix

Tokenize the segment with `/[:]([a-zA-Z_]\w*)/g` — the same tokenizer the OData-style branch already uses — and substitute every resolvable `:name` token, leaving unresolvable tokens literal. A single `:id` still resolves; comma-packed `:lat,:lon` now resolves both; an unsolvable token is left as-is.

## Tests

New `packages/bruno-electron/src/ipc/network/interpolate-vars.spec.js`: comma-packed substitution, single-param regression, unsolvable-token-left-literal. `npm run test:ci --workspace=packages/bruno-electron` → 52 suites, 736 passed, 1 skipped — including the 39 pre-existing path-param cases.

A known minor behavioral alignment: a single leading-colon segment whose name contains `-` or `.` (e.g. `:user-id`) no longer resolves as one token — but this already did not resolve in the OData branch, no existing test exercises a resolvable hyphenated single-segment name, and the fix deliberately reuses the OData tokenizer for consistency.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed URL path interpolation when multiple parameters appear within a single, comma-separated path segment.
  * Preserved unresolved parameter placeholders instead of removing them.
  * Maintained existing behavior for standard single-parameter paths.

* **Tests**
  * Added regression coverage for multiple parameters, single-parameter substitution, and unresolved tokens.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/interpolate-vars.js
+++ b/packages/bruno-electron/src/ipc/network/interpolate-vars.js
@@ -196,12 +196,16 @@ const interpolateVars = (request, envVariables = {}, runtimeVariables = {}, proc
       .map((path) => {
         // traditional path parameters
         if (path.startsWith(':')) {
-          const paramName = path.slice(1);
-          const existingPathParam = request.pathParams.find((param) => param.name === paramName);
-          if (!hasResolvablePathParamValue(existingPathParam)) {
-            return '/' + path;
+          const paramRegex = /[:]([a-zA-Z_]\w*)/g;
+          let match;
+          let result = path;
+          while ((match = paramRegex.exec(path))) {
+            const existingPathParam = request.pathParams.find((param) => param.name === match[1]);
+            if (hasResolvablePathParamValue(existingPathParam)) {
+              result = result.replace(':' + match[1], existingPathParam.value);
+            }
           }
-          return '/' + existingPathParam.value;
+          return '/' + result;
         }
 
         // for OData-style parameters (parameters inside parentheses)
--- a/packages/bruno-electron/src/ipc/network/interpolate-vars.spec.js
+++ b/packages/bruno-electron/src/ipc/network/interpolate-vars.spec.js
@@ -1,0 +1,42 @@
+const interpolateVars = require('./interpolate-vars');
+
+describe('interpolateVars - path params', () => {
+  it('substitutes multiple path params packed in one comma-separated segment', () => {
+    const request = {
+      url: 'https://api.test/points/:lat,:lon',
+      mode: 'get',
+      headers: {},
+      data: { mode: 'none' },
+      pathParams: [
+        { name: 'lat', value: '39.7', type: 'path', enabled: true },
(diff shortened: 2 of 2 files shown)


YuriNachos:
Force-pushed to drop an internal report file my tooling had committed into the branch by mistake — the diff is now only `interpolate-vars.js` and its spec. No production or test code changed.

Note on the red check: `security/snyk (Bruno)` reports `You have used your limit of private tests`, which is an account-quota error on the Snyk side rather than a finding against this PR — every currently open PR in this repo shows the same status.

YuriNachos:
Note on the red check, in case it makes this PR look unready — **`security/snyk (Bruno)` is not reporting a vulnerability in this diff.** Its own status text is:

> You have used your limit of private tests

That is the Snyk org account hitting its test quota, so the scan never ran. It is red on every open fork PR in this repo right now, not just mine — sampled just now: #8917, #8915, #8913, #8904, all with the same status.

`CodeRabbit`, the only other check here, is green. Nothing in this branch can turn Snyk green; that one needs someone with access to the Bruno Snyk org. Happy to rebase or re-trigger if it helps.

YuriNachos:
This PR has stayed intentionally narrow — one tokenizer change in `interpolate-vars.js` plus its spec, so `/points/:lat,:lon` resolves both params instead of being sent as a literal segment. Is there anything else you'd like from me before it can get a review pass — more interpolation edge cases, a rebase, or anything open in the CodeRabbit thread? The red Snyk job is the org-quota failure I noted in my comment above rather than a finding against this diff. I also have #8909, #8911 and #8941 queued in the same area, so if it helps triage I'm glad to prioritise whichever of those is most useful to the team first.

PR #8909, opened 2026-08-08
fix(editor): mask unquoted template fragments so the JSON editor stops flagging valid bodies

### LINKED ISSUE #8859: JSON editor reports "Unexpected token {" for an unquoted variable containing JSON, while the request is sent successfully
Internal - [BRU-4145]

### This bug is:

- [x] Annoying.

### Bruno version

4.0.0

### 💻 Operating System

Windows

Windows 11

### Description

Bruno's JSON body editor reports a parse error when an unquoted variable is used to insert a valid JSON fragment into an array.

The request itself is rendered and sent successfully. The error appears only in the JSON editor/UI validation.

### Expected behavior

The JSON editor should recognize an unquoted `{{variable}}` expression as a template placeholder and should not report a JSON syntax error when the interpolated request body is valid.

Alternatively, Bruno could validate the body after variable interpolation or mark the diagnostic as a template-related warning rather than a JSON parse error.

### Additional information

Quoting the variable is not a valid workaround:

```json
"{{DEDUPLICATION}}"
```

This changes the value into a JSON string instead of inserting the JSON fragment.

Changing the body type to plain text/raw or constructing the body in a pre-request script is also not suitable in this case. The request must remain a JSON body and the fragment must be managed through a variable.

The issue seems to be limited to editor validation because the actual request is sent successfully.

### Environment

- Bruno version: `4.0.0`
- Operating system: `Windows 11`
- Installation type: `Desktop]`
- Request body type: `JSON`

### Screenshot

Image

### Steps to reproduce

1. Create a variable containing a JSON fragment with multiple array elements:

```bru
vars {
  DEDUPLICATION: '''
    {
      "condition": {
        "not_condition": {
          "CURRENT_RUN_RETRY": 0
        }
      },
      "action": "External_DB",
      "database": "itsm",
      "query": "SELECT r.id AS guid_double_check FROM reqs r",
      "params": ["CURRENT_RUN_GUID"],
      "first": ["guid_double_check"]
    },
    {
      "condition": {
        "guid_double_check": true
      },
      "action": "response",
      "status": 409,
      "text": "Already exists",
      "end": true
    },
  '''
}
```

2. Use this variable as an unquoted JSON fragment in a JSON request body:

```json
[
  {
    "abc": 1
  },
  {
    "bbc": 2
  },
  {{DEDUPLICATION}}
  {
    "ccc": 3
  }
]
```

3. Open the request body in Bruno's JSON editor.

### Actual behavior

The JSON editor reports an error on `D` in the variable expression:

```text
Parse error on line ...
Unexpected token "D"
```

The editor appears to parse `{{DEDUPLICATION}}` as regular JSON before variable interpolation.

However, when the request is sent, the variable is interpolated correctly and the resulting request body is valid JSON.

### 📁 Collection to reproduce

_No response_

Image

YuriNachos:
### Description

The JSON body editor's lint pass reports a false **"Unexpected token"** parse error for a request body that Bruno itself sends successfully: an **unquoted** `{{variable}}` used to splice a JSON fragment into an array/object (e.g. several array elements passed in one variable).

### Problem

When a `{{variable}}` placeholder sits as a **standalone array element** (between `,` / `[` / `]`), the editor's JSON linter flags the body as invalid even though the interpolated request is perfectly valid JSON. The variable typically carries its own separators (its value ends in `,`), so the *sent* request is correct — only the editor's lint diagnostic is wrong.

Closes #8859.

The lint helper (`CodeEditor/index.js`) masked every unquoted placeholder with a literal `1` via an inline regex. That regex had two compounding flaws for this case:

1. Its negative lookahead `(?![^"},]*")` did **not** exclude `{`, so for a placeholder followed by `\n  {` (the next object) it ate the `{`, reached the next object's `"`, and wrongly concluded the placeholder was inside a quoted string — leaving it **un-masked**, so `jsonlint` choked on `{{variable}}` directly.
2. Even after fixing (1), masking a standalone element with `1` yields `1 {` with **no comma** between it and the next element — still a parse error. A single static per-placeholder mask cannot satisfy both the value-position case (`"k": {{OBJ}}`, which needs `1`) and the element-position case (which must vanish).

### Fix

Extract the masking into a small position-aware pure helper, `maskJsonTemplateVariables(text)`, and tighten the in-string guard:

- **Value position** (nearest non-whitespace char before the placeholder is `:`, or the placeholder is the whole body) → replaced with `1` (unchanged behaviour, correct for `"k": {{OBJ}}`).
- **Element position** (standalone between `[`/`{`/`,`/`]`/`}`) → marked for removal, then the commas its removal dislodges are cleaned up **at that spot only** (collapse `,,`, drop a leading `,` after `[`/`{`, drop a trailing `,` before `]`/`}`), looped until stable. Scoping the clean-up to the removal site means a genuine comma mistake elsewhere in the body is still reported.
- **Inside a quoted JSON string** → left untouched (its value may legitimately contain quotes/braces); the lookbehind/lookahead guard is preserved, with the lookahead tightened to also stop at `{` so a standalone fragment followed by the next object is no longer mistaken for a quoted value.

The helper is pure (no `window`/CodeMirror/React) and unit-tested directly against `@prantlf/jsonlint`. The call site keeps `stripJsonComments` exactly where it was: `jsonlint.parse(stripJsonComments(maskJsonTemplateVariables(text)))`. Real syntax errors (e.g. a missing value, or a stray comma away from a placeholder) are still reported — the masker masks template variables, it does not repair JSON.

### Screenshots

No UI change — editor lint diagnostic only. A body such as

```json
[
  { "abc": 1 },
  {{DEDUPLICATION}}
  { "ccc": 3 }
]
```

no longer shows a false "Unexpected token" error in the editor (it was already being sent correctly).

| Before | After |
| --- | --- |
| False "Unexpected token" lint error on a valid unquoted `{{fragment}}` | Lints clean |

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON linting for requests containing template variables.
  * Fixed validation of template variables used as values, array elements, and object fragments.
  * Preserved template variables inside quoted strings while maintaining accurate syntax-error reporting.

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -14,6 +14,7 @@ import { setupAutoComplete, showRootHints } from 'utils/codemirror/autocomplete'
 import { setupAiAutocomplete } from 'utils/codemirror/aiGhostText';
 import { buildAutocompleteContext } from 'utils/ai';
 import StyledWrapper from './StyledWrapper';
+import { maskJsonTemplateVariables } from './mask-json-variables';
 import * as jsonlint from '@prantlf/jsonlint';
 import { JSHINT } from 'jshint';
 import stripJsonComments from 'strip-json-comments';
@@ -179,7 +180,7 @@ class CodeEditor extends React.Component {
       }
       const jsonlint = window.jsonlint.parser || window.jsonlint;
       try {
-        jsonlint.parse(stripJsonComments(text.replace(/(?<!"[^":{]*){{[^}]*}}(?![^"},]*")/g, '1')));
+        jsonlint.parse(stripJsonComments(maskJsonTemplateVariables(text)));
       } catch (error) {
         const { message, location } = error;
         const line = location?.start?.line;
--- a/packages/bruno-app/src/components/CodeEditor/mask-json-variables.js
+++ b/packages/bruno-app/src/components/CodeEditor/mask-json-variables.js
@@ -1,0 +1,64 @@
+// Masks Bruno template variables ({{name}}) in a JSON body so the editor's lint
+// pass can syntax-check it *before* the request send path interpolates the real
+// values. A variable may splice either a single JSON value or a whole fragment
+// (several array elements / object members), so the mask must be position-aware:
+//
+//   - Value position  (nearest non-whitespace char before it is `:`, or the
+//     placeholder is the whole body) — the variable stands for one value, so it
+//     is replaced by `1`.
+//   - Element position (anything else: between `[`, `{`, `,`, `]`, `}`) — the
+//     variable may splice several elements, so it is marked for removal and the
+//     commas its removal dislodges are cleaned up at that spot only.
+//   - Inside a quoted JSON string — left untouched (its value may legitimately
+//     contain quotes/braces). The lookbehind/lookahead guard skips a placeholder
+//     that sits inside a quoted string; the lookahead also stops at `{` so a
+//     standalone fragment followed by the next object is not mistaken for a
+//     quoted value.
+//
+// Pure: no window, CodeMirror, or React — directly unit-testable with jsonlint.
+
+const UNQUOTED_PLACEHOLDER = /(?<!"[^":{]*){{[^}]*}}(?![^"},{]*")/g;
+const WHITESPACE = /\s/;
+
+// Marks where an element-position placeholder was removed, so the comma clean-up
+// below stays scoped to the removal site and never rewrites real commas the user
+// typed elsewhere or inside a string literal. A private-use char never occurs in
+// real JSON and is never produced by the masking regex above.
+const REMOVED_ELEMENT = '\uE000';
+
+function nearestNonWhitespaceBefore(text, end) {
+  for (let j = end - 1; j >= 0; j -= 1) {
+    if (!WHITESPACE.test(text[j])) {
+      return text[j];
+    }
+  }
+  return '';
+}
+
+export function maskJsonTemplateVariables(text) {
+  const masked = text.replace(UNQUOTED_PLACEHOLDER, (_match, offset) => {
+    const previous = nearestNonWhitespaceBefore(text, offset);
+    return previous === ':' || previous === '' ? '1' : REMOVED_ELEMENT;
+  });
+
+  // Nothing to clean up unless at least one element-position placeholder was removed.
+  if (!masked.includes(REMOVED_ELEMENT)) {
+    return masked;
+  }
+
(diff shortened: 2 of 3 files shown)


YuriNachos:
Thanks — I checked the CodeRabbit finding rather than taking it at face value, and I want to report what I actually found, because the mechanism is right but the impact is not Major.

**The mechanism is real.** In `{"msg":"prefix: {{name}}, suffix"}` the lookbehind `(?<!"[^":{]*)` cannot span the `:` inside the string, so the placeholder is *not* recognised as quoted and does get masked. Confirmed by running it.

**The impact is not.** The masked text never reaches the document — it exists only as the argument to `jsonlint.parse` inside the `lint` helper (`index.js:183`). So an over-eager in-string mask cannot corrupt a body; the only thing it can change is the lint verdict.

I enumerated the in-string cases to see whether any verdict actually flips. Oracle: *if the body is valid JSON once the placeholder is interpolated with a real value, the masked body must also be valid* — i.e. no false lint error. 64 combinations of leading/trailing in-string context (`a: `, `, `, ` [`, `{`, `a", "b`, `: , `, empty) around an in-string placeholder:

```
false-lint-error cases: 0
```

I also could not construct the reverse — a body that is genuinely invalid but lints clean because of the masking. Removal only ever deletes the marker and the commas immediately adjacent to it, all of which sit *inside* the quotes, so the string's delimiters survive and the document's structure is unchanged.

**The one residual effect, and it is Minor:** masking inside a string changes that line's length, so a genuine syntax error later on the same line is reported at a shifted column. That is inherent to the mask-then-lint design and predates this PR for value-position placeholders; this change widens it slightly to in-string ones.

I've left the regex as-is because a lexical string tracker is a materially larger change than this bug fix warrants and would be the third rewrite of this predicate in one PR. If you'd prefer the column-accurate version — pad the replacement to the original placeholder's width instead of collapsing it — I'm happy to add it here or in a follow-up; your call.

(The red `security/snyk` check is the org's `You have used your limit of private tests` quota error. It is currently red on every open PR in the repo, including the Bruno team's own — not something this branch can affect.)

YuriNachos:
Note on the red check, in case it makes this PR look unready — **`security/snyk (Bruno)` is not reporting a vulnerability in this diff.** Its own status text is:

> You have used your limit of private tests

That is the Snyk org account hitting its test quota, so the scan never ran. It is red on every open fork PR in this repo right now, not just mine — sampled just now: #8917, #8915, #8913, #8904, all with the same status.

`CodeRabbit`, the only other check here, is green. Nothing in this branch can turn Snyk green; that one needs someone with access to the Bruno Snyk org. Happy to rebase or re-trigger if it helps.

PR #8910, opened 2026-08-08
fix(converters): await pm.sendRequest promise chains as a post-pass

abhishekp-bruno:
BRU-3117

### Description

Migrating a postman collection which has a script containing pm.sendRequest we replace it with await bru.sendRequest while also handling responsePropertyMap accordingly ( eg: res.json() -> res.data )

To do all this we use jscodeshift library to create AST(Abstract Syntax tree) from the Postman script and then modify it to work in Bruno script environment.

### Problem

When we migrate a Postman script which looks like this pm.sendRequest(config) we check if there is an await that already exists, else we add an await to the script and replace pm.sendRequest -> bru.sendRequest for the migration.
But in our current implementation we are not checking for any chain implementation, if the script has .then .catch .finally handlers which ends up breaking the pm to bru script conversation.

The migration goes like this in the following steps

pm-script --> AST (Abstract syntax tree - using jscodeshift library)
AST --> modified AST (we modify the AST nodes to convert them to work in bruno)
the problem happens in this step modified AST --> converted back to source code during this step since now that we have added await when it is converted back to source code form, it checks precedence. Since (.) binds tighter than await it gets converted to

```
(await bru.sendRequest(config)).then((res) => {
  console.log(res.json());
});
```

### Fix

1. Skip statements that already have .then / .catch / .finally handlers
A chained call manages its own asynchrony — awaiting it would double-handle the response and change execution order.
2. Await the outermost expression, not the inner call
When we do await, await is added to the whole expression statement rather than being nested inside it, so the entire chain resolves before the script continues.
```
// before (broken)
(await sendRequest(options)).then(...);

// after
await sendRequest(options);
```

3. Map the response shape in the .then handler
Postman's pm.sendRequest callback and Bruno's resolved response expose different shapes like this 

Postman | Bruno
-- | --
res.json() / res.text() | res.data
res.code | res.status
res.status | res.statusText

This is handled in the first .then handler where response is consumed 

### Screenshots

| Before | After |
| --- | --- |
| image | image |

#### Contribution Checklist:

## Summary by CodeRabbit

* **Improvements**
  * Improved conversion of Postman `pm.sendRequest` promise chains.
  * Supports `.then`, `.catch`, and `.finally`, including nested and multiple chains.
  * Translates response properties, callbacks, request configuration variables, and bracket notation.
  * Preserves behavior in asynchronous contexts, including shadowed and destructured parameters.

* **Bug Fixes**
  * Prevents incorrect `await` insertion for chained requests.
  * Safely handles unsupported or ambiguous callback patterns.


prateek-bruno (commented):
Currently this, where code is a variable will gets transformed to res.status which is incorrect:
```js
pm.sendRequest({ url: 'https://x.com' }).then((res) => {
  console.log(res[code]);
});
```
```js
// Current output
await bru.sendRequest({ url: 'https://x.com' }).then((res) => {
  console.log(res.status);   // user's dynamic lookup silently replaced
});
```

prateek-bruno (commented):
`res['json']()` = `res.json()`, but currently this isn't handled:
```js
// Postman input
pm.sendRequest({ url: 'https://x.com' }).then((res) => {
  console.log(res['json']());
});
```
```js
// Current output
await bru.sendRequest({ url: 'https://x.com' }).then((res) => {
  console.log(res['json']());   // TypeError at runtime on the Bruno response
});
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -320,7 +327,7 @@ const sendRequestTransformer = (path, j) => {
     [requestOptions]
   );
 
-  return wasAwaited ? sendRequestCall : j.awaitExpression(sendRequestCall);
+  return wasAwaited || isChained ? sendRequestCall : j.awaitExpression(sendRequestCall);
```
prateek-bruno, line 330: We've introduced `isInAsyncContext` to ensure await isn't added to bru.sendRequest in sync context, but in main pass if it's not chained it will not do that, can we have some consistency since we're doing it in the PR:
```js
// Postman input
pm.sendRequest({ url: 'https://x.com' }).then((res) => {
  pm.sendRequest({ url: 'https://y.com' });
});
```
```js
// Current output
await bru.sendRequest({ url: 'https://x.com' }).then((res) => {
  await bru.sendRequest({ url: 'https://y.com' });   // SyntaxError: the arrow is not async
});
```

packages/bruno-converters/src/utils/send-request-chain-transformer.js
```diff
@@ -0,0 +1,179 @@
...
+ * access with a non-literal key (`p[someVar]`) has no static name — even an
+ * Identifier key named `then` is a variable there, not the method.
+ * @param {Object} memberExpr - MemberExpression node
+ * @returns {string|null}
+ */
+const getStaticPropertyName = (memberExpr) => {
+  const property = memberExpr.property;
+
+  if (memberExpr.computed) {
+  // .['then] => literal
```
prateek-bruno, line 40: The correct format is ['then'] and not .['then], it should be updated to this:

```suggestion
  // ['then'] => literal
```

prateek-bruno (commented):
Currently we only check the first chain link and if it's "then" we do the conversion. This misses many cases:
1) First link is not then (a `catch`/`finally` before it, on success the res just passes through them anyway)
2) First then doesn't consume res and returns it back, so the next then also gets the res

How about:
- We walk the chain, skipping `catch`/`finally` links
- On the first `then` with a param, do the conversion
- Check its return value, if it doesn't return back the res (or the res got reassigned somewhere in between) then break
- If it returns back the res then continue walking to the next `then`, its param is the new res

It might still miss some cases but most of them related to chained walk will get covered, and anything ambiguous just breaks the walk, so worst case we under-translate, never rewrite the wrong thing.

PR #8922, opened 2026-08-10
feat(CI-OAuth2): add Keycloak IdP setup and OAuth2 state validation e2e tests

abhishekp-bruno:
BRU-3817

### Description

Adds Keycloak to the CI pipeline so the OAuth 2.0 flows are tested against a real identity
provider on all three platforms (Linux, macOS, Windows)

### Problem

Until now the only way to test OAuth 2.0 was to stand up a mock server that imitated how an
IdP is *expected* to respond. That gave us coverage of our own code paths but not of the
protocol: the mock reflected our assumptions, so any misreading of the spec was baked into
both the implementation and the test. 

 It could not tell us whether Bruno is actually compliant  as per  RFC or PKCE

### Fix

We add Keycloak to the CI pipeline with a pre-configured realm setup so that the OAuth 2.0 test cases can run against a consistent authentication server.

GitHub Actions supports Docker-based setups on Linux, but Docker cannot be configured in the same way on macOS and Windows runners due to platform limitations. Since our test suite runs across Linux, macOS, and Windows, relying on Docker would make the Keycloak setup inconsistent across operating systems.

To address this, we now:

- Download the Keycloak tarball distribution directly in the CI pipeline and prepare it with the fixture realm.
- Start Keycloak in the same step that runs the tests, wait until the realm is ready, and shut it down when the tests finish.
- Use three shared composite actions (`setup-keycloak`, `run-oauth2-e2e-tests`, `collect-keycloak-logs`) written in bash, so Linux, macOS, and Windows run the same scripts with no per-OS copies.
- Capture Keycloak output to a log and upload it as an artifact even when the job fails, so startup or configuration issues can be debugged without rerunning CI.

Also in this PR

OAuth2.0 test cases 

#### Contribution Checklist:

## Summary by CodeRabbit

- **Authentication**
  - Expanded OAuth 2.0 coverage for authorization-code and implicit flows.
  - Supports validation of matching and mismatched callback state values, including user-supplied and generated states.

- **Quality Improvements**
  - Added cross-platform end-to-end coverage for OAuth 2.0 authentication.
  - Improved diagnostic reporting for authentication test failures across Linux, macOS, and Windows.

### DIFF
--- a/packages/bruno-electron/src/utils/oauth2.js
+++ b/packages/bruno-electron/src/utils/oauth2.js
@@ -713,8 +713,8 @@ const generateCodeVerifier = () => {
 };
 
 // Build an OAuth2 state string to help prevent CSRF and forged auth codes.
-// If the user passes a state, it goes first; we append random bytes after it.
-// The user keeps their custom data, and the random suffix keeps the flow secure.
+// If the user supplies a non-empty state, it is used as-is; otherwise a
+// cryptographically random value is generated.
 const generateState = ({ userState }) => {
   const trimmedUserState = userState?.trim();
   if (trimmedUserState && trimmedUserState.length > 0) {
--- a/packages/bruno-electron/src/utils/oauth2.js
+++ b/packages/bruno-electron/src/utils/oauth2.js
@@ -713,8 +713,8 @@ const generateCodeVerifier = () => {
 };
 
 // Build an OAuth2 state string to help prevent CSRF and forged auth codes.
-// If the user supplies a non-empty state, it is used as-is; otherwise a
-// cryptographically random value is generated.
+// A user-supplied state is trimmed and used if anything remains; a missing or
+// whitespace-only state falls back to a cryptographically random value.
 const generateState = ({ userState }) => {
   const trimmedUserState = userState?.trim();
   if (trimmedUserState && trimmedUserState.length > 0) {
--- a/tests/auth/oauth2/oauth2-state-validation.spec.ts
+++ b/tests/auth/oauth2/oauth2-state-validation.spec.ts
@@ -1,0 +1,351 @@
+// use this command to start the keycloak server: and make sure that you update the path of the realm-export.json file
+// docker run -d --name bruno-keycloak \
+//   -p 8090:8080 \
+//   -e KC_BOOTSTRAP_ADMIN_USERNAME=admin \
+//   -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
+//   -v <paste path of the file here>:/opt/keycloak/data/import/realm-export.json:ro \
+//   quay.io/keycloak/keycloak:26.0 \
+//   start-dev --http-port=8080 --import-realm
+
+import type { ElectronApplication } from 'playwright';
+import { test, expect, waitForReadyPage } from '../../../playwright';
+import { openRequest, selectRequestPaneTab, selectEnvironment } from '../../utils/page';
+import { buildCommonLocators } from '../../utils/page/locators';
+
+const COLLECTION = 'oauth2-state';
+const KEYCLOAK = 'http://127.0.0.1:8090';
+const KEYCLOAK_REALM = 'bruno';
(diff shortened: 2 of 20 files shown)


.github/actions/auth/oauth2/posix/collect-keycloak-logs/action.yml
```diff
@@ -0,0 +1,26 @@
...
+  using: 'composite'
+  steps:
+    # Keycloak logs live in /tmp (outside the workspace), so stage them into the workspace
+    # where upload-artifact can pick them up.
+    - name: Stage Keycloak logs
+      shell: bash
+      run: |
+        set -uo pipefail
+        mkdir -p keycloak-logs
+        cp /tmp/keycloak.log keycloak-logs/ 2>/dev/null || echo "No /tmp/keycloak.log to collect."
```
prateek-bruno, line 19: Wanted to know why are we only doing keycloak.log here but not keycloak.err.log like windows action do.

.github/actions/auth/oauth2/posix/collect-keycloak-logs/action.yml
prateek-bruno, line 1: ~Can we check if we can extract them to a common platform independent action, the two of them already seem to be possible this one and setup-keycloak, check for run-oauth2-e2e-tests also once.~

I think we can later see the optimization part, since anyways this also works and making them common isn't a blocker.

tests/auth/oauth2/oauth2-state-validation.spec.ts
```diff
@@ -0,0 +1,396 @@
...
+ *   - authorization code: callback URI + error/success toast + Keycloak token endpoint
+ *   - implicit grant: callback URI + error/success toast (no token exchange POST)
+ *
+ * Requires a running Keycloak (`bin/kc.sh start-dev --http-port=8090`) with the realm,
+ * client, and user configured (see the KEYCLOAK_* constants below).
+ *
+ * Fixtures (auto-loaded by Playwright):
+ *   - fixtures/collection  → oauth2-state collection (Authorization Code + Implicit requests)
+ *   - init-user-data         → preloads collection, enables system-browser OAuth2 mode
+ */
```
prateek-bruno, line 38: do we need these comments ?

abhishekp-bruno: updated

.github/actions/auth/oauth2/posix/collect-keycloak-logs/action.yml
abhishekp-bruno: updated to use a common file for all three platform

.github/actions/auth/oauth2/posix/collect-keycloak-logs/action.yml
abhishekp-bruno: That difference was an artifact of the two implementations rather than a decision. On POSIX the server was started with > keycloak.log 2>&1, so stderr already ended up in the same file. On Windows, PowerShell's Start-Process only supports redirecting stdout and stderr to separate files, hence the extra keycloak.err.log.

 With the actions now merged into a single bash script for all three OSes, Keycloak is launched the same way everywhere and both streams go into one keycloak.log. The collect action globs keycloak*.log from RUNNER_TEMP, so the asymmetry is gone and there is no separate error log to miss.

packages/bruno-electron/src/utils/oauth2.js
```diff
@@ -713,8 +713,8 @@ const generateCodeVerifier = () => {
 };
 
 // Build an OAuth2 state string to help prevent CSRF and forged auth codes.
-// If the user passes a state, it goes first; we append random bytes after it.
-// The user keeps their custom data, and the random suffix keeps the flow secure.
+// If the user supplies a non-empty state, it is used as-is; otherwise a
+// cryptographically random value is generated.
```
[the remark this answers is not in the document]
abhishekp-bruno: this comment has been updated

PR #8944, opened 2026-08-12
perf(runner): store runner request and response payloads in sqlite

chirag-bruno:
The collection runner puts every request and response into the redux store, so a big run holds all
of it in memory and reserialises it on every runner event.

Those payloads now go into sqlite instead, keyed on collection uid + iteration index + request uid.
The runner list only receives the status and status text it actually renders, and the response pane
reads the full payload back when you expand a row. Rows are cleared per collection on Reset and at
the start of every run, so the table does not grow across runs.

Ticket - BRU-1071

## Summary by CodeRabbit

* **New Features**
  * Runner results now retain complete request and response details for completed or failed items.
  * Response panels display stored results when available, with reliable fallback behavior.
  * Results remain available throughout collection-run workflows for more consistent viewing.
  * Test results now clearly distinguish passed and failed requests.

* **Bug Fixes**
  * Prevented outdated runner results from appearing after rerunning or resetting collections.
  * Collection runs now clear previous results before starting, ensuring displayed data matches the latest execution.
  * Improved handling of failed requests so their responses are displayed consistently.

### DIFF
--- a/packages/bruno-app/src/hooks/useClearStoredRunnerExchanges/index.js
+++ b/packages/bruno-app/src/hooks/useClearStoredRunnerExchanges/index.js
@@ -1,0 +1,16 @@
+import { useCallback } from 'react';
+import { useSqliteMutation } from '@usebruno/sqlite/web';
+
+const useClearStoredRunnerExchanges = (collectionUid) => {
+  const { mutateAsync } = useSqliteMutation('delete_runner_responses_for_collection');
+
+  return useCallback(async () => {
+    try {
+      await mutateAsync({ collection_uid: collectionUid });
+    } catch (error) {
+      console.error('Failed to clear stored runner payloads', error);
+    }
+  }, [mutateAsync, collectionUid]);
+};
+
+export default useClearStoredRunnerExchanges;
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -469,14 +469,52 @@ const registerNetworkIpc = (mainWindow) => {
     });
   };
 
-  const sendRunnerResponseReceived = ({ requestUid, responseReceived, error, eventData }) => {
+  const runnerRow = ({ requestUid, eventData }) => ({
+    request_uid: requestUid,
+    collection_uid: eventData.collectionUid,
+    iteration_index: eventData.iterationIndex,
+    item_uid: eventData.itemUid
+  });
+
+  const sendRunnerRequestSent = ({ requestUid, requestSent, eventData }) => {
+    const request = safeStringifyJSON(requestSent);
+
     // TODO (chirag): DB storage comes here
-    console.log('[runner:response-received]', {
+    console.log('[runner:request -> db]', {
+      ...runnerRow({ requestUid, eventData }),
+      bytes: Buffer.byteLength(request || ''),
+      request
+    });
+
+    // TODO (chirag): once requests are read from the DB, this is all the renderer gets
+    console.log('[runner:request -> renderer]', {
+      ...eventData,
+      requestUid
+    });
+
+    mainWindow.webContents.send('main:run-folder-event', {
+      type: 'request-sent',
+      requestSent,
+      ...eventData
+    });
+  };
+
+  const sendRunnerResponseReceived = ({ requestUid, responseReceived, error, eventData }) => {
+    const response = safeStringifyJSON(responseReceived);
+
+    // TODO (chirag): DB storage comes here
+    console.log('[runner:response -> db]', {
+      ...runnerRow({ requestUid, eventData }),
+      bytes: Buffer.byteLength(response || ''),
+      response
+    });
+
+    // TODO (chirag): once responses are read from the DB, this is all the renderer gets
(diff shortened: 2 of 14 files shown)


packages/bruno-app/src/hooks/useClearStoredRunnerExchanges/index.js
```diff
@@ -0,0 +1,16 @@
...
+const useClearStoredRunnerExchanges = (collectionUid) => {
+  const { mutateAsync } = useSqliteMutation('delete_runner_responses_for_collection');
+
+  return useCallback(async () => {
+    try {
+      await mutateAsync({ collection_uid: collectionUid });
+    } catch (error) {
+      console.error('Failed to clear stored runner payloads', error);
+    }
+  }, [mutateAsync, collectionUid]);
```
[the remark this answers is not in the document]
chirag-bruno: @coderabbitai, this seems fine. The failure case of this operation is harmless. Isn't it?

packages/bruno-sqlite/migrations/0000001_runner_responses.ts
```diff
@@ -0,0 +1,21 @@
...
+    );
+    CREATE INDEX idx_runner_responses_scope ON runner_responses (collection_uid, iteration_index);
+  `;
+};
+export const down = (): string => {
+  return `
+    DROP INDEX IF EXISTS idx_runner_responses_scope;
+    DROP TABLE IF EXISTS runner_responses;
+  `;
+};
```
sid-bruno, line 15: Drops unread columns, timestamps, and the unused iteration-scope index.

```suggestion
export const up = (): string => {
  return `
    CREATE TABLE runner_responses (
      request_uid TEXT PRIMARY KEY,
      collection_uid TEXT NOT NULL,
      request TEXT,
      response TEXT
    );
  `;
};
export const down = (): string => {
  return `
    DROP TABLE IF EXISTS runner_responses;
  `;
};
```

packages/bruno-sqlite/statements/runner_responses.sql
```diff
@@ -0,0 +1,16 @@
...
+  COALESCE(@response, (SELECT response FROM runner_responses WHERE request_uid = @request_uid)),
+  COALESCE((SELECT created_at FROM runner_responses WHERE request_uid = @request_uid), CURRENT_TIMESTAMP),
+  CURRENT_TIMESTAMP
+);
+
+-- name: get_runner_response :one
+SELECT request, response FROM runner_responses WHERE request_uid = @request_uid;
+
+-- name: delete_runner_responses_for_collection :exec
+DELETE FROM runner_responses WHERE collection_uid = @collection_uid;
```
sid-bruno, line 14: Matches the slimmer table: only fields that get and delete actually use.

```suggestion
-- name: upsert_runner_response :exec
INSERT OR REPLACE INTO runner_responses (
  request_uid, collection_uid, request, response
) VALUES (
  @request_uid, @collection_uid,
  COALESCE(@request, (SELECT request FROM runner_responses WHERE request_uid = @request_uid)),
  COALESCE(@response, (SELECT response FROM runner_responses WHERE request_uid = @request_uid))
);

-- name: get_runner_response :one
SELECT request, response FROM runner_responses WHERE request_uid = @request_uid;

-- name: delete_runner_responses_for_collection :exec
DELETE FROM runner_responses WHERE collection_uid = @collection_uid;
```

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -469,6 +470,53 @@ const registerNetworkIpc = (mainWindow) => {
...
+    try {
+      statements.execute('upsert_runner_response', {
+        ...runnerRow({ requestUid, eventData }),
+        request,
+        response
+      });
+    } catch (error) {
+      console.error('[runner] failed to store exchange', requestUid, error);
+    }
+  };
```
sid-bruno: Inlines the row helper and stops writing unused iteration/item fields.

```suggestion
  const storeRunnerExchange = ({ requestUid, eventData, request = null, response = null }) => {
    const statements = getStatements();
    if (!statements) return;

    try {
      statements.execute('upsert_runner_response', {
        request_uid: requestUid,
        collection_uid: eventData.collectionUid,
        request,
        response
      });
    } catch (error) {
      console.error('[runner] failed to store exchange', requestUid, error);
    }
  };
```

packages/bruno-sqlite/migrations/0000001_runner_responses.ts
chirag-bruno: Applied. Also dropped the `created_at`/`updated_at` commit that sat on top of this, since nothing reads those either.

packages/bruno-sqlite/statements/runner_responses.sql
chirag-bruno: Applied.

packages/bruno-electron/src/ipc/network/index.js
chirag-bruno: Applied.

sid-bruno (dismissed):

sid-bruno (approved):

PR #8946, opened 2026-08-12
fix(openAPI):multifile openAPI specs can be imported

adwait-bruno:
### Description
REF:BRU-3806

 OpenApi Specs having external references ($ref) to different files can now be resolved and previewed.

### Problem

Previously OpenAPI specs having external file references could not be resolved 

### Fix

Since Bruno runs locally, passing the Swagger URL directly to the renderer meant that external files could not be resolved. This PR introduces the external library apidevtools/json-schema-ref-parser(https://github.com/APIDevTools/json-schema-ref-parser), which resolves external files in the Electron app and sends the resolved spec to the renderer. The renderer then passes it to Swagger, which inlines the external references.

Few Reasons to use an External Library
1)Base Paths Change Per Spec File: An external reference is resolved  relative to the file and not root , so for each spec its path has to be tracked
2)Repeated refs needs to be cached so that they are not fetched again and again
3)Visited files need to be tracked so that circular references do not crash the app
4)Updating Pointers after inlining content
5)It covers many edge cases that we would otherwise have to handle ourselves.

### Screenshots

| Before | After |
| 
image
 |  
Screenshot 2026-08-12 at 4 30 28 PM
|

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * OpenAPI specifications now resolve local external references, including nested and chained references.
  * Referenced endpoints and schemas appear in the API specification viewer while the original source remains unchanged.
  * Multi-file specifications are supported, with updates reflected when referenced files change.

* **Bug Fixes**
  * Improved handling of local references to prevent unresolved-reference errors.

* **Tests**
  * Added coverage for reference resolution, parsing scenarios, file updates, and multi-file OpenAPI rendering.

### DIFF
--- a/packages/bruno-electron/src/app/apiSpecsWatcher.js
+++ b/packages/bruno-electron/src/app/apiSpecsWatcher.js
@@ -16,20 +16,15 @@ const hydrateApiSpecWithUuid = (apiSpec, pathname) => {
   return apiSpec;
 };
 
-const refFileWatchState = new Map();
-// Maps the root path of the spec to an object containing the watcher object and set of referenced files
-const syncRefFileWatchers = (rootPath, refFilePaths) => {
-  const state = refFileWatchState.get(rootPath);
-  if (!state) return;
-  // If any referenced file is not being watched,add it to watcher and update the state.
-  const unwatched = refFilePaths.filter((filePath) => !state.watchedRefFilePaths.has(filePath));
+const syncRefFileWatchers = ({ watcher, watchedRefFilePaths }, refFilePaths) => {
+  const unwatched = refFilePaths.filter((filePath) => !watchedRefFilePaths.has(filePath));
   if (!unwatched.length) return;
 
-  unwatched.forEach((filePath) => state.watchedRefFilePaths.add(filePath));
-  state.watcher.add(unwatched);
+  unwatched.forEach((filePath) => watchedRefFilePaths.add(filePath));
+  watcher.add(unwatched);
 };
 
-const add = async (win, pathname) => {
+const add = async (win, pathname, refWatchState) => {
   if (!hasApiSpecExtension(pathname)) return;
   try {
     const basename = path.basename(pathname);
@@ -47,13 +42,13 @@ const add = async (win, pathname) => {
     file.resolvedJson = resolvedJson;
     hydrateApiSpecWithUuid(file, pathname);
     win.webContents.send('main:apispec-tree-updated', 'addFile', file);
-    syncRefFileWatchers(pathname, refFilePaths);
+    syncRefFileWatchers(refWatchState, refFilePaths);
   } catch (err) {
     console.error(err);
   }
 };
 
-const change = async (win, pathname) => {
+const change = async (win, pathname, refWatchState) => {
   if (!hasApiSpecExtension(pathname)) return;
   try {
     const basename = path.basename(pathname);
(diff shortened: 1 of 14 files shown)


packages/bruno-electron/src/utils/apiSpecs.js
```diff
@@ -16,4 +26,34 @@ const parseApiSpecContent = (content, extension) => {
...
+  visited.add(node);
+
+  const refTarget = externalRefTarget(node.$ref, specDir);
+  if (refTarget && fs.existsSync(refTarget)) return true;
+
+  return Object.values(node).some((value) => containsExternalFileRef(value, specDir, visited));
+};
+
+const resolveExternalApiSpecRefs = async (json, apiSpecPath) => {
+  if (!containsExternalFileRef(json, path.dirname(apiSpecPath))) return null;
```
sachin-bruno: `containsExternalFileRef` here and `new $RefParser()` on the next line are outside the try below. The recursion can throw a RangeError on a deeply nested spec, and the constructor throws if the dep didn't install. Either one escapes this function, the watcher catch eats it, and the `apispec-tree-updated` send never runs, so a spec that opened fine before stops opening. Move the try up to cover both lines and return null on failure.

sachin-bruno (approved):

arpit-bruno (changes requested):
referenced files are nott watched still and  only the root spec triggers

packages/bruno-electron/src/app/apiSpecsWatcher.js
```diff
@@ -111,10 +131,24 @@ class ApiSpecWatcher {
...
-        .on('change', (pathname) => change(win, pathname, apiSpecUid, watchPath, workspacePath));
+      const isSpecItself = (pathname) => path.resolve(pathname) === path.resolve(watchPath);
 
+      // watch the changes in the spec file and referenced file.
+      watcher
+        .on('add', (pathname) => {
+          // if the added file is spec file then sends an addFile message to renderer.
+          if (isSpecItself(pathname)) add(win, watchPath);
+          // else if the added file is referenced file then refresh the spec file
+          else change(win, watchPath);
```
arpit-bruno: `watcher.add()` makes chokidar emit an event for every file we register and else block runs for every referenced file even after we have resolved so 10 ref files = 10 `change(...)` which is redundant here.

arpit-bruno: We only need this else for a deleted ref file that comes back, so lets skip it when the path is already in watchedRefFilePaths

packages/bruno-electron/src/app/apiSpecsWatcher.js
```diff
@@ -16,6 +16,19 @@ const hydrateApiSpecWithUuid = (apiSpec, pathname) => {
   return apiSpec;
 };
 
+const refFileWatchState = new Map();
```
arpit-bruno: This map re-stores the watcher we already have in this.watchers[watchPath], dont we need a cleanup for this?

packages/bruno-electron/src/app/apiSpecsWatcher.js
arpit-bruno, line 1: shouldwe add a unit test for the ref-file watching? Right now we have only for resolveExternalApiSpecRefs

arpit-bruno (changes requested):
a few typos in the new comments in apiSpecs.js and apiSpecsWatcher.js like"chceks, referencs, resoved".

Also most of them explains what the very next line already says like "if the added file is spec file then sends an addFile message".  
Lets try not add too many inline comments as  comments should only explain the "why" behind a code-block or any specific line like  the reason behind a decision or an edge case.

arpit-bruno (approved):

bijin-bruno (changes requested):
@adwait-bruno Please check the e2e test failure with `tests/variables-tab/variables-tab.spec.ts`

bijin-bruno (approved):

PR #8950, opened 2026-08-12
fix(app): keep app code when a collection mounts from the file cache

chirag-bruno:
### Description

App code disappeared from a collection once it was migrated from bru to yml with the file cache (mount path v2) enabled.

### Problem

The mount v2 tree builder never copied the `app` block onto the items it sends the renderer, so `app.enabled` read as `false` and the code editor was empty. Migration made this visible because it unmounts and reopens the collection, but any cold mount with the file cache on dropped app code the same way. The files on disk were always correct.

Separately, the large-file `.bru` parse path redacted every big text block except app code, because the app block is a dictionary (`enabled` + `code`) rather than a text block, so big apps were still parsed inline by the grammar.

### Fix

- Carry `app` on the mount tree node, and refresh it from disk on a tree merge like every other file-derived field.
- Redact the app `code` value on the large-file path, keeping its pair and `'''` delimiters so the skeleton still parses.

Ticket - BRU-4243

## Summary by CodeRabbit

- **Bug Fixes**
  - Application code and metadata now remain intact when collections are loaded or migrated.
  - Improved handling of standalone application-code items and requests without application code.
  - Large application-code blocks are safely redacted while preserving surrounding metadata and formatting.
  - Empty or incomplete code blocks are handled without introducing unwanted redactions.

- **Reliability**
  - Collection migration now preserves the application view and existing code.

### DIFF
--- a/packages/bruno-app/src/providers/ReduxStore/slices/collections/app-code-tree.spec.js
+++ b/packages/bruno-app/src/providers/ReduxStore/slices/collections/app-code-tree.spec.js
@@ -1,0 +1,30 @@
+import reducer, { collectionLoadedFromTree } from 'providers/ReduxStore/slices/collections';
+
+const COLLECTION_UID = 'col-1';
+
+const treeItem = (app) => ({
+  uid: 'item-1',
+  name: 'req',
+  type: 'http-request',
+  seq: 1,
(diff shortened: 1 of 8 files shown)


sachin-thakur-bruno (approved):
LGTM

bijin-bruno (approved):

PR #8951, opened 2026-08-12
fix(sqlite): write generated artifacts as TypeScript literals

chirag-bruno:
The generator built src/generated/*.ts with JSON.stringify, which double
quotes every string and key. That trips our lint rules, so a freshly
generated tree came out dirty until someone ran lint:fix over it.

Replace that with a small literal() helper that emits TS source directly:
single quoted strings, bare identifier keys, two space indent.

Tests cover the quoting and escaping round trip (single quotes inside
SQL, backslashes, newlines) and both __proto__ cases.

Ticket - BRU-4266

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQLite artifact generation to preserve strings, backslashes, line endings, and nested data accurately.
  - Added safer handling for special characters and object keys, reducing the risk of malformed generated SQL or migration data.
  - Improved formatting and readability of generated arrays and objects while keeping empty collections compact.
  - Added more consistent serialization for generated migration and statement artifacts.

### DIFF
--- a/packages/bruno-sqlite/scripts/lib/literal.ts
+++ b/packages/bruno-sqlite/scripts/lib/literal.ts
@@ -1,0 +1,28 @@
+const IDENTIFIER = /^[A-Za-z_$][A-Za-z0-9_$]*$/;
+
+const quote = (value: string): string =>
+  `'${value.replace(/\\/g, '\\\\').replace(/'/g, '\\\'').replace(/\n/g, '\\n').replace(/\r/g, '\\r')}'`;
+
+// A `__proto__` property name in an object initializer invokes the prototype setter, quoted or
+// not; only a computed key emits it as a data property.
+const key = (name: string): string => {
+  if (name === '__proto__') return `[${quote(name)}]`;
+  return IDENTIFIER.test(name) ? name : quote(name);
+};
+
+export const literal = (value: unknown, indent = ''): string => {
+  if (typeof value === 'string') return quote(value);
+
+  if (Array.isArray(value)) {
+    const items = value.map((item) => `${indent}  ${literal(item, `${indent}  `)}`);
+    return items.length ? `[\n${items.join(',\n')}\n${indent}]` : '[]';
+  }
+
+  if (typeof value === 'object' && value !== null) {
+    const properties = Object.entries(value).map(([name, item]) =>
+      `${indent}  ${key(name)}: ${literal(item, `${indent}  `)}`);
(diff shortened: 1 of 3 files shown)


packages/bruno-sqlite/scripts/lib/literal.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const literal = (value: unknown, indent = ''): string => {
+  if (typeof value === 'string') return quote(value);
+
+  if (Array.isArray(value)) {
+    const items = value.map((item) => `${indent}  ${literal(item, `${indent}  `)}`);
+    return items.length ? `[\n${items.join(',\n')}\n${indent}]` : '[]';
+  }
+
+  if (typeof value === 'object' && value !== null) {
```
sachin-thakur-bruno, line 21: can value also contains map, set, date etc? Should we need to handle them separatly or throw for them?

chirag-bruno: I mean, this comes in a codepath that is being called by us in the generator, we are not passing anything arbitrary like `Date`, `Map`, `Set`, or anything else like that. This was just a replacement for the default `JSON.stringify()` as it did not support generating strings with `single-quotes`. I don't think it should be a problem.

packages/bruno-sqlite/scripts/lib/literal.ts
```diff
@@ -0,0 +1,21 @@
...
+    const properties = Object.entries(value).map(([name, item]) =>
+      `${indent}  ${IDENTIFIER.test(name) && name !== '__proto__' ? name : quote(name)}: ${literal(item, `${indent}  `)}`);
```
sachin-thakur-bruno: for __proto__ should we do ['__proto__'] ? JS will treat this '__proto__' also as a prototype object instead of normal key

sid-bruno:
A couple of small test gaps to tighten this before merge:

1. **`__proto__`** - worth one assertion since that filter landed in its own commit. Something like expecting `{ __proto__: 'x' }` to emit a quoted `'__proto__'` key.
2. **Realistic shape** - one fixture that looks like a real migration/statement (SQL with a `'` in it) through `literal` + evaluate would sit closer to what generate actually writes.
3. **Optional:** a tiny generate smoke that the written artifact uses single quotes - unit tests never hit the writer path today.

sid-bruno (approved):

PR #8959, opened 2026-08-13
fix(mock-server): review fixes on mock-server

utkarsh-bruno:
JIRA - BRU-4283

### Description

After the review skill, there have been few issues that were flagged. These included 
- Any mock request kills the main process once a delay is set.
- Five write handlers skip the workspace-path guard their siblings use.
- A corrupt store file erases every mock server for the workspace
- A git pull of the store file is silently overwritten.
- BRU-4157 root cause found
- BRU-4190 root cause found
- Clone's port schema has no required

### Problem

- setTimeout(sendResponse, delay) moves response writing outside Express's error boundary, and there is no process.on('uncaughtException') anywhere in bruno-electron. 
- create/save/delete/replace-responses and delete omit validateWorkspacePath, while list-instances, save-instance and clone-responses in the same file call it
- A malformed mockserver.yml is swallowed into createEmptyStore(), cached, and written back by the next save
- The store is cached for the process lifetime with no invalidation, so the watcher's store-updated round trip re-reads stale data and the next save clobbers the pulled file
- routeCount is a second, independent derivation of a number the route table already computes from mockResponses
- an async suggestAvailableMockServerPort().then(setFieldValue) effect overwrites whatever the user typed and re-runs on any configuredInstances identity change.
- Clone's port schema has no required() or integer check and the taken-port test returns false for empty, so clearing the field clears the error and the clone persists on the default port

### Fix

- Wrap the setTimeout callback in try/catch so invalid headers/status codes return a 500 JSON error instead of an uncaught exception that kills the main process.
- Call validateWorkspacePath on the five write handlers that were missing it.
- Stop swallowing parse failures into createEmptyStore(). Throw Failed to read mock server store (...) so nothing is cached and disk is not overwritten on the next save.
- Add invalidateWorkspaceStoreCache to drop the in-memory store and cancel pending debounced writes
- Derive routeCount from mockResponses via countMockRoutes() and stop storing routeCount/exampleCount in servers state.
- Add typeError, required, and integer to the port schema; guard the duplicate-port test so empty values defer to required; run

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - Mock server cloning now checks that ports are valid, available, and not duplicated before proceeding.
  - Mock server dashboards now display accurate route and response counts.
- **Bug Fixes**
  - Improved synchronization when mock server data changes externally.
  - Invalid or corrupted mock server configuration now reports a clear error instead of silently resetting.
  - Delayed response failures now return a structured server error.
  - Added validation to prevent invalid mock server data changes.

### DIFF
--- a/packages/bruno-electron/src/ipc/mock-server/index.js
+++ b/packages/bruno-electron/src/ipc/mock-server/index.js
@@ -169,6 +169,8 @@ const registerMockServerIpc = (mainWindow) => {
 
   ipcMain.handle('renderer:mock-server-create-response', async (_event, payload) => {
     try {
+      validateWorkspacePath(payload?.workspacePath);
+
       const response = createEmptyMockResponse(payload?.name);
       if (payload?.description) {
         response.description = payload.description;
@@ -190,6 +192,9 @@ const registerMockServerIpc = (mainWindow) => {
   ipcMain.handle('renderer:mock-server-save-response', async (_event, payload) => {
     try {
       const { response, ...location } = payload;
+
+      validateWorkspacePath(location.workspacePath);
+
       const savedResponse = saveMockResponse(location, response);
       const routeResult = await mockServer.reloadRoutesFromStore(location.mockServerUid, location);
       return { success: true, response: savedResponse, routes: routeResult?.routes || [] };
@@ -201,6 +206,9 @@ const registerMockServerIpc = (mainWindow) => {
   ipcMain.handle('renderer:mock-server-delete-response', async (_event, payload) => {
     try {
       const { responseUid, ...location } = payload;
+
+      validateWorkspacePath(location.workspacePath);
+
       deleteMockResponse(location, responseUid);
       const routeResult = await mockServer.reloadRoutesFromStore(location.mockServerUid, location);
       return { success: true, responseUid, routes: routeResult?.routes || [] };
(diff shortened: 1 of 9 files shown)


packages/bruno-electron/src/ipc/mock-server/index.js
```diff
@@ -160,6 +160,8 @@ const registerMockServerIpc = (mainWindow) => {
 
   ipcMain.handle('renderer:mock-server-create-response', async (_event, payload) => {
     try {
+      validateWorkspacePath(payload?.workspacePath);
```
bijin-bruno: Can we avoid optional chaining for payload? Is there a case where payload is null?

packages/bruno-electron/src/ipc/mock-server/index.js
```diff
@@ -243,6 +254,8 @@ const registerMockServerIpc = (mainWindow) => {
 
   ipcMain.handle('renderer:mock-server-delete', async (_event, payload) => {
     try {
+      validateWorkspacePath(payload?.workspacePath);
```
bijin-bruno: Can we avoid optional chaining for payload? Is there a case where payload is null?
It would have failed in line 260

naman-bruno (approved):

PR #8960, opened 2026-08-13
feat(pm-migration):handle-pm.sendrequest-script-chaining

abhishekp-bruno:
BRU-3117

### Description

Promise chains are now supported while migrating a postman collection scripts. Previously only the callback form and await pm.sendRequest(...) were handled; .then()/.catch()/.finally() chains were left untransformed. The transformer now walks the chain from the sendRequest call outward.

### Problem

When we migrate a Postman script which looks like this pm.sendRequest(config) we check if there is an await that already exists, else we add an await to the script and replace pm.sendRequest -> bru.sendRequest for the migration.
But in our current implementation we are not checking for any chain implementation, if the script has .then .catch .finally handlers which ends up breaking the pm to bru script conversation.

The migration goes like this in the following steps

pm-script --> AST (Abstract syntax tree - using jscodeshift library)
AST --> modifies AST (we modify the AST nodes to convert them to work in bruno)
the problem happens in this step -  modified AST --> converted back to source code during this step since now that we have added await when it is converted back to source code form, it checks precedence. **Since `(.)` binds tighter than `await` it gets converted to**

```
(await bru.sendRequest(config)).then((res) => {
  console.log(res.json());
});
```

### Fix

-  Await the outermost expression, not the inner call. When we do await, await is added to the whole expression statement rather than being nested inside it, so the entire chain resolves before the script continues.
- 
```
// before (broken)
(await sendRequest(options)).then(...);

// after
await sendRequest(options);
```

3. Map the response shape in the .then handler
Postman's pm.sendRequest callback and Bruno's resolved response expose different shapes like this 

Postman | Bruno
-- | --
res.json() / res.text() | res.data
res.code | res.status
res.status | res.statusText

This is handled in the first .then handler where response is consumed 

### Screenshots

| Before | After |
| --- | --- |
| image | image |

#### Contribution Checklist:

## Summary by CodeRabbit

* **Improvements**
  * Improved conversion of Postman `pm.sendRequest` promise chains.
  * Supports `.then`, `.catch`, and `.finally`, including nested and multiple chains.
  * Translates response properties, callbacks, request variables, and bracket notation.
  * Preserves behavior in asynchronous contexts, including shadowed and destructured parameters.

* **Bug Fixes**
  * Prevents incorrect or unnecessary `await` insertion.
  * Safely handles unsupported, ambiguous, reassigned, and pass-through callback patterns.
  * Preserves existing asynchronous behavior while avoiding unsafe rewrites.

### DIFF
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -550,8 +550,7 @@ const sendRequestTransformer = (path, j) => {
 
   const wasParentAwaited = callPath.parent.value.type === 'AwaitExpression';
 
-  // the only shape that does is `(await bru.sendRequest(req)).then(h)`, which calls `.then` on
-  // a plain response object and throws at runtime regardless of what we emit.
+  // Already inside an `await`, so no promise chain can legitimately follow: `(await bru.sendRequest(req)).then(h)`
   if (wasParentAwaited) return sendRequestCall;
 
   const chainLinks = getPromiseChainLinks(callPath);
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -284,9 +284,9 @@ const rewriteThenHandlerResponseAccess = (j, handlerPath) => {
     }
   }).forEach((memberPath) => {
     const propertyName = getStaticPropertyName(memberPath.value);
-    if (!Object.hasOwn(responsePropertyMap, propertyName)) return;
-
+    if (!propertyName) return;
     const bruProperty = responsePropertyMap[propertyName];
+    if (typeof bruProperty !== 'string') return;
 
     // skip references shadowed by a nested re-declaration of the name
     if (!resolvesToHandlerParam(memberPath, responseVarName, handler)) return;
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -516,9 +516,6 @@ const sendRequestTransformer = (path, j) => {
   const requestOptions = args[0];
   const callback = args[1];
 
-  // Check if original call was awaited
-  const wasParentAwaited = callPath.parent.value.type === 'AwaitExpression';
-
   // transform the request config options
   if (requestOptions.type === 'ObjectExpression') {
     // Transform headers
@@ -550,6 +547,8 @@ const sendRequestTransformer = (path, j) => {
     j.identifier('bru.sendRequest'),
     transformedCallback ? [requestOptions, transformedCallback] : [requestOptions]
   );
+
+  const wasParentAwaited = callPath.parent.value.type === 'AwaitExpression';
 
   // the only shape that does is `(await bru.sendRequest(req)).then(h)`, which calls `.then` on
   // a plain response object and throws at runtime regardless of what we emit.
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -214,10 +214,7 @@ const getPromiseChainLinks = (callPath) => {
 };
 
 /**
- * Whether a function is a `.then`/`.catch`/`.finally` handler argument. Such a
- * handler's return value is already promise-wrapped by the chain, so turning it
- * async is transparent to its caller — unlike an arbitrary callback, where the
- * caller would start receiving a promise in place of the value it expects.
+ * Whether a function is a `.then`/`.catch`/`.finally` handler argument
  * @param {Object} functionPath - Path of the function to test
  * @returns {boolean}
  */
@@ -285,9 +282,6 @@ const rewriteThenHandlerResponseAccess = (j, handlerPath) => {
 
     const replacement = j.memberExpression(j.identifier(responseVarName), j.identifier(bruProperty));
 
-    // response.json() collapses to response.data — the call goes away with it.
-    // Only when the member is the callee: in console.log(response.code) the
-    // parent is also a CallExpression, and replacing it would eat the log.
     const parentPath = memberPath.parent;
     if (parentPath.value.type === 'CallExpression' && parentPath.value.callee === memberPath.value) {
       j(parentPath).replaceWith(replacement);
@@ -311,11 +305,8 @@ const resolvesToHandlerParam = (path, name, handler) => {
 };
 
 /**
- * The return statements belonging to the handler itself. `find` reaches into nested functions too,
- * so a `return res.data` inside a callback declared in the handler would be mistaken for the
- * handler's own return; keep only the returns whose nearest enclosing function is the handler.
- *
- * @example
+ * The return statements belonging to the handler itself.
+* @example
  * sendRequest(req).then((res) => {
  *   const extract = () => {
  *     return res.data;      // nested function's return — not the handler's
@@ -323,7 +314,6 @@ const resolvesToHandlerParam = (path, name, handler) => {
  *   console.log(extract());
  *   return res;             // the handler's actual return
  * });
- *
  * @param {Object} j - jscodeshift API
  * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
  * @returns {Object[]} Paths of the handler's own return statements
@@ -335,9 +325,7 @@ const findOwnReturns = (j, handlerPath) =>
     .filter((returnPath) => j(returnPath).closest(j.Function).get().value === handlerPath.value);
 
 /**
- * Whether every value the handler can return is its own response parameter, so the
- * next `.then` in the chain still receives the response. A path that falls through
- * to an implicit undefined does not qualify.
+ * Whether every value the handler can return is its own response parameter
  * @param {Object} j - jscodeshift API
  * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
  * @returns {boolean}
@@ -353,9 +341,8 @@ const returnsParamUnchanged = (j, handlerPath) => {
     return body.type === 'Identifier' && body.name === paramName;
   }
 
-  // a body that can complete without returning forwards an implicit undefined, so the last
-  // statement must be a return — this also rejects handlers with no return at all, which the
-  // `every` check below would treat as vacuously passing
+  // Every path must end in a return: a body that can fall through forwards an implicit undefined,
+  // and a handler with no returns at all would pass the `every` check below vacuously.
   const lastStatement = body.body[body.body.length - 1];
   if (!lastStatement || lastStatement.type !== 'ReturnStatement') return false;
 
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -223,8 +223,8 @@ const getPromiseChainLinks = (callPath) => {
  */
 const isPromiseChainHandler = (functionPath) => {
   const parent = functionPath.parent;
-  if (!parent || parent.value.type !== 'CallExpression') return false;
-  if (!parent.value.arguments.includes(functionPath.value)) return false;
+  if (!parent || parent.value.type !== 'CallExpression') return false; // means the function is not a handler
+  if (!parent.value.arguments.includes(functionPath.value)) return false; // means the function is not a handler
 
   const callee = parent.value.callee;
   return callee.type === 'MemberExpression' && PROMISE_CHAIN_METHODS.has(getStaticPropertyName(callee));
@@ -325,18 +325,19 @@ const returnsParamUnchanged = (j, handlerPath) => {
   const paramName = handler.params[0].name;
   const body = handler.body;
 
-  // if the body is not a block statement then check if the body is an
-  // identifier and the name is the same as the parameter name
+  // a concise arrow body is the returned expression itself, so it only forwards the
+  // response when that expression is the parameter — `res => res` yes, `res => res.data` no
   if (body.type !== 'BlockStatement') {
     return body.type === 'Identifier' && body.name === paramName;
   }
 
-  // if the body is a block statement then check if the last statement is
-  // a return statement and the argument is an identifier and the name is the same as the parameter name
+  // if the body is a block statement then check if the last statement isna return statement
+  // and the argument is an identifier and the name is the same as the parameter name
   const lastStatement = body.body[body.body.length - 1];
   if (!lastStatement || lastStatement.type !== 'ReturnStatement') return false;
 
-  // check if the return statement is the own return statement of the handler
+  // check that decides whether a .then(handler) is a pass-through (every value it can return is the untouched response param),
+  // so the rewrite can keep flowing down the chain.
   const ownReturns = j(handlerPath)
     .find(j.ReturnStatement)
     .paths()
@@ -362,22 +363,27 @@ const isResponseParamReassigned = (j, handlerPath) => {
   const handler = handlerPath.value;
   const paramName = handler.params[0].name;
 
+  // check if code overwriting the handler's response parameter itself
   const isRebind = (path, target) =>
     target.type === 'Identifier'
     && target.name === paramName
     && resolvesToHandlerParam(path, paramName, handler);
 
-  const assigned = j(handlerPath)
+  // `res = ...`, `res += ...`, `res ||= ...`
+  const assignedByAssignmentExpression = j(handlerPath)
     .find(j.AssignmentExpression)
     .paths()
     .some((path) => isRebind(path, path.value.left));
 
-  if (assigned) return true;
+  if (assignedByAssignmentExpression) return true;
 
-  return j(handlerPath)
+  // `res++` / `--res` — a separate node type, with its target under `argument`
+  const assignedByUpdateExpression = j(handlerPath)
     .find(j.UpdateExpression)
     .paths()
     .some((path) => isRebind(path, path.value.argument));
+
+  return assignedByUpdateExpression;
 };
 
 /**
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -238,10 +238,12 @@ const makeAwaitable = (j, path) => {
 
   const functionPath = enclosingFunction.get();
   if (functionPath.value.async === true) return true;
-  if (!isPromiseChainHandler(functionPath)) return false;
+  if (isPromiseChainHandler(functionPath)) {
+    functionPath.value.async = true;
+    return true;
+  }
 
-  functionPath.value.async = true;
-  return true;
+  return false;
 };
 
 /**
--- a/packages/bruno-converters/src/utils/send-request-transformer.js
+++ b/packages/bruno-converters/src/utils/send-request-transformer.js
@@ -291,6 +291,87 @@ const rewriteThenHandlerResponseAccess = (j, handlerPath) => {
 };
 
 /**
+ * Whether an identifier reference inside a handler resolves to the handler's own
+ * parameter rather than a nested re-declaration of the same name.
+ * @param {Object} path - Path of the reference
+ * @param {string} name - Parameter name
+ * @param {Object} handler - Handler function node owning the parameter
+ * @returns {boolean}
+ */
+const resolvesToHandlerParam = (path, name, handler) => {
+  const declaringScope = path.scope.lookup(name);
+  return Boolean(declaringScope) && declaringScope.node === handler;
+};
+
+/**
+ * Whether every value the handler can return is its own response parameter, so the
+ * next `.then` in the chain still receives the response. A path that falls through
+ * to an implicit undefined does not qualify.
+ * @param {Object} j - jscodeshift API
+ * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
+ * @returns {boolean}
+ */
+const returnsParamUnchanged = (j, handlerPath) => {
+  const handler = handlerPath.value;
+  const paramName = handler.params[0].name;
+  const body = handler.body;
+
+  // if the body is not a block statement then check if the body is an
+  // identifier and the name is the same as the parameter name
+  if (body.type !== 'BlockStatement') {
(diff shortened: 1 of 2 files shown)


packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,244 @@ const transformBody = (j, requestOptions) => {
...
+    // response.json() collapses to response.data — the call goes away with it.
+    // Only when the member is the callee: in console.log(response.code) the
+    // parent is also a CallExpression, and replacing it would eat the log.
+    const parentPath = memberPath.parent;
+    if (parentPath.value.type === 'CallExpression' && parentPath.value.callee === memberPath.value) {
+      j(parentPath).replaceWith(replacement);
+    } else {
+      j(memberPath).replaceWith(replacement);
+    }
+  });
```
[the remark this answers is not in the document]
prateek-bruno: Please check if it's relevant in latest code.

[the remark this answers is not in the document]
prateek-bruno: @abhishekp-bruno Please ensure res[code] is not translated, it isn't same as res['code'] or res.code

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -296,31 +533,62 @@ const sendRequestTransformer = (path, j) => {
...
+  if (!chainLinks.length) {
+    return makeAwaitable(j, callPath) ? j.awaitExpression(sendRequestCall) : sendRequestCall;
+  }
+
+  for (const link of chainLinks) {
+    const [handler] = link.callPath.value.arguments;
+
+    const hasFulfilledHandler = link.methodName === 'then' && Boolean(handler) && !isNullLiteral(handler);
+
+    if (!hasFulfilledHandler) continue;
```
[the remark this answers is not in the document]
prateek-bruno: This is resolved in latest.

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,238 @@ const transformBody = (j, requestOptions) => {
...
+
+/**
+ * Whether `await` may be emitted here. Bruno runs scripts in an async closure, so top level is
+ * always awaitable. Inside a non-async function `await` is a syntax error that breaks the whole
+ * script, so we emit it only after making that function async — safe only for promise-chain handlers.
+ * @param {Object} j - jscodeshift API
+ * @param {Object} path - Path the await would be emitted at
+ * @returns {boolean}
+ */
+const makeAwaitable = (j, path) => {
```
prateek-bruno: The function name is misleading, it could be named this:

```suggestion
const ensureAsyncContext = (j, path) => {
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -296,31 +518,63 @@ const sendRequestTransformer = (path, j) => {
...
+    const reassignsResponse = isResponseParamReassigned(j, handlerPath);
+
+    rewriteThenHandlerResponseAccess(j, handlerPath);
+
+    if (!returnsResponse) break;
+    if (reassignsResponse) break;
+  }
+
+  const outermostPath = chainLinks[chainLinks.length - 1].callPath;
+  if (outermostPath.parent.value.type !== 'AwaitExpression' && makeAwaitable(j, outermostPath)) {
```
prateek-bruno: Can extract this to a variable so this becomes more readable:
```
outermostPath.parent.value.type !== 'AwaitExpression'
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,238 @@ const transformBody = (j, requestOptions) => {
...
+  }
+  return property.type === 'Identifier' ? property.name : null;
+};
+
+/**
+ * Whether an argument node is an explicit absent value, as in `.then(null, onError)`.
+ * @param {Object} node - Argument node
+ * @returns {boolean}
+ */
+const isNullLiteral = (node) =>
```
prateek-bruno: Since this is scoped to handler usage and it covers identifiers as well and not only literals it can be named better.

```suggestion
const isNullishHandler = (node) =>
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,238 @@ const transformBody = (j, requestOptions) => {
...
+  }).forEach((memberPath) => {
+    const property = memberPath.value.property;
+    if (property.type !== 'Identifier') return;
+
+    const bruProperty = responsePropertyMap[property.name];
+    if (!bruProperty) return;
+
+    // skip references shadowed by a nested re-declaration of the name
+    const declaringScope = memberPath.scope.lookup(responseVarName);
+    if (!declaringScope || declaringScope.node !== handler) return;
```
prateek-bruno: This is can be rewritten like this:

```suggestion
    if (!resolvesToHandlerParam(memberPath, responseVarName, handler)) return;
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,238 @@ const transformBody = (j, requestOptions) => {
...
+
+    const replacement = j.memberExpression(j.identifier(responseVarName), j.identifier(bruProperty));
+
+    const parentPath = memberPath.parent;
+    if (parentPath.value.type === 'CallExpression' && parentPath.value.callee === memberPath.value) {
+      j(parentPath).replaceWith(replacement);
+    } else {
+      j(memberPath).replaceWith(replacement);
+    }
```
prateek-bruno, line 305: This if condition can be extracted to be more readable.

```suggestion
    const parent = memberPath.parent;
    const isMethodCall = parent.value.type === 'CallExpression' && parent.value.callee === memberPath.value;

    // `json()`/`text()` are methods on the Postman response but plain properties on Bruno's,
    // so the call has to lose its parentheses: `res.json()` -> `res.data`
    if (isMethodCall) {
      j(parent).replaceWith(replacement);
    } else {
      j(memberPath).replaceWith(replacement);
    }
```

packages/bruno-converters/src/utils/send-request-transformer.js
prateek-bruno: When doing this move that `resolvesToHandlerParam` method above this so that eslint doesn't complain about method being used before declaration.

packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js
```diff
@@ -803,4 +803,731 @@ describe('Send Request Translation', () => {
...
+  }
+})
+  .then((res) => {
+    console.log(res.json());
+  })
+  .catch((err) => {
+    console.error(err);
+  });
+      `;
+      const translatedCode = translateCode(code);
```
prateek-bruno, line 830: The variable doesn't needs to be extracted like this, since method name is self explanatory here. Line 1066 already does this inlining of translateCode.

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: updated to isNullish since it is not a handler

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: renamed

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: updated

packages/bruno-converters/src/utils/send-request-transformer.js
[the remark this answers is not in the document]
abhishekp-bruno: in our existing code we are not handling `res[code]`, will not be handling it in this ticket.

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: updated 

```  const isAlreadyAwaited = outermostPath.parent.value.type === 'AwaitExpression';
  const shouldAwaitChain = !isAlreadyAwaited && ensureAsyncContext(j, outermostPath);

  if (shouldAwaitChain) {
    outermostPath.parentPath.value[outermostPath.name] = j.awaitExpression(outermostPath.value);
  }```

File: packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js (hunk 8 above)
Comment (abhishekp-bruno) [reply packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js:830]: this is the pattern in the existing code, updating this will create inconsistency. Will update line 1066 code to follow the same.

File: packages/bruno-converters/src/utils/send-request-transformer.js (hunk 1 above)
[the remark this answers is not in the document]
Comment (prateek-bruno) [reply packages/bruno-converters/src/utils/send-request-transformer.js:306]: @abhishekp-bruno Can we document it so we know atleast. maybe a test.skip should be enough.

**Review by prateek-bruno — APPROVED** [review]
LGTM!

File: packages/bruno-converters/src/utils/send-request-transformer.js [hunk 9]
```diff
@@ -148,6 +148,240 @@ const transformBody = (j, requestOptions) => {
...
+      j(parent).replaceWith(replacement);
+    } else {
+      j(memberPath).replaceWith(replacement);
+    }
+  });
+};
+
+/**
+ * The return statements belonging to the handler itself.
+* @example
```
sanish-bruno, line 311: fix indent,

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,240 @@ const transformBody = (j, requestOptions) => {
...
+  if (!parent || parent.value.type !== 'CallExpression') return false; // means the function is not a handler
+  if (!parent.value.arguments.includes(functionPath.value)) return false; // means the function is not a handler
+
+  const callee = parent.value.callee;
+  return callee.type === 'MemberExpression' && PROMISE_CHAIN_METHODS.has(getStaticPropertyName(callee));
+};
+
+/**
+ * Whether `await` may be emitted here. Bruno runs scripts in an async closure, so top level is
+ * always awaitable. Inside a non-async function `await` is a syntax error that breaks the whole
```
sanish-bruno: Drop them or replace with the why.

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -296,31 +520,66 @@ const sendRequestTransformer = (path, j) => {
...
+    // a `.catch` handler substitutes its own return value for the rest of the chain, so links
+    // past it no longer receive the response we mapped.
+    if (link.methodName === 'catch') break;
+
+    const hasFulfilledHandler = link.methodName === 'then' && Boolean(handler) && !isNullish(handler);
+
+    if (!hasFulfilledHandler) continue;
+
+    if (handler.type !== 'FunctionExpression' && handler.type !== 'ArrowFunctionExpression') break;
+    if (handler.params[0]?.type !== 'Identifier') break; // if the handler does not have a parameter then break
```
sanish-bruno:
```suggestion
    if (handler.params.length && handler.params[0].type === 'Identifier') break;
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,240 @@ const transformBody = (j, requestOptions) => {
...
+    .paths()
+    .filter((returnPath) => j(returnPath).closest(j.Function).get().value === handlerPath.value);
+
+/**
+ * Whether every value the handler can return is its own response parameter
+ * @param {Object} j - jscodeshift API
+ * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
+ * @returns {boolean}
+ */
+const returnsParamUnchanged = (j, handlerPath) => {
```
sanish-bruno:
```suggestion
/**
 * Whether the handler forwards its response parameter untouched, so response-shape
 * rewriting can safely continue on the next `.then` in the chain.
 * @param {Object} j - jscodeshift API
 * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
 * @returns {boolean}
 const isPassThroughHandler = (j, handlerPath) => {
```

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,240 @@ const transformBody = (j, requestOptions) => {
...
+ * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
+ * @returns {Object[]} Paths of the handler's own return statements
+ */
+const findOwnReturns = (j, handlerPath) =>
+  j(handlerPath)
+    .find(j.ReturnStatement)
+    .paths()
+    .filter((returnPath) => j(returnPath).closest(j.Function).get().value === handlerPath.value);
+
+/**
```
sanish-bruno, line 329:
```suggestion
/**
 * Whether the handler forwards its response parameter untouched, so response-shape
 * rewriting can safely continue on the next `.then` in the chain.
 * @param {Object} j - jscodeshift API
 * @param {Object} handlerPath - Path of the `.then` fulfilled handler argument
 * @returns {boolean}
 const isPassThroughHandler = (j, handlerPath) => {
```

packages/bruno-converters/src/utils/send-request-transformer.js
sanish-bruno, line 336:
```suggestion
  // Whether control leaves this node only via return or throw — never by falling through.
  // Terminates the chain-mapping walk at fall-through bodies without rejecting bodies whose
  // paths all return via if/else or nested blocks.
  const alwaysExitsViaReturnOrThrow = (node) => {
    if (!node) return false;
    switch (node.type) {
      case 'ReturnStatement':
      case 'ThrowStatement':
        return true;
      case 'BlockStatement':
        return node.body.some(alwaysExitsViaReturnOrThrow);
      case 'IfStatement':
        return alwaysExitsViaReturnOrThrow(node.consequent)
          && alwaysExitsViaReturnOrThrow(node.alternate);
      default:
        return false;
    }
  };

  const isPassThroughHandler = (j, handlerPath) => {
    const handler = handlerPath.value;
    const paramName = handler.params[0].name;
    const body = handler.body;

    if (body.type !== 'BlockStatement') {
      return body.type === 'Identifier' && body.name === paramName;
    }

    // Reject fall-through bodies — an implicit `undefined` return isn't pass-through.
    // Try/switch/loops aren't modelled and are treated as fall-through; extend the helper
    // if those show up in real migrations.
    if (!alwaysExitsViaReturnOrThrow(body)) return false;

    const ownReturns = findOwnReturns(j, handlerPath);

    return ownReturns.every((returnPath) => {
      const argument = returnPath.value.argument;
      const returnsParamIdentifier = Boolean(argument) && argument.type === 'Identifier' && argument.name === paramName;
      return returnsParamIdentifier && resolvesToHandlerParam(returnPath, paramName, handler);
    });
  }{
```

sanish-bruno (changes requested):

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: we want to break the loop  if there are no params with this if there are no params its falsy and it will proceed 
` empty params → params[0]?.type is undefined → undefined !== 'Identifier' is true → breaks.`
`empty params → length is 0 (falsy) → whole condition false → does not break`

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: discussed keeping this as it is

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: updated

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: removed

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: fixed

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: updated to this for more clarity 
` if (handler.params.length === 0 || handler.params[0].type !== 'Identifier') break; `

sanish-bruno: sorry more like `if (!handler.params.length || handler.params[0].type !== 'Identifier') break;`

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,238 @@ const transformBody = (j, requestOptions) => {
...
+};
+
+/**
+ * Whether a function is a `.then`/`.catch`/`.finally` handler argument
+ * @param {Object} functionPath - Path of the function to test
+ * @returns {boolean}
+ */
+const isPromiseChainHandler = (functionPath) => {
+  const parent = functionPath.parent;
+  if (!parent || parent.value.type !== 'CallExpression') return false; // means the function is not a handler
```
sanish-bruno: rm these comments on both llines, i think jsdoc is explaining is well

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,238 @@ const transformBody = (j, requestOptions) => {
...
+  if (!parent.value.arguments.includes(functionPath.value)) return false; // means the function is not a handler
+
+  const callee = parent.value.callee;
+  return callee.type === 'MemberExpression' && PROMISE_CHAIN_METHODS.has(getStaticPropertyName(callee));
+};
+
+/**
+ * @param {Object} j - jscodeshift API
```
sanish-bruno, line 234: add a brief comments on why is it beinn used

sanish-bruno (changes requested):

packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js
```diff
@@ -803,4 +803,729 @@ describe('Send Request Translation', () => {
...
+      expect(translatedCode).toBe(`await bru.sendRequest({ url: 'https://echo.usebruno.com' }).then((res) => res.data);`);
+    });
+
+    it('should leave bracket-notation response access with a variable key alone', () => {
+      const code = `pm.sendRequest({ url: 'https://echo.usebruno.com' }).then((res) => res[method]());`;
+      const translatedCode = translateCode(code);
+      expect(translatedCode).toBe(`await bru.sendRequest({ url: 'https://echo.usebruno.com' }).then((res) => res[method]());`);
+    });
+
+    it('should not treat a computed variable key as a promise method', () => {
```
sanish-bruno: do we need the test case.

packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js
```diff
@@ -803,4 +803,729 @@ describe('Send Request Translation', () => {
...
+      expect(translatedCode).toBe(`await bru.sendRequest({ url: 'https://echo.usebruno.com' })['then']((res) => res.data);`);
+    });
+
+    it('should rewrite bracket-notation response access with a literal key', () => {
+      const code = `pm.sendRequest({ url: 'https://echo.usebruno.com' }).then((res) => res['json']());`;
+      const translatedCode = translateCode(code);
+      expect(translatedCode).toBe(`await bru.sendRequest({ url: 'https://echo.usebruno.com' }).then((res) => res.data);`);
+    });
+
+    it('should leave bracket-notation response access with a variable key alone', () => {
```
sanish-bruno: rm

sanish-bruno (changes requested):

packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js
abhishekp-bruno: removed

packages/bruno-converters/tests/postman/postman-translations/transpiler-tests/transformers/send-request.test.js
abhishekp-bruno: no, removed

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: updated

packages/bruno-converters/src/utils/send-request-transformer.js
abhishekp-bruno: removed

sanish-bruno (approved):

packages/bruno-converters/src/utils/send-request-transformer.js
```diff
@@ -148,6 +148,241 @@ const transformBody = (j, requestOptions) => {
...
+  }
+  return property.type === 'Identifier' ? property.name : null;
+};
+
+/**
+ * Whether an argument node is an explicit absent value, as in `.then(null, onError)`.
+ * @param {Object} node - Argument node
+ * @returns {boolean}
+ */
+const isNullish = (node) =>
```
vijayh-bruno, line 183: we have a util already for this purpose - https://github.com/usebruno/bruno/blob/aa5d85f71ffe04c2006d3924cff88ce0ccaeb41d/packages/bruno-converters/src/utils/ast-utils.js#L91

can we use that?

abhishekp-bruno: isNullLiteral only matches null, isNullish checks undefined (an Identifier node) as well, we need it here

vijayh-bruno (approved):

PR #8964, opened 2026-08-13
fix(mock-server): restricting port numbers by showing errors

utkarsh-bruno:
### Description

Mock server port numbers now have a single inline error for different validation fails.

### Problem
BRU-4188
Currently, the mock server port number field would show multiple inline errors for validations like range and availability. 

### Fix

Moved the port range rules into a single shared helper (getMockServerPortRangeError) so all three places check the same thing and show the same message. Both modals now run full form validation before submitting, so a bad port actually blocks the save instead of slipping through. Port errors go through one channel and show up inline under the field rather than as a toast plus a second error line. The create modal only checks the port when that field is actually visible. Also fixed the dashboard's useEffect that called validatePort before it was defined, and added unit tests for the new helper.

#### Contribution Checklist:

Note: The PR is created on top of https://github.com/usebruno/bruno/pull/8959.

## Summary by CodeRabbit

- **New Features**
  - Mock server ports now validate required values, whole numbers, valid ranges, and availability.
  - Mock server names support any characters when nonempty and within the length limit.
  - Collection-backed mock servers retain their collection location.
  - Dashboard statistics and route counts now reflect current mock responses.
  - Mock server changes automatically update responses, routes, statistics, and related interface state.

- **Bug Fixes**
  - Improved create and clone flows with more reliable saving, navigation, and response loading.
  - Validation errors now appear directly in the relevant form fields and clear when corrected.


packages/bruno-app/src/components/MockServer/MockServerDashboard/index.js
```diff
@@ -118,6 +106,10 @@ const MockServerDashboard = ({ instance, collection }) => {
     }
   };
 
+  useEffect(() => {
+    validatePort(activePort);
+  }, [activePort]);
```
sid-bruno: nit: Since `validatePort` can race for values, we probably want to add in a scoped check based on a value if it's possible.

packages/bruno-app/src/utils/mock-server/mock-server-instances.js
```diff
@@ -12,6 +12,32 @@ import {
...
+  if (normalizedPort < MOCK_SERVER_PORT_MIN) {
+    return 'Port must be at least 1';
+  }
+
+  if (normalizedPort > MOCK_SERVER_PORT_MAX) {
+    return 'Port must be 65535 or less';
+  }
+
+  return null;
+};
```
sid-bruno, line 40: Any specific reason to hand write this instead of sticking to `yup`?

utkarsh-bruno: Yes, since I am not using form on MockServerDashboard, it keeps the port in plain state and just wants an error string to show.

sid-bruno (dismissed):

packages/bruno-app/src/components/MockServer/MockServerDashboard/index.js
```diff
@@ -84,40 +85,36 @@ const MockServerDashboard = ({ instance, collection }) => {
...
+    resolvePortError(activePort).then((error) => {
+      if (isCurrent) {
+        setPortError(error);
+      }
+    });
+
+    return () => {
+      isCurrent = false;
+    };
+  }, [activePort]);
```
sid-bruno: Revalidates when other mock ports change without array-ref thrash.

```suggestion
  const conflictingPortsKey = workspaceInstances
    .filter((i) => i.uid !== storedInstance.uid)
    .map((i) => Number(i.port))
    .join(',');

  useEffect(() => {
    let isCurrent = true;

    resolvePortError(activePort).then((error) => {
      if (isCurrent) {
        setPortError(error);
      }
    });

    return () => {
      isCurrent = false;
    };
  }, [activePort, conflictingPortsKey]);
```

packages/bruno-app/src/components/MockServer/MockServerDashboard/index.js
```diff
@@ -84,40 +85,36 @@ const MockServerDashboard = ({ instance, collection }) => {
...
+
```
sid-bruno, line 125: If another workspace changes the ports, there's no way to find out currently.

```suggestion
    const conflictingPortsKey = workspaceInstances
    .filter((i) => i.uid !== storedInstance.uid)
    .map((i) => Number(i.port))
    .join(',');

  useEffect(() => {
    let isCurrent = true;
    resolvePortError(activePort).then((error) => {
      if (isCurrent) {
        setPortError(error);
      }
    });
    return () => {
      isCurrent = false;
    };
  }, [activePort, conflictingPortsKey]);

```

sid-bruno (approved):

PR #8966, opened 2026-08-13
fix(docs): add support for inline raw HTML nodes and disable forced line breaks

Gopi-bruno:
### Description

Fixes three related bugs in the Docs Rich Text Editor's raw-HTML handling (`packages/bruno-app/src/ui/RichTextEditor`), tracked as [BRU-4313](https://usebruno.atlassian.net/browse/BRU-4313), [BRU-4326](https://usebruno.atlassian.net/browse/BRU-4326), and [BRU-4329](https://usebruno.atlassian.net/browse/BRU-4329): wrapped prose was silently rewritten with forced line breaks, inline HTML with no dedicated schema node vanished on save, and content wrapped in a plain text-only tag (`<div>`, `<p>`, ...) couldn't be edited at all in Rich Text mode. Also closes a paste-based XSS gap and a lossy round-trip bug found while hardening the fix for the third issue.

### Problem

1. **[BRU-4313] Wrapped prose churn** — `Markdown.configure({ breaks: true })` parsed an ordinary soft line break (a `\n` inside a paragraph — meant to render as a space, per CommonMark) as a hard break. `EditorHardBreak`'s serializer then wrote that back out as a `\` line-continuation, so opening and saving *any* doc with hand-wrapped prose silently rewrote every wrapped line into a forced break — a whole-file diff plus a semantic change nobody asked for.
2. **[BRU-4326] Inline HTML silently dropped** — only block-level raw HTML (`<div>`, `<video>`, `<details>`, ...) had a preservation mechanism (`EditorRawHtmlBlock`). Inline HTML with no dedicated mark — `<u>`, `<mark>`, `<span>`, `<abbr>`, `<small>` — had no `html_inline` equivalent, so it vanished the moment the doc was edited and saved in Rich Text mode.
3. **[BRU-4329] Text-only tags weren't editable** — content wrapped in a plain text-only tag (`<div>`, `<p>`, ...) was preserved as an opaque, non-editable atom (same treatment as `<video>`/`<table>`), so users could see it in Rich Text mode but couldn't click into it and edit the text.
4. **Missing sanitize config on update** — the raw-HTML-block node view's `update()` path called `DOMPurify.sanitize()` without `SANITIZE_CONFIG` (which forbids `<style>`), unlike its initial-render path — so a `<style>` block in an untrusted collection's docs could restyle the whole app UI after the node re-rendered.

### Fix

- **BRU-4313** — Set `breaks: false` on the Markdown extension so a soft line break stays a reflowable space, matching how any other CommonMark renderer (e.g. GitHub) displays the same file.
- **BRU-4326** — Added `EditorRawHtmlInline`, the inline counterpart to `EditorRawHtmlBlock`: any inline HTML tag with no schema mapping is preserved as an opaque, sanitized atom instead of disappearing. An early pass swallowed *every* inline HTML tag indiscriminately, which broke tags that already have a dedicated mapping — `<br>` (hard breaks), `<kbd>`/`<sup>` (existing inline marks), and the `<input>` checkbox `markdown-it-task-lists` synthesizes for task items. Fixed by exempting those tag names (matched by name, not exact string, so attributes don't matter — including single- and double-quoted `type="checkbox"`) from the raw-HTML wrapping.
- **BRU-4329** — Added `EditorRawHtmlTextBlock`: a block whose HTML sanitizes down to a single wrapper with only plain text and `<br>`s becomes a real, editable ProseMirror node instead of an opaque atom, preserving its original tag and attributes. Anything with nested markup, or a self-contained/dangerous tag (`<video>`, `<script>`, `<table>`, ...), still falls back to the opaque `rawHtmlBlock` atom.
- Fixed the missing `SANITIZE_CONFIG` on the block node's `update()` call, and de-duplicated the block/inline/text-block node definitions into one `createRawHtmlNode` factory so the `DOMPurify.sanitize` call sites can't drift independently again.
- Split the shared markdown-it parser setup into two independent functions (one for `html_block`, one for `html_inline`), so registering one raw-HTML node no longer has the side effect of patching the other's renderer rule too.

**Hardening found while reviewing the above:**
- **Paste-based XSS** — `EditorRawHtmlTextBlock`'s ProseMirror `parseHTML` rule matched *any* element carrying its marker attribute, including one pasted in from an untrusted external page, and copied its attributes verbatim with no sanitization (unlike the markdown-load path). A crafted paste such as `<div data-raw-html-text-block="img" onerror="alert(1)">` would execute immediately and persist into the collection's `docs`. Fixed by re-running the same DOMPurify sanitization on the pasted element and rejecting a marker-attribute tag value that names a dangerous tag (`script`, `iframe`, ...).
- **Lossy round-trip on unedited legacy blocks** — making a simple HTML block editable meant its markup got rebuilt from the parsed DOM on every save (normalizing quote style, tag case, entities) even when the user never touched it, silently rewriting unrelated content already on disk. Fixed by capturing the exact original bytes at parse time and re-emitting them verbatim as long as the block's visible content is unchanged, only reconstructing once it's actually edited (with validation so a pasted/forged "original" can't be smuggled through).
- A `Node` naming collision (the module's `@tiptap/core` import shadowing the global DOM `Node`) in a follow-up whitespace-handling fix silently broke text-block parsing for every input; fixed by qualifying the DOM constants as `globalThis.Node.ELEMENT_NODE`/`TEXT_NODE`.

Added Playwright coverage in `tests/richtext-editor/`:
- Extended `raw-html-passthrough.spec.ts` with inline-HTML, editable-text-block, and legacy-round-trip cases.
- New `soft-line-breaks.spec.ts` covering the BRU-4313 fix.
- Extensive new Jest coverage in `editorMarkdownSerialize.spec.js` for the text-block feature, the paste-XSS fix (via a direct `ProseMirrorDOMParser` simulation of a hostile clipboard paste), and the byte-preservation fix.

##### Before 

https://github.com/user-attachments/assets/ec76502a-78c4-47ff-8f18-87a70f7ff612

##### After

https://github.com/user-attachments/assets/1c7fa1cf-f869-4a7c-bd40-72ef919d0394

#### Contribution Checklist:

- [x] **Create an issue and link to the pull request.** _(linked: [BRU-4313](https://usebruno.atlassian.net/browse/BRU-4313), [BRU-4326](https://usebruno.atlassian.net/browse/BRU-4326), [BRU-4329](https://usebruno.atlassian.net/browse/BRU-4329))_

## Summary by CodeRabbit

* **New Features**
  * Added editable support for simple raw HTML blocks, preserving tags, attributes, text, and line breaks.
  * Added support for preserving inline raw HTML in the Rich Text Editor.
  * Recognized inline tags and task-list checkboxes continue to use native editor behavior.
  * Soft-wrapped Markdown remains continuous paragraph text without forced line breaks.

* **Bug Fixes**
  * Improved Markdown round-tripping for HTML, attributes, text edits, and line breaks.
  * Unsupported or nested HTML is safely preserved without disrupting editor content.

### DIFF
--- a/packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
+++ b/packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
@@ -5,6 +5,7 @@ import runMarkdownitSetupOnce from '../../utils/markdownitSetupOnce';
 const RAW_HTML_BLOCK_ATTR = 'data-raw-html-block';
 const RAW_HTML_INLINE_ATTR = 'data-raw-html-inline';
 const TEXT_BLOCK_TAG_ATTR = 'data-raw-html-text-block';
+const ORIGINAL_HTML_ATTR = 'data-raw-html-original';
 
 // We forbid <style> in DOMPurify to prevent untrusted docs from injecting CSS that affects the whole app UI.
 const SANITIZE_CONFIG = { FORBID_TAGS: ['style'] };
@@ -19,7 +20,7 @@ const RECOGNIZED_INLINE_HTML_TAGS = new Set([
 ]);
 
 // Task list checkboxes are passed through as plain HTML since updateDOM reads them directly; all other <input> tags remain opaque raw-HTML atoms.
-const TASK_LIST_CHECKBOX_PATTERN = /^<input\b[^>]*\btype="checkbox"/i;
+const TASK_LIST_CHECKBOX_PATTERN = /^<input\b[^>]*\btype=["']checkbox["']/i;
 
 const getInlineHtmlTagName = (html) => {
   const match = html.trim().match(/^<\/?([a-zA-Z][\w-]*)/);
@@ -51,6 +52,22 @@ const NON_TEXT_BLOCK_TAGS = new Set([
 const escapeHtmlText = (text) => text.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
 
 const escapeHtmlAttrValue = (value) => escapeHtmlText(value).replace(/"/g, '&quot;');
+
+// Reads an element's attributes into a plain object, so callers can compare/serialize them without touching the DOM.
+const attributesToObject = (element, excludedNames = []) => {
+  const attrs = {};
+  [...element.attributes].forEach((attr) => {
+    if (!excludedNames.includes(attr.name)) {
+      attrs[attr.name] = attr.value;
+    }
+  });
+  return attrs;
+};
+
+const buildAttrString = (attrs) =>
+  Object.entries(attrs)
+    .map(([name, value]) => ` ${name}="${escapeHtmlAttrValue(value)}"`)
+    .join('');
 
 // Returns the top-level element if it sanitizes down to a single wrapper with only plain text and <br>s; otherwise returns null to fall back to an opaque raw-HTML atom.
 const parseSimpleTextBlockElement = (html) => {
@@ -105,11 +122,11 @@ const setupRawHtmlBlockParser = (markdownit) => {
       const tagName = textBlockElement?.tagName.toLowerCase();
 
       if (textBlockElement && !NON_TEXT_BLOCK_TAGS.has(tagName)) {
-        const attrs = [...textBlockElement.attributes]
-          .map((attr) => ` ${attr.name}="${escapeHtmlAttrValue(attr.value)}"`)
-          .join('');
+        const attrs = buildAttrString(attributesToObject(textBlockElement));
+        // The trailing newline(s) markdown-it includes at the end of an html_block token aren't part of the tag itself.
+        const originalHtml = encodeRawHtml(html.replace(/\n+$/, ''));
 
-        return `<${tagName} ${TEXT_BLOCK_TAG_ATTR}="${tagName}"${attrs}>${serializeTextBlockContent(textBlockElement)}</${tagName}>`;
+        return `<${tagName} ${TEXT_BLOCK_TAG_ATTR}="${tagName}" ${ORIGINAL_HTML_ATTR}="${originalHtml}"${attrs}>${serializeTextBlockContent(textBlockElement)}</${tagName}>`;
       }
 
       return `<div ${RAW_HTML_BLOCK_ATTR}="${encodeRawHtml(html)}"></div>`;
@@ -133,10 +150,10 @@ const setupRawHtmlInlineParser = (markdownit) => {
 };
 
 // Shared factory for raw HTML nodes, as block and inline placeholders only differ in tags, attributes, and markdown wiring.
-const createRawHtmlNode = ({ name, tag, attr, className, extra, serialize, setupParser }) =>
+const createRawHtmlNode = ({ name, tag, attr, className, schemaOptions, serialize, setupParser }) =>
   Node.create({
     name,
-    ...extra,
+    ...schemaOptions,
 
     addAttributes() {
       return {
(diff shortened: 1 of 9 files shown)


packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
```diff
@@ -16,6 +17,21 @@ const SANITIZE_CONFIG = { FORBID_TAGS: ['style'] };
 // raw HTML, so a blank line stays an editable paragraph on reload.
 const BLANK_LINE_MARKER_PATTERN = /^<br\s*\/?>$/i;
 
+// These inline tags are already consumed by something else in the pipeline
```
sachin-thakur-bruno: please refactor the comments

packages/bruno-app/src/ui/RichTextEditor/extensions/index.js
```diff
@@ -118,7 +118,15 @@ const createExtensions = ({ allowHtml = true, collectionPath = '' } = {}) => [
   }),
   Markdown.configure({
     html: allowHtml,
-    breaks: true,
+    // A lone `\n` inside a paragraph is a CommonMark soft break (renders as a
```
sachin-thakur-bruno: we don;t need this comment. maybe one line is enough

packages/bruno-app/src/ui/RichTextEditor/extensions/index.js
```diff
@@ -118,7 +118,15 @@ const createExtensions = ({ allowHtml = true, collectionPath = '' } = {}) => [
...
+    // space), not a line break. With `breaks: true`, tiptap-markdown parses
+    // it as a hardBreak node instead, and EditorHardBreak's serializer then
+    // writes that back out as an explicit `\` line-continuation — so loading
+    // and re-saving any doc with hand-wrapped prose silently turned every
+    // wrapped line into a forced break. `breaks: false` keeps soft breaks as
+    // plain reflowable text, matching how any other CommonMark renderer
+    // (e.g. GitHub) would display the same file.
+    breaks: false,
```
sachin-thakur-bruno, line 122: earlier it was true. Please check if its not breaking anything.

packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
```diff
@@ -16,6 +17,21 @@ const SANITIZE_CONFIG = { FORBID_TAGS: ['style'] };
...
+// before an unrecognized tag would need to fall back to an opaque raw-HTML
+// atom: `br` -> EditorHardBreak, `kbd`/`sup` -> EditorInlineHtmlMarks (both
+// via their own parseHTML rule matching the literal tag), and `input` -> the
+// task-list checkbox that markdown-it-task-lists renders as inline HTML and
+// editorMarkdownParse's updateDOM reads directly off the DOM. Matching by tag
+// name (not the full tag string) so this still applies regardless of the
+// attributes markdown-it-task-lists puts on a given checkbox.
+const RECOGNIZED_INLINE_HTML_TAGS = new Set(['br', 'kbd', 'sup', 'input']);
```
sachin-thakur-bruno: i think this will break existing inline tags. Please check if <b> for bold is working or not?

packages/bruno-app/src/ui/RichTextEditor/extensions/index.js
Gopi-bruno: It would fix the unnecessary back slash addition issue now, it shouldn't break anything else

sachin-thakur-bruno (approved):
changes looks good to me. Tested locally. working fine.

sachin-thakur-bruno:
Adding for a note, we need to thing about removing the hardcoded html tags. @Gopi-bruno

PR #8975, opened 2026-08-14
fix(environments): keep environment colors when mounting from the file cache

chirag-bruno:
### Problem

With the file cache enabled, an environment's color is gone the next time the collection mounts. The mount v2 tree builder rebuilds every environment from `uid`, `name` and `variables` alone, so the `color` (and `externalSecrets`) parsed from the file never reach the renderer. Colors look fine while you are editing because the watcher change event carries the whole environment.

### Fix

Carry the parsed environment onto the tree node and override only the fields derived from the file path.

Unit tests cover the color and external secrets; an e2e asserts the color badge after a mount served from the cache — without the fix the badge renders transparent.

Ticket: BRU-4287

## Bug Fixes
* Environment colors now remain visible when collections are loaded from the file cache.
* Environment names are consistently derived from their filenames.
* Additional environment settings, including external secrets, are preserved when environments are loaded.

## Tests
* Added coverage for environment properties, color persistence, and file-cached collections.
* Added test support for verifying environment color badges.

### DIFF
--- a/packages/bruno-electron/src/services/mount/tree-builder.js
+++ b/packages/bruno-electron/src/services/mount/tree-builder.js
@@ -147,6 +147,7 @@ const buildEnvironmentNode = (collectionPath, relativePath, entry, uidFor) => {
   const absolutePath = path.join(collectionPath, relativePath);
   const data = entry.data || {};
   return {
+    ...data,
     uid: uidFor(absolutePath),
     name: stripExt(basename),
     variables: data.variables || [],
--- a/tests/environments/color-file-cache/env-color-file-cache.spec.ts
+++ b/tests/environments/color-file-cache/env-color-file-cache.spec.ts
@@ -80,12 +80,10 @@ test.describe('[file-cache on] Environment colors', () => {
   });
 
   test('survive a mount served from the file cache', async () => {
-    const environments = buildCommonLocators(page).environment;
-
-    await page.locator('#sidebar-collection-name').filter({ hasText: COLLECTION_NAME }).click();
+    const locators = buildCommonLocators(page);
+    await locators.sidebar.collection(COLLECTION_NAME).click();
     await openEnvironmentSelector(page);
-
-    const colorBadge = environments.listItem(ENVIRONMENT_NAME).locator('.rounded-full').first();
+    const colorBadge = locators.environment.listOptionBadge(ENVIRONMENT_NAME).first();
     await expect(colorBadge).toHaveCSS('background-color', COLOR_RGB);
   });
 });
--- a/tests/environments/color-file-cache/env-color-file-cache.spec.ts
+++ b/tests/environments/color-file-cache/env-color-file-cache.spec.ts
@@ -1,85 +1,13 @@
-import * as os from 'os';
-import * as path from 'path';
-import * as fs from 'fs';
-import { test, expect, ElectronApplication, Page, waitForReadyPage, closeElectronApp } from '../../../playwright';
+import { test, expect } from '../../../playwright';
 import { buildCommonLocators } from '../../utils/page/locators';
 import { openEnvironmentSelector } from '../../utils/page/environments';
 
 const COLLECTION_NAME = 'Env Color Collection';
 const ENVIRONMENT_NAME = 'Local';
-const COLOR = '#CE4F3B';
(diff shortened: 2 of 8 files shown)


tests/environments/color-file-cache/env-color-file-cache.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  test.afterAll(async () => {
+    if (app) {
+      await closeElectronApp(app);
+    }
+    if (fixture) {
+      await fixture.cleanup();
+    }
+  });
+
+  test('survive a mount served from the file cache', async () => {
```
sachin-thakur-bruno: we should do a second mount (restartApp) to test?

tests/environments/color-file-cache/env-color-file-cache.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    const environments = buildCommonLocators(page).environment;
+
+    await page.locator('#sidebar-collection-name').filter({ hasText: COLLECTION_NAME }).click();
```
sachin-thakur-bruno: I think we have a locator for this. Please check

packages/bruno-electron/src/services/mount/tree-builder.js
```diff
@@ -146,6 +146,7 @@ const buildEnvironmentNode = (collectionPath, relativePath, entry, uidFor) => {
   const absolutePath = path.join(collectionPath, relativePath);
   const data = entry.data || {};
   return {
+    ...data,
```
sachin-thakur-bruno, line 150: do we need whole data? can;t we just pick color? `color: data.color,`

tests/environments/color-file-cache/env-color-file-cache.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    await openEnvironmentSelector(page);
+
+    const colorBadge = environments.listItem(ENVIRONMENT_NAME).locator('.rounded-full').first();
```
sachin-thakur-bruno: can we please avoid using the class selector?

tests/environments/color-file-cache/env-color-file-cache.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+const COLOR = '#CE4F3B';
+const COLOR_RGB = 'rgb(206, 79, 59)';
+
+interface Fixture {
+  userDataPath: string;
+  collectionPath: string;
+  cleanup: () => Promise<void>;
+}
+
+const buildFixture = async (): Promise<Fixture> => {
```
sachin-thakur-bruno: can we use pageWithUserData instead ??

tests/environments/color-file-cache/env-color-file-cache.spec.ts
chirag-bruno: Updated this.

packages/bruno-electron/src/services/mount/tree-builder.js
chirag-bruno: This is to ensure that all the other data points that might get added to the env node gets carried without explicit change. A change would only be needed if the key differs, else this works fine. This is just a safety measure to not drop any key:value pairs.

tests/environments/color-file-cache/env-color-file-cache.spec.ts
chirag-bruno: Done. This is changed

tests/environments/color-file-cache/env-color-file-cache.spec.ts
chirag-bruno: Not really. Since we are using the cache while the app starts, the source of the tree is the DB, even on the first run.

tests/environments/color-file-cache/env-color-file-cache.spec.ts
chirag-bruno: We could, but that would mean that I drop the fixture generation and have static fixtures. Will do that.

sachin-thakur-bruno (dismissed):
LGTM

sachin-thakur-bruno (approved):

PR #8980, opened 2026-08-14
fix (sqlite): updated the createDatabase method to handle failures

chirag-bruno:
Earlier the app would fail to open if the OS denied access to the DB file. The first fix here was to fall back to an in-memory DB whenever setup failed, but that turned out to be too blunt. A bad migration or a corrupt file would get quietly papered over with a throwaway DB, and the user would carry on with a store that disappears on quit.

So instead of one catch-all fallback, we now look at what actually broke:

- **The file cannot be opened** (permissions, missing folder): in-memory fallback, same as before. The app starts, just without persistence.
- **The file opens but the migrations fail**: the file is the problem, not us. This happens with a corrupt DB, one written by a newer version of Bruno, or a `down` that no longer applies. We move it into `sqlite-backup/` with a timestamp on it, then create a fresh file and migrate that instead.
- **The rebuild fails too**: our own migrations are broken, so renaming the file a second time will not help anyone. We log it and fall back to in-memory.
- **A single statement will not prepare**: drop only that statement, instead of letting it take the whole store down with it.

A few notes on the details:

- Nothing in this DB is authored by the user, it is all derived state, so rebuilding it is safe. We keep the old file rather than deleting it, so we can still look at it if someone reports something odd.
- The journal and WAL are moved along with the DB file, because sqlite pairs those to a DB by filename and a leftover one would otherwise get replayed into the fresh file.
- Migration errors are tagged with a `Symbol.for` brand rather than checked with `instanceof`. We ship separate CJS and ESM bundles and each inlines its own copy of the class, so `instanceof` can come back false for what is clearly the same error. Moving the user's file aside is the destructive branch, so it only runs on a positive match, and anything we cannot classify goes to the in-memory fallback instead of renaming data.
- Executing a dropped statement now tells you it could not be prepared, instead of the old and fairly confusing "Unknown statement".
- `npm run migration:verify` now prepares every generated statement against the migrated schema and lists all of the failures in one run, so a statement pointing at a column that no migration creates fails there rather than at runtime.

Ticket: BRU-4300

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database startup reliability by safely handling initialization and migration failures.
  * Added automatic in-memory fallback when a database file cannot be opened or recovered.
  * Preserved failed database files through backup before migration recovery.
  * Prevented cleanup errors when database setup is incomplete.
  * Improved error messages for unavailable or invalid SQL statements.
  * Allowed valid statements to remain usable when another statement fails to prepare.
  * Added clearer warnings and error reporting when database creation or migrations fail.

### DIFF
--- a/packages/bruno-sqlite/src/node/index.ts
+++ b/packages/bruno-sqlite/src/node/index.ts
@@ -1,10 +1,10 @@
-import { DB } from './db';
-import type { DatabaseOptions } from './db';
-import { Statements } from './statements';
-import type { OnMutation } from './statements';
+import { existsSync, mkdirSync, renameSync } from 'node:fs';
+import { basename, dirname, join } from 'node:path';
+import { DB, DatabaseOptions, isDatabaseMigrationError } from './db';
+import { Statements, OnMutation } from './statements';
 import { migrations } from '../generated/node/migrations';
 
-export { DB } from './db';
+export { DB, DatabaseMigrationError, isDatabaseMigrationError } from './db';
 export type { DatabaseOptions } from './db';
 export { Statements } from './statements';
 export type { OnMutation } from './statements';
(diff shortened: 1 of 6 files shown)


packages/bruno-sqlite/src/node/index.ts
```diff
@@ -18,12 +16,28 @@ export type CreateDatabaseOptions = DatabaseOptions & {
...
+    console.warn('failed to open the database file: ', err);
+  };
+
+  try {
+    db = new DB(':memory:', migrations, dbOptions);
+    statements = new Statements(db._db!, onMutation);
+    return { db, statements };
+  } catch (err) {
+    console.error('failed to open in-memory database: ', err);
   }
```
sid-bruno, line 52: I know I asked for this to be the solution but instead can we just let the app run without cache since the DB is not opening up, we can hold a log of some kind in the app data path for why the DB open failed but if we end up caching everything in memory for a large set of collections that wouldn't be optimal

packages/bruno-sqlite/src/node/index.ts
```diff
@@ -18,12 +16,28 @@ export type CreateDatabaseOptions = DatabaseOptions & {
...
-  const statements = new Statements(db._db, onMutation);
+
   return { db, statements };
```
sachin-thakur-bruno: On fail both these will be undefined. 
In sqlite.js we are passing the `statements` to registerSQLiteIpc. Should we guard it?

packages/bruno-sqlite/src/node/index.ts
```diff
@@ -18,12 +16,28 @@ export type CreateDatabaseOptions = DatabaseOptions & {
...
-  if (db._db === undefined) {
-    throw new Error('Failed to open the database.');
+  let db: DB | undefined = undefined;
+  let statements: Statements | undefined = undefined;
+  try {
+    db = new DB(path, migrations, dbOptions);
+    statements = new Statements(db._db!, onMutation);
+    return { db, statements };
+  } catch (err) {
+    console.warn('failed to open the database file: ', err);
```
sachin-thakur-bruno: should we close the db connection? error might be from statements.

packages/bruno-sqlite/src/node/index.ts
chirag-bruno: Yes, the issue there is that anybody using should be mindful that their feature should work without DB. Memory fallback gives us ability to use DB, even if file system permission fails. I understand that keeping things in memory is not optimal but failing on FS permissions and having no DB handle could be a much bigger failure point.

sachin-thakur-bruno (dismissed):

sid-bruno (approved):

PR #8984, opened 2026-08-14
fix(oauth2): resolve token to a string before header injection

### LINKED ISSUE #8940: OAuth 2.0 Auto-Injection Sends "[object Object]" Leading to 401 Unauthorized

### This bug is:

- [ ] Annoying.
- [x] This feature was working in a previous version but is broken in the current release.

### Bruno version

4.0.0

### 💻 Operating System

macOS

Tahoe 26.5.2

### 🐞 Describe the bug

After updating the Bruno GUI client to v4.0.0, making requests using OAuth 2.0 authorization results in a 401 Unauthorized error. While the GUI successfully fetches, displays, and refreshes the access token inside the Auth panel, the underlying core engine fails to stringify or resolve the token object during the actual HTTP request dispatch. Instead of injecting the plain-text token string into the Authorization header, Bruno evaluates the variable as a raw JavaScript object, transmitting the literal text "[object Object]" to the backend server.

### 📝 Steps to reproduce

.

### 📁 Collection to reproduce

_No response_

Image

YuriNachos:
### Description

OAuth2 token injection now resolves the token to a string before placing it into the `Authorization` header (or the request URL), so a non-string token can never reach the wire as the literal `[object Object]`.

### Problem

Closes #8940. With OAuth 2.0 client credentials (AWS Cognito) on Bruno 4.0.0, requests come back 401 because the outgoing `Authorization` header carries the literal string `[object Object]` instead of the token. The Auth panel itself fetches and displays the token fine — only the request engine's injection is broken.

### Fix

Every injection site interpolated the token into a template string without asserting that it resolved to a string:

```js
const tokenValue = tokenSource === 'id_token' ? credentials?.id_token : credentials?.access_token;
request.headers['Authorization'] = `${tokenHeaderPrefix} ${tokenValue}`.trim();
```

`credentials` is the parsed token-endpoint response (and the cached copy of it) — external data. Whenever `access_token`/`id_token` resolves to a non-string (a non-conforming provider response such as `{"access_token": {...}}`, or a corrupted cached credential), JS coercion stringified it to `[object Object]`. The same unguarded interpolation also produced `Bearer undefined` on the WebSocket and gRPC paths whenever no token resolved, and `undefined <token>` / `null <token>` when the header prefix was missing (a `.bru`/`.yml` round-trip can leave it `null`).

This PR adds two helpers in `packages/bruno-electron/src/utils/oauth2.js`:

- `resolveOAuth2TokenValue(credentials, tokenSource)` — resolves the configured token to a trimmed string: a string token is trimmed; an object holding a string `.token`/`.value` property (non-conforming provider shapes) is unwrapped; anything else resolves to `undefined`.
- `applyOAuth2TokenToRequest(request, { credentials, tokenPlacement, tokenHeaderPrefix, tokenQueryKey, tokenSource })` — resolves the token and places it in the `Authorization` header or as a query param; when the token cannot be resolved to a string, injection is skipped entirely so nothing degenerate is sent.

All ten former injection sites — 4 HTTP grant types in `ipc/network/index.js`, 3 WebSocket grant types in `ws-event-handlers.js`, and 3 gRPC grant types in `prepare-grpc-request.js` (which had its own local copy of the logic) — now go through the shared helper. Behavior for well-formed tokens is unchanged. Related fixes surfaced while wiring the sites through the helper:

- The WebSocket query-placement was a **dead write**: it mutated the item's `request.url` while the connection uses `wsRequest.url` (snapshotted earlier), so a query-placed token never reached a WS connection. The token now lands on the URL the connection actually uses, and it is placed **after** `interpolateVars` — `new URL()` percent-encodes raw `{{var}}` templates beyond the interpolator's reach — matching the HTTP and gRPC paths, which already interpolate first.
- `tokenSource` (Access Token / ID Token) is now honored on WebSocket and gRPC too; those paths previously always sent `access_token` even when `id_token` was configured.
- A non-string, non-unwrappable token (e.g. a JSON number) now skips injection entirely instead of being coerced onto the wire — consistent with the fix's invariant that only a real string token is ever sent.

### Screenshots

No UI change.

| Before | After |
| --- | --- |
| `Authorization: Bearer [object Object]` → 401 | `Authorization: Bearer <token>` |

#### Verification

- `npm run test:ci --workspace=packages/bruno-electron` — 54 suites, 784 passed, including the new `tests/network/oauth2-token-injection.spec.js` (36 tests): the helper matrix, integration through `configureRequest` for all four grant types, and through `prepareWsRequest` (header, query, templated-url interpolation order, id_token source, no-token) — asserting that garbage token shapes (nested object, number, null, missing) never produce `[object Object]` in the header or URL.
- `npm run lint` — 0 errors.
- End-to-end against a mock provider returning `{"access_token":{"token":"…"}}`: pre-fix the header was `Bearer [object Object]`; post-fix it is `Bearer <unwrapped token>`, and an unwrappable object leaves the header unset.

#### Contribution Checklist:

- [x] **Create an issue and link to the pull request.** (#8940)

## Summary by CodeRabbit

- **New Features**
  - OAuth2 tokens are now applied consistently across HTTP, gRPC, and WebSocket requests.
  - Supports placing tokens in authorization headers or URL query parameters.
  - Supports configurable token sources, header prefixes, and query parameter names.
  - Handles tokens returned in multiple credential formats.

- **Bug Fixes**
  - Improved OAuth2 handling for interpolated URLs and WebSocket requests.
  - Prevents invalid, missing, or stringified credential values from being added to requests.
  - Safely ignores malformed URLs during query-parameter token injection.

### DIFF
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -28,7 +28,7 @@ const { chooseFileToSave, writeFile, getCollectionFormat, hasRequestExtension } 
 const { addCookieToJar, getDomainsWithCookies, getCookieStringForUrl } = require('../../utils/cookies');
 const { createFormData } = require('../../utils/form-data');
 const { findItemInCollectionByPathname, sortFolder, getAllRequestsInFolderRecursively, getEnvVars, getTreePathFromCollectionToItem, mergeVars, sortByNameThenSequence } = require('../../utils/collection');
-const { getOAuth2TokenUsingAuthorizationCode, getOAuth2TokenUsingClientCredentials, getOAuth2TokenUsingPasswordCredentials, getOAuth2TokenUsingImplicitGrant, updateCollectionOauth2Credentials, clearOauth2CredentialsByCredentialsId } = require('../../utils/oauth2');
+const { getOAuth2TokenUsingAuthorizationCode, getOAuth2TokenUsingClientCredentials, getOAuth2TokenUsingPasswordCredentials, getOAuth2TokenUsingImplicitGrant, updateCollectionOauth2Credentials, clearOauth2CredentialsByCredentialsId, applyOAuth2TokenToRequest } = require('../../utils/oauth2');
 const { preferencesUtil } = require('../../store/preferences');
 const { getProcessEnvVars } = require('../../store/process-env');
 const { getBrunoConfig } = require('../../store/bruno-config');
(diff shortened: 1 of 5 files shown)


YuriNachos:
All three review points addressed in 9ab03a6f0:

- **WS prompt variables** — `prepareWsRequest` now forwards `collection.promptVariables` to the final `interpolateVars` call, matching the HTTP path. Added a regression test with a `{{?roomId}}` prompt variable in the WebSocket URL (resolves before the token is appended to the query).
- **gRPC coverage** — new test driving `configureRequest` end-to-end: a nested `{token: "…"}` provider shape is unwrapped and lands as `Bearer …` in the gRPC metadata, never `[object Object]`.
- **OS-neutral temp paths** — the spec now uses `os.tmpdir()`/`path.join()` instead of hardcoded `/tmp`.

(Also dropped an unused `path` import the first commit had introduced.) Suite: 38/38 in the oauth2 spec; the two remaining workspace failures (axios-instance getSentHeaders, fetch-gql-schema) fail identically on current main — pre-existing, untouched by this PR.

PR #9011, opened 2026-08-18
fix(docs): improve raw HTML block rendering and styling security 

Gopi-bruno:
### Description

Follow-up hardening for the raw-HTML passthrough in the Docs rich text editor (introduced in #8966): fixes three regressions in how a raw/plain-text HTML block is parsed, rendered and serialized back to markdown, plus a style-injection gap. Scoped entirely to `RichTextEditor`'s raw-HTML extension — no changes to the request/response pipeline.

### Problem

Four issues in `EditorRawHtmlBlock` / list-item handling, found via manual review and confirmed by reproducing each one against the editor directly (parse → live DOM → serialize):

1. **List item content silently dropped on save.** A list item whose only content was a block-level raw HTML tag (e.g. a `<div>` as a second block inside a "loose" list item) had its tag and attributes stripped down to bare text on markdown serialization, because the shared inline-serialization helper treated any text-only block — including the raw HTML text block — the same as a plain paragraph.
2. **List item content ejected from the list entirely.** Separately, `listItem`/`taskItem` required their first child to be a `paragraph` (tiptap's default `content: 'paragraph block*'`). A list item whose only content was a raw HTML block (text or opaque) couldn't satisfy that during parsing, so ProseMirror synthesized an empty paragraph in its place and pushed the actual raw HTML block out as a sibling after the whole list.
3. **Whitespace-significant content mangled on save.** The "was this block edited?" check compared the live text against the original bytes without accounting for ProseMirror's whitespace collapsing, so multi-line raw HTML (e.g. indented `<div>` content) had its original formatting rewritten on the very first save — not just after a later edit. `<pre>` was also missing from the set of tags that must stay opaque, so whitespace-significant `<pre>` blocks could be promoted to an editable node and have their formatting collapsed.
4. **Style-based UI takeover.** A `style` attribute on raw HTML (e.g. `position:fixed;width:100vw;height:100vh;z-index:99999`) wasn't stripped by the sanitizer, so a hostile or compromised collection's docs could render a full-viewport overlay inside the app.

None of these affect request execution, environments, or scripts — they're isolated to markdown docs content (collection/folder/request `docs`).

### Fix

- `EditorTaskList`: loosened `listItem`/`taskItem` content model from `paragraph block*` to `block+` so a non-paragraph block can lead a list item without being ejected.
- `editorMarkdownSerialize.js`: scoped the "inline it directly" fast path to actual `paragraph` nodes only; every other text-only block (raw HTML included) now goes through the tag-preserving fallback.
- `EditorRawHtmlBlock`: normalize whitespace on the *original* bytes (not the live text) before the unedited-comparison, so unedited whitespace-significant content round-trips byte-for-byte from the first save; added `pre` to the set of tags that always stay opaque.
- `SANITIZE_CONFIG`: added `FORBID_ATTR: ['style']` so DOMPurify strips the `style` attribute from raw HTML; also added `contain: paint` on the editor's scroll wrapper as a second layer of defense against fixed/absolute-positioned content escaping the panel bounds.

### Testing

- Unit tests (`editorMarkdownSerialize.spec.js`): added/updated cases for each of the four issues above, run against the production extension set (`createExtensions()`), not a filtered subset — 61/61 passing.
- E2E (`tests/richtext-editor/raw-html-passthrough.spec.ts`): new test verifying a bare `<div>` and a bare `<video>` as the sole content of a list item stay nested under the list and round-trip through markdown, instead of being ejected.
- Verified each fix by reproducing the bug against the unfixed code first (via `git stash`) and confirming the same repro passes only with the fix applied.

#### Contribution Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved rich-text editor handling of raw HTML, including nested blocks in lists and tables.
  - Prevented positioned HTML content from escaping the editor panel.
  - Preserved original HTML more reliably when changes are limited to whitespace.
  - Kept videos, divs, and other block content correctly nested within list items.
  - Improved handling of preformatted content and unsafe style elements.
  - Improved Markdown conversion for complex lists and tables.
  - Ensured keyboard saves use the latest documentation content.
  - Improved documentation settings when no folder is selected.

- **Tests**
  - Added coverage for raw HTML, Markdown conversion, saving, and folder-state handling.

### DIFF
--- a/packages/bruno-app/src/components/Documentation/DocsEditor/index.js
+++ b/packages/bruno-app/src/components/Documentation/DocsEditor/index.js
@@ -44,12 +44,20 @@ const DocsEditor = ({
   const lastEmittedDocsRef = useRef(DOCS_NOT_YET_EMITTED);
   const isMarkdownModeRef = useRef(isMarkdownMode);
   const isEditingRef = useRef(isEditing);
-  // Plain assignment during render — these refs exist only so the onUpdate
-  // closure below can read the latest mode/editing flags without being
-  // recreated on every change; there's no external system to synchronize
-  // with, so an effect would just add an unnecessary extra render pass.
+  const onSaveRef = useRef(onSave);
+  // Plain assignment during render — these refs exist only so the onUpdate/
+  // handleKeyDown closures below can read the latest mode/editing flags and
+  // onSave without being recreated on every change; there's no external
+  // system to synchronize with, so an effect would just add an unnecessary
+  // extra render pass. onSave specifically needs this because useEditor's
+  // options (including editorProps.handleKeyDown) are only synced into the
+  // live editor when a dep in [collectionPath] changes — a re-render driven
+  // by the parent handing down a new onSave (e.g. after an edit updates its
+  // local docs state) does not, so a stale ref here would silently save
+  // pre-edit content on Ctrl+S.
   isMarkdownModeRef.current = isMarkdownMode;
   isEditingRef.current = isEditing;
+  onSaveRef.current = onSave;
 
   const editor = useEditor(
     {
@@ -71,7 +79,7 @@ const DocsEditor = ({
         handleKeyDown: (_view, event) => {
           if (event.key === 's' && (event.ctrlKey || event.metaKey)) {
             event.preventDefault();
-            onSave();
+            onSaveRef.current();
             return true;
           }
           return false;
--- a/packages/bruno-app/src/components/Documentation/DocsEditor/index.js
+++ b/packages/bruno-app/src/components/Documentation/DocsEditor/index.js
@@ -45,16 +45,7 @@ const DocsEditor = ({
   const isMarkdownModeRef = useRef(isMarkdownMode);
   const isEditingRef = useRef(isEditing);
   const onSaveRef = useRef(onSave);
-  // Plain assignment during render — these refs exist only so the onUpdate/
-  // handleKeyDown closures below can read the latest mode/editing flags and
-  // onSave without being recreated on every change; there's no external
-  // system to synchronize with, so an effect would just add an unnecessary
-  // extra render pass. onSave specifically needs this because useEditor's
-  // options (including editorProps.handleKeyDown) are only synced into the
-  // live editor when a dep in [collectionPath] changes — a re-render driven
-  // by the parent handing down a new onSave (e.g. after an edit updates its
-  // local docs state) does not, so a stale ref here would silently save
-  // pre-edit content on Ctrl+S.
+  // Refs so onUpdate/handleKeyDown stay fresh, useEditor only re-syncs on [collectionPath] changes.
   isMarkdownModeRef.current = isMarkdownMode;
   isEditingRef.current = isEditing;
   onSaveRef.current = onSave;
--- a/packages/bruno-app/src/ui/RichTextEditor/components/StyledWrapper.js
+++ b/packages/bruno-app/src/ui/RichTextEditor/components/StyledWrapper.js
@@ -7,6 +7,8 @@ const StyledWrapper = styled.div`
     min-height: 0;
     overflow-y: auto;
     position: relative;
+    /* Uses paint containment to prevent untrusted raw HTML from using fixed/absolute positioning to escape the panel bounds. */
+    contain: paint;
   }
 
   .tiptap {
--- a/packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
+++ b/packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
@@ -7,8 +7,9 @@ const RAW_HTML_INLINE_ATTR = 'data-raw-html-inline';
 const TEXT_BLOCK_TAG_ATTR = 'data-raw-html-text-block';
 const ORIGINAL_HTML_ATTR = 'data-raw-html-original';
 
-// We forbid <style> in DOMPurify to prevent untrusted docs from injecting CSS that affects the whole app UI.
-const SANITIZE_CONFIG = { FORBID_TAGS: ['style'] };
+// We forbid <style> and the style attribute in DOMPurify to prevent untrusted docs from injecting CSS
+// (e.g. position:fixed;width:100vw;height:100vh) that affects the whole app UI.
+const SANITIZE_CONFIG = { FORBID_TAGS: ['style'], FORBID_ATTR: ['style'] };
 
 // A standalone <br/> marks an empty paragraph; treating it as a hard break ensures blank lines remain editable paragraphs.
 const BLANK_LINE_MARKER_PATTERN = /^<br\s*\/?>$/i;
@@ -50,6 +51,9 @@ const NON_TEXT_BLOCK_TAGS = new Set([
 ]);
 
 const escapeHtmlText = (text) => text.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
+
+// Mimics ProseMirror's whitespace collapse to correct the original bytes for the "unedited" comparison below.
+const normalizeWhitespace = (text) => text.replace(/\s+/g, ' ').trim();
 
 const escapeHtmlAttrValue = (value) => escapeHtmlText(value).replace(/"/g, '&quot;');
 
@@ -349,8 +353,10 @@ export const EditorRawHtmlTextBlock = Node.create({
             inner += child.type.name === 'hardBreak' ? '<br>' : escapeHtmlText(child.text || '');
           });
 
+          // Normalizes only the original to offset parse-time whitespace collapse; comparing against unmodified 'inner' ensures real edits remain detectable.
           const originalElement = originalHtml ? parseSimpleTextBlockElement(originalHtml) : null;
-          const isUnedited = originalElement && serializeTextBlockContent(originalElement) === inner;
+          const isUnedited = originalElement
+            && normalizeWhitespace(serializeTextBlockContent(originalElement)) === inner;
 
           state.write(isUnedited ? originalHtml : `<${tag}${buildAttrString(htmlAttrs)}>${inner}</${tag}>`);
           state.closeBlock(node);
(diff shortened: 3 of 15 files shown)


packages/bruno-app/src/ui/RichTextEditor/components/StyledWrapper.js
```diff
@@ -7,6 +7,8 @@ const StyledWrapper = styled.div`
     min-height: 0;
     overflow-y: auto;
     position: relative;
+    /* Uses paint containment to prevent untrusted raw HTML from using fixed/absolute positioning to escape the panel bounds. */
+    contain: paint;
```
sachin-thakur-bruno, line 11: what was the problem without this?

packages/bruno-app/src/ui/RichTextEditor/utils/editorMarkdownSerialize.spec.js
```diff
@@ -601,5 +601,128 @@ describe('Editor markdown serialization', () => {
...
+      let rawHtmlBlockCount = 0;
+      let rawHtmlTextBlockCount = 0;
+      editor.state.doc.descendants((node) => {
+        if (node.type.name === 'rawHtmlBlock') rawHtmlBlockCount += 1;
+        if (node.type.name === 'rawHtmlTextBlock') rawHtmlTextBlockCount += 1;
+      });
+
+      expect(rawHtmlBlockCount).toBe(1);
+      expect(rawHtmlTextBlockCount).toBe(0);
+      expect(getMarkdown(editor)).toBe(source);
```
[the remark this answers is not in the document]
Gopi-bruno: It passes even with the `pre` in `rawHtmlBlock `.

packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
```diff
@@ -7,8 +7,9 @@ const RAW_HTML_INLINE_ATTR = 'data-raw-html-inline';
 const TEXT_BLOCK_TAG_ATTR = 'data-raw-html-text-block';
 const ORIGINAL_HTML_ATTR = 'data-raw-html-original';
 
-// We forbid <style> in DOMPurify to prevent untrusted docs from injecting CSS that affects the whole app UI.
-const SANITIZE_CONFIG = { FORBID_TAGS: ['style'] };
+// We forbid <style> and the style attribute in DOMPurify to prevent untrusted docs from injecting CSS
+// (e.g. position:fixed;width:100vw;height:100vh) that affects the whole app UI.
+const SANITIZE_CONFIG = { FORBID_TAGS: ['style'], FORBID_ATTR: ['style'] };
```
sachin-thakur-bruno: will this drop the style attribute from existing docs?

packages/bruno-app/src/ui/RichTextEditor/components/StyledWrapper.js
Gopi-bruno: Without `contain: paint`, any raw HTML block (e.g., from a pasted or maliciously crafted markdown document) that includes inline CSS like `position: fixed` or `position: absolute` would be positioned relative to the browser viewport.

Gopi-bruno: Since we're not strpping class attribute and by chance if the snippet uses tailwind and properties like
```fixed inset-0 w-screen h-screen z-50 bg-red-500 bg-opacity-50 backdrop-blur flex items-center justify-center pointer-events-auto text-4xl text-white font-bold``` will take over the whole app, this `contain: paint` would prevent that.

File: packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js (hunk 3 above)
Comment (Gopi-bruno) [reply packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js]: Yes, it'll drop them from the existing docs.

File: packages/bruno-app/src/ui/RichTextEditor/utils/editorMarkdownSerialize.js [hunk 4]
```diff
@@ -44,6 +51,20 @@ const flattenListItemParagraphs = (listNode) => {
   return entries;
 };
 
+// Safely inlines plain paragraphs; returns true if handled, or false to preserve custom tags/attrs via state.render fallback.
+const renderInlineParagraph = (state, block) => {
+  if (!block.isTextblock || block.type.name !== 'paragraph') {
```
sachin-thakur-bruno, line 54: what happens to elements other than 'paragraph' ? I guess this will break if we have heading.

packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
sachin-thakur-bruno: wouldn't that be a problem?

packages/bruno-app/src/components/Documentation/DocsEditor/index.js
```diff
@@ -44,12 +44,20 @@ const DocsEditor = ({
...
+  // system to synchronize with, so an effect would just add an unnecessary
+  // extra render pass. onSave specifically needs this because useEditor's
+  // options (including editorProps.handleKeyDown) are only synced into the
+  // live editor when a dep in [collectionPath] changes — a re-render driven
+  // by the parent handing down a new onSave (e.g. after an edit updates its
+  // local docs state) does not, so a stale ref here would silently save
+  // pre-edit content on Ctrl+S.
   isMarkdownModeRef.current = isMarkdownMode;
   isEditingRef.current = isEditing;
+  onSaveRef.current = onSave;
```
sachin-thakur-bruno, line 51: why we need a ref for this?

packages/bruno-app/src/components/Documentation/DocsEditor/index.js
```diff
@@ -44,12 +44,20 @@ const DocsEditor = ({
   const lastEmittedDocsRef = useRef(DOCS_NOT_YET_EMITTED);
   const isMarkdownModeRef = useRef(isMarkdownMode);
   const isEditingRef = useRef(isEditing);
-  // Plain assignment during render — these refs exist only so the onUpdate
-  // closure below can read the latest mode/editing flags without being
-  // recreated on every change; there's no external system to synchronize
-  // with, so an effect would just add an unnecessary extra render pass.
+  const onSaveRef = useRef(onSave);
+  // Plain assignment during render — these refs exist only so the onUpdate/
```
sachin-thakur-bruno: do we need this lengthy comment? Please make it readable

packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
Gopi-bruno: In that case, we should stop stripping the attributes and rely on the `contain: paint` but the docs area will be affected because of those styles. I'll revert this if we're ready for that trade-off.

Gopi-bruno: I'll continue to strip off the tags though

sachin-thakur-bruno: okay, its fine to strip the style attribute also. Should we strip class also?

Gopi-bruno: It contains/prevents any styles from markdown from overtaking the whole app, and confines them to the Docs container

packages/bruno-app/src/components/Documentation/DocsEditor/index.js
Gopi-bruno: shortened

packages/bruno-app/src/ui/RichTextEditor/components/StyledWrapper.js
Gopi-bruno: Resolving for now

packages/bruno-app/src/ui/RichTextEditor/extensions/EditorRawHtmlBlock/index.js
Gopi-bruno: Resolving this for now

packages/bruno-app/src/ui/RichTextEditor/utils/editorMarkdownSerialize.js
Gopi-bruno: Verified — it doesn't break. `renderInlineParagraph` only fast-paths actual `paragraph` nodes; a heading falls through to `state.render(block, ...)`, which uses heading's own registered markdown serializer (`## text`) on its own line. List items are naturally multi-line (each block gets its own indented line via `wrapBlock`), so this round-trips fine:

`<li><p>first</p><h2>a heading</h2></li>` → `- first  \n  ## a heading` → reparses back to the same two blocks.

This is different from the table-cell case, where a row has to stay single-line — that's the actual bug (heading/codeBlock in a cell corrupting the row), now fixed in fc7d279e4.

packages/bruno-app/src/components/Documentation/DocsEditor/index.js
Gopi-bruno: `useEditor`'s options (including `editorProps.handleKeyDown`) only get re-synced into the live ProseMirror view when a dependency in `[collectionPath]` changes — a re-render driven purely by new props (like a fresh `onSave` after `WorkspaceDocs` updates `localDocs`) doesn't trigger that resync. Confirmed this with a reproduction: without the ref, Ctrl+S kept calling the very first render's `onSave`, saving stale pre-edit content. The ref (matching the existing `isMarkdownModeRef`/`isEditingRef` pattern already in this file) keeps it fresh without touching `useEditor`'s deps array.

packages/bruno-app/src/ui/RichTextEditor/components/StyledWrapper.js
Gopi-bruno: Confirmed this actually works — tested `contain: paint` against a real `position: fixed; width: 100vw; height: 100vh` element in an actual browser (not just reading the spec). The hostile element was fully clipped to the container's bounds; content below the container stayed untouched.

That's also why we ended up reverting `FORBID_ATTR: ['style']` in a follow-up (b6738ac4a) — containment alone was enough, so we didn't need to strip the style attribute and silently break existing docs' styling.

packages/bruno-app/src/ui/RichTextEditor/utils/editorMarkdownSerialize.js
```diff
@@ -150,10 +168,7 @@ const serializeInlineBlocks = (state, parent, separator = '<br/>') => {
 
     first = false;
 
-    if (block.isTextblock) {
-      if (block.textContent.length || block.content.size > 0) {
-        state.renderInline(block);
-      }
+    if (renderInlineParagraph(state, block)) {
```
sachin-thakur-bruno, line 171: Table is broken now. Its adding <br > to in table. 

Image

PR #9043, opened 2026-08-20
feat(dynamic-variables): generate UUID v7 for $randomUUID

### LINKED ISSUE #9040: update the inbuilt dynamic variable {{$randomUUID}} to uuid version 7

### This feature

- [ ] Blocks me from using Bruno
- [x] Would improve my quality of life in Bruno
- [ ] Is something I've never seen an API client do before

### 🚀 Describe the feature you want to add, and how it would change your usage of Bruno

It would be niceto support the latest version of uuid. Im using websockets and im not able to script uuid version 7 via fakerjs or the uuid library. Sticking to {{$randomUUID}} is my only choice and its uuid version 4. Updating to version 7 would not be a problem since the length of the uuid will be the same.

### 📷 Mockups or Images of the feature

_No response_

Max-Reisinger:
### Description

Updates the built-in `{{$randomUUID}}` dynamic variable to generate UUID version 7 values.

### Problem

`{{$randomUUID}}` currently uses `faker.string.uuid()`, which generates UUID v4 values. This prevents users from using a time-ordered UUID v7 through the built-in dynamic variable, including in WebSocket messages.

Fixes #9040

### Fix

- Added `uuid@^10.0.0` and its TypeScript definitions to the `bruno-common` build dependencies.
- Replaced the `faker.string.uuid()` call for `randomUUID` with `uuid.v7()`.
- Updated the existing dynamic-variable test to validate the UUID v7 version and variant bits.
- Kept the existing `{{$guid}}` behavior unchanged.

### Testing

- `npm test --workspace=packages/bruno-common -- --runInBand`
- `npm run lint -- packages/bruno-common/src/utils/faker-functions.ts packages/bruno-common/src/utils/faker-functions.spec.ts`
- `npm run build --workspace=packages/bruno-common`
- Verified the built CommonJS artifact generates a UUID reported as version 7.

### Screenshots

Not applicable. This change does not modify the UI.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Enhancements**
  * Updated generated random UUID values to use the newer UUIDv7 format.
  * Improved UUID validation to ensure generated values follow the expected standard.

### DIFF
--- a/packages/bruno-common/package.json
+++ b/packages/bruno-common/package.json
@@ -47,6 +47,7 @@
     "@rollup/plugin-terser": "^1.0.0",
     "@rollup/plugin-typescript": "^12.1.2",
     "@types/jest": "^29.5.14",
+    "@types/uuid": "^10.0.0",
     "babel-jest": "^29.7.0",
     "form-data": "4.0.4",
     "is-ip": "^5.0.1",
@@ -54,7 +55,8 @@
     "rollup": "3.30.0",
     "rollup-plugin-dts": "^5.0.0",
     "rollup-plugin-peer-deps-external": "^2.2.4",
-    "typescript": "^5.8.3"
+    "typescript": "^5.8.3",
+    "uuid": "^10.0.0"
   },
   "overrides": {
     "rollup": "3.30.0"
--- a/packages/bruno-common/src/utils/faker-functions.ts
+++ b/packages/bruno-common/src/utils/faker-functions.ts
@@ -1,4 +1,5 @@
 import { faker } from '@faker-js/faker';
+import { v7 as uuidv7 } from 'uuid';
 
 export const timeBasedDynamicVars = new Set(['timestamp', 'isoTimestamp']);
 
@@ -6,7 +7,7 @@ export const mockDataFunctions = {
   guid: () => faker.string.uuid(),
   timestamp: () => Math.floor(Date.now() / 1000).toString(),
   isoTimestamp: () => new Date().toISOString(),
-  randomUUID: () => faker.string.uuid(),
+  randomUUID: () => uuidv7(),
   randomNanoId: () => faker.string.nanoid(),
   randomAlphaNumeric: () => faker.string.alphanumeric(),
   randomBoolean: () => faker.datatype.boolean(),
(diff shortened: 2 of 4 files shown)


packages/bruno-common/src/utils/faker-functions.ts
```diff
@@ -1,12 +1,13 @@
 import { faker } from '@faker-js/faker';
+import { v7 as uuidv7 } from 'uuid';
```
saravana512, line 2: const crypto = require('node:crypto');
const uuid = () => {
	try {
		return crypto.randomUUID();
	} catch (error) {
		throw new Error(error);
	}
};

function validate(uuid) {

    if (!uuid) {
        throw new Error(
            'Invalid UUID: UUID is required and cannot be empty'
        )
    }

	const _regex = /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}|00000000-0000-0000-0000-000000000000)$/i;
	return typeof uuid === 'string' && _regex.test(uuid);
}
module.exports = { uuid, validate };

Note: My suggestion only 
Why we need to install new package ? shall we use crypto module to generate UUID.
Reference: npm i node-uuid-unique ( Check this package code - above code is available in npm packkage )

packages/bruno-common/package.json
```diff
@@ -47,14 +47,16 @@
     "@rollup/plugin-terser": "^1.0.0",
     "@rollup/plugin-typescript": "^12.1.2",
     "@types/jest": "^29.5.14",
+    "@types/uuid": "^10.0.0",
```
saravana512, line 50: If use crypto dont need this types uuid package and uuid package

saravana512 (changes requested):

packages/bruno-common/package.json
Max-Reisinger: `crypto.randomUUID()` cannot replace these dependencies because it generates UUID v4 rather than UUID v7. Also, `uuid@10` does not bundle TypeScript declarations, so `@types/uuid` is required for the `bruno-common` TypeScript build. Since `uuid@^10.0.0` is already used elsewhere in this repository, I propose retaining both build dependencies here.

packages/bruno-common/src/utils/faker-functions.ts
Max-Reisinger: Thanks for the suggestion. `crypto.randomUUID()` only generates UUID v4 values, so it cannot satisfy the UUID v7 requirement in #9040. The repository already uses `uuid@^10.0.0` in `bruno-js` and `bruno-electron`, and that version provides the required `v7()` implementation. For those reasons, I propose retaining the current `uuid.v7()` implementation.

Max-Reisinger:
@saravana512 I’ve replied to both review threads with the UUID v7 and TypeScript dependency details. Could you please re-review this PR when you have a chance? Also, the Windows Playwright job failed only on unrelated E2E timeouts, but GitHub does not allow the PR author to rerun it without repository admin rights. Could a maintainer please rerun the failed job? Thanks!

PR #9044, opened 2026-08-20
fix: use NtlmClient as an adapter of axios instance

### LINKED ISSUE #4895: NTLM fails with self-signed certificate
[JIRA](https://usebruno.atlassian.net/browse/BRU-1448)

### This bug is:

- [ ] annoying

### Bruno version

2.5.0

### Operating System

Windows 10 19045.5965

### Describe the bug

> Also, `axios-ntlm` is [setting new instance of `httpAgent` and `httpsAgents`](https://github.com/Catbuttes/axios-ntlm/blob/0fcf14a7bcc836f5540cb800abdd0584b96f6f09/src/ntlmClient.ts#L31-L37) and it will interfere with proxy setup. So, proxy might not work when NTLM Auth is used.
> 
> _Originally posted by @ramki-bruno in https://github.com/usebruno/bruno/pull/3329#discussion_r1907412652_
            

I'm encountering issues with certificate validation because a new httpsAgent is created by axios-ntlm that doesnt receive `httpsAgentRequestFields`.

Setting the httpsAgent before the call to NtlmClient() fixes the issue for me, but is probably not a complete solution for all users. I'm not familiar enough to know how you would prefer to fix this to include the proxy settings and for the cli as well.

```js
  if (request.ntlmConfig) {
    const axiosRequestConfig = axiosInstance.defaults;
    axiosRequestConfig.httpsAgent = new https.Agent({
      ...httpsAgentRequestFields,
      keepAlive: true,
    });

    axiosInstance=NtlmClient(request.ntlmConfig, axiosRequestConfig);
    delete request.ntlmConfig;
  }
```

_No response_

Image

prateek-bruno:
### Description

Ref: BRU-1448
Fixes #4895

Updates NtlmClient to be an adapter instead of replacing our axios instance, causing the interceptors to not be attached to it.

Also on cli since https agent request field didn't had keepAlive, so this also adds that so that the handshake connection remains alive throughout.

### Problem

NTLM failed because we replaced the axiosInstance that included Bruno interceptors for timeline, settings, etc., creating an inconsistent anomaly in Bruno.

### Fix

- Use NtlmClient as an adapter instead of replacing our axiosInstance
- This preserves settings and interceptors in the NTLM flow
- CLI doesn’t keep the connection alive while running single request via HTTPS, this adds keepAlive to CLI for ntlm requests

Also adds Playwright and CLI tests for NTLM behavior over TLS and to verify interceptors apply.

### Screenshots

| Before | After |
| --- | --- |
| NTLM HTTPS request gives SSL issue, HTTP opens correct | NTLM HTTPS request picks the certificate and works, HTTP remains same in behaviour |
| <video src="https://github.com/user-attachments/assets/15b0b40c-3dad-45f3-98b1-f3f9bcb14fb5" width="100%" /> | <video src="https://github.com/user-attachments/assets/e9de023e-c9fe-47e1-a590-a34f71a1e646" width="100%" /> |

#### Contribution Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved NTLM authentication reliability by maintaining a single connection throughout the handshake.
  - Preserved request interceptors during NTLM-authenticated requests.
  - Improved NTLM support across redirects, streamed responses, request bodies, cookies, proxies, and HTTPS connections.
  - Added support for custom certificate authorities and configurable TLS verification.

- **Tests**
  - Added comprehensive coverage for successful authentication, connection reuse, proxy scenarios, redirects, TLS, client certificates, and authentication failures.

### DIFF
--- a/packages/bruno-cli/src/runner/run-single-request.js
+++ b/packages/bruno-cli/src/runner/run-single-request.js
@@ -18,6 +18,7 @@ const path = require('path');
 const { parseDataFromResponse } = require('../utils/common');
 const { getCookieStringForUrl, saveCookies } = require('../utils/cookies');
 const { createFormData } = require('../utils/form-data');
+const axios = require('axios');
 const { NtlmClient } = require('axios-ntlm');
 const { addDigestInterceptor, addEdgeGridInterceptor, getHttpHttpsAgents, makeAxiosInstance: makeAxiosInstanceForOauth2, applyOAuth1ToRequest } = require('@usebruno/requests');
 const { getCACertificates, transformProxyConfig, applySentHeadersToRequest } = require('@usebruno/requests');
@@ -397,7 +398,9 @@ const runSingleRequest = async function (
     const noproxy = get(options, 'noproxy', false);
     const cachedSystemProxy = get(options, 'cachedSystemProxy', null);
     const disableCache = !get(options, 'cacheSslSession', false);
-    const httpsAgentRequestFields = {};
+    // NTLM authenticates the connection rather than the request, so its handshake only completes
+    // when every leg reuses one socket.
+    const httpsAgentRequestFields = request.ntlmConfig ? { keepAlive: true } : {};
 
     if (insecure) {
       httpsAgentRequestFields['rejectUnauthorized'] = false;
@@ -678,7 +681,7 @@ const runSingleRequest = async function (
         request.timeout = requestTimeout;
       }
 
-      let axiosInstance = makeAxiosInstance({
+      const axiosInstance = makeAxiosInstance({
         requestMaxRedirects: requestMaxRedirects,
         disableCookies: options.disableCookies,
         followRedirects: followRedirects,
@@ -692,7 +695,8 @@ const runSingleRequest = async function (
       });
 
       if (request.ntlmConfig) {
-        axiosInstance = NtlmClient(request.ntlmConfig, axiosInstance.defaults);
+        const ntlmInstance = NtlmClient(request.ntlmConfig, {});
+        axiosInstance.defaults.adapter = (config) => ntlmInstance.request({ ...config, adapter: axios.getAdapter('http') });
         delete request.ntlmConfig;
       }
 
--- a/packages/bruno-cli/src/utils/axios-instance.js
+++ b/packages/bruno-cli/src/utils/axios-instance.js
@@ -5,7 +5,7 @@ const { createFormData } = require('./form-data');
 const { setupProxyAgents } = require('./proxy-util');
 const { isSameOrigin, DEFAULT_MAX_REDIRECTS } = require('@usebruno/common').utils;
 const { applyOmitHeaders, shouldOmitConnection } = require('@usebruno/common');
-const { getSentHeaders, applyOmitConnectionToAxiosConfig } = require('@usebruno/requests');
+const { getSentHeaders, applyOmitConnectionToAxiosConfig, handleNtlmRedirect } = require('@usebruno/requests');
 
 const redirectResponseCodes = [301, 302, 303, 307, 308];
 const METHOD_CHANGING_REDIRECTS = [301, 302, 303];
@@ -183,6 +183,8 @@ function makeAxiosInstance({
           }
 
           const requestConfig = createRedirectConfig(error, redirectUrl);
+
+          handleNtlmRedirect(requestConfig, error.config.url, redirectUrl, forwardAuthorizationHeader);
 
           if (!isSameOrigin(error.config.url, redirectUrl)) {
             /* AWS SigV4 signs a request for a specific host; re-signing after a cross-origin
--- a/packages/bruno-electron/src/ipc/network/axios-instance.js
+++ b/packages/bruno-electron/src/ipc/network/axios-instance.js
@@ -8,7 +8,7 @@ const { addCookieToJar, getCookieStringForUrl } = require('../../utils/cookies')
 const { preferencesUtil } = require('../../store/preferences');
 const { safeStringifyJSON } = require('../../utils/common');
 const { createFormData } = require('../../utils/form-data');
-const { getSentHeaders, applyOmitConnectionToAxiosConfig } = require('@usebruno/requests');
+const { getSentHeaders, applyOmitConnectionToAxiosConfig, handleNtlmRedirect } = require('@usebruno/requests');
 const { isSameOrigin, DEFAULT_MAX_REDIRECTS } = require('@usebruno/common').utils;
 const { applyOmitHeaders } = require('@usebruno/common');
 
@@ -377,6 +377,8 @@ function makeAxiosInstance({
               ...error.config.headers
             }
           };
+
+          handleNtlmRedirect(requestConfig, error.config.url, redirectUrl, forwardAuthorizationHeader);
 
           if (!isSameOrigin(error.config.url, redirectUrl)) {
             /* AWS SigV4 signs a request for a specific host; re-signing after a cross-origin
--- a/packages/bruno-electron/src/ipc/network/index.js
+++ b/packages/bruno-electron/src/ipc/network/index.js
@@ -157,7 +157,7 @@ const configureRequest = async (
 
   const { promptVariables = {} } = collection;
   let { proxyMode, proxyModeReason, proxyConfig, httpsAgentRequestFields, interpolationOptions } = certsAndProxyConfig;
-  let axiosInstance = makeAxiosInstance({
+  const axiosInstance = makeAxiosInstance({
     proxyMode,
     proxyModeReason,
     proxyConfig,
@@ -169,7 +169,8 @@ const configureRequest = async (
   });
 
   if (request.ntlmConfig) {
-    axiosInstance = NtlmClient(request.ntlmConfig, axiosInstance.defaults);
+    const ntlmInstance = NtlmClient(request.ntlmConfig, {});
+    axiosInstance.defaults.adapter = (config) => ntlmInstance.request({ ...config, adapter: axios.getAdapter('http') });
     delete request.ntlmConfig;
   }
 
--- a/packages/bruno-tests/package.json
+++ b/packages/bruno-tests/package.json
@@ -3,6 +3,13 @@
   "version": "0.0.1",
   "description": "",
   "main": "src/index.js",
+  "exports": {
+    ".": "./src/index.js",
+    "./ntlm": {
+      "types": "./src/ntlm/index.d.ts",
+      "default": "./src/ntlm/index.js"
+    }
+  },
   "scripts": {
     "start": "node .",
     "start:grpc": "node ./src/grpc/index.js"
@@ -27,11 +34,12 @@
     "cookie-parser": "^1.4.6",
     "cors": "^2.8.5",
     "express": "^4.21.2",
-    "graphql": "^16.10.0",
-    "graphql-yoga": "^5.10.6",
     "express-basic-auth": "^1.2.1",
     "fast-xml-parser": "^5.7.0",
+    "graphql": "^16.10.0",
+    "graphql-yoga": "^5.10.6",
     "http-proxy": "^1.18.1",
+    "js-md4": "^0.3.2",
     "js-yaml": "4.3.1",
     "jsonwebtoken": "^9.0.3",
     "lodash": "4.18.1",
--- a/packages/bruno-tests/src/ntlm/certificates.js
+++ b/packages/bruno-tests/src/ntlm/certificates.js
@@ -1,0 +1,39 @@
+const path = require('node:path');
+const { execFileSync } = require('node:child_process');
+
+let opensslChecked = false;
+
+const assertOpensslAvailable = () => {
+  if (opensslChecked) return;
+  try {
+    execFileSync('openssl', ['version'], { stdio: 'ignore' });
+  } catch {
+    throw new Error('The NTLM TLS test server needs `openssl` on PATH to generate its certificates; install openssl or skip the TLS specs.');
+  }
+  opensslChecked = true;
+};
+
+const generateSelfSignedCert = (dir, name, commonName) => {
+  assertOpensslAvailable();
+
+  const keyPath = path.join(dir, `${name}-key.pem`);
+  const certPath = path.join(dir, `${name}-cert.pem`);
+
+  try {
+    execFileSync('openssl', [
+      'req', '-x509', '-newkey', 'rsa:2048', '-nodes',
+      '-keyout', keyPath,
+      '-out', certPath,
+      '-days', '1',
+      '-subj', `/CN=${commonName}`,
+      '-addext', 'subjectAltName=DNS:localhost,IP:127.0.0.1'
+    ], { stdio: ['ignore', 'ignore', 'pipe'] });
+  } catch (error) {
+    const detail = error.stderr?.toString().trim() || error.message;
+    throw new Error(`Could not generate the ${name} certificate with openssl: ${detail}`);
+  }
+
+  return { keyPath, certPath };
+};
+
+module.exports = { generateSelfSignedCert };
--- a/packages/bruno-tests/src/ntlm/server.js
+++ b/packages/bruno-tests/src/ntlm/server.js
@@ -1,0 +1,128 @@
+const crypto = require('node:crypto');
+const fs = require('node:fs');
+const http = require('node:http');
+const https = require('node:https');
+const os = require('node:os');
+const path = require('node:path');
+
+const { type2Message, messageType, provesPassword } = require('./messages');
+const { generateSelfSignedCert } = require('./certificates');
+
+const CLIENT_CERT_NAME = 'bruno-ntlm-client';
+
+const readBody = (req) =>
+  new Promise((resolve) => {
+    let body = '';
+    req.on('data', (chunk) => {
+      body += chunk;
+    });
+    req.on('end', () => resolve(body));
+  });
+
+/**
+ * Authentication is bound to the connection, the way IIS does it, so a handshake split across
+ * sockets fails here instead of passing quietly. Every request is recorded to make that reuse
+ * observable, and `/redirect?to=<url>` takes its target from the request so no test configures it.
+ */
+const handleRequest = ({ password, requests }) => async (req, res) => {
+  const body = await readBody(req);
+  const authorization = req.headers.authorization;
+  const [pathname, query] = req.url.split('?');
+
+  const record = {
+    socketId: req.socket.ntlmSocketId,
+    type: messageType(authorization),
+    url: req.url,
+    body,
+    headers: req.headers,
+    clientCertName: req.socket.getPeerCertificate?.()?.subject?.CN
+  };
+  requests.push(record);
+
+  const challengeWith = (header) => {
+    res.writeHead(401, { 'www-authenticate': header, 'content-length': 0 });
+    return res.end();
+  };
+
+  switch (messageType(authorization)) {
+    case 1: {
+      const nonce = crypto.randomBytes(8);
+      req.socket.ntlmChallenge = nonce;
+      return challengeWith(type2Message(nonce));
+    }
+    case 3: {
+      const nonce = req.socket.ntlmChallenge;
+      record.provedPassword = Boolean(nonce) && provesPassword(authorization, password, nonce);
+
+      if (!record.provedPassword) {
+        return challengeWith('Negotiate, NTLM');
+      }
+      if (pathname === '/redirect') {
+        res.writeHead(302, { 'location': new URLSearchParams(query).get('to'), 'content-length': 0 });
+        return res.end();
+      }
+      res.writeHead(200, { 'content-type': 'application/json' });
+      return res.end(JSON.stringify({ authenticated: true, url: pathname }));
+    }
+    default:
+      return challengeWith('NTLM');
+  }
+};
+
+const createCertificates = (requireClientCert) => {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'bruno-ntlm-tls-'));
+
+  return {
+    dir,
+    server: generateSelfSignedCert(dir, 'server', 'localhost'),
+    client: requireClientCert ? generateSelfSignedCert(dir, 'client', CLIENT_CERT_NAME) : null
+  };
+};
+
+const tlsOptionsFor = ({ server, client }) => ({
+  key: fs.readFileSync(server.keyPath),
+  cert: fs.readFileSync(server.certPath),
+  ...(client ? { requestCert: true, rejectUnauthorized: true, ca: [fs.readFileSync(client.certPath)] } : {})
+});
+
+const startNtlmServer = async ({ tls = false, password = 'pass', requireClientCert = false } = {}) => {
+  const requests = [];
+  const sockets = [];
+  const handler = handleRequest({ password, requests });
+
+  const certificates = tls ? createCertificates(requireClientCert) : null;
+  const listener = certificates
+    ? https.createServer(tlsOptionsFor(certificates), handler)
+    : http.createServer(handler);
+
+  let socketId = 0;
+  listener.on(tls ? 'secureConnection' : 'connection', (socket) => {
+    socket.ntlmSocketId = ++socketId;
+    sockets.push(socket);
+  });
+
+  await new Promise((resolve) => listener.listen(0, '127.0.0.1', resolve));
+  const { port } = listener.address();
+  const scheme = tls ? 'https' : 'http';
+
+  return {
+    baseUrl: `${scheme}://127.0.0.1:${port}`,
+    requests,
+    certPath: certificates?.server.certPath,
+    clientCertPath: certificates?.client?.certPath,
+    clientKeyPath: certificates?.client?.keyPath,
+    clientCertName: CLIENT_CERT_NAME,
+    messageTypesSeen: () => requests.map((request) => request.type),
+    negotiations: () => requests.filter((request) => request.type === 1),
+    connectionsUsed: () => new Set(requests.map((request) => request.socketId)).size,
+    close: async () => {
+      sockets.forEach((socket) => socket.destroy());
+      await new Promise((resolve) => listener.close(resolve));
+      if (certificates) {
+        fs.rmSync(certificates.dir, { recursive: true, force: true, maxRetries: 3, retryDelay: 100 });
+      }
+    }
+  };
+};
+
+module.exports = { startNtlmServer };
--- a/tests/auth/ntlm/init-user-data-client-cert/preferences.json
+++ b/tests/auth/ntlm/init-user-data-client-cert/preferences.json
@@ -1,0 +1,20 @@
+{
+  "lastOpenedCollections": [
+    "{{collectionPath}}/ntlm-tls"
+  ],
+  "preferences": {
+    "request": {
+      "sslVerification": false,
+      "clientCertificates": {
+        "certs": [
+          {
+            "domain": "127.0.0.1",
+            "type": "cert",
+            "certFilePath": "{{clientCertPath}}",
+            "keyFilePath": "{{clientKeyPath}}"
+          }
+        ]
+      }
+    }
+  }
+}
--- a/tests/auth/ntlm/init-user-data-tls-off/preferences.json
+++ b/tests/auth/ntlm/init-user-data-tls-off/preferences.json
@@ -1,0 +1,10 @@
+{
+  "lastOpenedCollections": [
+    "{{collectionPath}}/ntlm-tls"
+  ],
+  "preferences": {
+    "request": {
+      "sslVerification": false
+    }
+  }
+}
--- a/tests/auth/ntlm/ntlm-tls.spec.ts
+++ b/tests/auth/ntlm/ntlm-tls.spec.ts
@@ -1,0 +1,128 @@
+import * as path from 'path';
+import { test, expect, closeElectronApp, type ElectronApplication } from '../../../playwright';
+import { buildCommonLocators, openRequest, sendRequest, waitForReadyPage } from '../../utils/page';
+import { startNtlmServer, type NtlmEndpoint } from '@usebruno/tests/ntlm';
+import { SELF_SIGNED_CERTIFICATE, TLS_HANDSHAKE_FAILURE } from '../../utils/constants';
+
+const PASSWORD = 'Br!unoNtlm';
+
+test.describe('ntlm against an endpoint with a self signed certificate', () => {
+  let endpoint: NtlmEndpoint;
+  let app: ElectronApplication;
+
+  test.afterEach(async () => {
+    await endpoint?.close();
+    if (app) await closeElectronApp(app);
+  });
+
+  test('authenticates when certificate verification is off', async ({ launchElectronApp, collectionFixturePath }) => {
+    endpoint = await startNtlmServer({ password: PASSWORD, tls: true });
+
+    app = await launchElectronApp({
+      initUserDataPath: path.join(__dirname, 'init-user-data-tls-off'),
+      templateVars: { collectionPath: collectionFixturePath! },
+      dotEnv: { NTLM_BASE_URL: endpoint.baseUrl }
+    });
+    const page = await waitForReadyPage(app);
+
+    await test.step('send a request with ntlm auth', async () => {
+      await openRequest(page, 'ntlm-tls', 'api');
+      await sendRequest(page, 200);
+    });
+
+    await test.step('the server saw one full handshake on one connection', async () => {
+      expect(endpoint.messageTypesSeen()).toEqual([null, 1, 3]);
+      expect(endpoint.connectionsUsed()).toBe(1);
+    });
+  });
+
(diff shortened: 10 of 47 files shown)


packages/bruno-tests/src/ntlm/certificates.js
```diff
@@ -0,0 +1,20 @@
...
+  const certPath = path.join(dir, `${name}-cert.pem`);
+
+  execFileSync('openssl', [
+    'req', '-x509', '-newkey', 'rsa:2048', '-nodes',
+    '-keyout', keyPath,
+    '-out', certPath,
+    '-days', '1',
+    '-subj', `/CN=${commonName}`,
+    '-addext', 'subjectAltName=DNS:localhost,IP:127.0.0.1'
+  ], { stdio: 'ignore' });
```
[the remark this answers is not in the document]
prateek-bruno: OpenSSL is available on all of our CI runners.

packages/bruno-cli/tests/integration/ntlm/run-ntlm.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+  const run = (args) => runCli(['run', ...args, '--noproxy', '--env-var', `baseUrl=${endpoint.baseUrl}`], collectionDir);
+
+  beforeEach(() => {
+    collectionDir = createCollectionFixture(FIXTURE_COLLECTION);
+  });
+
+  afterEach(async () => {
+    await endpoint.close();
+    fs.rmSync(collectionDir, { recursive: true, force: true });
```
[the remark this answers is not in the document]
prateek-bruno: Resolved!

packages/bruno-electron/tests/network/ntlm-connection.spec.js
```diff
@@ -0,0 +1,118 @@
...
+    expect(response.status).toBe(200);
+  });
+
+  test('keeps a streamed response on that single connection', async () => {
+    const response = await send({ responseType: 'stream' });
+    response.data.resume();
+
+    expect(server.messageTypesSeen()).toEqual([null, 1, 3]);
+    expect(server.connectionsUsed()).toBe(1);
+  });
```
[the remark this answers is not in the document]
prateek-bruno: Resolved!

packages/bruno-tests/src/ntlm/server.js
```diff
@@ -0,0 +1,126 @@
...
+  const certificates = tls ? createCertificates(requireClientCert) : null;
+  const listener = certificates
+    ? https.createServer(tlsOptionsFor(certificates), handler)
+    : http.createServer(handler);
+
+  let socketId = 0;
+  listener.on('connection', (socket) => {
+    socket.ntlmSocketId = ++socketId;
+    sockets.push(socket);
+  });
```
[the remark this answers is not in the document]
prateek-bruno: Resolved!

packages/bruno-electron/tests/network/ntlm-tls.spec.js
```diff
@@ -0,0 +1,120 @@
...
+    expect(response.status).toBe(200);
+    expect(response.data).toMatchObject({ authenticated: true });
+    expect(server.messageTypesSeen()).toEqual([null, 1, 3]);
+    expect(server.legs.map((leg) => leg.clientCertName)).toEqual(Array(3).fill(server.clientCertName));
+    expect(server.connectionsUsed()).toBe(1);
+  });
+
+  test('fails when no client certificate is configured', async () => {
+    appState.globalClientCertificates = [];
+
```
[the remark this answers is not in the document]
prateek-bruno: It already does this in beforeEach.

packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -383,7 +384,9 @@ const runSingleRequest = async function (
     const noproxy = get(options, 'noproxy', false);
     const cachedSystemProxy = get(options, 'cachedSystemProxy', null);
     const disableCache = !get(options, 'cacheSslSession', false);
-    const httpsAgentRequestFields = {};
+    // NTLM authenticates the connection rather than the request, so its handshake only completes
+    // when every leg reuses one socket.
+    const httpsAgentRequestFields = request.ntlmConfig ? { keepAlive: true } : {};
```
Pragadesh-45, line 403: Nit. Electron applies `keepAlive: true` unconditionally (`axios-instance.js:198`) and `getHttpHttpsAgents` hardcodes it. please check the ee repo

```
    const agentOptions = {
      ...httpsAgentRequestFields,
      keepAlive: true
    };
```

packages/bruno-electron/tests/network/helpers/stub-adapter.js
```diff
@@ -0,0 +1,60 @@
...
+
+const sendThroughStubAdapter = async ({ request, responses }) => {
+  const calls = [];
+  const axiosInstance = await configureFor(request);
+  const response = await axiosInstance({ ...request, adapter: stubAdapter(responses, calls) });
+
+  return { response, calls };
+};
+
+// A stub in the request's adapter would shadow the ntlm adapter, so this one sits beneath it.
```
Pragadesh-45: The same shadowing happens in production on every redirect, because the redirect config inherits the adapter

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -168,7 +168,9 @@ const configureRequest = async (
   });
 
   if (request.ntlmConfig) {
-    axiosInstance = NtlmClient(request.ntlmConfig, axiosInstance.defaults);
+    // Keeps Bruno's interceptors and the single prepared connection across the handshake.
+    const ntlmInstance = NtlmClient(request.ntlmConfig, {});
+    axiosInstance.defaults.adapter = (config) => ntlmInstance.request({ ...config, adapter: axios.getAdapter('http') });
```
Pragadesh-45, line 173: The redirect legs will break from here. and there are two separate causes stacked on top of each other. `ntlm-handshake.spec.js:76` and `run-ntlm.spec.js:41` both assert the resulting failure as expected behaviour.

1. connection-scoped headers ride the redirect. `{ ...config }` shares the same headers object with the outer config, so `axios-ntlm's ` writes at `ntlmClient.js:116-117` land on it. The redirect handler copies `Authorization: NTLM <T3> `and `X-retry: false` forward. `X-retry === 'false'` is exactly the guard at `ntlmClient.js:100` that stops re-negotiation, so the new socket gets a replayed stale `T3` and a `401`.

This also means the `NTLMv2` proof is forwarded cross-origin: `forwardAuthorizationHeader` defaults to `true` (`index.js:140`), so `axios instance.js:401` only strips Authorization when a user has opted out. NTLM authenticates a connection, not a request - forwarding it is never useful, and a redirect to a hostile host gets an offline-crackable, relayable proof. Consistent with how Basic/Bearer are already handled, but this is not right for the NTLM.

2. The redirect leg never reaches the NTLM shim. The shim passes `adapter: axios.getAdapter('http')` into `ntlmInstance.request`, so the resolved config carries adapter as an own property. The redirect handler spreads that config, and a per-request adapter shadows `defaults.adapter`, the redirect goes straight down the raw `http` adapter with no NTLM interceptor attached. Fixing only Cause A gets you a clean 401 with no handshake; both are needed.

packages/bruno-electron/tests/network/ntlm-connection.spec.js
Pragadesh-45, line 1: can we add the test for proxies too?

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: But on CLI it is never passed, even in ee for https.

packages/bruno-electron/tests/network/helpers/stub-adapter.js
prateek-bruno: not relevant, stub adapter is used only for interceptors and ntlm one is the only one that uses the overridden adapter.

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: We'll keep it like this for this PR and will open a separate PR for unconditionally passing keep alive in HTTPS agent once I get more context. Don't want to regress other untested flows.

packages/bruno-electron/tests/network/ntlm-connection.spec.js
prateek-bruno: Will skip this for now as even if we add it it would be mocking proxy. And since we don't have proxy tests for interceptors in this convention it would be better if we manually check it.

packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -17,6 +17,7 @@ const path = require('path');
 const { parseDataFromResponse } = require('../utils/common');
 const { getCookieStringForUrl, saveCookies } = require('../utils/cookies');
 const { createFormData } = require('../utils/form-data');
+const axios = require('axios');
```
sanish-bruno, line 21: I feels like, we should ideally put external library imports at the top, followed by in-house library imports then relative imports from the files. Something to keep in mind for future PRs, This file is not following it so the comment can be ignored!

packages/bruno-cli/tests/integration/ntlm/run-ntlm.spec.js
```diff
@@ -0,0 +1,70 @@
...
+const FIXTURE_COLLECTION = path.join(__dirname, 'fixtures', 'collection');
+const FIXTURE_PASSWORD = 'Br!unoNtlm';
+
+describe('CLI run — ntlm', () => {
+  let endpoint;
+  let collectionDir;
+
+  const startEndpoint = (options) => startNtlmServer({ ...options, password: FIXTURE_PASSWORD });
+
+  const run = (args) => runCli(['run', ...args, '--noproxy', '--env-var', `baseUrl=${endpoint.baseUrl}`], collectionDir);
```
sanish-bruno: can't we use `runCli` directly?

sanish-bruno: I would also think about whether we needed this abstraction in the firstplace, we are techinically only save 3 default arguments from repeating, thoughts?

```suggestion
  const run = (args, { collectionDir, baseUrl }) => runCli(['run', ...args, '--noproxy', '--env-var', `baseUrl=${baseUrl}`], collectionDir);
```

packages/bruno-tests/package.json
```diff
@@ -32,6 +32,7 @@
     "express-basic-auth": "^1.2.1",
     "fast-xml-parser": "^5.7.0",
     "http-proxy": "^1.18.1",
+    "js-md4": "^0.3.2",
```
sanish-bruno, line 42: reasoning behind we chose to add this library(`js-md4`)?, can be moved to dev-dependecies?

packages/bruno-electron/tests/network/helpers/app-state.js
```diff
@@ -0,0 +1,71 @@
+const productDefaults = () => ({
```
sanish-bruno: can just be an object, not args for the fn, can can be renamed to `defaultState` may be

packages/bruno-electron/tests/network/helpers/stub-adapter.js
```diff
@@ -0,0 +1,60 @@
+const axios = require('axios');
+const { AxiosError } = axios;
+const { messageType } = require('../../../../bruno-tests/src/ntlm');
```
sanish-bruno: do we ever used cross package dependencies like this, i guess we can add it as require('@usebruno/tests) and add it as a dev dependency, same goes for all such references ?

packages/bruno-tests/src/ntlm/certificates.js
```diff
@@ -0,0 +1,20 @@
+const path = require('node:path');
+const { execFileSync } = require('node:child_process');
+
+const generateSelfSignedCert = (dir, name, commonName) => {
+  const keyPath = path.join(dir, `${name}-key.pem`);
+  const certPath = path.join(dir, `${name}-cert.pem`);
+
+  execFileSync('openssl', [
```
sanish-bruno: `openssl` can have issues while running in windows CI, may need to gate it depending on the availability of `openssl`, can skip TLS tests if the package is absent

packages/bruno-tests/src/ntlm/server.js
```diff
@@ -0,0 +1,126 @@
...
+ * Authentication is bound to the connection, the way IIS does it, so a handshake split across
+ * sockets fails here instead of passing quietly. Every leg is recorded to make that reuse
+ * observable, and `/redirect?to=<url>` takes its target from the request so no test configures it.
+ */
+const handleRequest = ({ password, legs }) => async (req, res) => {
+  const body = await readBody(req);
+  const authorization = req.headers.authorization;
+  const [pathname, query] = req.url.split('?');
+
+  legs.push({
```
sanish-bruno: `legs` ---> `handshakeLegs`

packages/bruno-electron/tests/network/ntlm-handshake.spec.js
```diff
@@ -0,0 +1,82 @@
...
+jest.mock('../../src/store/bruno-config', () => require('./helpers/app-state').brunoConfig);
+jest.mock('../../src/store/system-proxy', () => require('./helpers/app-state').systemProxy);
+jest.mock('../../src/utils/cookies', () => require('./helpers/app-state').cookies);
+
+const { appState, resetAppState } = require('./helpers/app-state');
+const { startNtlmEndpoint, sendRequest } = require('./helpers/ntlm-request');
+
+let server;
+let baseUrl;
+let legs;
```
sanish-bruno: rename for better readability

packages/bruno-electron/src/ipc/network/index.js
prateek-bruno: Have fixed all three!

packages/bruno-electron/tests/network/ntlm-handshake.spec.js
```diff
@@ -0,0 +1,82 @@
...
+
+const BODY = '{"hello":"world"}';
+
+const send = (overrides = {}) => sendRequest({ url: `${baseUrl}/api`, ...overrides });
```
sanish-bruno: i think we may not need this extraction,  we can directly use `sendRequest`, not many defaults to deal with here

sanish-bruno (changes requested):

packages/bruno-tests/src/ntlm/certificates.js
prateek-bruno: It is actually passing on windows CI as it is bundled with all windows runner, same reason why Windows SSL Tests pass.

packages/bruno-electron/tests/network/helpers/app-state.js
prateek-bruno: I believe this is better cause appState is not keeping a reference to productDefaults but applying transformations on it. If we make it a variable then we would have to handle Object copy and all.

packages/bruno-tests/package.json
prateek-bruno: js-md4 is already a transitive dependency, it is used by NTLM for challenge validation in handshake. It uses MD4 and not MD5, so had to import it.

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: Ignoring for now as discussed.

packages/bruno-electron/src/ipc/network/axios-instance.js
```diff
@@ -414,6 +418,21 @@ function makeAxiosInstance({
             }
           }
 
+          if (carriesNtlm) {
+            const staysOnOrigin = isSameOrigin(error.config.url, redirectUrl);
+
+            Object.keys(requestConfig.headers).forEach((key) => {
+              const lowerKey = key.toLowerCase();
+              if (lowerKey === 'x-retry' || (lowerKey === 'authorization' && !staysOnOrigin)) {
```
Pragadesh-45:
```suggestion
              if (lowerKey === 'x-retry' || (lowerKey === 'authorization')) {
```

packages/bruno-electron/src/ipc/network/axios-instance.js
```diff
@@ -414,6 +418,21 @@ function makeAxiosInstance({
...
+                delete requestConfig.headers[key];
+              }
+            });
+
+            if (staysOnOrigin) {
+              delete requestConfig.adapter;
+            }
+          }
```
Pragadesh-45: `staysOnOrigin` is needed here as it is the negotiation defense which is needed for the cross-origin calls

Pragadesh-45:
```suggestion
          if (carriesNtlm) {
            const staysOnOrigin = isSameOrigin(error.config.url, redirectUrl);

            // A finished message proves nothing on the socket a redirect opens, and X-retry
            // would stop the library negotiating a new one.
            Object.keys(requestConfig.headers).forEach((key) => {
              const lowerKey = key.toLowerCase();
              if (lowerKey === 'x-retry' || lowerKey === 'authorization') {
                delete requestConfig.headers[key];
              }
            });

            // Only this origin gets a fresh handshake; another host would choose the challenge
            // our password signs, so its redirect keeps the inherited adapter and is offered none.
            if (staysOnOrigin) {
              delete requestConfig.adapter;
            }
          }
```

packages/bruno-cli/src/utils/axios-instance.js
```diff
@@ -207,6 +211,21 @@ function makeAxiosInstance({
...
+              const lowerKey = key.toLowerCase();
+              if (lowerKey === 'x-retry' || (lowerKey === 'authorization' && !staysOnOrigin)) {
+                delete requestConfig.headers[key];
+              }
+            });
+
+            if (staysOnOrigin) {
+              delete requestConfig.adapter;
+            }
+          }
```
Pragadesh-45: same as the electron

```suggestion
          if (carriesNtlm) {
            const staysOnOrigin = isSameOrigin(error.config.url, redirectUrl);

            // A finished message proves nothing on the socket a redirect opens, and X-retry
            // would stop the library negotiating a new one.
            Object.keys(requestConfig.headers).forEach((key) => {
              const lowerKey = key.toLowerCase();
              if (lowerKey === 'x-retry' || lowerKey === 'authorization') {
                delete requestConfig.headers[key];
              }
            });

            // Only this origin gets a fresh handshake; another host would choose the challenge
            // our password signs, so its redirect keeps the inherited adapter and is offered none.
            if (staysOnOrigin) {
              delete requestConfig.adapter;
            }
          }
```

Pragadesh-45 (changes requested):

packages/bruno-electron/tests/network/ntlm-connection.spec.js
```diff
@@ -0,0 +1,93 @@
...
+
+    expect(server.messageTypesSeen()).toEqual([null, 1, 3]);
+    expect(secondOrigin.messageTypesSeen()).toEqual([null]);
+  });
+
+  test('negotiates again when a redirect moves onto a new connection', async () => {
+    const response = await sendRequest({ url: `${baseUrl}/redirect?to=${baseUrl}/landing` });
+
+    expect(response.status).toBe(200);
+    expect(server.messageTypesSeen()).toEqual([null, 1, 3, 3, 1, 3]);
```
Pragadesh-45:
```suggestion
    expect(server.messageTypesSeen()).toEqual([null, 1, 3, null, 1, 3]);
```

prateek-bruno: This is not correct, current one is correct.

prateek-bruno: When it is same origin redirect it will try to send the Authorization code once, if it rejects then it will begin re-negotiation.

prateek-bruno: Done!

Pragadesh-45 (approved):

Pragadesh44-Bruno (approved):

Pragadesh44-Bruno (dismissed):

tests/utils/cli.ts
```diff
@@ -0,0 +1,13 @@
+import { exec } from 'child_process';
+import * as path from 'path';
+
+const BRU_BIN = path.resolve(__dirname, '../../packages/bruno-cli/bin/bru.js');
+
+// execSync blocks the node event loop, so any other js handler will remain frozen
+// That's why here we are running by exec, this also makes it async
+export const runCLIAsync = (cwd: string, args: string): Promise<{ code: number; stdout: string }> =>
+  new Promise((resolve) => {
+    exec(`node ${BRU_BIN} ${args}`, { cwd, env: { ...process.env, FORCE_COLOR: '0' } }, (error, stdout) =>
```
sanish-bruno: why do we have ``FORCE_COLOR` here?

packages/bruno-cli/src/utils/axios-instance.js
```diff
@@ -207,6 +211,28 @@ function makeAxiosInstance({
             }
           }
 
+          if (carriesNtlm) {
+            const staysOnOrigin = isSameOrigin(error.config.url, redirectUrl);
+
+            // A finished message proves nothing on the socket a redirect opens, and X-retry
+            // would stop the library negotiating a new one.
+            Object.keys(requestConfig.headers).forEach((key) => {
```
sanish-bruno: should we use a helper here? something like.

```js
deleteHeaders(headers, ['x-retry', 'authorization'])
```

packages/bruno-cli/src/runner/run-single-request.js
```diff
@@ -681,7 +684,8 @@ const runSingleRequest = async function (
       });
 
       if (request.ntlmConfig) {
-        axiosInstance = NtlmClient(request.ntlmConfig, axiosInstance.defaults);
+        const ntlmInstance = NtlmClient(request.ntlmConfig, {});
+        axiosInstance.defaults.adapter = (config) => ntlmInstance.request({ ...config, adapter: axios.getAdapter('http') });
```
sanish-bruno, line 699: confirm if we need to provide the http adapter explicitly

packages/bruno-tests/src/ntlm/server.js
```diff
@@ -0,0 +1,128 @@
...
+    : http.createServer(handler);
+
+  let socketId = 0;
+  listener.on(tls ? 'secureConnection' : 'connection', (socket) => {
+    socket.ntlmSocketId = ++socketId;
+    sockets.push(socket);
+  });
+
+  await new Promise((resolve) => listener.listen(port, '127.0.0.1', resolve));
+  const listeningPort = listener.address().port;
```
sanish-bruno: check if we need to extract this port ,can both port and listeningPort can be a same value?

packages/bruno-tests/src/ntlm/server.js
```diff
@@ -0,0 +1,128 @@
...
+  const scheme = tls ? 'https' : 'http';
+
+  return {
+    baseUrl: `${scheme}://127.0.0.1:${listeningPort}`,
+    urlFor: (hostname) => `${scheme}://${hostname}:${listeningPort}`,
```
sanish-bruno: check how the urlFor is being used, why baseUrl doesn't get the same customisability as that of urlFor

tests/auth/ntlm/init-user-data/collection-security.json
```diff
@@ -0,0 +1,10 @@
+{
+  "collections": [
```
sanish-bruno: can we use `setSandBoxMode` and remove this file, either way, collection always default to `safe` mode if i am not wrong

tests/auth/ntlm/ntlm-tls.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+import { test, expect, type Page } from '../../../playwright';
+import { buildCommonLocators, openRequest, waitForReadyPage } from '../../utils/page';
+import type { NtlmEndpoint } from '../../../packages/bruno-tests/src/ntlm';
+
+const { startNtlmServer } = require('@usebruno/tests/ntlm');
+
+const PASSWORD = 'Br!unoNtlm';
+const PORT = 8101;
+
+test.describe.serial('ntlm against an endpoint with a self signed certificate', () => {
```
sanish-bruno: do we have to make the test serial?

tests/auth/ntlm/ntlm-tls.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+  let endpoint: NtlmEndpoint;
+
+  test.afterEach(async () => {
+    await endpoint.close();
+  });
+
+  const send = async (page: Page) => {
+    await openRequest(page, 'ntlm-tls', 'api');
+    await page.getByTestId('send-arrow-icon').click();
```
sanish-bruno: can we use `sendRequest` util for this? feels a bit odd it is returning a locator

tests/auth/ntlm/ntlm-tls.spec.ts
sanish-bruno: add a comment why the test should be serial

tests/auth/ntlm/ntlm-tls.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    return buildCommonLocators(page).response;
+  };
+
+  test('authenticates when certificate verification is off', async ({ launchElectronApp }) => {
+    endpoint = await startNtlmServer({ password: PASSWORD, port: PORT, tls: true });
+
+    const response = await send(
```
sanish-bruno: this structure feels a bit odd , can you take a look at this

tests/auth/ntlm/ntlm.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+const SECOND_PORT = 8100;
+const INIT_USER_DATA = path.join(__dirname, 'init-user-data');
+
+test.describe.serial('ntlm over a real connection', () => {
+  let primary: NtlmEndpoint;
+  let second: NtlmEndpoint;
+  let page: Page;
+
+  test.beforeAll(async ({ launchElectronApp }) => {
+    page = await waitForReadyPage(await launchElectronApp({ initUserDataPath: INIT_USER_DATA }));
```
sanish-bruno: break it in to steps

tests/auth/ntlm/ntlm.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const hops = await timelineHeaders.requestHops();
+
+    expect(hops.map((hop) => hop.request)).toEqual(['GET http://127.0.0.1:8099/api']);
+    expect(hops.map((hop) => hop.headerLines.join('\n'))).toEqual([expect.stringMatching(/authorization: NTLM/i)]);
+  });
+
+  test('reports how long the answer took', async () => {
+    await send('api', 200);
+
+    await expect(buildCommonLocators(page).response.pane()).toContainText(/\d+(ms|\.\d+s)/);
```
sanish-bruno: split into multiple line

sanish-bruno (changes requested):

packages/bruno-cli/src/utils/axios-instance.js
prateek-bruno: It is existing pattern so better to leave it like this for now.

tests/utils/cli.ts
prateek-bruno: FORCE_COLOR is inherited from playwright's own reporter, and chalk would then colour the run summary a caller reads out of stdout. So we're disabling it here.

packages/bruno-cli/src/runner/run-single-request.js
prateek-bruno: Yes, without it ntlmInstance will try to adopt itself as it's adapter recursively.

tests/auth/ntlm/ntlm.spec.ts
prateek-bruno: Resolved!

tests/auth/ntlm/ntlm.spec.ts
prateek-bruno: Resolved!

tests/auth/ntlm/ntlm-tls.spec.ts
prateek-bruno: Resolved!

tests/auth/ntlm/ntlm-tls.spec.ts
prateek-bruno: Have made it parallel!

tests/auth/ntlm/ntlm-tls.spec.ts
prateek-bruno: Resolved!

tests/auth/ntlm/init-user-data/collection-security.json
prateek-bruno: Resolved!

tests/auth/ntlm/ntlm-tls.spec.ts
prateek-bruno: Have made it parallel!

packages/bruno-tests/src/ntlm/server.js
prateek-bruno: Resolved, have dropped it!

packages/bruno-tests/src/ntlm/server.js
prateek-bruno: The sent port is requested port and listening port is what the system actually gave us, we've utilised it properly and made the tests run in parallel.

tests/auth/ntlm/init-user-data-tls-off/preferences.json
```diff
@@ -0,0 +1,5 @@
+{
+  "lastOpenedCollections": [
+    "{{collectionPath}}/ntlm-tls"
+  ]
```
vijayh-bruno: No preferences block here, and sslVerification defaults to true. So this fixture runs with verification on.

tests/auth/ntlm/init-user-data-client-cert/preferences.json
```diff
@@ -0,0 +1,19 @@
+{
+  "lastOpenedCollections": [
+    "{{collectionPath}}/ntlm-tls"
+  ],
+  "preferences": {
+    "request": {
```
vijayh-bruno, line 6: sslVerification is missing here too, so TLS would fail against the self-signed cert before the client certificate ever comes into play?

packages/bruno-electron/src/ipc/network/index.js
```diff
@@ -168,7 +168,8 @@ const configureRequest = async (
   });
 
   if (request.ntlmConfig) {
-    axiosInstance = NtlmClient(request.ntlmConfig, axiosInstance.defaults);
+    const ntlmInstance = NtlmClient(request.ntlmConfig, {});
+    axiosInstance.defaults.adapter = (config) => ntlmInstance.request({ ...config, adapter: axios.getAdapter('http') });
```
vijayh-bruno, line 173: will the NTLM auth header end up in response.sentHeaders, along with X-retry?

tests/auth/ntlm/init-user-data-tls-off/preferences.json
prateek-bruno: Default ssl verification is off in bruno and playwright doesn't modify it either:

```
// packages/bruno-electron/src/store/preferences.js:15
const defaultPreferences = { request: { sslVerification: false, … } };
```

tests/auth/ntlm/init-user-data-client-cert/preferences.json
prateek-bruno: Default ssl verification is off in bruno and playwright doesn't modify it either:

```
// packages/bruno-electron/src/store/preferences.js:15
const defaultPreferences = { request: { sslVerification: false, … } };
```

packages/bruno-electron/src/ipc/network/index.js
prateek-bruno: Screenshot 2026-08-27 at 12 07 26 PM

Yes it will, NTLM is supposed to show and X-retry: false is used by the NtlmClient library internally to track the last request.

sanish-bruno (dismissed):

tests/auth/ntlm/init-user-data-tls-off/preferences.json
prateek-bruno: I would like to clear up a divergence in this repo and EE repo, in this repo we default sslVerification to true instead of false, so this point does makes sense. I will push again with this fixed.

I can see that Linux OAuth Tests broke because of this.

tests/auth/ntlm/init-user-data-client-cert/preferences.json
prateek-bruno: I would like to clear up a divergence in this repo and EE repo, in this repo we default sslVerification to true instead of false, so this point does makes sense. I will push again with this fixed.

I can see that Linux OAuth Tests broke because of this.

sanish-bruno (approved):

vijayh-bruno (approved):

PR #9072, opened 2026-08-24
feat(generate-docs): advanced filtering options in generate documentation flow

sachin-bruno:
Ref: BRU - 4022

### Description

Adds a Generate Documentation option to collections so you can create a shareable HTML page of your API docs right from the app. It also has advanced controls to filter requests by tags, choose which environments to include, and optionally add the git repo link so readers can open the collection in Bruno.

### Problem

Git repo URL is always embedded. Git-backed collections embed gitCollectionUrl in every generated doc. It powers the Open in Bruno clone deeplink, but users distributing docs publicly may not want a repo URL exposed, especially a private one. No opt-out today.

Every request is always rendered. Teams publishing docs need to leave out WIP, internal-only, or unfinished requests. Today that means cleaning the collection first or leaking those endpoints.

### Fix

Added a Generate Documentation modal that lets you filter by tags, pick which environments to include, and turn on a git repo link, then builds the HTML from those choices. The core generation logic (building the docs, tag filtering, environment selection, and git url handling) was moved into bruno-common so the same code can be reused by the CLI in a follow up PR.

### Screenshots

| Before | 

image

After |

image

image

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added advanced documentation options to select requests or filter by included and excluded tags.
  - Added tag autocomplete, validation, duplicate detection, keyboard navigation, and accessible controls.
  - Added optional Git repository URL inclusion in generated documentation.
  - Improved environment selection visuals with an accompanying icon.

- **Bug Fixes**
  - Prevented invalid Git remote values from appearing in generated documentation.
  - Improved handling of unset collection versions and tag metadata.

- **Tests**
  - Added comprehensive coverage for filtering, Git links, tag controls, and generated documentation.


packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+  const menuId = `${baseId}-listbox`;
+  const optionId = (index) => `${baseId}-option-${index}`;
+  const [text, setText] = useState('');
+  const [error, setError] = useState('');
+  const [isOpen, setIsOpen] = useState(false);
+  const [highlighted, setHighlighted] = useState(-1);
+  const [menuStyle, setMenuStyle] = useState(null);
+  const fieldRef = useRef(null);
+
+  const suggestions = tagsHintList.filter((tag) => tag.toLowerCase().includes(text.trim().toLowerCase()));
```
vasharma05-bruno, line 27: Can `tagsHintList` be undefined in any case?

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+  useLayoutEffect(() => {
+    if (!isOpen) return undefined;
+    positionMenu();
+    window.addEventListener('scroll', positionMenu, true);
+    window.addEventListener('resize', positionMenu);
+    return () => {
+      window.removeEventListener('scroll', positionMenu, true);
+      window.removeEventListener('resize', positionMenu);
+    };
+  }, [isOpen, positionMenu]);
```
vasharma05-bruno, line 48: Adding event listeners in the `useLayoutEffect` is not needed, since this will work on the syncronous and blocking way. So the initial positioning must happen with layoutEffect, and other event listeners can go into `useEffect`

```react
useLayoutEffect(() => {
    if (!isOpen) return undefined;
    positionMenu();
}, [isOpen, positionMenu])

useEffect(() => {
    window.addEventListener('scroll', positionMenu, true);
    window.addEventListener('resize', positionMenu);
    return () => {
      window.removeEventListener('scroll', positionMenu, true);
      window.removeEventListener('resize', positionMenu);
    };
  }, [positionMenu]);
```

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+          onChange={handleInputChange}
+          onKeyDown={handleKeyDown}
+          onFocus={() => setIsOpen(true)}
+          onBlur={() => {
+            setIsOpen(false);
+            setHighlighted(-1);
+          }}
+          role="combobox"
+          aria-label={ariaLabel}
+          aria-expanded={!!showMenu}
```
vasharma05-bruno:
```suggestion
          aria-expanded={Boolean(showMenu)}
```

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+          aria-controls={showMenu ? menuId : undefined}
+          aria-activedescendant={showMenu && highlighted >= 0 ? optionId(highlighted) : undefined}
```
vasharma05-bruno, line 133:
```suggestion
          aria-controls={showMenu && menuId}
          aria-activedescendant={showMenu && highlighted >= 0 && optionId(highlighted)}
```

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+                onMouseDown={(e) => {
+                  e.preventDefault();
+                  addTag(suggestion);
+                }}
+                onMouseEnter={() => setHighlighted(index)}
+              >
+                {suggestion}
+              </li>
+            ))}
+          </Menu>,
```
vasharma05-bruno, line 164: Should we use the `MenuDropdown` component here?

vasharma05-bruno: Highlighting, checking, and more is already implemented in it.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+        <span className="docs-tag-error" role="alert">
+          {error}
+        </span>
+      )}
+
+      {tags.length > 0 && (
+        <ul className="docs-tag-list">
+          {tags.map((tag) => (
+            <li key={tag} className="docs-tag-item">
+              <IconTag size={16} className="docs-tag-icon" aria-hidden="true" strokeWidth={1.33} />
```
vasharma05-bruno, line 178: We use `1.5` in most of the places. Please confirm with the designs.
Thanks!

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
```diff
@@ -0,0 +1,195 @@
...
+                {tag}
+              </span>
+              <button
+                type="button"
+                className="docs-tag-remove"
+                title="Remove tag"
+                aria-label={`Remove ${tag}`}
+                onClick={() => handleRemoveTag(tag)}
+              >
+                <IconX size={12} strokeWidth={2} aria-hidden="true" />
```
vasharma05-bruno: Above we are using strokeWidth = 1.33 and here we are using 2.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/StyledWrapper.js
```diff
@@ -0,0 +1,129 @@
...
+    border: 1px solid ${(props) => props.theme.border.border0};
+    border-radius: ${(props) => props.theme.border.radius.base};
+  }
+
+  .docs-tag-icon {
+    width: 1rem;
+    height: 1rem;
+    color: ${(props) => props.theme.colors.text.subtext0};
+    stroke-width: 1.33;
+    flex-shrink: 0;
```
vasharma05-bruno: Aren't these property already coming from the Tabler Icon? We are defining the size and strokeWidth, so I believe we can remove the `width`, `height` and `fiex-shrink` and `stroke-width` from here.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/StyledWrapper.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  .docs-tag-text {
+    font-size: ${(props) => props.theme.font.size.base};
+    font-weight: 400;
```
vasharma05-bruno: No need to define font-weight 400

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/StyledWrapper.js
```diff
@@ -0,0 +1,129 @@
...
+    background: none;
+    border: none;
+    cursor: pointer;
+    color: ${(props) => props.theme.colors.text.subtext0};
+    flex-shrink: 0;
+
+    svg {
+      width: 0.75rem;
+      height: 0.75rem;
+    }
```
vasharma05-bruno: Ditto, why are we doing the same thing from React component and also from stykes

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/IncludeExcludeTags/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+    if (include.includes(trimmedTag)) {
+      return 'Tag is already in the include list';
+    }
+    if (exclude.includes(trimmedTag)) {
+      return 'Tag is already in the exclude list';
+    }
+  };
+
+  const handleAddTag = ({ tag, to }) => {
```
vasharma05-bruno: We can take the values as args


```suggestion
  const handleAddTag = (tag, to) => {
```

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/IncludeExcludeTags/index.spec.js
```diff
@@ -0,0 +1,68 @@
+import '@testing-library/jest-dom';
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { ThemeProvider } from 'styled-components';
+import themes from 'themes/index';
+import IncludeExcludeTags from './index';
+
+const renderWithTheme = (ui) => render(<ThemeProvider theme={themes.light}>{ui}</ThemeProvider>);
```
vasharma05-bruno, line 8: I discussed with Utkarsh, we don't need to use the `renderWithTheme`, and can directly use the `render`

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <StyledWrapper className="advanced">
+      <button
+        type="button"
+        className="advanced-toggle"
+        onClick={() => setOpen((prev) => !prev)}
+        aria-expanded={open}
+        data-testid="docs-advanced-toggle"
+      >
+        <IconChevronRight className="advanced-chevron" size={16} strokeWidth={1.36} aria-hidden="true" />
```
vasharma05-bruno, line 31: These strokeWidth values are varying very much

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+                      data-tooltip-id="docs-tags-hint"
+                      data-tooltip-content={TAGS_HINT}
+                    >
+                      <IconInfoCircle size={16} aria-hidden="true" />
+                    </span>
+                  </button>
+                </div>
+              </div>
+
+              <Tooltip id="docs-tags-hint" style={{ maxWidth: '16rem' }} />
```
vasharma05-bruno: Can we move this to style file

packages/bruno-app/src/hooks/useCollectionGitRemoteUrl/index.js
```diff
@@ -0,0 +1,35 @@
...
+      .then((url) => {
+        if (!active) return;
+        setGitCollectionUrl(url || null);
+        setIsResolved(true);
+      })
+      .catch(() => {
+        if (!active) return;
+        setGitCollectionUrl(null);
+        setIsResolved(true);
+      });
```
vasharma05-bruno, line 26: Is there a cleanup for this invokation function @sid-bruno, that we can use in the return statement of the useEffect?

packages/bruno-common/src/api-docs/index.ts
```diff
@@ -0,0 +1,197 @@
+import isRequestTagsIncluded from '../tags';
+
+export interface TaggedItem {
+  type?: string;
+  tags?: string[];
+  items?: TaggedItem[];
+  [key: string]: unknown;
+}
+
+export interface NamedEnvironment {
```
vasharma05-bruno, line 10: Should we use the `Environment` type from @opencollection/types`? @sid-bruno

packages/bruno-common/src/api-docs/index.ts
```diff
@@ -0,0 +1,197 @@
...
+const isFolder = (item: TaggedItem): boolean => item.type === 'folder' || Array.isArray(item.items);
+const isRequestItem = (item: TaggedItem): boolean =>
+  typeof item.type === 'string' && item.type.endsWith('-request');
+
+export const filterRequestItemsByTags = <T extends TaggedItem>(
+  items: T[],
+  includeTags: string[] = [],
+  excludeTags: string[] = []
+): T[] => {
+  if (includeTags.length === 0 && excludeTags.length === 0) {
```
vasharma05-bruno, line 65:
```suggestion
  if (!includeTags && !excludeTags) {
```

sachin-bruno: includeTags and excludeTags are always arrays, never undefined.
The params default to [] (includeTags: string[] = []), so a missing argument becomes [].
The only caller passes options.tags?.include ?? [] and options.tags?.exclude ?? [], so even when tags is not set they arrive as `[]`

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/index.jsx
sachin-bruno: The value is from figma.

packages/bruno-common/src/api-docs/index.ts
sachin-bruno: bruno-common is a zero-dependency leaf, and this module stays decoupled from the collection types on purpose.

packages/bruno-app/src/hooks/useCollectionGitRemoteUrl/index.js
sachin-bruno: The effect already returns a cleanup a couple of lines down: return () => { active = false; }.
Electron's ipcRenderer.invoke has no cancel or abort API (no AbortSignal), so we can't tear down the call itself

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/IncludeExcludeTags/index.spec.js
sachin-bruno: I gave this a try and plain `render` fails all the tests here with `TypeError: Cannot read properties of undefined (reading 'size')`. The component's StyledWrapper reads props.theme.font.size.sm and props.theme.colors.text.subtext2, and since there's no global ThemeProvider in jest.setup, styled-components falls back to an empty theme and throws.

The sibling specs (DocsTagList, Advanced) wrap in ThemeProvider the same way, and around 22 specs across bruno-app follow this pattern, so I would like to keep renderWithTheme here.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/IncludeExcludeTags/index.jsx
sachin-bruno: Both work, so mostly preference, but I would keep the object form: the named key documents the intent (`{ tag, to: 'include' }` vs `(tag, 'include')`) and keeps handleAddTag/handleRemoveTag symmetric. Also the four call sites pass `{ tag, to }`, so changing just the signature would break them.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
sachin-bruno: Values taken from figma

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
sachin-bruno: Value taken from figma only

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
sachin-bruno: MenuDropdown is a menu widget role="menu", opened by a trigger with a static items list, whereas this is a combobox/typeahead: a text input with live filtered suggestions. MenuDropdown has no text input or filtering, so we would be bolting those on and exposing menu semantics for what should be a combobox, which is an accessibility regression. The highlighting/checkmark bits do overlap, but those are menu features, the filtering and combobox wiring (the bulk here) isn't something MenuDropdown provides.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
sachin-bruno: When the menu is closed, showMenu is `false` (not undefined), and React renders `aria-controls={false}` as `aria-controls="false"`, a broken reference to a non-existent element. The `? menuId : undefined` form omits the attribute instead, which is what we want. Hence this is valid in this case

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
sachin-bruno: Good point. Done

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
sachin-bruno: It can't be undefined here. The prop defaults to [] tagsHintList = [], so an omitted/undefined value becomes an empty array before the .filter. And the only caller (IncludeExcludeTags) always passes availableTags.filter(), which is always an array.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
vasharma05-bruno: I missed that `showMenu` is boolean value. Thanks!

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/DocsTagList/index.jsx
vasharma05-bruno: I still feel that we can update the `role` in the menudropdown, if it's a multicheck, since we are copying the whole component

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/IncludeExcludeTags/index.jsx
vasharma05-bruno: I mean to say, make the change similar for both functions, we are creating a new objkect, then destructing it to get the values, the same can be done with function arguments

packages/bruno-app/src/hooks/useCollectionGitRemoteUrl/index.js
vasharma05-bruno: I meant if we have something like `removeEventListener`, but I understand it's an invoking function, not a eventListener.

packages/bruno-common/src/api-docs/index.ts
vasharma05-bruno: Are we using the exact same type as from opencollection

packages/bruno-common/src/api-docs/index.ts
vasharma05-bruno: I remembered [] is Falsy, but that's in Python, not JS. My bad 🤦

packages/bruno-common/src/api-docs/index.ts
sachin-bruno: yes same type decalred in bruno common

vasharma05-bruno (approved):
Looks good to me!

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/Advanced/IncludeExcludeTags/index.jsx
sachin-bruno: Done

packages/bruno-common/src/api-docs/index.ts
```diff
@@ -0,0 +1,199 @@
...
+  return openCollection;
+};
+
+export const buildApiDocsHtml = (
+  collectionName: string,
+  embeddedCollectionData: string,
+  options: ApiDocsHtmlOptions = {}
+): string => {
+  const baseUrl = options.rendererBaseUrl ?? DEFAULT_RENDERER_BASE_URL;
+  const gitLine = options.gitCollectionUrl
```
arpit-bruno, line 144: The remote url goes into the html as is. If the remote has a token in it like `https://ghp_xxx@github.com/org/repo.git` the token ends up in the generated file.

The AC doesnt cover this case so checking if this is what we want. @abhishek-bruno  should we strip the username and password from the url before embedding or skip the link when they are present?

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/index.js
```diff
@@ -105,63 +77,45 @@ const GenerateDocumentation = ({ onClose, collectionUid }) => {
...
     [environments]
   );
 
+  const availableTags = useMemo(
+    () => getUniqueTagsFromItems(collection?.items || [], { includeDrafts: false }),
+    [collection?.items]
+  );
+  const [filterByTags, setFilterByTags] = useState(false);
+  const [docTags, setDocTags] = useState({ include: [], exclude: [] });
+  const [includeGitLink, setIncludeGitLink] = useState(true);
```
arpit-bruno, line 91: This resets every time the modal opens. I turned the git link off and generated doc it reopened and it was back on. 
Same case for the tag filters.

arpit-bruno (changes requested):

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/index.js
sachin-bruno: Confirmed with @abhishek-bruno and we are not persisting the states user selected in the modal. The tags and git link will get reset after closing.

packages/bruno-common/src/api-docs/index.ts
abhishek-bruno: We need to strip it at all cases. No sensitive data should be leaked into the exported file.

packages/bruno-app/src/components/Sidebar/Collections/Collection/GenerateDocumentation/index.js
arpit-bruno: Marking this as resolved

arpit-bruno (approved):

bijin-bruno (approved):

PR #9073, opened 2026-08-24
feat(cli): Generate docs through cli

sachin-bruno:
Ref: BRU-2546

### Problem

Documentation generation is UI-only today (Collection > Generate Documentation). There is no way to produce the HTML docs from the command line, so doc generation cannot run in scripts or release/CI pipelines.

### Fix

Added the bru docs generate command that reuses the shared generator from bruno-common, so the CLI and the app produce the exact same docs. It reads the collection from disk, applies the tag and environment filters, orders the output to match the app sidebar, and writes the HTML to the chosen output file (defaults to collectionname-documentation.html).

### Screenshots

image

image

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Added `bru docs generate` for standalone HTML API documentation.
  * Supports environment and tag filtering, repeated or comma-separated values, custom output paths, all environments, and optional Git links.
  * Loads environments from Bruno, YAML, and JSON files with alphabetical ordering.

* **Bug Fixes**
  * Improved validation, duplicate handling, filtering, and error messages.
  * Added clearer handling for malformed, unsupported, or missing environment files.
  * Standardized collection folder ordering.

* **Tests**
  * Added comprehensive coverage for documentation generation and environment handling.


packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,138 @@
...
+    .option('format', {
+      alias: 'f',
+      type: 'string',
+      default: 'html',
+      description: 'Output format; only "html" is supported today'
+    })
+    .option('envs', {
+      type: 'string',
+      description: 'Comma-separated environment names to embed'
+    })
```
vasharma05-bruno, line 27: Suggestion, instead of comma separated names, we can go with an array: `bru generate docs --env local --env uat --env prod` as well.

vasharma05-bruno: Can be used for every option which requires comma separated strings

packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,138 @@
...
+    .example('$0 docs generate --exclude-tags WIP --no-git-link', 'Drop WIP requests and omit the git link');
+};
+
+const handler = async (argv) => {
+  try {
+    const format = String(argv.format || 'html').toLowerCase();
+    if (format !== 'html') {
+      console.error(chalk.red(`Invalid output format "${format}". Only "html" is supported.`));
+      process.exit(EXIT_STATUS.ERROR_INCORRECT_OUTPUT_FORMAT);
+    }
```
vasharma05-bruno: Why are we giving the format as an option, if this is the only option present.

packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,138 @@
...
+      process.exit(EXIT_STATUS.ERROR_GENERIC);
+    }
+
+    const availableEnvNames = new Set(collection.environments.map((env) => env.name));
+    const missingEnv = [...includeEnvs, ...excludeEnvs].find((name) => !availableEnvNames.has(name));
+    if (missingEnv) {
+      console.error(chalk.red('Environment not found: ') + chalk.dim(missingEnv));
+      process.exit(EXIT_STATUS.ERROR_ENV_NOT_FOUND);
+    }
+
```
vasharma05-bruno:
```suggestion
    const availableEnvNames = new Set(collection.environments.map((env) => env.name));
    const unconfiguredEnv = [...includeEnvs, ...excludeEnvs].find((name) => !availableEnvNames.has(name));
    if (unconfiguredEnv) {
      console.error(chalk.red('Environment not found: ') + chalk.dim(unconfiguredEnv));
      process.exit(EXIT_STATUS.ERROR_ENV_NOT_FOUND);
    }
```

Suggestion: Can't we find all the missing/ unconfigured envs in one go, and list them? Let's say a user has spelling errors in 3 names, they will make the change, and will get to know about each error on every run.

packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,138 @@
...
+    if (includeEnvs.length > 0) {
+      const envByName = new Map(collection.environments.map((env) => [env.name, env]));
+      collection.environments = includeEnvs.map((name) => envByName.get(name));
+    } else if (excludeEnvs.length > 0) {
+      const excluded = new Set(excludeEnvs);
+      collection.environments = collection.environments.filter((env) => !excluded.has(env.name));
+    } else {
+      collection.environments = [];
```
vasharma05-bruno: I didn't understand this, if I want to import all environments, shouldn't it be easy for me to not define `envs` as a list of all envs and no `exclude-envs`, hence, collection.environments should be all environments?

Or give an option like `--envs=all` to the user.

packages/bruno-cli/src/utils/collection.js
```diff
@@ -39,7 +39,7 @@ const getFolderRoot = (dir, format) => {
   return parseFolder(fs.readFileSync(folderPath, 'utf8'), { format });
 };
 
-const createCollectionJsonFromPathname = (collectionPath) => {
+const createCollectionJsonFromPathname = (collectionPath, { sortFolders = sortByNameThenSequence } = {}) => {
```
vasharma05-bruno: You made this change to allow passing different sorting functions, then from generate.js, you are passing the same `sortByNameThenSequence` as the `sortFolders`, I didn't understand the reason for this change.

// generate.js
const { generateApiDocsHtml, getApiDocsFileName, resolveCollectionVersion, sortByNameThenSequence: sortFolders } = require('@usebruno/common');
    const collection = createCollectionJsonFromPathname(collectionPath, { sortFolders });

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: Comma separated is the requirement from the product side.

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: Currently we are supporting only one format which is html. In future there can be multiple formats, so accomodated the code for this with html check. This is also confirmed with the product

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: In the docs modal, the current behaviour is all the environments are unselected by default. User has to select the environments. Some environments may contain secrets or values that the user dont want to share. So the same behaviour is kept in cli as well. User has to include envs by name to generate docs with the selected envs. if no envs are mentioned then the generated docs will not contain any envs.

Giving an option something like --envs=all to the user is a good option. This can be checked with @abhishek-bruno. I can raise this as a separate PR if needed.

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: Good point. Done

packages/bruno-cli/src/commands/docs/generate.js
vasharma05-bruno: We can ask @abhishek-bruno to remove this in future for now.

packages/bruno-cli/src/commands/docs/generate.js
vasharma05-bruno: The same review is from code rabbit, and in constrast to UI, where the User has a list of environments they can directly view, the CLI doesn't provide that.

Still there are 2 ways:
1. Either allow not passing either `envs` and `excluded-envs` and then assume it should be `all` by default. OR
2. Enforce passing either of the args, and alongside allow `envs=all`

Thanks!
CC: @abhishek-bruno!

packages/bruno-cli/src/utils/collection.js
sachin-bruno: done

packages/bruno-cli/src/commands/docs/generate.js
abhishek-bruno: Giving an option like `--envs-all` or `'--all-envs'` to the user is much needed. Thanks for the suggestion @vasharma05-bruno. We can include this in the scope of the same ticket @sachin-bruno

packages/bruno-cli/src/commands/docs/generate.js
abhishek-bruno: @sachin-bruno No hard requirement from my side for comma separated. We can go with the best approach for our usecase.

@sreelakshmi-bruno @bijin-bruno @sid-bruno @arpit-bruno  What do you think?

packages/bruno-cli/src/commands/docs/generate.js
abhishek-bruno: We can remove the --format option for now. We can add it once we support multiple output formats.

packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,144 @@
...
+      process.exit(EXIT_STATUS.ERROR_INVALID_FILE);
+    }
+
+    const includeTags = splitCsv(argv.tags);
+    const excludeTags = splitCsv(argv.excludeTags);
+    const includeEnvs = [...new Set(splitCsv(argv.envs))];
+    const excludeEnvs = [...new Set(splitCsv(argv.excludeEnvs))];
+    const allEnvs = Boolean(argv.allEnvs);
+
+    const conflictingTag = findConflict(includeTags, excludeTags);
```
vasharma05-bruno: Ditto, we can list all the conflicting tags altogether

vasharma05-bruno: We can update the `findConflict` to use `filter` instead of `find`

vasharma05-bruno (approved):
Just one clarification needed, else looks good to me!

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: done

packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,175 @@
...
+    const availableEnvNames = new Set(collection.environments.map((env) => env.name));
+    const missingEnvs = [...includeEnvs, ...excludeEnvs].filter((name) => !availableEnvNames.has(name));
+    if (missingEnvs.length > 0) {
+      console.error(chalk.red('Environments not found: ') + chalk.dim(missingEnvs.join(', ')));
+      process.exit(EXIT_STATUS.ERROR_ENV_NOT_FOUND);
+    }
+
+    if (includeEnvs.length > 0) {
+      const envByName = new Map(collection.environments.map((env) => [env.name, env]));
+      collection.environments = includeEnvs.map((name) => envByName.get(name));
```
arpit-bruno: Doubt here: env ordering isnt consistent across the flags. `--envs` keeps the order I list them for example: 
 With `--envs Development,QA,Staging,Production` the docs keep my order but `--all-envs` comes out alphabetical -> Development, Production, QA, Staging .

since `loadEnvironments` sorts by name and only the include path maps over my given flag order. 
Should all the flags order the same way for consistency ?

Image
Image

packages/bruno-cli/src/utils/environment.js
```diff
@@ -21,6 +25,43 @@ const parseEnvironmentJson = (parsed = {}) => {
...
+      try {
+        const envJson = parseEnvFile(filePath);
+        const name = envJson.name || path.basename(file, path.extname(file));
+        return { ...envJson, name, variables: envJson.variables || [] };
+      } catch (err) {
+        throw new Error(`environments/${file}: ${err.message}`);
+      }
+    });
+
+  return environments.sort((a, b) => a.name.localeCompare(b.name));
```
arpit-bruno, line 61: If two env files end up with the same name there is no dedup or warning shown. Lets say we have two prod envs with different values. `--all-envs` embeds both so the env picker shows prod twice and `--env prod ` silently picks one so now we could get the wrong prod's values without knowing.

arpit-bruno: Should we warn or show error on duplicate names when loading it?

packages/bruno-cli/src/utils/environment.js
```diff
@@ -21,6 +25,43 @@ const parseEnvironmentJson = (parsed = {}) => {
   return normalized;
 };
 
+const parseEnvFile = (filePath) => {
```
arpit-bruno, line 28: https://github.com/usebruno/bruno/blob/main/packages/bruno-cli/src/commands/run.js#L392
We already have this same parser `loadEnvFromFile` which does the same thing.

Can we keep just one copy? export this function and let run.js call it instead of having its own version

arpit-bruno (changes requested):

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: In the app all the envs are alphabetically ordered. When generating docs through the app through select all it gets downloaded in the same alphabhetical order. So followed the same for consistency in the cli when selecting all. When the user lists them one by one then the envs are listed by user's order.

packages/bruno-cli/src/utils/environment.js
sachin-bruno: In app we cannot create two envs with the same name. This case is not valid. I have handled one case where 
if the user enters `--envs prod, staging --env prod` in this case duplicate is handled and the docs will be generated with two envs prod and staging.

packages/bruno-cli/src/utils/environment.js
arpit-bruno: resolved

arpit-bruno (approved):

packages/bruno-cli/src/commands/docs/generate.js
```diff
@@ -0,0 +1,145 @@
...
+
+    const includeTags = splitCsv(argv.tags);
+    const excludeTags = splitCsv(argv.excludeTags);
+    const includeEnvs = [...new Set(splitCsv(argv.envs))];
+    const excludeEnvs = [...new Set(splitCsv(argv.excludeEnvs))];
+    const allEnvs = Boolean(argv.allEnvs);
+
+    const conflictingTags = findConflicts(includeTags, excludeTags);
+    if (conflictingTags.length > 0) {
+      console.error(chalk.red(`${pluralizeWord(conflictingTags.length, 'Tag')} cannot be both included and excluded: `) + chalk.dim(conflictingTags.join(', ')));
```
bijin-bruno: I think it is fine to have conflicts within include and exclude envs/tags.
We can just give precedence for `exclude-*` flag. eg: include: ['a','b','c'], exclude: ['b'] just means 'b' is excluded.

packages/bruno-cli/src/utils/common.js
```diff
@@ -42,8 +43,42 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) =>
...
+};
+
+const findConflicts = (include, exclude) => {
+  const excluded = new Set(exclude);
+  return [...new Set(include)].filter((name) => excluded.has(name));
+};
+
+const pluralizeWord = (count, word) => `${word}${count === 1 ? '' : 's'}`;
+
+const getGitRemoteUrl = (collectionPath) => {
```
bijin-bruno: Move to `utils/git.js` to be consistent with electron package structure.

packages/bruno-cli/src/utils/common.js
```diff
@@ -42,8 +43,42 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) =>
...
+  try {
+    const url = execSync('git remote get-url origin', {
```
bijin-bruno: We might need to handle the remote name instead of hardcoding to origin. Does the app currently hard codes the remote as 'origin'?

packages/bruno-cli/src/utils/common.js
```diff
@@ -42,8 +43,42 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) =>
...
+      cwd: collectionPath,
+      stdio: ['ignore', 'pipe', 'ignore']
+    })
```
bijin-bruno: Good to have a timeout in options

packages/bruno-cli/src/utils/common.js
```diff
@@ -42,8 +43,42 @@ const parseDataFromResponse = (response, disableParsingResponseJson = false) =>
   return { data, dataBuffer };
 };
 
+const splitCsv = (value) => {
```
bijin-bruno:
```suggestion
const parseListOption = (value) => {
```

packages/bruno-cli/tests/commands/docs-generate.spec.js
```diff
@@ -0,0 +1,486 @@
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+const { execSync } = require('child_process');
+const { describe, it, expect, beforeAll, afterAll, afterEach } = require('@jest/globals');
+const generate = require('../../src/commands/docs/generate');
+
+const FIXTURE = path.resolve(__dirname, '../runner/fixtures/collection-json-from-pathname/collection');
```
bijin-bruno: It's better to have own fixture to test different cases of generate docs. Lets add a collection fixture under `tests/commands/docs/fixtures/collection`.
This way cross test conflicts can also be avoided. It's okay to duplicate collection.

packages/bruno-cli/tests/commands/docs/generate.spec.js
bijin-bruno, line 1: Move to `tests/commands/docs/generate.js`

bijin-bruno (changes requested):

packages/bruno-cli/src/utils/common.js
sachin-bruno: Yes the app hardcodes origin as well. getCollectionGitRepoUrl in bruno-electron/src/utils/git.js uses git listRemote --get-url origin

packages/bruno-cli/src/utils/common.js
sachin-bruno: done

packages/bruno-cli/src/utils/common.js
sachin-bruno: done

packages/bruno-cli/tests/commands/docs-generate.spec.js
sachin-bruno: done

packages/bruno-cli/tests/commands/docs/generate.spec.js
sachin-bruno: done

packages/bruno-cli/src/utils/common.js
sachin-bruno: done

packages/bruno-cli/src/commands/docs/generate.js
sachin-bruno: done

arpit-bruno (approved):

bijin-bruno (dismissed):

arpit-bruno (approved):

bijin-bruno (approved):

PR #9081, opened 2026-08-25
fix(bruno-js): await async test() callbacks in developer mode

shubh-bruno:
### Description

BRU-3714

Fixes a bug where async `test()` callbacks were being dropped in Developer Mode (NodeVM sandbox) when the script didn't await them itself.

### Problem

In Developer Mode, if a `test()` callback was async and the script moved on without awaiting it, its result could be lost entirely, so it never showed up in the Tests tab. This happened in pre-request, post-response, and test scripts, at the request, folder, and collection level.

### Fix

Every `test()` call is now remembered, and before the run reports its final results, it waits for all of those calls to actually finish - even the ones the script itself never waited on. This is the same approach Safe Mode (QuickJS) uses internally keep draining in a loop with `splice` + `Promise.all` - take everything currently pending, wait for it, check again in case a test registered another test while waiting, and repeat until nothing's left - instead of a single one-time wait.

This wait is set up separately for each script phase - pre-request, post-response, and the tests tab each get their own independent `pendingTestPromises` array.

### Before vs After

https://github.com/user-attachments/assets/74da6c1d-c098-461c-b083-1a9d70ec2892

#### Contribution Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Asynchronous test callbacks are now properly awaited before results are displayed.
  - Async tests work reliably across pre-request, post-response, test, folder-level, and collection-level scripts.
  - Tests that await additional requests now report accurate results.
  - Multiple concurrent asynchronous tests are handled correctly, including delayed assertions and failures.

- **UI Improvements**
  - Response-pane sections and overflow tabs now provide more reliable identification for automated interactions.


packages/bruno-js/src/utils/results.js
```diff
@@ -20,12 +19,87 @@ const getResultsSummary = (results) => {
   return summary;
 };
 
+const TEST_AWAIT_TIMEOUT_MS = 5000;
```
sid-bruno: change this to polling duration (eg: 2000, 2 seconds) for now instead of a timeout.

We don't have a concept of cancelling the tests in QuickJS either, we wait for the queue to drain, I'd prefer the same thing here.

packages/bruno-cli/tests/integration/nodevm-async-callbacks/run-nodevm-async-callbacks.spec.js
```diff
@@ -0,0 +1,81 @@
...
+
+const FIXTURE_COLLECTION = path.join(__dirname, 'fixtures', 'collection');
+
+// `logResults` prints a checkmark and its message as two separately-chalked pieces
+// (`chalk.green('✓ ') + chalk.dim(message)`), so if the invoking shell has color forced
+// on (e.g. FORCE_COLOR set), the raw output has ANSI codes spliced between them - breaking
+// a literal `toContain('✓ message')` check even though the text reads the same on screen.
+// Stripping codes here keeps the assertions below agnostic to the runner's environment.
+
+const stripAnsi = (text) => text.replace(/\x1b\[[0-9;]*m/g, '');
```
sid-bruno: A util for this should already be there, if it's inlined in another tests, move it out into a shared one instead of inlining here again

packages/bruno-js/src/utils/results.js
```diff
@@ -20,12 +20,51 @@ const getResultsSummary = (results) => {
...
+      let settled;
+      try {
+        settled = await Promise.race([Promise.all(batch).then(() => true), poll]);
+      } finally {
+        clearTimeout(pollId);
+      }
+
+      if (settled) cursor = batchEnd;
+    }
+  };
```
sid-bruno: QuickJS already drains with splice + Promise.all; polling adds timers without new behavior.

```suggestion
  /**
   * Waits for every test() call registered so far to settle, including ones registered
   * mid-wait. Mirrors QuickJS waitForPendingDeferreds().
   */
  const waitForPendingTests = async () => {
    while (pendingTestPromises.length) {
      const batch = pendingTestPromises.splice(0);
      await Promise.all(batch);
    }
  };
```

sid-bruno (approved):

PR #9123, opened 2026-08-31
feat(sidebar): allow reordering response examples with drag and drop

### LINKED ISSUE #9105: Allow reordering of saved example results with drag and drop

### This feature

- [ ] Blocks me from using Bruno
- [x] Would improve my quality of life in Bruno
- [ ] Is something I've never seen an API client do before

### 🚀 Describe the feature you want to add, and how it would change your usage of Bruno

I often save multiple results (called "examples") from previous calls, since it's useful for me to have snapshots of what the results were at different points of time. I often name these on the dates, but not necessarily. It seems that all new saved results ("examples") are systematically placed at the bottom of the list. 

I would prefer to be able to reorder these and have not found a way to do so (short of modifying the yml), either by drag-and-drop or by going into the "..." options. 

Image

For context, the previous saved examples I have on my screenshot were previously created, reordered, then imported from Postman (that's why the order is not chronological). 
(This feature was possible with Postman via drag and drop). 

### 📷 Mockups or Images of the feature

_No response_

fantpmas:
## Summary

Response examples could only be created, renamed, cloned and deleted — their order in the sidebar was fixed to creation order. This makes them draggable so they can be reordered within their request.

Fixes [#9105](https://github.com/usebruno/bruno/issues/9105)

Examples are stored as an ordered array inside the request file and both formats persist that order (`bru` writes consecutive `example` blocks, `yml` an `examples` list), so a reorder is just moving the entry and saving. `renderer:save-request` already re-syncs the position-derived example uid cache, so an open example tab keeps pointing at the same example after a reorder.

Dragging an example onto a *different* request is deliberately out of scope — that moves it between files and needs its own design.

## Changes

- `moveResponseExample` reducer plus a shared `getReorderedExampleUids` helper, so the reducer and the drop handler cannot disagree about what a drop means
- `determineExampleDrop` for the above/below hover split, alongside the existing `determineCollectionItemDrop`
- drag/drop wiring on `ExampleItem` under its own `response-example` dnd type, so example drags and request/folder drags stay invisible to each other
- drops originating from another request are rejected, and a drop that lands an example back in its own slot is a no-op — leaving the request clean rather than dirtying it and rewriting the file
- the dragged item carries `name`/`type`, without which the shared drag layer renders no preview at all
- no new CSS — example rows already inherit the drop-indicator styles from `CollectionItem`'s wrapper
- `findCollectionDir` moved into `tests/utils` (first commit) so more than one test area can use it

## Demo

https://github.com/user-attachments/assets/1b2f7dbb-8c18-4f1f-b588-48765a5bef07

## Testing

**Unit — 28 new** (`packages/bruno-app`): reducer placement above/below, the remove-then-resolve off-by-one, draft seeding from saved examples, unsaved draft edits preserved, unknown/self uids and no-change drops leaving the request clean, uid preservation; plus both new helpers.

**E2E — 5 new** (`tests/response-examples/reorder-examples.spec.ts`):
- reorder + resulting on-disk order, parameterised over both `bru` and `yml`
- a no-change drop leaves the request file byte-identical
- an open example tab still shows the same example after a reorder
- dragging an example onto another request leaves both requests untouched

No `page.waitForTimeout` — the negative assertions synchronise on a drop that *does* save, which cannot complete before an earlier save would have.

**Suites run:** `npm run lint` 0 errors; `bruno-app` unit 1929/1929; `tests/response-examples/` + `tests/naming-collisions/` 87 passed.

## Summary by CodeRabbit

- **New Features**
  - Added drag-and-drop reordering for response examples within a request.
  - Examples can be placed above or below another example.
  - Reordered examples are saved across supported collection formats.
  - Invalid cross-request and no-op drops are prevented.

- **Bug Fixes**
  - Preserved the active example when response examples are reordered.
  - Updated the displayed order immediately after saving.

- **Tests**
  - Added coverage for reordering, persistence, invalid drops, and unchanged no-op operations.


packages/bruno-app/src/providers/ReduxStore/slices/collections/exampleReducers.js
```diff
@@ -163,6 +164,34 @@ export const deleteResponseExample = (state, action) => {
...
+  if (!item.draft) {
+    item.draft = cloneDeep(item);
+  }
+
+  if (!item.draft.examples) {
+    item.draft.examples = item.examples ? cloneDeep(item.examples) : [];
+  }
+
+  const draftExamples = item.draft.examples;
+  item.draft.examples = reorderedUids.map((uid) => draftExamples.find((e) => e.uid === uid));
```
saravana512, line 192: shall we use 

const examplesByUid = new Map(draftExamples.map((example) => [example.uid, example]));

item.draft.examples = reorderedUids.map((uid) => examplesByUid.get(uid));

packages/bruno-app/src/utils/collections/index.js
```diff
@@ -1604,6 +1604,51 @@ export const determineCollectionItemDrop = ({ item, hoverBoundingRect, clientOff
...
+ * `targetExampleUid`.
+ *
+ * Returns null when the drop should be ignored: either uid missing from this request's
+ * examples (which is how a drag that started under a different request is rejected), or a
+ * drop that lands the example back in the slot it already occupies. Callers rely on null to
+ * mean "do nothing" — no draft, no save, no file write.
+ *
+ * @returns {string[]|null} example uids in their new order.
+ */
+export const getReorderedExampleUids = ({ examples, draggedExampleUid, targetExampleUid, dropType }) => {
```
saravana512, line 1636: Too much expensive operations in this function 

export const getReorderedExampleUids = ({
  examples,
  draggedExampleUid,
  targetExampleUid,
  dropType
}) => {
  if (!examples?.length || draggedExampleUid === targetExampleUid) {
    return null;
  }

  const uids = new Array(examples.length);
  let draggedIndex = -1;
  let targetIndex = -1;

  for (let i = 0; i < examples.length; i++) {
    const uid = examples[i].uid;
    uids[i] = uid;

    if (uid === draggedExampleUid) draggedIndex = i;
    if (uid === targetExampleUid) targetIndex = i;
  }

  if (draggedIndex === -1 || targetIndex === -1) {
    return null;
  }

  // Calculate the final position after removing the dragged item.
  let insertAt = targetIndex - (draggedIndex < targetIndex ? 1 : 0);

  if (dropType === 'below') {
    insertAt++;
  }

  if (insertAt === draggedIndex) {
    return null;
  }

  const reorderedUids = uids.slice();
  reorderedUids.splice(draggedIndex, 1);
  reorderedUids.splice(insertAt, 0, draggedExampleUid);

  return reorderedUids;
};

tests/naming-collisions/utils.ts
```diff
@@ -29,17 +31,6 @@ export const listRequestFiles = (dir: string, ext = '.bru'): string[] => {
   return out;
 };
 
-export const findCollectionDir = (testDir: string): string => {
```
saravana512, line 32: good to remove nice one, fs.readdirSync this takes verymuch expensive it blocks the io

tests/response-examples/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Both formats persist examples as an ordered sequence — `bru` writes consecutive
+ * `example { }` blocks, `yml` writes an `examples:` list — so first-appearance order in the
+ * raw text is the saved order. Comparing raw positions rather than parsing keeps this helper
+ * honest about what actually landed on disk and works unchanged for both formats.
+ *
+ * Requires example names that are not substrings of each other or of the request name.
+ */
+export const exampleOrderOnDisk = (filePath: string, names: string[]): string[] => {
+  const content = fs.readFileSync(filePath, 'utf8');
```
saravana512, line 22: readFileSync this is blocks the io please use async file read oerations or stream



import { promises as fs } from 'node:fs';

export const exampleOrderOnDisk = async (
  filePath: string,
  names: string[]
): Promise<string[]> => {
  const content = await fs.readFile(filePath, 'utf8');

  return names
    .map((name) => {
      const index = content.indexOf(name);

      if (index === -1) {
        throw Object.assign(
          new Error(`Example "${name}" is missing from ${filePath}`),
          { code: MISSING_EXAMPLE }
        );
      }

      return { name, index };
    })
    .sort((a, b) => a.index - b.index)
    .map(({ name }) => name);
};

tests/response-examples/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Use this to assert a file was **not** rewritten. Content alone cannot show that: a reorder
+ * resolving to the order already on disk would be saved with byte-identical content, so only the
+ * mtime separates "never written" from "written again the same". Autosave is off in tests, so
+ * nothing else touches these files mid-test.
+ *
+ * Don't invert it to prove a file *was* rewritten — mtime granularity varies by filesystem, so a
+ * timestamp is not guaranteed to advance between two nearby writes. Assert the persisted order or
+ * content changed instead.
+ */
+export const fileSnapshot = (filePath: string): { content: string; mtimeMs: number } => ({
```
saravana512, line 48: dont use sync

tests/utils/collection-files.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Locate the collection directory Bruno created inside a throwaway test dir.
+ *
+ * `createTmpDir` hands back the parent, and the collection lands in a subdirectory named
+ * after itself, so specs that assert on files have to find it by its marker file. The marker
+ * differs per format: `bru` collections carry `bruno.json`, `yml` collections
+ * `opencollection.yml`.
+ */
+export const findCollectionDir = (testDir: string): string => {
+  const markers = ['bruno.json', 'opencollection.yml'];
+  for (const entry of fs.readdirSync(testDir, { withFileTypes: true })) {
```
saravana512, line 14: dont use sync

saravana512 (changes requested):

packages/bruno-app/src/providers/ReduxStore/slices/collections/exampleReducers.js
fantpmas: At the sizes that actually occur, building a Map is slower. Hashing 2n strings costs more than a linear scan over a tiny array.

packages/bruno-app/src/utils/collections/index.js
fantpmas: 8 declarative lines vs ~25 imperative lines with two mutable index accumulators, for no measurable gain.

tests/response-examples/utils.ts
fantpmas: These are Playwright test helpers, not app or Electron-main code. There's no event loop serving anything. A test body is strictly sequential, and the helper runs between UI actions.

tests/utils/collection-files.ts
fantpmas: These are Playwright test helpers, not app or Electron-main code. There's no event loop serving anything. A test body is strictly sequential, and the helper runs between UI actions.

tests/response-examples/utils.ts
fantpmas: These are Playwright test helpers, not app or Electron-main code. There's no event loop serving anything. A test body is strictly sequential, and the helper runs between UI actions. There is nothing to unblock.

PR #9129, opened 2026-08-31
feat(headers): show collection and folder headers in the inherited accordian

pooja-bruno:
BRU-4027

 ### Description

Show collection and folder in the request Headers tab so you can see what will be sent without opening collection or folder settings.

### Problem

Collection and folder headers are applied at send time, but they are not visible on the request. You have to open collection or folder settings to find the source, and it is easy to miss overrides, duplicates, and Bruno’s runtime defaults.

### Fix

On HTTP requests, the Headers tab can show an Inherited Headers accordion (collection + folder headers, plus Bruno defaults).

### Screenshots

Before

image

After

https://github.com/user-attachments/assets/96b6c04f-2813-4a0e-8419-8924671e86f0

#### Contribution Checklist:

## New Features
- Added inherited HTTP headers from collection and folder settings to the request headers pane.
- Added controls to show, hide, and count inherited headers.
- Added navigation from inherited headers to their source settings rows, with focus highlighting.
- Added sticky table headers and section rows, plus reduced-motion support for row highlights.
- Added automatic scrolling and pane repositioning when viewing inherited headers.

## Bug Fixes
- Resolved inherited headers by nearest source, honoring case-insensitive names and disabled entries.
- Ensured request-defined headers take precedence and only effective inherited headers are sent.


packages/bruno-app/src/components/RequestPane/RequestHeaders/getInheritedHeaders.js
```diff
@@ -0,0 +1,56 @@
...
+      headers: collection.draft?.root
+        ? get(collection, 'draft.root.request.headers', [])
+        : get(collection, 'root.request.headers', [])
+    }
+  ];
+};
+
+// All inherited headers are listed, so a name defined at several levels resolves
+// to the nearest enabled value. Within one source the last definition wins.
+export const getInheritedHeaders = (collection, item) => {
```
saravana512: export const getInheritedHeaders = (collection, item) => {
  const claimedNames = new Set();

  return getInheritedHeaderSources(collection, item).flatMap(({ headers, ...source }) => {
    const inheritedHeaders = [];

    for (const header of headers) {
      const normalizedName = header.name?.toLowerCase();

      if (
        !header.enabled ||
        !normalizedName ||
        claimedNames.has(normalizedName)
      ) {
        continue;
      }

      claimedNames.add(normalizedName);

      inheritedHeaders.push({
        ...header,
        uid: `inherited-${source.type}-${source.uid}-${header.uid}`,
        sourceRowUid: header.uid,
        rowType: 'inherited',
        source,
      });
    }

    return inheritedHeaders;
  });
};

avoid 
    1. effectiveHeaders.forEach((_, normalizedName) => claimedNames.add(normalizedName));
    2.  Array.from(effectiveHeaders.values());

packages/bruno-app/src/components/RequestPane/RequestHeaders/getInheritedHeaders.js
```diff
@@ -0,0 +1,56 @@
+import get from 'lodash/get';
+import { getTreePathFromCollectionToItem } from 'utils/collections';
+
+// Sources ordered by precedence: nearest folder first, collection last.
+export const getInheritedHeaderSources = (collection, item) => {
```
saravana512, line 5: const folders = [];

for (let i = treePath.length - 1; i >= 0; i--) {
  const item = treePath[i];

  if (item.type !== 'folder') continue;

  folders.push({
    type: 'folder',
    uid: item.uid,
    name: item.name,
    headers: item.draft
      ? get(item, 'draft.request.headers', [])
      : get(item, 'root.request.headers', []),
  });
}

folders.push({
  type: 'collection',
  uid: collection.uid,
  name: collection.name,
  headers: collection.draft?.root
    ? get(collection, 'draft.root.request.headers', [])
    : get(collection, 'root.request.headers', []),
});

return folders;



shall we use like this ?
avoid 2 loop

reverse();
folders.map(

packages/bruno-app/src/hooks/useFocusTableRow/index.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @param {object} params
+ * @param {string} params.uid      Tab uid hosting the table
+ * @param {string} params.tableId  Table the signal must target to be consumed
+ * @returns {{ focusRow: { uid?: string, name?: string, requestedAt: number } | null, onFocusRowHandled: () => void }}
+ */
+export const useFocusTableRow = ({ uid, tableId }) => {
+  const dispatch = useDispatch();
+  const focusTableRow = useSelector((state) => {
+    const tab = find(state.tabs.tabs, (t) => t.uid === uid);
```
saravana512, line 10: shall we use hashmap here ? avoid find

packages/bruno-app/src/providers/ReduxStore/slices/tabs.js
```diff
@@ -366,6 +366,25 @@ export const tabsSlice = createSlice({
         tab.focusErrorLine = null;
       }
     },
+    setFocusTableRow: (state, action) => {
+      const tab = find(state.tabs, (t) => t.uid === action.payload.uid);
```
saravana512, line 370: i saw some find shall we use hashmap ? or else change the structure as object or hashmap (new Map())

tests/request/headers/default-headers.spec.ts
```diff
@@ -7,7 +7,7 @@ import {
   selectRequestPaneTab,
```
saravana512, line 7: in this file saw somay await it takes long time to run to reduce the test case execution time 

try to use promise.all

packages/bruno-app/src/components/RequestPane/RequestHeaders/getInheritedHeaders.js
pooja-bruno: It would show the wrong value.  If a folder has `X-Token` twice, Bruno sends the last one. The Map does the same thing later rows overwrite earlier ones. If we keep the first match instead, the Headers tab would show the first value, but the request would still send the last one.

packages/bruno-app/src/components/RequestPane/RequestHeaders/getInheritedHeaders.js
pooja-bruno: I’d keep `filter().reverse()` it reads as “folders, nearest first,” then collection last.

tests/request/headers/default-headers.spec.ts
pooja-bruno: These steps have to stay in order create collection, then the request, then open Headers, then toggle inherited. `Promise.all` would start them together and race

packages/bruno-app/src/hooks/useFocusTableRow/index.js
pooja-bruno: Makes sense in general, but tabs are still an array everywhere in the slice, and people only have a handful open. find here matches the rest of the app.

packages/bruno-app/src/components/EditableTable/index.js
```diff
@@ -36,23 +37,25 @@ const TableRow = React.memo(
...
+      'row-focus-flash': !!item?.uid && flashedRowUid === item.uid
+    }, getRowClassName?.(item));
     const rowName = keyColumn ? item?.[keyColumn.key] : undefined;
 
     return (
       <tr
         {...rest}
         className={className}
-        data-testid={getRowTestId(item)}
+        data-testid={getRowTestId?.(item)}
```
shubh-bruno: `getRowTestId?.(item)` - was this needed change

packages/bruno-app/src/components/EditableTable/index.js
```diff
@@ -36,23 +37,25 @@ const TableRow = React.memo(
       handleDragHandleMouseDown,
       isRowEditable,
       getRowClassName,
-      getRowTestId
+      getRowTestId,
+      flashedRowUid
     } = context;
     const isEmpty = isLastEmptyRow(item, rowIndex);
-    const canDrag = reorderable && !isEmpty && rowIndex < reorderableRowCount && isRowEditable(item);
+    const canDrag = reorderable && !isEmpty && rowIndex < reorderableRowCount && (isRowEditable?.(item) ?? true);
```
shubh-bruno: `isRowEditable?.(item)` - was this needed change

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
```diff
@@ -26,17 +33,49 @@ import { headerNameRegex, headerValueRegex } from 'utils/common/regex';
...
+  overflowWrap: 'break-word',
+  wordWrap: 'break-word'
+};
+
+const HeaderHint = ({ id, text, className, place = 'top', testId, tooltipTestId, children }) => (
+  <>
+    <span id={id} className={className} data-testid={testId}>
+      {children}
+    </span>
+    {createPortal(
```
sachin-thakur-bruno: nit: can we used from `packages/bruno-app/src/ui/Portal/index.js`

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
```diff
@@ -26,17 +33,49 @@ import { headerNameRegex, headerValueRegex } from 'utils/common/regex';
...
 
+/** Portaled so overflow-hidden table cells cannot clip the hint. */
+const HEADER_HINT_STYLE = {
+  maxWidth: 220,
+  whiteSpace: 'normal',
+  overflowWrap: 'break-word',
+  wordWrap: 'break-word'
+};
+
+const HeaderHint = ({ id, text, className, place = 'top', testId, tooltipTestId, children }) => (
```
sachin-thakur-bruno, line 57: can we use ToolHint from components?

sachin-thakur-bruno (changes requested):
**BUG**:
When the same header exists in both the Inherited and Request sections, the inherited header is not shown in the table because the request header takes precedence.

If the request header is then unchecked, the inherited header still does not appear in the table, even though it is still being sent with the request.

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
```diff
@@ -79,6 +118,24 @@ const RequestHeaders = ({ item, collection, addHeaderText }) => {
   const [scroll, setScroll] = usePersistedState({ key: `request-headers-scroll-${item.uid}`, default: 0 });
   useTrackScroll({ ref: wrapperRef, selector: '.flex-boundary', onChange: setScroll, initialValue: scroll });
 
+  const pinHeadersToTop = useCallback(() => {
+    const pane = wrapperRef.current?.closest('.flex-boundary');
```
sachin-thakur-bruno: Instead of using `requestAnimationFrame`, I think we can simply use `virtuosoRef.current?.scrollToIndex({ index: 0, align: 'start' })`.
Please try this once.

Pass a ref from this component to `EditableTable`, accept a forwardRef in EditableTable.
In `RequestHeaders`
  const pinHeadersToTop = useCallback(() => {
    tableRef.current?.scrollToTop();
    setScroll(0);
  }, [setScroll]);

In `EditableTable` accept ref and
  useImperativeHandle(ref, () => ({
    scrollToTop: () => virtuosoRef.current?.scrollToIndex({ index: 0, align: 'start' })
  }), []);

shubh-bruno:
nit: PR looks ok but because we are using `getTreePathFromCollectionToItem` its bit slow when we switch the tabs and come to headers table

shubh-bruno (approved):

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
pooja-bruno: I’m using a Tooltip here because we’re using the same approach in the variable table. I’m keeping it consistent with the existing implementation.

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
pooja-bruno: It is not working

packages/bruno-app/src/hooks/useRevealFocusedTableRow/index.js
```diff
@@ -0,0 +1,144 @@
...
+
+    let cancelled = false;
+    let frame = 0;
+    let attempts = 0;
+
+    const finish = () => {
+      if (!cancelled) onFocusRowHandledRef.current?.();
+    };
+
+    const revealRow = () => {
```
sachin-thakur-bruno: instead of calling scrollToIndex in a requestAnimationFrame loop , we can virtusos's `scrollIntoView`.
it already knows which rows it has and hasn't measured, and it keeps adjusting on its own until the target row is confirmed on-screen, then calls a done callback you provide. So instead of our own frame-by-frame loop reading the DOM, we make one call and let Virtuoso tell us when it's finished (done: finish)

pooja-bruno: done

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
pooja-bruno: done

sachin-thakur-bruno:
> nit: PR looks ok but because we are using `getTreePathFromCollectionToItem` its bit slow when we switch the tabs and come to headers table
@pooja-bruno can we please use instead of using `getTreePathFromCollectionToItem` . 
we can collect folder ancestors while traversing the tree, avoiding a separate filter pass.
Use DFS to find the target and collect its folder ancestors. Each tree item is visited at most once. Then we can just call this and reverse. No need to run filter again.

const getAncestorFoldersForItem = (collection, item) => {
  const targetUid = item?.uid;
  if (!targetUid || !collection?.items) {
    return [];
  }

  const findFolders = (items, ancestors) => {
    for (const treeItem of items) {
      if (treeItem.uid === targetUid) {
        return treeItem.type === 'folder' ? [...ancestors, treeItem] : ancestors;
      }
      if (treeItem.items?.length) {
        const nextAncestors = treeItem.type === 'folder' ? [...ancestors, treeItem] : ancestors;
        const found = findFolders(treeItem.items, nextAncestors);
        if (found) {
          return found;
        }
      }
    }
    return null;
  };

  return findFolders(collection.items, []) ?? [];
};

then   const folders = getAncestorFoldersForItem(collection, item).reverse();

packages/bruno-app/src/components/RequestPane/RequestHeaders/index.js
sachin-thakur-bruno: we can use the ToolHint here. We can update in variable table later.

packages/bruno-app/src/components/EditableTable/index.js
```diff
@@ -560,6 +588,7 @@ const EditableTable = ({
           components={{ TableRow }}
           context={virtuosoContext}
           defaultItemHeight={ROW_HEIGHT}
+          {...(renderFullWidthRow ? { increaseViewportBy: 2000 } : {})}
```
sachin-thakur-bruno: please check its not degrading table performance.  each inherited row mounts two SingleLineEditor CodeMirror instances which will degrade the performance and adds a lag mainly on scroll

pooja-bruno: removed it

sachin-thakur-bruno (approved):
LGTM

PR #9080, opened 2026-08-25
fix(CLI): fixed junit reporters omit iterations and requests skipped due to bail

sharan-bruno:
BRU-3508

### Description

When --bail stopped a run, the JUnit report did not clearly show which requests were skipped. Requests that never started were missing from the report, and bail-skipped requests appeared as empty suites, which could make CI tools treat them as passing.
The HTML report also showed skipped requests without the method, URL, or reason for the skip.

### Problem

- JUnit only used the requests that the runner actually processed.
- Iterations that never started because of bail were missing.
- Skipped suites had skipped="1" but no <skipped> element.
- HTML was checking response.status, which did not work correctly for bail-skipped requests.
- Skipped requests were missing their method and UR

### Fix

- All planned requests and iterations are now included in the reports.
- A common helper creates the bail-skipped result.
- JUnit now adds <skipped message="Request Skipped due to bail"/>.
- Bail-skipped requests keep their actual method and URL.
- Skipped requests show - as the response status.
- HTML now checks result.status correctly.
- HTML shows "Request skipped due to bail" and highlights it correctly.
- bru.runner.skipRequest() also returns - as the response status.

Here is the CSV file I used for testing:
[test.csv](https://github.com/user-attachments/files/31220972/test.csv)

Here is the sample collection:
[CLI-skipped-request-due-to-bail.yml](https://github.com/user-attachments/files/31221144/CLI-skipped-request-due-to-bail.yml)

Commands to run
for Html  -----   `bru run --csv-file-path test.csv --bail --reporter-html results.html`
for Junit -----  `bru run --csv-file-path test.csv --bail --reporter-junit results.xml`
for json -----   `bru run --csv-file-path test.csv --bail --reporter-json results.json`

Here are the before and after generated report files for the reference

Before:
[before-results.html](https://github.com/user-attachments/files/31366554/before-results.html)
[before-results.json](https://github.com/user-attachments/files/31366555/before-results.json)
[before-results.xml](https://github.com/user-attachments/files/31366557/before-results.xml)

After:
[after-results.html](https://github.com/user-attachments/files/31366536/after-results.html)
[after-results.json](https://github.com/user-attachments/files/31366543/after-results.json)
[after-results.xml](https://github.com/user-attachments/files/31366544/after-results.xml)

### Screenshots

HTML Report
| Before | After |
| --- | --- |
| image | image |

JUnit Report
| Before | After |
| --- | --- |
| image | image |

JSON Report
| Before | After |
| --- | --- |
| image | image |

[...]

Issue #9074, opened 2026-08-25
CLI 4.0.0: `bru run` retains ~13 MB per executed request (9.5 GB over a 2,170-request run; 3.0.3 peaks at 2 GB)


dgyesbreghs:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue. (Closest is #8137, which reports steady memory growth with CLI 3.4.x; this is 4.0.0 and an order of magnitude larger.)

### This bug is:

- [x] Making Bruno unusable for me.
- [x] This feature was working in a previous version but is broken in the current release.

### Bruno version

@usebruno/cli 4.0.0 (Node 24.12.0). Compared against @usebruno/cli 3.0.3 from the same lockfile.

### 💻 Operating System

Linux

### Specific OS Version / Distro

Ubuntu (dev machine, 64 GB RAM) and Ubuntu CI VMs (Azure Standard_D2ds_v6: 2 vCPU, 7.9 GB RAM, no swap)

### 🐞 Describe the bug

`bru run` with **@usebruno/cli 4.0.0** retains roughly **13 MB of memory per executed request** for the lifetime of the run. On a ~2,170-request run the runner process reaches **~6 GB after ~500 requests and ~9.5 GB by the end**. On an 8 GB CI VM the kernel OOM-kills the machine ~2 minutes into the run (the CI agent itself dies). **3.0.3 on the exact same collection peaks at ~2 GB** and finishes the run ~8× faster.

Measured by sampling the RSS of the runner process (`node …/@usebruno/cli/bin/bru.js`; note Node 24 names it `MainThread` in `ps`) every 5 s while running the same folder list:

| CLI | flags | peak RSS | duration | result |
|---|---|---|---|---|
| 4.0.0 | `--max-old-space-size=8192` | **9.5 GB** (6.1 GB at ~500 requests, 7.7 GB at ~1,000) | 514 s | 2170/2170 pass (64 GB machine) |
| 4.0.0 | `--max-old-space-size=2048` | 5.7 GB at 63 s, still climbing | — | cap has no effect → the growth is not in the V8 old space (external/native memory) |
| 4.0.0 | `--reporter-skip-response-body` | 2.4 GB at 182 requests, same slope | — | not the response bodies |
| 3.0.3 | `--max-old-space-size=8192` | **2.0 GB** | 65 s | 2170/2170 pass |

Additional data points:

[...]

PR #9170, opened 2026-09-03
fix(nodevm): add `--experimental-cache-modules` to cache npm modules during script execution

sid-bruno:
### Description

Extends on #9078 to add in a marker based system for all the context sensitive modules. This is being done to make sure that modules that do access the bruno globals either top level or in a nested module are not a part of the shared module cache that's being added. 

Also adds a `--experimental-cache-modules` flag in to make sure that whoever is using this understands the behavioral risk that it might bring. 

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added an experimental CLI option to cache npm modules across developer sandbox script runs.
  - Cached modules preserve shared state while refreshing runtime variables for each script, including asynchronous executions.
- **Bug Fixes**
  - Improved runtime context, request attribution, and variable handling for requests originating from cached modules.
  - The option now displays a warning and is ignored unless the developer sandbox is selected.
- **Tests**
  - Added comprehensive coverage for caching, isolation, globals, validation, and script execution behavior.

### DIFF
--- a/.github/actions/tests/run-cli-tests/action.yml
+++ b/.github/actions/tests/run-cli-tests/action.yml
@@ -5,6 +5,10 @@ inputs:
     description: 'Shell to use (bash, pwsh)'
     required: false
     default: 'bash'
+  experimental_cache_modules:
+    description: 'Pass --experimental-cache-modules to bru run (requires developer sandbox)'
+    required: false
+    default: 'false'
 runs:
   using: 'composite'
   steps:
@@ -23,7 +27,11 @@ runs:
         npm start --workspace=packages/bruno-tests &
         sleep 5
         cd packages/bruno-tests/collection
-        node ../../bruno-cli/bin/bru.js run --env Prod --output junit.xml --format junit --sandbox developer
+        EXTRA_ARGS=""
+        if [ "${{ inputs.experimental_cache_modules }}" = "true" ]; then
+          EXTRA_ARGS="--experimental-cache-modules"
+        fi
+        node ../../bruno-cli/bin/bru.js run --env Prod --output junit.xml --format junit --sandbox developer $EXTRA_ARGS
 
     - name: Run Local Testbench and CLI Tests - Windows
       if: inputs.shell == 'pwsh'
@@ -42,4 +50,8 @@ runs:
         }
 
         cd packages/bruno-tests/collection
-        node ../../bruno-cli/bin/bru.js run --env Prod --output junit.xml --format junit --sandbox developer
+        $bruArgs = @("run", "--env", "Prod", "--output", "junit.xml", "--format", "junit", "--sandbox", "developer")
+        if ("${{ inputs.experimental_cache_modules }}" -eq "true") {
+          $bruArgs += "--experimental-cache-modules"
+        }
+        & node ../../bruno-cli/bin/bru.js @bruArgs
--- a/.github/workflows/tests-linux.yml
+++ b/.github/workflows/tests-linux.yml
@@ -55,6 +55,35 @@ jobs:
           comment_mode: always
           check_run: false
 
+  cli-test-cache-modules:
+    name: CLI Tests with experimental-cache-modules (Linux)
+    runs-on:
+      - ubuntu-latest
+      - self-hosted
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Node Dependencies
+        uses: ./.github/actions/common/setup-node-deps
+
+      - name: Run CLI Tests
+        uses: ./.github/actions/tests/run-cli-tests
+        with:
+          experimental_cache_modules: 'true'
+
+      - name: Publish Test Report
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        if: always()
+        with:
+          check_name: CLI Test Results with experimental-cache-modules (Linux)
+          files: packages/bruno-tests/collection/junit.xml
+          comment_mode: always
+          check_run: false
+
   e2e-test:
     name: Playwright E2E Tests (Linux) ${{ matrix.shard }}/${{ matrix.shardTotal }}
     timeout-minutes: 240
--- a/.github/workflows/tests-macos.yml
+++ b/.github/workflows/tests-macos.yml
@@ -49,6 +49,33 @@ jobs:
           comment_mode: off
           check_run: false
 
+  cli-test-cache-modules:
+    name: CLI Tests with experimental-cache-modules (macOS)
+    runs-on: macos-latest
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Node Dependencies
+        uses: ./.github/actions/common/setup-node-deps
+
+      - name: Run CLI Tests
+        uses: ./.github/actions/tests/run-cli-tests
+        with:
+          experimental_cache_modules: 'true'
+
+      - name: Publish Test Report
+        uses: EnricoMi/publish-unit-test-result-action/macos@v2
+        if: always()
+        with:
+          check_name: CLI Test Results with experimental-cache-modules (macOS)
+          files: packages/bruno-tests/collection/junit.xml
+          comment_mode: off
+          check_run: false
+
   e2e-test:
     name: Playwright E2E Tests (macOS)
     timeout-minutes: 240
--- a/.github/workflows/tests-windows.yml
+++ b/.github/workflows/tests-windows.yml
@@ -59,6 +59,36 @@ jobs:
           comment_mode: off
           check_run: false
 
+  cli-test-cache-modules:
+    name: CLI Tests with experimental-cache-modules (Windows)
+    runs-on: windows-latest
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Setup Node Dependencies
+        uses: ./.github/actions/common/setup-node-deps
+        with:
+          shell: pwsh
+
+      - name: Run CLI Tests
+        uses: ./.github/actions/tests/run-cli-tests
+        with:
+          shell: pwsh
+          experimental_cache_modules: 'true'
+
+      - name: Publish Test Report
+        uses: EnricoMi/publish-unit-test-result-action/windows@v2
+        if: always()
+        with:
+          check_name: CLI Test Results with experimental-cache-modules (Windows)
+          files: packages/bruno-tests/collection/junit.xml
+          comment_mode: off
+          check_run: false
+
   e2e-test:
     name: Playwright E2E Tests (Windows)
     timeout-minutes: 240
--- a/packages/bruno-cli/src/commands/run.js
+++ b/packages/bruno-cli/src/commands/run.js
@@ -138,6 +138,11 @@ const builder = async (yargs) => {
       default: 'safe',
       type: 'string'
     })
+    .option('experimental-cache-modules', {
+      type: 'boolean',
+      default: false,
+      describe: 'Share npm modules across script runs in the developer sandbox (experimental)'
+    })
     .option('output', {
       alias: 'o',
       describe: 'Path to write file results to',
@@ -313,6 +318,7 @@ const handler = async function (argv) {
       reporterJunit,
       reporterHtml,
       sandbox,
+      experimentalCacheModules,
       testsOnly,
       bail,
       reporterSkipAllHeaders,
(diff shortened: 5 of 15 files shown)


dgyesbreghs:
@sid-bruno as requested, I ran this branch against the real 2,170-request collection from #9074 (16 top-level folders, collection-level scripts requiring `@faker-js/faker`/`moment`/`nanoid` before every request, heavy `bru.runRequest` setup chains, `--sandbox=developer`, sampling the runner's RSS every 5 s). Four findings, roughly in order of importance:

## 1. Breaking change: skipped requests no longer resolve with `status: 'skipped'`

This branch renames the synthesized status for skipped requests from `'skipped'` to `'-'` (`run-single-request.js` ×2, `run.js` ×1). That shape reaches `bru.runRequest()` resolvers, and scripts in the wild branch on it — our collection's setup helpers do:

```js
const response = await bru.runRequest(createPath);
if (response && response.status === 'skipped') return; // requests skipped via their pre-request script
```

Result: every skip-aware helper chain broke and **1,210 of 2,170 requests failed** ("Setup step … returned status -"). This is independent of `--shared-script-modules` — it reproduces with the flag off too. Reverting just the rename locally makes the collection pass again (run in §3). If the rename is wanted for reporter display, could it stay `'skipped'` on the object handed to scripts?

## 2. Default mode (flag off) is heavier than 4.0.0

Sharing is opt-in, so out of the box this branch keeps the per-context module evaluation: our run peaked at **20.6 GB RSS** (4.0.0 peaked at 9.5 GB on the same collection; an 8 GB CI agent dies either way unless users discover the flag). RSS also shows a huge sawtooth (20.6 GB → 8 GB on a major GC), i.e. it's collectable garbage that V8 only reclaims near `--max-old-space-size` — the same pathology as #9074.

## 3. With the flag on, the static classifier defeats itself for the packages that matter

`isContextSensitiveModule` marks a module context-sensitive if its source (transitively, whole `require` tree) contains any of `bru/req/res/test/expect/assert/console/…` as a bare word, or any dynamic `require`. Measured against this collection's dependencies:

| module | classification | why |
|---|---|---|
| `moment` | **SENSITIVE** → per-context | its own source uses `res`, `test`, `console` as identifiers (deprecation warnings use `console.warn`) |
| `@faker-js/faker` | **SENSITIVE** → per-context | poisoned transitively via `dist/chunk-ZKNYQOPP.cjs` |
| `nanoid` | shared | clean |

So the two packages that caused #9074 are exactly the ones the heuristic cannot share. Minified/bundled dists make this common — almost any non-trivial package mentions `console`, `test`, `assert` or `res` somewhere in its tree. With the rename from §1 reverted locally so the collection actually passes, the flag-on run gives: **2170/2170 passed, peak 2.53 GB RSS, 345 s** — functional, but ~3× the memory and ~1.7× the runtime of the original #9078 approach on the same run, almost entirely because faker/moment keep getting re-evaluated per script.

## 4. One leaked socket per request (pre-existing, #9079)

Both runs held one open TCP connection per executed request (~1,900 sockets open near the end) — that's the throwaway keep-alive agent issue from #9079, orthogonal to this branch but worth remembering for CI boxes with a 1,024 fd limit.

## Comparison on this collection (same targets, same machine)

| build | flag | result | peak runner RSS |
|---|---|---|---|
| 4.0.0 (baseline) | — | 2170/2170 | 9.5 GB |
| #9078 as merged+reworked → this branch | off | 926/2170 (§1) | 20.6 GB |
| this branch | `--shared-script-modules` | 960/2170 (§1) | 1.7 GB (not comparable, most requests failed early) |
| this branch + §1 rename reverted | `--shared-script-modules` | 2170/2170 | 2.53 GB |
| #9078 original (ALS + facades) | — | 2170/2170 | 0.9 GB |
| 3.0.3 | — | 2170/2170 | 2.0 GB |

Happy to re-run any variant (or a tweaked classifier) against this collection — turnaround is ~10 minutes.

sid-bruno:
> 1. Breaking change: skipped requests no longer resolve with status: 'skipped'

Good catch, thanks, this is from https://github.com/usebruno/bruno/pull/9080 so that's out of whatever I was changing but we'll get to it as well 

> 2. Default mode (flag off) is heavier than 4.0.0

Hmm, this would need a deeper check what all got added into the CLI process 

> 3. With the flag on, the static classifier defeats itself for the packages that matter

A more robust one would end up needing a lot more analysis and the `async_hooks` path causes identity mismatches and since we can't guarantee how or what people's scripts are doing, it's a risk I'm not sure would be smart to take.

> 4. One leaked socket per request

The `--cache-ssl-session` was specifically for this but yeah we should add in a cleanup phase after the request processing is done, thanks for catching that will review that in the next few days

dgyesbreghs:
Thanks for the detailed follow-up!

On (3), completely your call — one clarification on what the merged #9078 approach actually risks, since "identity" covers a few different things: the facades keep `typeof` and method identity stable (`bru.setVar === bru.setVar` holds, `typeof bru === 'object'`), and `AsyncLocalStorage` only decides *which* script's objects a facade resolves to — scripts themselves never see facades, only npm modules do. The real identity caveat is narrower: inside an npm module, `bru` is not reference-equal to the raw per-script object, so exotic patterns (e.g. using `bru` as a Map key across executions) would notice. If the sniffing route stays, a middle ground that would rescue most real packages: treat only `bru`/`req`/`res`/`expect`/`__brunoTestResults`/`__bruSetScope` as poison words and drop `console`/`test`/`assert` (my sample: `moment` trips on `res`/`test`/`console` as plain identifiers, faker on a bundled chunk — happy to re-run the 2,170-request suite against any tweaked word list).

On (4): #9079 already implements exactly that cleanup phase — throwaway agents destroyed once the response is in (success, error, and before each redirect hop re-creates them, OAuth2 token agents included), keep-alive preserved on the wire, with an integration test that fails on main (6 sockets open at the 6th request) and passes with the fix. Feel free to take it as-is or fold it in here.

And whenever a build with the skip-status fix is up, say the word — the collection run takes me ~10 minutes.

sid-bruno:
> On (3), ....

fair enough

> And whenever a build with the skip-status fix is up, say the word

Accidental breaking implementation from https://github.com/usebruno/bruno/pull/9080, shouldn't have caused it, I wouldn't want to fix it in this PR causing a scope leak in the PR but have informed this internally for it to be addressed

dgyesbreghs:
Makes sense — and confirmed `main` doesn't have the rename, so it's contained to this lineage. Standing offer stays: ping me on any iteration of this branch (or the #9080 fix) and I'll have the 2,170-request numbers back within the hour.

dgyesbreghs:
Re-ran the 2,170-request collection against the current head (7622ec2, the AsyncLocalStorage + facade approach): **2170/2170 passed, peak runner RSS 1.08 GB** — the memory regression is gone and the skip-status break is resolved. 👍

The one remaining thing this run shows is the socket side: it held ~2,178 open connections (one per request) for the whole run — i.e. #9079 is still needed on top of this. Not a blocker for this PR, just confirming the two are independent as expected. Nice work on the rework!

node-vm unit suite is also green (63/63) with the primitive-value guard you added.

sid-bruno:
@dgyesbreghs 

`tests/runtime.spec.js` has failures, specifically `should expose each request URL to a cached module across sequential runs` since we added VM level isolation for modules so that people using nested modules accessing bruno's global context would still work. 

I'm still looking into ways to fix that without having to do the scan that I was doing previously

dgyesbreghs:
Reproduced `tests/runtime.spec.js › should expose each request URL to a cached module across sequential runs` on the current head and dug into it — sharing my findings in case they're useful, no pressure to take any of it.

## Why it fails

The failing assertion is `captured()` (a module doing `const urlAtLoad = req.getUrl()` at top level). `dynamic()` works because method calls late-bind through the facade; but `urlAtLoad` is a **plain string captured once at module-load**, and since the module is shared, that load only ever happens during run 1 → every later run sees run 1's URL. No facade can fix a captured primitive — the module genuinely has to be re-evaluated per run.

The old static scan solved this by detecting such modules, but (as we saw earlier) it also mis-flags faker/moment on incidental identifier matches, so they lose sharing.

## An approach that avoids the scan: detect the touch at runtime

Instead of scanning source, watch whether a module actually **reads a Bruno global while it is loading**. The facade getter is the single choke point, so it's one extra line there plus a load-stack:

- Keep a `moduleLoadStack` (a frame per module currently being evaluated).
- In the context-global getter, if a load is in progress, flag every frame on the stack → transitive sensitivity, precise, zero false positives.
- After a module loads: if it was flagged, it's context-sensitive → don't put it in the shared cache; re-evaluate it per script context (cached per-run). Otherwise share it once as today.

faker/moment never read `bru`/`req`/`res` *at load* (their earlier "sensitive" verdict was purely the regex hitting `res`/`test`/`console` as minified identifiers), so at runtime they stay shared — while `sequential-request-reader`, which really does call `req.getUrl()` at load, gets re-evaluated per run.

## Results on my 2,170-request collection

| | result | peak RSS |
|---|---|---|
| current head (share-all) | 2170/2170 but the seq-capture test fails | 1.08 GB |
| head + this approach | 2170/2170, **seq-capture test passes** | **1.07 GB** |

So the correctness fix costs ~nothing in memory — faker/moment stay shared. `bruno-js` suite: the runtime.spec.js case goes green; the two `index.spec.js` failures (`global identity/realm` and `call-site mapping for concurrent scripts`) are **pre-existing on this branch, unchanged** by this — the realm one is separate: `bru instanceof Object` is false because the facade's Proxy target is created in the host realm, not the shared VM context (creating the target inside the VM context would fix that one, happy to look if useful).

Prototype branch (on top of your head, one file changed): https://github.com/dgyesbreghs/bruno/commit/40f4a8e — take it, adapt it, or ignore it. Glad to iterate or re-benchmark any variant.

sid-bruno:
@dgyesbreghs 

A mark and re-evaluate is the same thing I'm going through, it does reduce the overall heap but might still be a tad bit larger than what was in 3.0.3, I'll push the implementation in a bit, should be able to get better numbers post that

sid-bruno:
@dgyesbreghs 

when you get time, another run with the current changes should help confirm the fix.

dgyesbreghs:
Ran the 2,170-request collection against `c25a8b7`. Headline: there's a blocking bug that makes every request fail, and once patched, the numbers below.

## Blocking: `options` used before declaration (TDZ)

`packages/bruno-cli/src/runner/run-single-request.js` reads `options` on the `scriptingConfig.cacheModules = get(options, …)` line, three lines *above* its `const options = getOptions()` declaration. Every request throws `Cannot access 'options' before initialization` → **2170/2170 failed**. It's also why CLI Tests + Unit Tests are red on this PR (Copilot's review flagged the same thing). Trivial reorder fixes it:

```js
const options = getOptions();
scriptingConfig.cacheModules = get(options, 'cacheModules', false) === true;
```

## Results with that one-line reorder applied locally

| mode | result | peak runner RSS | duration |
|---|---|---|---|
| default (no flag) | 2170/2170 | **8.86 GB** | 455 s |
| `--experimental-cache-modules` | 2170/2170 | **0.87 GB** | 193 s |

So the sharing path works well (0.87 GB, and the load-time touch-detection handles the `req.getUrl()`-at-load case correctly). The thing worth confirming is the **default**: with caching opt-in, an out-of-the-box `bru run` is back at ~8.9 GB on this collection — i.e. #9074 still bites anyone who hasn't found the flag, which for CI is the common case. Is opt-in the intended end state, or a staging step while the shared path bakes? If it's meant to default-on eventually, happy to re-run once that flips.

(Socket count was ~2.8k open, i.e. #9079 still applies on top — unchanged from before.)

sid-bruno:
@dgyesbreghs https://github.com/usebruno/bruno/pull/9170#issuecomment-5568134763 was already fixed, you'll have to re-run with current head

dgyesbreghs:
Re-ran on the current head (`68f74c5`) — the TDZ is fixed, `bru run` executes cleanly again. Numbers on the 2,170-request collection:

| mode | result | peak runner RSS |
|---|---|---|
| `--experimental-cache-modules` | **2170/2170** | **1.57 GB** |
| default (no flag) | 2170/2170 so far | climbing past 5.1 GB at request ~420, then **OOM-killed my box** before finishing |

So the sharing path is solid — 2170/2170 and the load-time touch-detection handles the `req.getUrl()`-at-load case. The default is the open question I flagged before: with caching opt-in, an out-of-the-box run still grows unbounded (it OOM'd here), i.e. #9074 is unchanged for anyone who doesn't pass the flag — which for CI runners is the default path. Is opt-in the intended end state, or is `--experimental-cache-modules` a staging gate before it becomes the default? If the latter, ping me when it flips and I'll confirm the default path.

(Socket count ~2.7k again — #9079 still applies on top, unchanged.)

sid-bruno:
@dgyesbreghs is used by quite a few people so changing the default behaviour isn't exactly a practical solution so it'll behind a experimental flag for a while. 

Thanks for validating it. 

cc: @bijin-bruno feel free to review and merge

bijin-bruno (dismissed):

bijin-bruno (approved):

PR #9177, opened 2026-09-04
fix(CLI): (changes reverted) update response status to 'skipped' for omitted requests

sharan-bruno:
### Description

Reverted the status skipped related changes - The HTML runner report showed `skipped` in the Response Code field for skipped requests.

### Problem

A skipped request never reaches the server, so it has no status code — but the report printed the literal `skipped` where a number belongs, unlike the summary table which already used `-`.

### Fix

Response Code now renders `-` when the request was skipped, and the status code unchanged otherwise, via a `getResponseCode` helper extracted from the template so it can be unit tested.

### Screenshots

| Before | After |
| --- | --- |
|  |  |

#### Contribution Checklist:

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized skipped request results to display a clear **“skipped”** status in CLI output and HTML reports.
  - Fixed HTML reports so skipped requests are identified correctly, including requests skipped because of prompt variables or pre-request scripts.
  - Preserved the response code display as **“-”** for skipped requests while retaining existing duration, size, timing, and skip-message details.

### DIFF
--- a/packages/bruno-common/src/runner/reports/html/template.ts
+++ b/packages/bruno-common/src/runner/reports/html/template.ts
@@ -327,7 +327,7 @@ export const htmlTemplateString = (resutsJsonString: string) => `<!DOCTYPE html>
             :bordered="false"
           >
             <template #header>
-              {{result.path}} - {{resultSummary}} {{hasError && result.status !== 'skipped' ? " - (request failed)" : "" }}
+              {{result.path}} - {{resultSummary}} {{hasError && result.response.status !== 'skipped' ? " - (request failed)" : "" }}
             </template>
           </n-alert>
         </template>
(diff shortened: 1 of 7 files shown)


packages/bruno-common/src/runner/reports/html/template.ts
```diff
@@ -363,7 +363,7 @@ export const htmlTemplateString = (resutsJsonString: string) => `<!DOCTYPE html>
                   <n-list-item>
                     <n-thing
                       title="Response Code"
-                      :description="'' + result.response.status"
+                      :description="responseCode"
```
sanish-bruno: we need to make sure this is a string!  or we can keep the old one itself, just worried why we have the `'' + result.response.status`

sanish-bruno (approved):

vijayh-bruno (approved):

PR #9180, opened 2026-09-04
fix(auth): Inherit 'none' Authentication from Folder

lhardt:
### Description

Hi! 

This is more of a question to the bruno devs than a PR in itself. 
If you find this change worth it, I can bring it to a better state and contribute it.

### Problem

Imagine you have a collection for your webapp `SystemX`, which, for 90% of endpoints, requires Oauth authentication. You then create environments for `SystemX`,  with data from `prod`, `testing`, `devel`, etc, and organize your requests in many subcollections, in the interest of reusing those Environments.

Then, your team creates a subset of endpoints, let's say `SystemX Public API`, which don't require authentication at all, but still uses the environments (such as for URL, port, etc).  Your first instinct would be to set the `Public API` folder authentication to `none`, and expect the `SystemX Public API` endpoints to not have any auth, right?

As per the code right now, setting `auth=none` in a subcollection has **no discernible effect** from setting `auth=inherit`. They are the same option. Notice that this behaviour is inconsistent with how the inheritance **text** is displayed on the UI.  You will read that your endpoints are inheriting auth=none from  `Public API`, but actually they are inheriting `auth=Oauth` from `SystemX`. 

The workaround is going through each request inside that subcollection and setting auth=none individually.

### Fix

Removing that condition should be enough.
I can test it and verify it if you find that this is the behaviour change you want

### Screenshots

TBD once agreed that this is a good solution. 

| Before | After |
| --- | --- |
|  |  |

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * Folder-level authentication now correctly applies explicit authentication modes, including “none,” instead of skipping them.


lhardt:
Update: squashed and formatted commit message

lhardt:
Update: rebased to the latest main

PR #9190, opened 2026-09-07
feat(sidebar): virtualize sidebar nested items

### LINKED ISSUE #8928: Performance issues with large collection variables
Internal - [BRU-4312]

### This bug is:

- [x] Annoying.

### Bruno version

4.0.0

### 💻 Operating System

macOS

26.6.1 (25G76)

### 🐞 Describe the bug

Hi team,

I am currently in the process of migrating my workflows from Postman to Bruno, but I am hitting a performance bottleneck with a large collection.
- Collection size: ~3,850 requests
- Collection variables: ~360 variables

Editing, scrolling, and navigating the Collection Variables tab in the GUI is extremely slow and glitchy. In the attached video, you can see the scrolling issues, and around the 30-second mark, I try to type—even typing very slowly, it takes almost 30 seconds for the characters to reflect in the UI.

To isolate the issue, I tried importing the same set of variables into a clean, empty collection, and the UI ran completely fine with no lag. The issue appears to be specific to the combination of having a large collection and a large number of collection variables simultaneously. Furthermore, using these same variables at the **Global level** works fast and smoothly.

In my previous setup (Postman), I used workspace variables to share context across requests during runner/CLI executions. Since Bruno CLI currently does not support passing Global variables via command line arguments, I am stuck between GUI performance issues and CLI limitations.

Is there a known workaround or planned fix for the Collection Variables tab performance with large datasets? Alternatively, is there a recommended way to pass global/environment variables dynamically in the Bruno CLI?

Thanks a lot in advance!

### 📝 Steps to reproduce

Use large collection with many folders and subfolders
Use large environment

### 📁 Collection to reproduce

_No response_

https://github.com/user-attachments/assets/7586ecc8-1d7e-405a-ab43-f5f8633d52f0

sachin-thakur-bruno:
### Description

Virtualizes the collections sidebar so it renders only the rows currently in view instead of the entire nested tree.
It keeps render cost flat no matter how many collections, folders, requests, or examples exist.

Added the missing tests here https://github.com/usebruno/bruno/pull/8891
closes https://github.com/usebruno/bruno/issues/8928

BRU-4448
BRU-4312

### Problem

The old sidebar rendered the collection tree recursively so every collection, folder, request, and example are mounted even though they are not in the viewport Rendering cost scaled with the total number of items, so large collections were slow and heavy.

### Benefits After Refactor

1. Heap size decreases when large collection are opened in the sidebar. 
2. Number of nodes in the DOM decreases from 38k to 2k.
3. Editor and codemirror type lags significantly becomes faster.
4. Memory usage dropped when a large collection is opened and we have variable table rendered on the screen.

### Fix

Used virtuoso to virtualize the sidebar items. Since virtuoso works on flat data, we are flattening the sidebar nested data and feeds it to virtuoso. Now the rows which are currently visible on viewport are mounted and in the DOM. Also added a overlook scan to avoid a flicker on fast scroll.
This fixes problems with slow typing in Editors, codemirrors whenever user have large number of items in the sidebar.

Used four maps to store the data and for fast lookup

1. **The object maps: itemsByUid / collectionsByUid**: These map a uid to the live folder/app/request or collection object in the Redux store. 
Rows are structural only, a row carries id, kind, depth, uids, sortName, parentName, but not the actual object. Since Rows are rebuilt on every flatten, and if each row embedded the full item object, then (a) rows would be fat, and (b) memoization would be hard.

5. **The index maps: rowIndexByItemUid / rowIndexByCollectionUid**: These map a uid to the row's position (its integer index in the rows array), and they exist for exactly one feature: scroll-to-active-tab.

### Changes

1. `utils/collections/flattenSidebarTree.js`: This contains the functions to flat the sidebar nested items.
       `flattenSidebarTree`: This takes the nested sidebar items and returns a flat row containing all items.
       `flattenCollection`: Adds a collection and its visible children to the flat sidebar row list.
       `walkChildren`: Flattens the children of a collection or folder into sidebar rows.

6. `Sidebar/Collections/index.js`: This used the `flattenSidebarTree` function and feed the flat rows to `Virtuoso`.
       It renders the rows using `SidebarRow` component.

7. `SidebarRow/index.jsx`: This renders the items using the `itemsByUid, collectionsByUid` maps. Based on the kind of row item it maps them to the right presentation componentCollectionRow, CollectionItemRow, EmptyCtaRow, ExampleItem, GitRemoteCollectionRow).

8. `slices/collections/index.js`: Since Virtuoso unmounts offscreen rows, per-row "examples expanded" React state wouldn't survive a scroll. Added the toggleRequestExamples reducer and an examplesExpanded flag on the item so expansion state lives in the store and examples are emitted as their own flat rows.

9. `Test-side changes`: the DOM is now flat, every helper and spec that relied on nesting (.locator('..'), #collection-<slug> descendant scoping) was rewritten to scope by data-collection-id/data-parent-name.

### Screenshots

|Memory usage  Before |Memory usage  After Virtualization |
| --- | --- |
|image| image|

|Heap Size and DOM Nodes Before |Heap Size and DOM Nodes After |
| --- | --- |
| image| image|
#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - Added a virtualized collections sidebar for smoother navigation through large collections.
  - Added “Add request” actions for empty collections and folders.
  - Added automatic scrolling to the active item.
  - Added consistent expansion and collapse behavior for request examples.

- **Improvements**
  - Improved sidebar search across nested folders, requests, and collections.
  - Improved drag-and-drop and navigation reliability across collections and formats.
  - Improved environment-variable error handling and editor updates.
  - Improved sidebar behavior when interacting with menus and dialogs.


packages/bruno-app/src/components/Environments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
```diff
@@ -15,13 +15,16 @@ const EnvironmentVariables = ({ environment, setIsModified, collection, searchQu
   const environmentsDraft = collection?.environmentsDraft;
   const hasDraftForThisEnv = environmentsDraft?.environmentUid === environment.uid;
 
-  // Check for non-secret variables used in sensitive fields
+  const collectionItems = collection?.items;
+  const collectionRoot = collection?.root;
+  const environmentVariables = environment?.variables;
```
chirag-bruno, line 20: Can we have these default values? This can probably helps us to not have other `null` and `undefined` checks?
```suggestion
  const collectionItems = collection?.items ?? [];
  const collectionRoot = collection?.root ?? {};
  const environmentVariables = environment?.variables ?? [];
```

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/CollectionItemRow/index.jsx
```diff
@@ -132,7 +128,7 @@ const CollectionItem = ({ item, collectionUid, collectionPathname, searchText, o
   const [newAppModalOpen, setNewAppModalOpen] = useState(false);
   const [runCollectionModalOpen, setRunCollectionModalOpen] = useState(false);
   const [itemInfoModalOpen, setItemInfoModalOpen] = useState(false);
-  const [examplesExpanded, setExamplesExpanded] = useState(false);
+  const examplesExpanded = !!item.examplesExpanded;
```
chirag-bruno: Makes it a little redable than double negation.
```suggestion
  const examplesExpanded = !Boolean(item.examplesExpanded);
```

packages/bruno-app/src/components/Environments/EnvironmentSettings/EnvironmentList/EnvironmentDetails/EnvironmentVariables/index.js
sachin-thakur-bruno: This is kept to keep the deps stable. these three are used as useMemo dependencies.
We already have 
line 24: if (!environmentVariables) { return … } guards before .filter,
line 56: flattenItems(collectionItems || []) defaults locally,
collectionRoot is passed on as-is (its consumers handle undefined).

packages/bruno-app/src/components/Sidebar/Collections/Collection/CollectionItem/CollectionItemRow/index.jsx
sachin-thakur-bruno: Changed it to use `Boolean(item.examplesExpanded);`

packages/bruno-app/src/utils/collections/flattenSidebarTree.js
```diff
@@ -0,0 +1,342 @@
...
+      itemUid: request.uid,
+      sortName: request.name || null
+    });
+
+    addItemToIndex(request.uid, request);
+
+    const hasExamples
+      = request.type === 'http-request' && Array.isArray(request.examples);
+
+    if (!hasSearch && hasExamples && request.examplesExpanded) {
```
sid-bruno: Example rows are dropped whenever a search is active, but `getVisibleSidebarUidsInOrder` still pushes example uids while searching (`packages/bruno-app/src/utils/collections/index.js`, `examplesVisible = request.type === 'http-request' && (hasSearchText || !isCollectionItemCollapsed(request))`). Shift-click ranges and multi-drag ordering then include uids that have no rendered row. Pre-virtualization `CollectionItem` also rendered examples under a matching request during search, because `itemIsCollapsed` was forced to `false`.

```suggestion
    if (hasExamples && (hasSearch || request.examplesExpanded)) {
```

packages/bruno-app/src/utils/collections/flattenSidebarTree.js
```diff
@@ -0,0 +1,342 @@
...
+    addItemToIndex
+  };
+
+  const visibleChildCount = walkChildren(collectionContext, {
+    collectionItems: collection.items,
+    depth: 1,
+    parentName: null
+  });
+
+  // append emtry row cta.
```
sid-bruno: Typo: "emtry row cta".

```suggestion
  // Append the collection-root empty-state CTA row.
```

packages/bruno-app/src/components/Sidebar/Collections/SidebarRow/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+          openBulkMenu={openBulkMenu}
+          isItemMultiDragDisabled={isItemMultiDragDisabled}
+          multiDragCollections={multiDragCollections}
+          multiDragItems={multiDragItems}
+        />
+      );
+    }
+    case 'empty-cta': {
+      return <EmptyCtaRow collection={resolved} itemUid={row.itemUid} depth={row.depth} />;
+    }
```
sid-bruno, line 65: Unlike every other branch, `empty-cta` has no `resolved` guard. `EmptyCtaRow` bails internally on a missing collection, but `renderRow` still returns an element, so `SidebarRow` renders an empty `[data-sidebar-row]` wrapper: a blank virtual row that also swallows the background click, since `handleContainerClick` skips anything matching `[data-sidebar-row]` and the multi-selection is not cleared.

```suggestion
    case 'empty-cta': {
      if (!resolved) return null;
      return <EmptyCtaRow collection={resolved} itemUid={row.itemUid} depth={row.depth} />;
    }
```

packages/bruno-app/src/components/Sidebar/Collections/index.js
```diff
@@ -58,10 +77,27 @@ const Collections = ({ showSearch, isCreatingCollection, onCreateClick, onDismis
...
+  const rowIndex = rowIndexByItemUid.get(activeTabUid);
+  const activeRowIndex = activeTabUid !== null
+    ? (rowIndex ?? rowIndexByCollectionUid.get(activeTabUid) ?? null)
+    : null;
+
+  useEffect(() => {
+    if (activeRowIndex === null) return;
+    virtuosoRef.current?.scrollIntoView({ index: activeRowIndex, behavior: 'smooth' });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [activeTabUid]);
```
sid-bruno: The effect only fires on `activeTabUid`, so when the row does not exist yet on that pass the item is never scrolled to: opening a request from global search / a deep link mounts the collection (or expands an ancestor) after the tab is already active, `activeRowIndex` is `null` at that point, and nothing re-runs once the row appears. On `main` the per-row `isTabForItemActive` effect covered this because the row itself mounted later. Depending on the resolved index and remembering the last scrolled tab keeps it to one scroll per tab and drops the eslint suppression.

```suggestion
  // Resolve the active tab's row index (item rows first, then collection headers).
  const rowIndex = rowIndexByItemUid.get(activeTabUid);
  const activeRowIndex = activeTabUid
    ? (rowIndex ?? rowIndexByCollectionUid.get(activeTabUid) ?? null)
    : null;

  // The row for a freshly activated tab can be missing on the first pass (collection
  // still mounting, or an ancestor collapsed), so scroll once the index resolves.
  const scrolledTabUidRef = useRef(null);

  useEffect(() => {
    if (activeRowIndex === null) return;
    if (scrolledTabUidRef.current === activeTabUid) return;
    scrolledTabUidRef.current = activeTabUid;
    virtuosoRef.current?.scrollIntoView({ index: activeRowIndex, behavior: 'smooth' });
  }, [activeTabUid, activeRowIndex]);
```

packages/bruno-app/src/providers/ReduxStore/slices/collections/index.js
```diff
@@ -1149,6 +1146,17 @@ export const collectionsSlice = createSlice({
...
     },
+    toggleRequestExamples: (state, action) => {
+      const collection = findCollectionByUid(state.collections, action.payload.collectionUid);
+
+      if (collection) {
+        const item = findItemInCollection(collection, action.payload.itemUid);
+
+        if (item && item.type === 'http-request') {
+          item.examplesExpanded = !item.examplesExpanded;
+        }
```
sid-bruno: `examplesExpanded` becomes a second source of truth for example visibility. The rendered rows read it (`flattenSidebarTree`), but `getVisibleSidebarUidsInOrder` — used by `selectSidebarRange` and `getSortedDraggedItems` — still reads `isCollectionItemCollapsed(item)`, and bulk Collapse/Expand plus `GlobalSearchModal`'s `expandItemPath` still write `item.collapsed`. So after using the chevron, shift-range selection and multi-drag order disagree with what is on screen, and bulk expand no longer reveals examples. On `main` the chevron dispatched `expandItem`/`collapseItem`, which is what kept these aligned.

```suggestion
        if (item && item.type === 'http-request') {
          item.examplesExpanded = !item.examplesExpanded;
          // `collapsed` is still read by getVisibleSidebarUidsInOrder (shift-range
          // selection, multi-drag ordering) and written by bulk collapse/expand.
          item.collapsed = !item.examplesExpanded;
        }
```

packages/bruno-app/src/components/Sidebar/Collections/SidebarRow/index.jsx
sachin-thakur-bruno: fixed it

sid-bruno (approved):

PR #9195, opened 2026-09-07
feat(bruno-sqlite): move cache collection fetch to bruno-sqlite

shubh-bruno:
### Description

BRU-4045

Moves the collection file cache (`FileIndex`) off its own SQLite wrapper and onto the `@usebruno/sqlite` SDK: schema as migrations, queries as generated statements, and the Preferences UI reading them through the React Query hooks.

The old `mount-snapshots.db` is deleted on first launch. Cache rows are derived from files on disk, so each collection just re-caches on its next mount.

### Problem

The file cache had a parallel SQLite stack that duplicated what `@usebruno/sqlite` already does:

- Its own connection wrapper (`services/storage/index.js`), with SQL inlined into JS and migrations tracked only by a version counter. It recorded how far it had migrated, never what it had run, so editing a migration that had already applied was a silent no-op. Two machines could end up with different schemas and no error anywhere.
- Its own database file (`mount-snapshots.db`), separate from `bruno.db`.
- One-off IPC handlers for cache size and clearing, with the size read as `fs.statSync(dbPath).size`, a number that only meant anything while that file held nothing but the cache.
- No renderer access at all, which Phase 2 (SQLite-backed Sidebar and Global search) needs.

### Fix

- Schema and queries moved into the SDK: `migrations/0000002_file_index_entries.ts` plus 8 statements in `statements/file-index.sql`. `FileIndex` keeps its logic but owns no database, it calls statements by name.
- `services/storage/` deleted, `FileIndex` was its only caller.
- `mount-snapshots.db` and its sidecars removed at startup (`removeLegacyFileIndex`).
- `content_bytes` column stores each row's byte size so the cache size in Preferences sums a small index instead of reading every cached file.
- `application_version` column records the build that parsed a row. `status()` checks it before mtime and hash, so a new build re-parses files it would otherwise consider unchanged.
- Indexes added: a covering index for the change-detection query, plus three `json_extract` indexes on request `url` / `method` / `name`.
- WAL enabled via a new pragma hook on `DB`. It can't live in a migration, SQLite refuses to change `journal_mode` inside a transaction.
- BigInt support added to the SDK. `mtime` is a nanosecond timestamp, too large for a JS number to hold exactly. Statements opt in with `-- name: x :many :bigints`.
- Preferences -> Cache uses the hooks: `useSqliteQuery('file_index_size')` and `useSqliteMutation`, replacing the two IPC handlers and a `useEffect`.

#### Note: the "Cache size" number changed meaning

It now reports cached content instead of file size on disk, so with nothing cached it reads `0B` where it previously read `12.0KB` (on mac). That 12 KB was never cache, it was three 4 KB SQLite pages present in any empty database. The old reading only worked because `mount-snapshots.db` held nothing else; `bruno.db` is shared with `runner_responses`.

| Before | After |
| --- | --- |
|  |  |

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - File-cache management now provides more consistent cache size reporting and clearing.
  - Cache entries track the application version, allowing outdated cached data to refresh automatically.
  - Database handling now supports improved journaling and reliable large-number values.
  - Legacy cache data is cleaned up automatically during database initialization.

- **Bug Fixes**
  - File-cache clearing now reports success or failure through notifications.
  - Disk space can be reclaimed after clearing the file cache.


chirag-bruno:
@shubh-bruno, @sid-bruno mentioned to add a version column in the `file cache` table. This can then be used to purge the cache entries on different versions. Discuss with @sid-bruno for more info.

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -12,49 +12,17 @@ const {
...
-    up: `
-      ALTER TABLE file_index_entries ADD COLUMN raw TEXT;
-      UPDATE file_index_entries SET mtime = 0, hash = '';
-      ALTER TABLE file_index_entries ADD COLUMN created_at INTEGER;
-      ALTER TABLE file_index_entries ADD COLUMN updated_at INTEGER;
-      UPDATE file_index_entries SET created_at = unixepoch(), updated_at = unixepoch();
-    `
-  }
-];
+const statements = () => {
```
chirag-bruno: This can move into the constructor of FileIndex class and assign it once during the construction than calling it multiple times.

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -12,49 +12,17 @@ const {
...
+  const ready = getStatements();
+  if (!ready) throw new Error('the file cache is unavailable: the sqlite database is not open');
+  return ready;
+};
+
+const applicationVersion = () => require('electron').app.getVersion();
```
chirag-bruno: Same here, can move into the constructor and can be accessed with `this.applicationVersion`

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -134,35 +89,21 @@ class FileIndex {
     const { op, relativePath } = entry;
 
     if (op === 'remove') {
-      this.#db.run(
-        'DELETE FROM file_index_entries WHERE collection_path = ? AND relative_path = ?',
-        root,
-        relativePath
-      );
+      statements().execute('file_index_delete_entry', { collection_path: root, relative_path: relativePath });
```
chirag-bruno: Why aren't we normalising the paths here?

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -134,35 +89,21 @@ class FileIndex {
...
-        mtime = excluded.mtime,
-        hash = excluded.hash,
-        data = excluded.data,
-        raw = excluded.raw,
-        updated_at = unixepoch()
-    `,
-      root,
-      relativePath,
-      id,
+    statements().execute('file_index_upsert', {
```
chirag-bruno: Need to normalise the paths.

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -134,35 +89,21 @@ class FileIndex {
...
-      JSON.stringify(data),
-      raw ?? null
-    );
+      data: JSON.stringify(data),
+      raw: raw ?? null,
+      application_version: applicationVersion()
+    });
   }
 
   stageParsed(collectionPath, absolutePath, data) {
```
chirag-bruno, line 119: Normalise the paths

packages/bruno-electron/src/services/mount/file-index.js
chirag-bruno, line 133: Normalisation needed.

packages/bruno-sqlite/statements/file-index.sql
```diff
@@ -0,0 +1,42 @@
...
+DELETE FROM file_index_entries
+WHERE collection_path = @collection_path AND relative_path = @relative_path;
+
+-- name: file_index_clear_collection :exec
+DELETE FROM file_index_entries WHERE collection_path = @collection_path;
+
+-- name: file_index_clear :exec
+DELETE FROM file_index_entries;
+
+-- name: file_index_vacuum :exec
```
chirag-bruno: Can you please change the name of this statement? Vaccum is not table specific. It cleans all the free pages in the Database. Also, if possible, can you move this to another statement file called `utils.sql`?

packages/bruno-sqlite/statements/file-index.sql
```diff
@@ -0,0 +1,42 @@
...
+FROM file_index_entries
+WHERE collection_path = @collection_path;
+
+-- name: file_index_upsert :exec
+INSERT INTO file_index_entries
+  (collection_path, relative_path, id, mtime, hash, data, raw, content_bytes, application_version,
+   created_at, updated_at)
+VALUES
+  (@collection_path, @relative_path, @id, @mtime, @hash, @data, @raw,
+   LENGTH(@data) + LENGTH(COALESCE(@raw, '')), @application_version, unixepoch(), unixepoch())
```
chirag-bruno: If I am not wrong, `LENGTH()` on a `TEXT` column gives us the length of the string (character by character), and not the number of bytes. Can you check this and update it if necessary?

packages/bruno-sqlite/statements/file-index.sql
```diff
@@ -0,0 +1,42 @@
+-- name: file_index_stored :many :bigints
+SELECT relative_path AS relativePath, id, mtime, hash, application_version AS applicationVersion
+FROM file_index_entries
+WHERE collection_path = @collection_path;
+
+-- name: file_index_entries_for_collection :many
+SELECT relative_path AS relativePath, data, raw
+FROM file_index_entries
+WHERE collection_path = @collection_path;
```
chirag-bruno: nit: can we have better names here? `file_index_stored` and `file_index_entries_for_collection` is not telling me anything specific about the statements.

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -181,21 +131,27 @@ class FileIndex {
...
   }
 
-  #loadStored(collectionPath) {
-    const rows = this.#db.all(
-      'SELECT relative_path AS relativePath, id, mtime, hash FROM file_index_entries WHERE collection_path = ?',
-      collectionPath
-    );
+  #loadMetadata(collectionPath) {
+    const rows = this.#statements.execute('file_index_metadata_for_collection', {
+      collection_path: collectionPath
```
chirag-bruno, line 152: Normalisation is not necessary as we are not writing anything that is being read by another machine which has another OS. Having said that, there are a lot of places where we have introduced normalisation (because of my previous comments, my bad). Can we either normalise all paths or remove normalisation in all paths?

packages/bruno-sqlite/migrations/0000002_file_index_entries.ts
```diff
@@ -0,0 +1,55 @@
+/**
```
chirag-bruno: Is this big comment needed?

packages/bruno-sqlite/migrations/0000002_file_index_entries.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    CREATE INDEX IF NOT EXISTS idx_file_index_lookup
+      ON file_index_entries(collection_path, relative_path, mtime, hash, id);
+
+    CREATE INDEX IF NOT EXISTS idx_file_index_url
+      ON file_index_entries(json_extract(data, '$.request.url'));
+    CREATE INDEX IF NOT EXISTS idx_file_index_method
+      ON file_index_entries(json_extract(data, '$.request.method'));
+    CREATE INDEX IF NOT EXISTS idx_file_index_name
+      ON file_index_entries(json_extract(data, '$.name'));
```
chirag-bruno: Are these indexes needed? They seem to help in queries where we are querying `url`, `method`, and `name` of the request. None of our statements need it, currently.

chirag-bruno (commented):
Just minor nitpicks. No blockers. Just check and let me know.

packages/bruno-electron/src/services/mount/file-index.js
```diff
@@ -68,12 +40,16 @@ class FileIndex {
...
       const stat = await fs.promises.stat(absolutePath, { bigint: true });
       const mtime = stat.mtimeNs;
-      const prior = stored.get(relativePath);
+      const prior = metadata.get(relativePath);
 
       if (!prior) {
         const hash = await hashFileAsync(absolutePath);
         return { kind: 'added', entry: { relativePath, absolutePath, mtime, hash } };
       }
+      if (prior.applicationVersion !== this.#applicationVersion) {
```
chirag-bruno: Please remove this condition, we have a different approach.

shubh-bruno: removed

chirag-bruno (approved):
LGTM

@sid-bruno, can you please make a final review on this?

packages/bruno-app/src/components/Preferences/Cache/index.js
```diff
@@ -57,15 +49,14 @@ const Cache = () => {
...
-      .invoke('renderer:clear-file-cache')
-      .then((size) => {
-        setFileCacheSize(size);
-        toast.success('File cache cleared');
-      })
-      .catch(() => toast.error('Failed to clear file cache'));
+  const handleClearFileCache = async () => {
+    try {
+      await clearFileCache.mutateAsync({});
+      await reclaimDiskSpace.mutateAsync({});
```
sid-bruno: `reclaimDiskSpace` executes a `VACCUM` on the entire db file, this is problematic considering the same DB file is used for other actions and is going to be used for other coming features. 

I'd remove this for now. As the cache grows with other data the vaccum will hold a lock on the DB tills it's done so if I clear cache and then go do a run on another collection, my DB will be locked and won't be able to write responses to the runner_responses.

sid-bruno: I'd shift to using an incremental vaccum instead of a full vaccum every cache clear, also this can be moved to the app's shutdown phase instead of doing it here

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: moved to constructor

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: moved to constructor

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: Removed normalizing completely

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: Removed normalizing completely

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: Removed normalizing completely

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: Removed normalizing completely

packages/bruno-sqlite/statements/file-index.sql
shubh-bruno: moved to`vacuum.sql`

packages/bruno-sqlite/statements/file-index.sql
shubh-bruno: renamed

packages/bruno-electron/src/services/mount/file-index.js
shubh-bruno: removed in all places

packages/bruno-sqlite/migrations/0000002_file_index_entries.ts
shubh-bruno: removed the comment

packages/bruno-sqlite/migrations/0000002_file_index_entries.ts
shubh-bruno: removed non-needed indexes

packages/bruno-app/src/components/Preferences/Cache/index.js
shubh-bruno: moved to incremental vaccum

packages/bruno-electron/src/services/mount/manager.js
```diff
@@ -137,7 +137,7 @@ class MountManager {
     await Promise.all(
       Array.from(this.#mounts.keys()).map((uid) => this.unmount(uid).catch(() => {}))
     );
-    await destroyPool({ force }).catch(() => {});
+    await destroyPool().catch(() => {});
```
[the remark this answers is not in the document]
shubh-bruno: reverted conflicts issue

sid-bruno (approved):

PR #9231, opened 2026-09-11
fix(editor): disable Alt-G jumpToLine on macOS so Option+G can type @ on German keyboards

### LINKED ISSUE #9114: [Monaco Editor] Alt+G / Option+G triggers "Jump to Line" instead of typing "@" on macOS (German layout)
REF - BRU-4540

### This bug is:

- [x] Making Bruno unusable for me.
- [ ] Annoying.

### Bruno version

4.1.0

### 💻 Operating System

macOS

_No response_

### Description
When using Bruno on macOS with a German or Swiss-German keyboard layout, typing `⌥ Option + G` (the standard shortcut for the `@` character) does not input the symbol. Instead, it triggers the built-in Monaco Editor command "Jump to line". 

Since this shortcut is hardcoded within the embedded editor engine, it completely overrides the macOS system layout behavior, making it impossible to type `@` symbols (e.g., in scripts or documentation) without using external workarounds.

### Expected Behavior
The `@` character should be inserted into the text field.

### Actual Behavior
A "Jump to line" overlay menu pops up, and no character is typed. There is currently no way to unbind or change this shortcut within Bruno's Keybindings preferences.

### Steps to Reproduce
1. Use macOS with a German or Swiss-German keyboard layout.
2. Open any code or script editor window inside Bruno.
3. Press `⌥ Option + G` to type an `@` symbol.

### 📁 Collection to reproduce

_No response_

Image

Turbidity-coder:
BRU-4540

## Summary

On macOS with a German or Swiss-German keyboard layout, pressing `Option+G` (the standard system shortcut for the `@` character) triggers CodeMirror's "Jump to Line" command instead of inserting `@`. This makes it impossible to type `@` symbols in any code/script editor inside Bruno.

## Root Cause

Bruno uses CodeMirror 5 with the `sublime` keyMap. The sublime keymap binds `Alt-G` to the `jumpToLine` command. On macOS, `Option` is equivalent to `Alt`, so `Option+G` is intercepted by CodeMirror before the OS input method can convert it to `@`.

## Fix

Added a platform-conditional keybinding override in `CodeEditor/index.js`:

- On macOS (`navigator.platform.includes('Mac')`), set `'Alt-G': false` to disable CodeMirror's jumpToLine binding
- This allows the macOS input method to handle `Option+G` normally and insert `@`
- Windows/Linux behavior is completely unchanged — `Alt-G` still triggers "Jump to Line" there

## Testing

- The fix is scoped to macOS only via `navigator.platform` detection
- No other keybindings are affected
- The change is minimal (7 lines including comments)

Closes #9114

## Summary by CodeRabbit

* **Bug Fixes**
  * On macOS, Option+G in the code editor can now insert the `@` character without triggering the previous shortcut.
  * XML folding indicators now display arrow characters correctly.
  * XML folding behavior now more accurately handles document children.

### DIFF
--- a/packages/bruno-app/src/components/CodeEditor/index.js
+++ b/packages/bruno-app/src/components/CodeEditor/index.js
@@ -220,7 +220,14 @@ class CodeEditor extends React.Component {
           } else {
             this.editor.toggleComment();
           }
-        }
+        },
+        // On macOS with German/Swiss-German keyboard, Option+G is the standard
+        // shortcut for "@". CodeMirror's sublime keymap binds Alt-G to "jumpToLine",
+        // which overrides the system input method. Disable it on macOS so the OS
+        // can handle Option+G and insert "@" correctly.
+        ...(typeof navigator !== 'undefined' && navigator.platform && navigator.platform.includes('Mac')
+          ? { 'Alt-G': false }
+          : {})
       },
       foldOptions: {
         widget: (from, to) => {
@@ -241,10 +248,10 @@ class CodeEditor extends React.Component {
                 '<a> ' + internal.replace(/(?<=\<|<\/)\w+:/g, '') + '</a>',
                 'application/xml'
               );
-              count = dcm.documentElement.children.length;
+              count = dcm.document.children.length;
             } catch (e) { }
           }
-          return count ? `\u21A4${count}\u21A6` : '\u2194';
+          return count ? `↤${count}↦` : '↔';
         }
       }
     })));


aditya-bruno:
Hey @Turbidity-coder , thanks for the fix. Tested this on macOS with a German layout and can confirm Option+G now types @ correctly in the Script tab and the raw Body editor.

A few spots still have the original problem though: the URL bar, Headers value field, Auth tab, and Params value field. Alt-G still triggers "Jump to Line" first, and @ ends up typed into that prompt instead of the actual field.

Worth prioritizing these, since they're actually where users are more likely to need @ day to day, things like emails in query params, credentials in a URL, or auth headers, more so than in the Script or Body editor. Would be great to apply the same Alt-G: false override wherever that other component sets its keymap, so the fix covers the fields people actually hit this in most.

Issue #9234, opened 2026-09-11
req.getHost() lower cases variables in hostname


harristom:
### I have checked the following:

- [x] I have searched existing issues and found nothing related to my issue.

### This bug is:

- [ ] Making Bruno unusable for me.
- [ ] Slowing me down but I'm able to continue working.
- [ ] Annoying.
- [ ] This feature was working in a previous version but is broken in the current release.

### Bruno version

4.1.0

### 💻 Operating System

Windows

### 🐞 Describe the bug

Variables in the hostname are lowercased by req.getHost(), breaking interpolation

### 📝 Steps to reproduce

Set variable HOST to "example.com"
Set request URL to "https://{{HOST}}/test"
In pre-request script: console.log(bru.interpolate(req.getHost()));
"{{host}}" is logged (the variable has been changed to lowercase and is not interpolated as there is no {{host}}, only {{HOST}})

### 📷 Screenshots/Live demo link

n/a

PR #9164, opened 2026-09-03
feat(scripts): enhanced pre-request scripts handling with environment variable interpolation and added tests

sharan-bruno:
BRU-2778

Fixes https://github.com/usebruno/bruno/issues/9233
Fixes https://github.com/usebruno/bruno/issues/9234

Related PR's:
https://github.com/usebruno/bruno/pull/9240 
https://github.com/usebruno/bruno/pull/9241 
https://github.com/usebruno/bruno/pull/7280

### Description

`req.getPath()`, `req.getHost()` and `req.getQueryString()` now work in pre-request scripts when the request URL contains template variables.

### Problem

Pre-request scripts run before URL interpolation, so `req.url` is still a raw template (`{{BASEURL}}/path/:p1/:p2`). `new URL()` throws on it, the error is swallowed, and all three helpers return `''`.

### Fix

`BrunoRequest` now resolves the URL itself before parsing, via an `interpolate` method mirroring `bru.interpolate` (same nine variable scopes, same precedence). Path param values are interpolated too, matching what `interpolateVars` sends. Behaviour for already-resolved URLs is unchanged.

### Screenshots

<table> <tr> <th>Before</th> <th>After</th> </tr> <tr> <td> <video src="https://github.com/user-attachments/assets/1c927db3-c003-4723-b854-e5bcc0940eca" controls></video> </td> <td> <video src="https://github.com/user-attachments/assets/bc0af045-39c3-4bc7-9acd-c18fb41239e9" controls></video> </td> </tr> </table>

#### Contribution Checklist:

- [ ] **I've used AI significantly to create this pull request**
- [x] **The pull request only addresses one issue or adds one feature.**
- [x] **The pull request does not introduce any breaking changes**
- [x] **I have added screenshots or gifs to help explain the change if applicable.**
- [x] **I have read the [contribution guidelines](https://github.com/usebruno/bruno/blob/main/contributing.md).**
- [ ] **Create an issue and link to the pull request.**
- [ ] **I've run the claude code review skill locally.**

Note: Keeping the PR small and focused helps make it easier to review and merge. If you have multiple changes you want to make, please consider submitting them as separate pull requests.

#### Publishing to New Package Managers

Please see [here](../publishing.md) for more information.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URL handling when template variables remain unresolved in different URL components.
  - Preserved URL fragments during parsing.
  - Preserved complete URL authority information, including credentials.
  - Maintained existing path-parameter substitution behavior.

[...]

PR #9241, opened 2026-09-12
fix(scripts): resolve variables in req.getHost() during pre-request

### LINKED ISSUE #9233: req.getHost() doesn't work if hostname is a variable

### This bug is:

- [ ] Annoying.

### Bruno version

4.1.0

### 💻 Operating System

Windows

_No response_

### 🐞 Describe the bug

req.getHost() returns an empty string if hostname is a variable.

Works if scheme is separate to the variable but has a different issue with case sensitivity that I will file separately

### 📝 Steps to reproduce

Set HOST variable to "https://example.com"
Set request URL to "{{HOST}}/test"
In pre-request script: console.log(req.getHost());
Empty string is logged

### 📁 Collection to reproduce

_No response_

n/a

brbousnguar:
### Description

In a pre-request script, `req.getHost()` returns an empty string when the host comes from a variable. For example, with `HOST=https://example.com` and URL `{{HOST}}/test`, it logs `''`. `req.getPath()` and `req.getQueryString()` have the same problem.

### Problem

Closes #9233

Pre-request scripts run before the request is interpolated. The Electron `runPreRequest` path and the CLI runner both call `interpolateVars` after the script. The URL helpers in `BrunoRequest` call `new URL(this.req.url)` on the raw template:

- `new URL('{{HOST}}/test')` throws, and the helper's `catch` returns `''`.
- `new URL('https://{{HOST}}/test')` parses, but the WHATWG parser lowercases the host to `{{host}}`. This is the scenario in #9234, which has the same root cause.

In post-response scripts and tests the URL is already interpolated, so the helpers work there. The existing `getHost.bru` sanity test (`{{host}}/ping`) only passes because it runs in the `tests` block.

### Fix

- `BrunoRequest` takes an optional `{ interpolate }` option. `getHost()`, `getPath()` and `getQueryString()` now parse the interpolated URL instead of the raw one. Resolution happens at call time, so a variable set earlier in the same script with `bru.setVar()` is picked up, the same way the network layer will resolve it.
- `ScriptRuntime.runRequestScript` passes `bru.interpolate`. This is the same pattern `CookieList` already uses to resolve the request URL. The post-response, test, assert and vars runtimes don't pass it, because their URL is already resolved.
- `req.url` and `req.getUrl()` are unchanged and still return the raw template, and the request object is not mutated.
- Safe mode (QuickJS) is covered too, since its shim calls the host-side `req.getHost()`.

This also fixes the lowercased `{{host}}` case reported in #9234, because the placeholder never reaches `new URL`. I've left that issue open for you to confirm.

One case this PR does not cover: a variable holding a host without a scheme (`HOST=example.com`) still makes `getHost()` return `''`. The network layer prepends `http://` only at send time. Changing that here would also change how literal scheme-less URLs are parsed, so I kept it out of scope.

### Testing

- New `packages/bruno-js/tests/bruno-request-url-helpers.spec.js`. It covers a scheme inside the variable, a scheme outside the variable (#9234), host plus port variables, path params and a query string, call-time resolution, the raw URL staying untouched, and the unresolvable fallback.
- `packages/bruno-js/tests/runtime.spec.js`: `runRequestScript` with `{{HOST}}/test?page=1` in both `nodevm` and `quickjs`, plus a case where the script sets `HOST` before calling `getHost()`.
- With the `src` changes reverted, 9 of these tests fail (`Received: ""` / `Received: "{{host}}"`). With the fix, all pass.

```
cd packages/bruno-js && node --experimental-vm-modules ../../node_modules/jest/bin/jest.js tests/bruno-request-url-helpers.spec.js tests/runtime.spec.js
npm test --workspace=packages/bruno-js        # 34 suites, 815 tests passing
npx eslint packages/bruno-js/src/bruno-request.js packages/bruno-js/src/runtime/script-runtime.js packages/bruno-js/tests/bruno-request-url-helpers.spec.js packages/bruno-js/tests/runtime.spec.js
```

### Screenshots

N/A. This is a scripting API change, covered by unit tests.

#### Contribution Checklist:

## Summary by CodeRabbit

* **Bug Fixes**
  * URL helper methods in pre-request scripts now resolve environment and script-defined variables before extracting the host, path, or query string.
  * Request URLs remain unchanged while their parsed components reflect the latest variable values.

* **Tests**
  * Added coverage for URL parsing with variables, multiple runtimes, dynamic updates, and invalid URLs.

### DIFF
--- a/packages/bruno-js/src/bruno-request.js
+++ b/packages/bruno-js/src/bruno-request.js
@@ -14,8 +14,9 @@ class BrunoRequest {
    * It must be noted that the user cannot set these properties directly.
    * They should use the respective setter methods to set these properties.
    */
-  constructor(req) {
+  constructor(req, { interpolate } = {}) {
     this.req = req;
+    this.__interpolate = interpolate;
     this.url = req.url;
     this.method = req.method;
     this.headers = req.headers;
@@ -48,7 +49,7 @@ class BrunoRequest {
 
   getHost() {
     try {
-      const url = new URL(this.req.url);
+      const url = new URL(this.__getInterpolatedUrl());
       return url.host;
     } catch (e) {
       return '';
@@ -57,7 +58,7 @@ class BrunoRequest {
 
   getPath() {
     try {
-      const url = new URL(this.req.url);
+      const url = new URL(this.__getInterpolatedUrl());
       let pathname = url.pathname;
 
       // If path params exist, interpolate them into the pathname
@@ -91,7 +92,7 @@ class BrunoRequest {
 
   getQueryString() {
     try {
-      const url = new URL(this.req.url);
+      const url = new URL(this.__getInterpolatedUrl());
       // Return query string without the leading '?'
       return url.search ? url.search.substring(1) : '';
     } catch (e) {
@@ -244,6 +245,10 @@ class BrunoRequest {
     } else if (callback) {
       throw new Error(`${callback} is not a function`);
     }
+  }
+
+  __getInterpolatedUrl() {
+    return this.__interpolate ? this.__interpolate(this.req.url) : this.req.url;
   }
 
   __safeParseJSON(str) {
--- a/packages/bruno-js/src/runtime/script-runtime.js
+++ b/packages/bruno-js/src/runtime/script-runtime.js
@@ -53,7 +53,9 @@ class ScriptRuntime {
       certsAndProxyConfig,
       requestUrl: request?.url
     });
-    const req = new BrunoRequest(request);
+    // Pre-request scripts run before the request is interpolated, so the URL helpers
+    // (getHost, getPath, getQueryString) resolve variables themselves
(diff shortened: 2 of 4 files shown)


sanish-bruno:
It is being worked on https://github.com/usebruno/bruno/pull/9164, we do not have an interpolation option there, to those whole need it will still need to use bru.interpolate! Let us know your thoughts @brbousnguar

brbousnguar:
Thanks for flagging it — #9164 looks like the better fix: it resolves the URL with its own `interpolate()` (matching `bru.interpolate`'s scopes/precedence) and also interpolates path-param values, which mine doesn't touch (I called that out as an open gap in my PR description). No reason to carry two competing fixes for the same issues. Happy to close this one in favor of #9164 — let me know if you'd rather I do that, or if there's anything here worth cherry-picking first (e.g. the `getPath()`/`getQueryString()` test coverage).

Issue #9236, opened 2026-09-11
Make Timeline URL selectable by double click


maxbobse:
### I have checked the following:

- [x] I've searched existing issues and found nothing related to my issue.

### This feature

- [ ] Blocks me from using Bruno
- [x] Would improve my quality of life in Bruno
- [ ] Is something I've never seen an API client do before

### 🚀 Describe the feature you want to add, and how it would change your usage of Bruno

My URL is composed of different variables so I can only see the actual URL that was used in the Timeline. If I double click on the timeline to highlight and copy it, it also automatically selects the HTTP method on the left side.

This is annoying because now I have to manually remove that method after pasting the value in my text editor.

<img width="739" height="421" alt="Image" src="https://github.com/user-attachments/assets/28bd7d03-dbfd-4a30-a3e8-63506e2f4f3c" />

ArpanMondalGITHUB:
in my case when i double clicking it only select a word and when i click three time it select everything , i claude it and found it can be a browser's own word-segmentation, which varies.
The real fix: stop relying on native double-click behavior entirely. Add an explicit double-click handler that always selects the entire URL text, deterministically, regardless of browser quirks — and nothing else.

maxbobse:
@ArpanMondalGITHUB thanks, you are correct - double-click for a word, tripple-click selects everythign at the moment. So improvement from my point of view to handle

- double click for single word as is
- triple click for URL

ArpanMondalGITHUB:
@maxbobse  Span have doubletap props so I added a doubletap function on url span and div

PR #9242, opened 2026-09-12
Feature:Make_Timeline_URL_selectable_by_double_click

ArpanMondalGITHUB:
BRU-#9236
### Description

Double-clicking the request URL in the Timeline (both the collapsed row and the expanded detail view) also selected the HTTP method sitting next to it, forcing users to manually delete the method after pasting the copied URL elsewhere. This is especially painful for URLs built from variables — Timeline is often the fastest place to grab the resolved URL mid-debugging, so getting the method tagged along on every double-click is annoying.

solve - #9236

### Fix

- Added an explicit `onDoubleClick` handler (`selectAllTextOnDoubleClick`) that selects the full text content of the clicked element via `Range`/`Selection`, instead of relying on the browser's native double-click word-boundary detection — which was inconsistent (sometimes grabbing the method too, sometimes only a punctuation-delimited fragment of the URL).
- Wired this handler onto the URL element in both places it's rendered: the collapsed row (`.tl-col-url`) and the expanded detail header (`.tl-header-url-text`). It's attached only to the URL element itself, so it's structurally impossible for it to include the adjacent method text.
- In the expanded detail view, the method and URL were previously rendered as two adjacent `<span>`s with no actual space character between them (only a CSS `margin-right`), so their underlying text ran together (e.g. `GEThttps://...`). Added a real space between them and set both spans to `display: inline-block` so each has its own isolated selection boundary.

Double-clicking the URL now always selects exactly the URL, nothing more.
https://github.com/user-attachments/assets/60164d32-fc86-44b4-8a31-b088a31ad7d7

#### Contribution Checklist:

## Summary by CodeRabbit

* **New Features**
  * Double-clicking a timeline request URL selects the full URL text for easier copying.

* **Style**
  * Improved timeline header layout for request methods and URLs.

### DIFF
--- a/packages/bruno-app/src/components/ResponsePane/Timeline/TimelineItem/StyledWrapper.js
+++ b/packages/bruno-app/src/components/ResponsePane/Timeline/TimelineItem/StyledWrapper.js
@@ -121,9 +121,13 @@ const StyledWrapper = styled.div`
     color: ${(props) => props.theme.text};
   }
   .tl-header-url-method {
+    display: inline-block;
     font-weight: 600;
     margin-right: 6px;
     text-transform: uppercase;
+  }
+  .tl-header-url-text {
+    display: inline-block;
   }
   .tl-header-src {
     display: inline-flex;
--- a/packages/bruno-app/src/components/ResponsePane/Timeline/TimelineItem/index.js
+++ b/packages/bruno-app/src/components/ResponsePane/Timeline/TimelineItem/index.js
@@ -165,7 +165,7 @@ const TimelineItem = ({
           <div className="tl-col-method">
             <Method method={method} />
           </div>
-          <div className="tl-col-url" title={url} data-testid="timeline-url">{url} onDoubleClick={selectAllTextOnDoubleClick}</div>
+          <div className="tl-col-url" title={url} data-testid="timeline-url" onDoubleClick={selectAllTextOnDoubleClick}>{url} </div>
           <div className="tl-col-badge">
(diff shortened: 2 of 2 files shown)


aditya-bruno:
Hey @ArpanMondalGITHUB , thanks for putting this together!

Quick thing first though. Generate Code already gives you a fully-resolved URL when "Interpolate Variables" is checked (screenshot attached), so `{{baseUrl}}` turns into the real URL and it's a one-click copy. The PR says Timeline is "the only place" to see that, which doesn't seem right.

Screenshot 2026-09-15 at 2 08 28 PM

What's your use case for this?

ArpanMondalGITHUB:
> Hey @ArpanMondalGITHUB , thanks for putting this together!
> 
> Quick thing first though. Generate Code already gives you a fully-resolved URL when "Interpolate Variables" is checked (screenshot attached), so `{{baseUrl}}` turns into the real URL and it's a one-click copy. The PR says Timeline is "the only place" to see that, which doesn't seem right.
> 
> Screenshot 2026-09-15 at 2 08 28 PM
> What's your use case for this?

Good catch, thanks — you're right, I wasn't aware Generate Code already covers that with Interpolate Variables on. I'll fix the PR description,
The underlying bug still stands though — double-clicking the URL in the Timeline to select+copy it also grabbed the HTTP method, which is annoying since Timeline is still the fastest way to grab the URL without opening a separate Generate Code view. i will frame it as "a quicker way to copy what Timeline already shows" rather than "the only way."
is this will be good ?

PR #9261, opened 2026-09-15
fix(filestore): canonicalise http verbs to upper case on read

### LINKED ISSUE #9249: Generate Code emits the HTTP verb in lowercase, producing a command servers reject
Internal - [BRU-4550]

### This bug is:

- [ ] Annoying.

### Bruno version

4.1.0

### 💻 Operating System

macOS

_No response_

### 🐞 Describe the bug

Generate Code copies the request's method into the snippet exactly as it is held
in memory. When that value is lowercase, the generated command is not runnable:

```bash
curl --request post --url https://api.example.com/users   # server responds 400
```

```
post /users HTTP/1.1
```

HTTP methods are case-sensitive tokens (RFC 9110), so `post` is not `POST`.
Strict servers and CDNs reject it — Auth0's `/oauth/token` and Cloudflare are the
two reported in the wild. The request works inside Bruno but the snippet the user
copies out does not, which is the worst shape for this bug: it looks like a
server problem, not a Bruno one.

This affects every target, not just curl — `wget --method post`, and the raw
`HTTP/1.1` request line — because `HTTPSnippet` copies the HAR `method` through
verbatim and each target renders whatever it is handed.

**Where the lowercase value comes from.** Two paths, both reachable without
hand-editing anything:

1. **`.bru` response examples.** `bruno-filestore`'s bru reader upper-cases the
   parent request's method but passes the raw block header into the example
   (`packages/bruno-filestore/src/formats/bru/index.ts:85` vs the `.toUpperCase()`
   three lines above at `:73`). Generate Code is reachable from an example, so the
   lowercase verb goes straight into the snippet.
2. **`.yml` collections** — the default format. The yml layer normalises in
   neither direction (`formats/yml/items/parseHttpRequest.ts:25`,
   `stringifyHttpRequest.ts:48`), so a lowercase verb is persisted and returned
   verbatim forever. `.bru`, by contrast, is symmetric and safe: lowercase on
   disk is the format's convention (the verb *is* the block header) and it is
   upper-cased on read.

Importers are not a source — Postman and OpenAPI both upper-case on the way in.
Insomnia passes the verb through unchanged, so it inherits whatever the export
contained.

### 📝 Steps to reproduce

Using a collection already committed to this repo, so no hand-written input is
needed:

1. Open `tests/mock-server/fixtures/collections/mock-server-test-collection`.
2. Open the `users/create-user` request — its `.bru` block header is `post {`,
   and the request itself shows **POST** in the UI (correctly upper-cased on read).
3. Open one of that request's **response examples**.
4. Right-click the example → **Generate Code** → **Shell / curl**.

Expected: `curl --request POST …`
Actual: `curl --request post …`

Same result for **wget** (`--method post`) and **HTTP / HTTP1.1**
(`post /users HTTP/1.1`).

For path 2: create a `.yml` collection, set any request's method to `post` in
`opencollection.yml`, reopen the collection, and generate code — the lowercase
verb survives the round trip and reaches the snippet.

### 📁 Collection to reproduce

`tests/mock-server/fixtures/collections/mock-server-test-collection`

Image

Mark007-R:
## Problem

Generate Code copies the request's method into the snippet exactly as it is held in memory. When that value is lower-case, the generated command is not runnable:

```bash
curl --request post --url https://api.example.com/users   # server responds 400
```

HTTP methods are case-sensitive tokens ([RFC 9110 §9](https://www.rfc-editor.org/rfc/rfc9110#section-9)) and every IANA-registered method is upper case, so `post` is not `POST`. Strict servers and CDNs reject it. The request works inside Bruno but the snippet the user copies out does not, which is the worst shape for this bug — it looks like a server problem, not a Bruno one.

This affects every target, not just curl (`wget --method post`, the raw `post /users HTTP/1.1` request line), because `HTTPSnippet` copies the HAR `method` through verbatim.

Two read paths let a lower-case verb reach the model:

1. **`.bru` response examples.** `parseBruRequest` upper-cases the parent request's method, but hands the raw block header to `bruExampleToJson` — so an example inherits `post` where its parent correctly shows `POST`. A `.bru` block header *is* the verb and is lower-case by the format's convention, so this is reachable without hand-editing anything.
2. **`.yml` collections** (the default format). The yml layer normalised in neither direction, so a lower-case verb in `opencollection.yml` survived every round trip and reached the snippet forever.

Fixes #9249

## Fix

Normalise at the **read boundary** so the in-memory model is always canonical, rather than patching the snippet generator. That fixes Generate Code and everything else that renders the method verbatim, and it repairs existing collections on load without rewriting anyone's files — the canonical value is simply what gets written on the next save.

- New shared `normalizeHttpMethod` helper in `bruno-filestore/src/utils`.
- `bruExampleToJson` now normalises the inherited/own method, closing the parent-vs-example asymmetry.
- `parseHttpRequest` normalises both the request and its examples; `parseGraphQLRequest` gets the same treatment, since it is the same class of hole.

Two things deliberately left alone:

- **`toUpperCase()`, not lodash `_.upperCase`** — matching the comment already in `parseBruRequest`, because `_.upperCase` strips the special characters custom methods are allowed to contain (`m-search` → `M SEARCH`).
- **gRPC** stores a case-sensitive fully-qualified path (`/pkg.Service/Method`) in the same field, so it is explicitly excluded.

I did not add write-side normalisation: it would rewrite users' committed collection files on save for no behavioural gain, and read-side alone already fixes the reported symptom.

## Tests

Two new specs, 12 cases:

- `bru/tests/example-method-case.spec.js` — example inherits the parent verb upper-cased (the reported path), example's own lower-case verb, already-upper-case left alone, custom `m-search` keeps its hyphen, missing verb falls back to `GET`, and a gRPC method path is untouched.
- `yml/method-case.spec.ts` — same matrix driven end-to-end from `opencollection.yml` text through `parseItem`, covering the request, its examples, graphql, and gRPC.

Full `bruno-filestore` suite passes (28 suites, 259 tests). `npx eslint` on the changed files reports no new warnings.

## Summary by CodeRabbit

- **Improvements**
  - HTTP and GraphQL request methods are now consistently normalized to uppercase when importing collections and examples.
  - Missing HTTP methods default to `GET`.
  - Existing uppercase methods and custom HTTP method punctuation remain unchanged.
  - gRPC method paths retain their original case-sensitive formatting.
  - gRPC examples without a method now inherit the parent method path correctly.


Mark007-R:
Closing in favour of #9250, which fixes #9249 and was opened first — I missed it when I picked up the issue.

One thing this PR also fixed that #9250 doesn't touch: a gRPC example with no method of its own inherits `GET` instead of the parent's method path, because `parseBruRequest` passes `http.method` to every example regardless of request type. The fix is still on this branch if it's wanted.

PR #9262, opened 2026-09-15
fix(scripts): keep template variables intact in req.getHost()

### LINKED ISSUE #9233: req.getHost() doesn't work if hostname is a variable

### LINKED ISSUE #9234: req.getHost() lower cases variables in hostname

Mark007-R:
## Problem

`getHost()` round-tripped the URL through `new URL()` and returned `url.host`. That throws away the template text a pre-request script is still looking at — scripts run *before* interpolation, so `req.url` holds the user's `{{VAR}}` placeholders at that point. Two reported symptoms, one cause:

**#9234 — the variable is lower-cased.** With `HOST = example.com` and a URL of `https://{{HOST}}/test`:

```js
console.log(bru.interpolate(req.getHost()));   // "{{host}}"
```

For a special scheme the WHATWG host parser runs domain-to-ASCII, which lower-cases the host. `{{HOST}}` becomes `{{host}}`, which matches no variable, so interpolation silently leaves it unresolved.

**#9233 — an empty string.** With `HOST = https://example.com` and a URL of `{{HOST}}/test`, `new URL('{{HOST}}/test')` throws and the `catch` returns `''`.

Fixes #9233
Fixes #9234

## Fix

Read the authority off the raw URL text with plain string work, and return it verbatim when it still contains a `{{var}}`. Callers then interpolate the result themselves, which is how the reporter is already using it.

Plain URLs are deliberately left on the old path — when the authority has no placeholder, the value still comes from `new URL().host`, so the existing behaviour (lower-cased hostname, port included, userinfo excluded, IPv6 brackets kept) is unchanged. The fix only engages where the old code was destructive.

One extra case fell out while testing, and it is the one judgement call in this PR: `new URL()` also returns an **empty** host for a schemeless `localhost:3000/api`, because `localhost:` parses as a scheme and the rest as an opaque path — no exception is thrown, so the old `catch` never fired and `getHost()` returned `''`. Bruno supports schemeless URLs (`DEFAULT_SCHEME` in `bruno-common` exists for exactly this), so I fall back to the raw authority there too. Happy to drop that hunk if you would rather keep this PR strictly to the two reported cases.

## Tests

New `tests/bruno-request-get-host.spec.js`, 14 cases:

- **Templated** — case preserved (`{{HOST}}`, the #9234 regression), variable supplying the scheme (#9233), variable with no path, partially templated host (`{{ENV}}.example.com`), explicit port, and a query string with no path.
- **Plain URLs unchanged** — host, `host:port`, userinfo excluded, literal hostname still lower-cased, IPv6 literal.
- **Edge cases** — schemeless `host:port`, empty URL, missing URL.

Full `bruno-js` suite passes (34 suites, 813 tests). `npx eslint` on both files is clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved request host detection for protocol-relative URLs, such as `//example.com/api`.
  - Preserved template placeholders in hostnames while continuing to handle schemes, ports, query strings, IPv6 addresses, and authentication information correctly.
  - Improved results for schemeless, path-only, empty, or missing URLs.

- **Tests**
  - Added coverage for templated hostnames, protocol-relative URLs, and common URL edge cases.


Mark007-R:
Closing this in favour of #9164, which landed on `main` and fixes #9233 and #9234 by reading the host from the raw URL whenever it's templated. I ran the cases this PR covered against `main` — `https://{{HOST}}/test`, `{{HOST}}/test` and a schemeless `localhost:3000/api` — and they all pass there.

The only thing this still handled is a protocol-relative URL: `//{{HOST}}/api` gives `''` on `main`. Nobody has reported that, so it doesn't seem worth a change of its own.

PR #9264, opened 2026-09-15
fix(filestore): declare nanoid as a runtime dependency

### LINKED ISSUE #9020: @usebruno/filestore requires nanoid at runtime but declares it as a devDependency, so installed consumers can fail with "Cannot find module 'nanoid'"
### Summary

`@usebruno/filestore` requires `nanoid` at runtime, but declares it only as a
**devDependency** — so it is missing from the installed package's dependency tree. Consumers only
work today when `nanoid` happens to be hoisted somewhere the package can reach.

This affects the currently published `@usebruno/filestore@0.11.0` as well as `main`.

### Details

`packages/bruno-filestore/src/utils/index.ts:1`:

```js
const { customAlphabet } = require('nanoid');
```

That call is **not** bundled away. `rollup.config.js` does not list `nanoid` in `externalDeps`, so
the natural reading is that rollup inlines it — but `@rollup/plugin-commonjs` leaves `require()`
inside an ES module untransformed unless `transformMixedEsModules` is set, and this config does not
set it. The call therefore passes through verbatim into the published bundles:

```
$ grep -oE '.{20}nanoid.{10}' dist/cjs/index.js
{customAlphabet:C}=require("nanoid")
```

Meanwhile the manifest has it backwards — `nanoid` under `devDependencies`, and `@types/nanoid`
(build-only) under `dependencies`.

### Why it is not visible in the repo

Five sibling packages declare `nanoid` correctly in `dependencies` — `bruno-app`,
`bruno-converters`, `bruno-electron`, `bruno-js`, `bruno-schema`, all pinning `3.3.18` — so npm
hoists it to the root `node_modules` and filestore resolves it by accident.

It surfaces as soon as a consumer's tree nests `nanoid` under the packages that declare it instead
of hoisting it. The first call into filestore then dies:

```
Error: Cannot find module 'nanoid'
Require stack:
- .../node_modules/@usebruno/filestore/dist/cjs/index.js
- .../node_modules/@usebruno/cli/src/utils/collection.js
- .../node_modules/@usebruno/cli/src/runner/prepare-request.js
- .../node_modules/@usebruno/cli/src/runner/run-single-request.js
```

I hit this installing `@usebruno/cli` + `@usebruno/filestore` from local tarballs, where npm nested
`nanoid` under `@usebruno/js` and `@usebruno/converters` rather than hoisting it.

`uuid()` — the helper built on `customAlphabet` — is called from ~20 sites across
`src/formats/yml/**` on the parse path, so this is hit by ordinary collection reads, not an edge case.

### Suggested fix

Move `nanoid` to `dependencies` (pinned `3.3.18`, matching the rest of the repo).

Two related things worth folding in or filing separately:

1. The same bare `require` also lands in `dist/esm/index.js`, where `require` is undefined under
   native ESM. Declaring the dependency fixes CJS consumers (what `main` resolves to, so all Node
   consumers) but not that. The fix there is `import { customAlphabet } from 'nanoid'` plus adding
   `nanoid` to `externalDeps` in `rollup.config.js` — the pattern `bruno-app` and `bruno-converters`
   already use.
2. `@types/nanoid: ^2.1.0` is types-only, describes nanoid 2's API, and ships to consumers for no
   benefit — nanoid 3.x bundles its own types.

Mark007-R:
## Problem

`@usebruno/filestore` uses `nanoid` at runtime but declared it as a **devDependency**, so it is missing from the installed package's dependency tree.

`src/utils/index.ts` builds `uuid()` on it:

```js
const { customAlphabet } = require('nanoid');
```

`uuid()` is called from ~20 sites across `src/formats/yml/**` on the parse path, so this is hit by ordinary collection reads.

It is not bundled away, either. `rollup.config.js` did not list `nanoid` in `externalDeps`, so the natural reading is that rollup inlines it — but `@rollup/plugin-commonjs` leaves `require()` inside an ES module untransformed unless `transformMixedEsModules` is set, and this config does not set it. I confirmed the call passes through verbatim into **both** published bundles:

```
dist/cjs/index.js:  {customAlphabet:w}=require("nanoid")
dist/esm/index.js:  {customAlphabet:F}=require("nanoid")
```

The ESM one is doubly broken — `require` is undefined under native ESM.

This is invisible in the repo because five sibling packages (`bruno-app`, `bruno-converters`, `bruno-electron`, `bruno-js`, `bruno-schema`) declare `nanoid` correctly, so npm hoists it to the root `node_modules` and filestore resolves it by accident. It surfaces as soon as a consumer's tree nests `nanoid` under the packages that declare it:

```
Error: Cannot find module 'nanoid'
Require stack:
- .../node_modules/@usebruno/filestore/dist/cjs/index.js
- .../node_modules/@usebruno/cli/src/utils/collection.js
```

Affects the published `@usebruno/filestore@0.11.0` as well as `main`.

Fixes #9020

## Fix

- Move `nanoid` to `dependencies`, pinned `3.3.18` to match the rest of the repo.
- Switch the bare `require` to `import { customAlphabet } from 'nanoid'` and add `nanoid` to `externalDeps` — the pattern `bruno-app` and `bruno-converters` already use. Verified against a rebuild:

  | bundle | before | after |
  | --- | --- | --- |
  | `dist/cjs/index.js` | `require("nanoid")` | `require("nanoid")` |
  | `dist/esm/index.js` | `require("nanoid")` | `import{customAlphabet as u}from"nanoid"` |

- Drop `@types/nanoid: ^2.1.0` from `dependencies`. It is build-only, describes nanoid 2's API, and nanoid 3.3.18 ships its own `index.d.ts`. It is referenced nowhere else in the repo, and `tsc --emitDeclarationOnly` still passes without it.
- `package-lock.json` regenerated; the diff is just those two moves.

## Tests

New `src/utils/index.spec.ts`, 4 cases — real coverage for `uuid()` (length, alphabet, no collisions over 500 draws), plus a guard asserting `nanoid` is declared as a runtime and not a dev dependency, so this cannot quietly regress.

Full `bruno-filestore` suite passes (27 suites, 251 tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filestore identifier generation reliability by ensuring the required runtime dependency is available in packaged builds.
  * Confirmed generated identifiers remain 21 characters long, URL-safe, and unique.


Mark007-R:
Closing as a duplicate of #9021, which makes the same fix and was opened first — I missed it when I picked up the issue.

Issue #9018, opened 2026-08-19
bru CLI reads the login shell environment before parsing argv, costing ~480ms on every invocation (including --version and --help)


W-Floyd:
### Summary

`bru` starts the user's login shell to read its environment **before argv is parsed**, so every
invocation pays for it — including `bru --version` and `bru --help`, which never use the result.
On macOS that is roughly **480ms**, more than half the total wall clock of `bru --version`.

Because it is blocking I/O rather than CPU work, it does not appear in a CPU profile, which makes
it easy to miss when profiling CLI startup.

### Where

`packages/bruno-cli/src/index.js` begins `run()` with:

```js
const run = async () => {
  // Fetch shell environment (useful when CLI is run as subprocess from GUI app or cron)
  await initializeShellEnv();
```

That is awaited before yargs parses argv, so it runs for every command and every flag.
`initializeShellEnv` → `fetchShellEnv` → `shellEnv()` from the ESM-only `shell-env` package, which
starts `$SHELL -ilc` and dumps its environment.

### Measurements

M-series Mac, Node 22, measured on `packages/bruno-cli`:

| | wall clock |
| --- | --- |
| `bru --version` as-is | ~890ms |
| `bru --version` without the shell read | ~410ms |

Confirmed it really is a shell start, by watching the process table during `bru --version`:

```
CHILD: /bin/zsh -ilc echo -n "_SHELL_ENV_DELIMITER_"; env; echo -n "_SHELL_ENV_DELIMITER_"; exit
```

For reference, `$SHELL -ilc true` alone takes ~520ms on this machine, so the cost tracks how heavy
the user's shell config is.

Two things worth noting for anyone reproducing this:

- `/usr/bin/time` on macOS folds a waited-for child's CPU into `user`+`sys`, so `user`+`sys` ≈ `real`
  here and looks like the process is CPU-bound when it is actually waiting on a subprocess.
- Setting `SHELL=/bin/sh` does **not** change it. `default-shell` prefers `os.userInfo().shell`
  (from `/etc/passwd`) over `$SHELL`, so that override is ignored.

### Secondary issue: no timeout or memoisation

`bruno-electron` wraps the same call in `src/store/shell-env-state.js` with a 60s timeout and
promise memoisation. The CLI awaits the raw function with neither. `fetchShellEnv` catches
*failures*, but not a shell that never returns — so a `.zshrc` that blocks on input will hang `bru`
indefinitely. The guard already exists in the codebase; it just is not on this path.

[...]

PR #9265, opened 2026-09-15
perf(cli): load command implementations only when selected

### LINKED ISSUE #9022: bru --version and --help load the entire bru run runtime (5182 modules) because .commandDir() requires every command eagerly
### Summary

`bru --version` and `bru --help` load **5182** modules — the entire `bru run` execution runtime —
before printing a single line. The cause is yargs' `.commandDir()`, which eagerly requires every
module in `src/commands` regardless of which command was invoked.

### Details

`packages/bruno-cli/src/index.js`:

```js
const { argv } = yargs
  .strict()
  .commandDir('commands')
```

`.commandDir()` scans the directory and requires each file at registration time, so `commands/run.js`
is always loaded. Its top-level requires pull in the whole runtime — `../runner/run-single-request`
at line 8 transitively brings in `@usebruno/js`, axios, recast, handlebars, jsonwebtoken, yup,
`@usebruno/converters`, `@usebruno/filestore` and ohm.

So `--version`, whose only real need is `CLI_VERSION` from `src/constants.js` (which needs nothing but
`package.json`), pays for the full `run` graph.

### Measurements

M-series Mac, Node 22:

| | module loads |
| --- | --- |
| `bru --version` | 5182 |
| `require('yargs')` alone, for scale | 70 |

Reproduce:

```bash
NODE_DEBUG=module node packages/bruno-cli/bin/bru.js --version 2>&1 | grep -c load
NODE_DEBUG=module node packages/bruno-cli/bin/bru.js --version 2>&1 | grep -c run-single-request
```

The second command returns a non-zero count, confirming `--version` really does load the request
runner.

In wall-clock terms `.commandDir()` accounts for ~380ms of `bru --version`, and it is CPU-bound
(V8 compile plus `node:internal/modules/cjs/loader`), so unlike I/O waits it does show up in a CPU
profile:

```bash
node --cpu-prof --cpu-prof-dir=/tmp/bruprof packages/bruno-cli/bin/bru.js --version
```

### Impact

Mostly felt by short-lived invocations — `--version`/`--help` in shell completions, version checks in
CI, and any tooling that shells out to `bru` for metadata. `bru run` is unaffected either way, since
it needs that graph.

### Suggested fix

Register the commands explicitly and require each module inside its builder/handler, so a command's
runtime loads only when that command is invoked. The trade-off is that `command`/`desc` then have to
be written out rather than read from the module — worth guarding with a test that asserts they stay
in step and that no command module goes unregistered.

Mark007-R:
## Problem

`bru --version` and `bru --help` loaded the entire `bru run` execution runtime before printing a single line.

`src/index.js` registered commands with yargs' `.commandDir()`:

```js
const { argv } = yargs
  .strict()
  .commandDir('commands')
```

`.commandDir()` scans the directory and requires each file **at registration time**, so `commands/run.js` was always loaded. Its top-level requires pull in `../runner/run-single-request`, and transitively `@usebruno/js`, axios, recast, handlebars, jsonwebtoken, yup, `@usebruno/converters`, `@usebruno/filestore` and ohm. Meanwhile `--version` needs only `CLI_VERSION` from `src/constants.js`, which needs nothing but `package.json`.

Measured on this branch's parent (Windows, Node 22):

| | module loads | wall clock (avg of 5) |
| --- | --- | --- |
| `bru --version` before | 4895 | 1667 ms |
| `bru --version` after | **696** | **552 ms** |

It also has a sharper failure mode than just being slow: in a workspace where the sibling packages have not been built yet, `bru --version` does not print a version at all — it dies with `Cannot find module '@usebruno/converters/dist/cjs/index.js'`, because registration reaches for the whole `run` graph.

Mostly felt by short-lived invocations — `--version`/`--help` in shell completions, version checks in CI, and tooling that shells out to `bru` for metadata. `bru run` is unaffected either way, since it needs that graph.

Fixes #9022

## Fix

Register the three commands explicitly and defer each `require` into the builder/handler, so a command's implementation loads only when that command is actually selected.

Command metadata (`command`, `desc`) moves into a new require-free `src/commands/manifest.js`, which `index.js` and the three command modules all read — so the strings stay in exactly one place rather than being duplicated between the registration site and the module.

Two details worth calling out:

- **`--help` output is byte-identical.** Commands are registered in the order `.commandDir()` walked the directory (`docs`, `import`, `run`), so the listing does not reorder.
- **`docs` has no handler of its own** — its builder demands a subcommand — so the wrapper uses `load().handler?.(argv)` to mirror yargs' own no-op default instead of calling `undefined`.

Out of scope: `initializeShellEnv()` still runs before argv is parsed, which is the remaining chunk of `--version` time. That is #9018.

## Tests

New `tests/integration/cli-lazy-command-loading.spec.js`, 9 cases. It runs the real binary under `NODE_DEBUG=module` and asserts against the module trace:

- `bru --version` prints a version, and loads neither `commands/run.js`, nor `run-single-request`, nor `@usebruno/js`.
- `bru --help` still lists all three commands, and loads none of those three either.
- `bru run --help` **does** load `commands/run.js` and prints run's options — the positive control, so the test cannot pass by simply breaking command loading.

I checked the guard actually bites: reverting just the four source files to their previous state and re-running this spec fails all six "does not load" assertions while the three behavioural ones still pass.

Full `bruno-cli` suite passes (35 suites, 383 tests), including the integration tests that drive `bru run` end to end. Manually verified `bru --help`, `bru run --help`, `bru import --help`, `bru docs --help` and `bru docs generate --help` all render as before.

## Summary by CodeRabbit

* **Performance**
  * Improved CLI startup for `bru --version` and `bru --help` by loading command functionality only when needed.

* **CLI**
  * Preserved the existing `docs`, `import`, and `run` commands and their help descriptions.
  * Ensured command-specific help remains available through `bru run --help`.

* **Tests**
  * Added coverage verifying startup behavior, command loading, and help output.
  * Improved test failure detection for unsuccessful or timed-out CLI commands.


Mark007-R:
Closing in favour of #9023, which takes the same approach and was opened first — I missed it when I picked up the issue.

PR #9291, opened 2026-09-18
feat: open sidebar openapi specs as workspace tabs

sachin-bruno:
BRU-3299

### Problem

Clicking an API Spec in the sidebar replaces the whole workspace area with the API Spec panel. The panel renders instead of the tab strip, not inside it, so a spec is never a tab. Only one spec is open at a time, switching specs loses the previous view, and nothing about an open spec survives a restart.

### Fix

API specs used to take over the whole window one at a time, so opening a second one replaced the first.
They now open as workspace tabs and can switch between, they come back after a restart, and unsaved edits are either saved or confirmed on every way of closing a tab.

### Screenshots

| Before | 

image

|After |
image

image

#### Contribution Checklist:

## Summary by CodeRabbit

- **New Features**
  - API specifications now open in dedicated tabs alongside workspace tabs.
  - Edit API specs with draft changes, keyboard saving, and save-on-close support.
  - Unsaved changes are clearly indicated, with options to save, discard, or cancel before closing.
  - API spec tabs and the active tab are restored when reopening or switching workspaces.
  - Newly created API specifications open automatically in their dedicated tab.

- **Bug Fixes**
  - Improved API spec matching across path separators and Windows path casing.
  - Missing or removed specification files are excluded from restored tabs.


packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
```diff
@@ -45,16 +45,35 @@ const MIN_RIGHT_PANE_WIDTH = 450;
...
+  onLeftPaneWidthChange,
+  draftContent,
+  onDraftChange
+}) => {
   const { displayedTheme, theme } = useTheme();
   const preferences = useSelector((state) => state.app.preferences);
 
-  const [editorContent, setEditorContent] = useState(content);
+  const isDraftControlled = typeof onDraftChange === 'function';
+  const [localContent, setLocalContent] = useState(content);
```
sundram-bruno: the component is controlled as well as uncontrolled (via the props draftContent and onDraftChange). now if you see the uncontrolled branch is dead. specviewer has two callers : 1. apispectab which passes ondraftchange. 2. openapispectab which passes readonly. 
and in case 2, the editor renders value=content directly. so the localcontent is never read. 

that localcontent was used by the old fullscreen apispecpanel. which is now removed. 
I guess we can make the component controlled only here. 
can you please check this.

sachin-bruno: done

packages/bruno-app/src/components/ApiSpecTab/index.js
```diff
@@ -0,0 +1,63 @@
...
+  );
+
+  const handleLeftPaneWidthChange = useCallback(
+    (width) => dispatch(updateApiSpecTabLeftPaneWidth({ uid: tabUid, apiSpecLeftPaneWidth: width })),
+    [dispatch, tabUid]
+  );
+
+  if (!apiSpec) {
+    return (
+      <div className="flex items-center justify-center h-full gap-2 opacity-50">
```
sundram-bruno: can we add data testids for this.

packages/bruno-app/src/components/ApiSpecTab/index.js
```diff
@@ -0,0 +1,63 @@
...
+        <IconLoader2 size={20} className="animate-spin" />
+        <span>Loading API spec...</span>
+      </div>
+    );
+  }
+
+  return (
+    <StyledWrapper className="flex flex-col flex-grow relative">
+      <SpecViewer
```
sundram-bruno: data-testid for this as well

packages/bruno-app/src/providers/ReduxStore/slices/workspaces/actions.js
```diff
@@ -13,7 +13,9 @@ import { removeCollection, addTransientDirectory, updateCollectionMountStatus, e
 import { sanitizeName } from 'utils/common/regex';
 import { clearCollectionState } from '../openapi-sync';
 import { updateGlobalEnvironments } from '../global-environments';
-import { addTab, restoreTabs } from '../tabs';
+import { addTab, focusTab, restoreTabs } from '../tabs';
+import { dropApiSpecTabsMissingFrom } from '../apiSpec';
```
sundram-bruno: this creates a circular dependency with apispec.js. since it requires ./workspaces/actions from this file. inside openApiSpecTab and closeApiSpecFile .
dropApiSpecTabsMissingFrom only reads workspaces/tabs state and calls closeTabs, nothing from the apiSpec slice. Can we move it into workspaces/actions.js?

packages/bruno-app/src/components/ApiSpecTab/index.js
```diff
@@ -0,0 +1,63 @@
+import React, { useCallback } from 'react';
+import find from 'lodash/find';
+import { IconLoader2 } from '@tabler/icons';
+import { useDispatch, useSelector } from 'react-redux';
+import SpecViewer from 'components/ApiSpecPanel/SpecViewer';
+import StyledWrapper from 'components/ApiSpecPanel/StyledWrapper';
```
sundram-bruno, line 6: ApiSpecPanel/index.js is deleted in this pr, so components/ApiSpecPanel/ now has no component of its own, but ApiSpecTab and OpenAPISpecTab still import SpecViewer.js and StyledWrapper.js from it.

we can move SpecViewer into its own components/SpecViewer/index.js with its own styledwrapper.js, per the one component per directory rule.

packages/bruno-app/src/components/RequestTabs/RequestTab/index.js
```diff
@@ -343,7 +372,16 @@ const RequestTab = ({ tab, collection, tabIndex, collectionRequestTabs, folderUi
...
         data-testid={`request-tab-${tab.type}`}
         onMouseDown={handleMouseDown}
-        onMouseUp={handleMouseUp}
+        onMouseUp={(e) => {
+          if (e.button === 1 && hasApiSpecDraft) {
+            e.stopPropagation();
+            e.preventDefault();
+            setShowConfirmApiSpecClose(true);
+            return;
+          }
```
sundram-bruno: The middle click check for api spec drafts is added as an inline arrow in JSX that wraps handleMouseUp. handleMouseUp at line 146 already handles the e.button === 1 case, so now the middle-click close logic lives in two places.

Can we move the hasApiSpecDraft check inside handleMouseUp next to the existing middle click branch, and pass onMouseUp={handleMouseUp} as before? Keeps all tab close logic in one handler.

packages/bruno-app/src/components/RequestTabs/RequestTab/ConfirmApiSpecClose/index.js
```diff
@@ -0,0 +1,53 @@
...
+import Portal from 'ui/Portal';
+
+const ConfirmApiSpecClose = ({ name, onCancel, onCloseWithoutSave, onSaveAndClose }) => {
+  return (
+    <Portal>
+      <Modal
+        size="md"
+        title="Unsaved changes"
+        confirmText="Save and Close"
+        cancelText="Close without saving"
```
sundram-bruno: hidefooter is passed as true here. so these will go unused. 
also ConfirmApiSpecClose is same as ConfirmCollectionClose (which already has copy for folder, request , making a total of 4). should we make it a common component or something, can you please check ?

packages/bruno-app/src/components/RequestTabs/RequestTab/ConfirmApiSpecClose/index.js
```diff
@@ -0,0 +1,53 @@
...
+              Don't Save
+            </Button>
+          </div>
+          <div className="flex gap-2">
+            <Button color="secondary" variant="ghost" onClick={onCancel}>
+              Cancel
+            </Button>
+            <Button onClick={onSaveAndClose}>
+              Save
+            </Button>
```
sundram-bruno: please add data-testid for these buttons. e2e takes them by text

packages/bruno-app/src/components/ApiSpecPanel/FileEditor/CodeEditor/index.js
```diff
@@ -96,7 +101,12 @@ export default class CodeEditor extends React.Component {
   componentWillUnmount() {
     if (this.editor) {
       this.editor.off('change', this._onEdit);
+      const editorElement = this.editor.getWrapperElement();
+      if (editorElement && editorElement.parentNode) {
+        editorElement.parentNode.removeChild(editorElement);
+      }
```
sundram-bruno, line 107: componentWillUnmount now removes the CodeMirror wrapper from its parent by hand. The wrapper lives inside this component's own ref'd node, which React removes on unmount anyway, so this is kind of a duplicate work. The shared components/CodeEditor does not do this either.

If there is a case where this is needed,  a comment might help.

packages/bruno-app/src/components/ApiSpecPanel/SpecViewer.js
```diff
@@ -45,16 +45,35 @@ const MIN_RIGHT_PANE_WIDTH = 450;
  *  - onSave                (fn)      Called with current editor content on save (editable mode only)
  *  - leftPaneWidth         (number|null) Persisted left pane width in px; null = use 50/50 default
  *  - onLeftPaneWidthChange (fn)      Persist the new width (called on mouseup / double-click / resize-clamp)
+ *  - draftContent          (string|null) Unsaved content owned by the caller. Pass with `onDraftChange`
+ *                          to hold the draft outside this component, so the caller can read whether
+ *                          there are unsaved edits (tab indicator, save shortcut, close confirmation).
+ *  - onDraftChange         (fn)      Receives every edit; its presence makes the editor controlled.
  */
-const SpecViewer = ({ content, resolvedSpec, readOnly, onSave, leftPaneWidth, onLeftPaneWidthChange }) => {
+const SpecViewer = ({
```
sundram-bruno, line 54: can we add unit test for this ?

packages/bruno-app/src/utils/snapshot/index.js
```diff
@@ -167,6 +167,12 @@ const normalizeWorkspaceSnapshotEntry = (pathname, entry = {}) => {
     activeWorkspaceTabType: WORKSPACE_TAB_TYPES.has(entry.activeWorkspaceTabType)
       ? entry.activeWorkspaceTabType
       : null,
+    apiSpecTabs: Array.isArray(entry.apiSpecTabs)
+      ? entry.apiSpecTabs.map((pathname) => normalizePath(pathname)).filter(Boolean)
+      : [],
+    activeApiSpecTabPathname: typeof entry.activeApiSpecTabPathname === 'string'
```
sundram-bruno, line 173: normalizeWorkspaceSnapshotEntry now handles apiSpecTabs and activeApiSpecTabPathname but this spec has no case for them. Sibling fields like activeWorkspaceTabType are asserted here. Please add a case covering: array of paths gets normalised, non-array becomes [], non-string active path becomes null, and a path that normalises to empty is dropped.

packages/bruno-app/src/providers/ReduxStore/slices/workspaces/apiSpecTabRestore.spec.js
```diff
@@ -0,0 +1,255 @@
...
+
+const WORKSPACE_UID = 'workspace-a';
+const SCRATCH_UID = 'scratch-a';
+const WORKSPACE_PATH = '/workspaces/team';
+const PETSTORE = `${WORKSPACE_PATH}/petstore.yaml`;
+const ORDERS = `${WORKSPACE_PATH}/orders.yaml`;
+
+let workspaceApiSpecPaths = [];
+
+const snapshotWith = ({ apiSpecTabs = [], activeApiSpecTabPathname = null }) => ({
```
sundram-bruno, line 44: Every fixture here goes through snapshotWith, which always sets apiSpecTabs and activeApiSpecTabPathname (even when empty). So nothing proves that a snapshot written before this PR, where those keys do not exist at all, still restores cleanly. Please add one case with a workspace entry that has neither key and assert the workspace opens with no spec tabs and no error.

tests/import/openapi/api-spec-tabs.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    const tmpDir = await createTmpDir('api-spec-bulk-close');
+    const specPath = path.join(tmpDir, EDITABLE_SPEC.file);
+    fs.copyFileSync(fixture(EDITABLE_SPEC.file), specPath);
+
+    await openApiSpecFromDialog(page, electronApp, specPath);
+    await openApiSpecSidebarItem(page, EDITABLE_SPEC.name);
+    await typeIntoApiSpecEditor(page, EDIT_MARKER);
+    await expect(tabUnsavedMarker(EDITABLE_SPEC.file)).toHaveClass(/has-changes/);
+
+    await createTransientRequest(page);
```
sundram-bruno, line 135: This creates an "Untitled" transient request tab and never closes it. afterEach only removes specs, so the tab stays open in the shared window for every later test in this worker. It can also trigger the unsaved transient prompt if a later test does a close-all. Please close it at the end of this test or in afterEach

tests/import/openapi/api-spec-tabs.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+
+    fs.rmSync(secondPath);
+
+    const restarted = await launchElectronApp({ userDataPath });
+    const restartedPage = await waitForReadyPage(restarted);
+    const restartedLocators = buildCommonLocators(restartedPage);
+
+    await expect(restartedLocators.tabs.requestTab(FIRST_SPEC.file)).toHaveCount(1, { timeout: 15000 });
+    await expect(restartedLocators.tabs.requestTab(SECOND_SPEC.file)).toHaveCount(0);
+    await expect(restartedPage.locator('.api-spec-item')).toHaveCount(1, { timeout: 30000 });
```
sundram-bruno: buildApiSpecPanelLocators(restartedPage).sidebarItems() already exists and is used in the same describe , please use the locator.

tests/utils/page/openapi/render-spec.ts
```diff
@@ -28,3 +35,71 @@ export const openApiSpecSidebarItem = async (page: Page, name: string): Promise<
...
+    let remaining = await sidebarItems().count();
+    while (remaining > 0) {
+      const item = sidebarItems().first();
+      await item.hover();
+      await item.locator('.menu-icon').click();
+      await page.locator('.dropdown-item.close-item').click();
+      await page.getByTestId('modal-submit-btn').click();
+      await expect(sidebarItems()).toHaveCount(remaining - 1);
+      remaining -= 1;
+    }
```
sundram-bruno: the loop body is a copy of removeApiSpecFromWorkspace above. can we call that helper from the loop instead ?

tests/import/openapi/api-spec-tabs.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    await expect(sidebarItem(EDITABLE_SPEC.name)).toHaveCount(0);
+    await expect(locators.tabs.requestTab(EDITABLE_SPEC.file)).toHaveCount(0);
+    expect(fs.existsSync(specPath)).toBe(true);
+  });
+});
+
+test.describe('API spec tabs come back after a restart', () => {
+  test.setTimeout(90000);
+
+  const openBothSpecs = async (page: any, app: any, first: string, second: string) => {
```
sundram-bruno: playwright already exports the right types for page and app , refer render-spec.ts which already imports them. 
suggested change : 
import { Page, ElectronApplication } from 'playwright';
const openBothSpecs = async (page: Page, app: ElectronApplication, first: string, second: string) => { ... }

packages/bruno-app/src/providers/ReduxStore/slices/apiSpec.js
```diff
@@ -57,37 +65,120 @@ export const apiSpecSlice = createSlice({
...
+  const tabUids = state.tabs.tabs
+    .filter((tab) => (
+      tab.type === API_SPEC_TAB_TYPE
+      && tab.collectionUid === scratchCollectionUid
+      && !workspacePathKeys.has(getApiSpecPathKey(tab.apiSpecPathname))
+    ))
+    .map((tab) => tab.uid);
+
+  if (tabUids.length) {
+    dispatch(closeTabs({ tabUids, reopenable: false }));
```
sundram-bruno: dropApiSpecTabsMissingFrom closes spec tabs without checking for the unsaved edits.
This runs on every workspace.yml change via loadWorkspaceApiSpecs. if a spec entry is removed from workspace.yml outside Bruno while that spec tab has a unsaved chagnes, the tab just closes. there is no prompt and the edit is just lost. 
Also, it never calls removeApiSpec, so the spec stays in state.apiSpec.apiSpecs with its draft but the sidebar no longer shows it.
can you check it by opening a spec, type something and then remove its entry from workspace.yml in a terminal and save it. tab will disappear silently.

packages/bruno-app/src/components/ApiSpecPanel/FileEditor/CodeEditor/index.js
sachin-bruno: StrictMode remounts on the same node in dev, so the wrapper has to be detached or the remount stacks a second editor on top of the first, and the shared components/CodeEditor does the same thing so this is needed

sundram-bruno (approved):

